PR #14, opened 2018-01-15
Fix Compile: Templated Member Func

ax3l:
Fix compile error:

- missing `template` in call of templated member functions

### DIFF
--- a/include/backend/BaseRecord.hpp
+++ b/include/backend/BaseRecord.hpp
@@ -145,7 +145,7 @@ template< typename T_elem >
 inline std::array< double, 7 >
 BaseRecord< T_elem >::unitDimension() const
 {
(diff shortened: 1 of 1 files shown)


ax3l:
@C0nsultant I feel free to merge this already, `dev` has even more compile errors in `ParallelIOTests` during `loadChunk` and opening of a Series for read (changed APIs).

anokfireball (commented):
The syntax is completely unfamiliar to me. I'm assuming this is required in similar situations and works like ```typename```, but for template member functions?
Approved, since it fixes the issue and is a no-brainer.

ax3l:
@C0nsultant That's a nasty syntax if you call a templated member function when the template can not be derived from a parameter: https://stackoverflow.com/questions/12676190/how-to-call-a-template-member-function

PR #21, opened 2018-01-30
Travis: New Env

ax3l:
Fix and build a new travis environment.
Build dependencies via [spack](https://spack.io/).

- compilers: Clang 5 & GCC 4.8.5
- dependencies: with and without MPI


.travis.yml
```diff
@@ -1,214 +1,114 @@
...
- - cd ADIOS2/v2.${ADIOS2_MINOR}.${ADIOS2_PATCH}
- - sudo make install -j4
- - cd ..
+            - *common_deps
+            - libopenmpi-dev
+            - openmpi-bin
+    - env:
+        - openPMD_USE_MPI=OFF openPMD_USE_HDF5=ON openPMD_USE_ADIOS1=OFF openPMD_USE_ADIOS2=OFF
+      compiler: clang
+    # GCC 4.8.5
```
ax3l, line 40: if gcc 4.8 does not cut it (e.g. missing `<regex>` impl.) we just go for gcc 4.9+: https://docs.travis-ci.com/user/languages/c/

ax3l:
main left-over issues are failing tests, will merge

Issue #19, opened 2018-01-16
Series Constructor: Path & File?


ax3l:
Do we really need both a path and a file (prefix) attribute in the (`Series`) constructors?

Probably it is also possible just to pass one string and split to `basename` with boost, similar to `yt` and `openPMD-viewer`.

anokfireball:
Changed in #24 .

PR #24, opened 2018-02-01
Simplify Series APIs.

anokfireball:
User now only supplies a string encoding previously included variables plus an optional AccessType.
Now you can focus on what you want your output to look like rather than what the API expects.


anokfireball:
Closes #19 .

PR #25, opened 2018-02-01
Split the library into serial and parallel.

anokfireball:
Building and using a built version of libopenPMD now only offers either serial or parallel backends.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -103,11 +103,11 @@ script:
   - mkdir -p $HOME/build 
   - cd $HOME/build
   - if [ ! -d samples/git-sample/ ]; then
-       mkdir -p samples/git-sample/
-       wget -nv https://github.com/openPMD/openPMD-example-datasets/raw/draft/example-3d.tar.gz
-       tar -xf example-3d.tar.gz
-       cp example-3d/hdf5/* samples/git-sample/
-       chmod 777 samples/
+       mkdir -p samples/git-sample/ &&
+       wget -nv https://github.com/openPMD/openPMD-example-datasets/raw/draft/example-3d.tar.gz &&
+       tar -xf example-3d.tar.gz &&
+       cp example-3d/hdf5/* samples/git-sample/ &&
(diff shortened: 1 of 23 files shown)


CMakeLists.txt
```diff
@@ -232,30 +232,18 @@ if(openPMD_HAVE_HDF5)
     target_link_libraries(openPMD.io PUBLIC ${HDF5_LIBRARIES})
     target_include_directories(openPMD.io SYSTEM PUBLIC ${HDF5_INCLUDE_DIRS})
     target_compile_definitions(openPMD.io PUBLIC ${HDF5_DEFINITIONS})
-    target_compile_definitions(openPMD.io PUBLIC "-DLIBOPENPMD_WITH_HDF5=ON")
-    if(openPMD_HAVE_MPI)
-        # TODO: remove and just rely on "WITH_MPI"
-        target_compile_definitions(openPMD.io PUBLIC "-DLIBOPENPMD_WITH_PARALLEL_HDF5=ON")
-    endif()
+    target_compile_definitions(openPMD.io PUBLIC "-DopenPMD_HAVE_HDF5=ON")
```
ax3l: @C0nsultant as pre-processor macros, `1` (ON) and `0` (OFF) are usually better than strings since you can build `#if`s on those

anokfireball: Currently the preprocessor logic relies on a variable being defined.
I agree that a boolean check is better since it conveys the right intention (present or not)

.travis.yml
```diff
@@ -102,6 +102,14 @@ install:
 script:
   - mkdir -p $HOME/build 
   - cd $HOME/build
+  - if [ ! -d samples/git-sample/ ]; then
+       mkdir -p samples/git-sample/
```
ax3l: @C0nsultant this needs concatenation via `&&` (or `;`) per line, but needs no `\`.

`before_script` into the `$HOME/build` dir would be fine, imho

PR #40, opened 2018-02-09
Series: Add 1.1.0 Standard

ax3l:
Add support for version 1.1.0 of the standard in the series object.


anokfireball:
As a sidenote: 1.1.0  is not fully supported yet. The same [UB](https://github.com/yt-project/yt/pull/1645) that resulted 1.0.1 is not yet handled.

ax3l:
Ok, so I will just open an issue for that!

PR #42, opened 2018-02-09
Examples: 1.1.0

ax3l:
The example files on the [example repo](https://github.com/openPMD/openPMD-example-datasets) have changed to openPMD 1.1.0


anokfireball:
Well, that is unfortunate. The new files have different particle counts. 85625 instead of 85000.

ax3l:
it's also a bit of a weird thing to check from those files ^^
such things might change when we update them xD

PR #52, opened 2018-02-11
Iteration: Add Missing template spec.

ax3l:
Add missing template specializations to `Iteration` member functions.


ax3l:
@C0nsultant there are a few more missing, e.g. in `Mesh::setGridSpacing` and `MeshRecordComponent::setPosition`

Issue #50, opened 2018-02-10
bool: h5py compatible


ax3l:
Representing `bool` types in HDF5 is [not pre-defined in HDF5](https://support.hdfgroup.org/HDF5/hdf5-quest.html#bool) and can e.g. be implemented via [enums](https://support.hdfgroup.org/HDF5/doc/H5.user/Datatypes.html#Datatypes_Enum).

Those are a little tricky and even case sensitive but can be implemented in a widely accepted way, that e.g. [`h5py`](http://docs.h5py.org/en/latest/faq.html) understands. See the implementation in libSplash for reference:

https://github.com/ComputationalRadiationPhysics/libSplash/blob/v1.7.0/src/include/splash/basetypes/ColTypeBool.hpp#L35-L89

anokfireball:
I took the guts of the problem straight from splash and integrated them into the HDF5 [backend](https://github.com/ComputationalRadiationPhysics/libopenPMD/blob/dev/src/IO/HDF5/HDF5IOHandler.cpp#L46).

Writing a bool works such that the value is h5py compliant. Reading [apparently](https://github.com/ComputationalRadiationPhysics/libopenPMD/blob/dev/src/IO/HDF5/HDF5IOHandler.cpp#L1122) does not. I'll close the issue once I've verified that reading this non-native bool works ([here](https://github.com/ComputationalRadiationPhysics/libopenPMD/blob/dev/test/SerialIOTest.cpp#L901)).

ax3l:
In the link above, we check for `bools` as follows in splash during reads:
```C++
        // in: hid_t datatype_id

        bool found = false;
        H5T_class_t h5_class = H5Tget_class(datatype_id);
        if(h5_class == H5T_ENUM)
        {
            if(H5Tget_nmembers(datatype_id) == 2)
            {
                char* m0 = H5Tget_member_name(datatype_id,0);
                char* m1 = H5Tget_member_name(datatype_id,1);
                if(strcmp("TRUE" , m0) == 0 && strcmp("FALSE", m1) == 0)
                    found = true;
                free(m1);
                free(m0);
            }
        }
```

If I remember correctly, it's important to check a datatype for `ints` before "bools" before other enums:
https://github.com/ComputationalRadiationPhysics/libSplash/blob/v1.7.0/src/generateCollectionType.cpp#L67-L127

anokfireball:
[...]

PR #56, opened 2018-02-12
Verify h5py compliant bool.

anokfireball:
As bool is not a default type in HDF5, work around this in a way
compliant with h5py. Boolean values are stored in an int8-sized enum,
holding true (1) and false (0).
Verify the correct reading and writing of both values.


anokfireball:
Fixes #50

ax3l:
@C0nsultant tests did not pass. probably the `%T` handling is not flexible enough yet to understand prefix zeros?

anokfireball:
The problem is that I changed from group-based reading to file-based reading (the encoding is fileBased) without reflecting that change in the assertions.
Should be fixed now.

As a sidenote, ```%T``` is treated as the regex ```\d+``` when looking for files, so zerp-prefixes are recognized.

ax3l (approved):

ax3l:
nice, thanks!

Issue #48, opened 2018-02-10
openPMD 1.1.0 Support: optional paths


ax3l:
As a minor improvement of the `1.0.X` series of openPMD, the `meshesPath` and `particlesPath` are not optional and interpreted as "no meshes/particles" if missing. We need to add this logic in both writers and readers.

Changelog of openPMD 1.1.0:

https://github.com/openPMD/openPMD-standard/releases/tag/1.1.0

PR #57, opened 2018-02-12
Comply with 1.0.1 and 1.1.0.

anokfireball:
Gracefully andle the undefined case when meshesPath / particlesPath are
not present as paths, but are in the root path as attributes (1.0.1).
Also enfoce the new 1.1.0 rule that enforces a path when the attribute
exists.


anokfireball:
Fixes #48

.travis/download_samples.sh
```diff
@@ -7,3 +7,13 @@ tar -xf example-3d.tar.gz
...
+
+#https://github.com/yt-project/yt/pull/1645
+mkdir -p samples/issue-sample/
+wget -nv https://github.com/yt-project/yt/files/1542668/no_fields.zip
+unzip no_fields.zip
+mv no_fields samples/issue-sample/
+wget -nv https://github.com/yt-project/yt/files/1542670/no_particles.zip
+unzip no_particles.zip
+mv no_particles samples/issue-sample/
+rm -rf no_fields.zip no_particles.zip;
```
ax3l: missing EOF empty newline :)

src/Iteration.cpp
```diff
@@ -148,15 +148,29 @@ Iteration::flush()
     Writable *w = this;
     while( w->parent )
         w = w->parent;
-    Series* o = dynamic_cast<Series *>(w);
+    Series* s = dynamic_cast<Series *>(w);
 
-    meshes.flush(o->meshesPath());
-    for( auto& m : meshes )
-        m.second.flush(m.first);
+    //TODO warn if openPMD >= 1.1.0, mp is set and no meshes
```
ax3l, line 153: warning on a valid meshesPath without meshes is not necessary, it's just zero meshes.

the only thing that 1.1.0 enforces is to throw an error if meshesPath is set and the path is invalid (group does not exist)

the same for particles below

test/CoreTest.cpp
```diff
@@ -86,35 +84,23 @@ BOOST_AUTO_TEST_CASE(output_default_test)
...
+
+    BOOST_TEST(o.openPMD() == "1.1.0");
+    BOOST_TEST(o.openPMDextension() == static_cast<uint32_t>(0));
+    BOOST_TEST(o.basePath() == "/data/%T/");
+    BOOST_TEST(o.meshesPath() == "customMeshesPath/");
+    BOOST_TEST(o.particlesPath() == "customParticlesPath/");
+    BOOST_TEST(o.iterationEncoding() == IE::groupBased);
+    BOOST_TEST(o.iterationFormat() == "/data/%T/");
+    BOOST_TEST(o.iterations.size() == 1);
+    BOOST_TEST(o.numAttributes() == 7); /* openPMD, openPMDextension, basePath, meshesPath, particlesPath, iterationEncoding, iterationFormat */
```
ax3l, line 102: it's always allowed to have more attributes than openPMD defines, so this is a sub-ideal test. maybe make it `>=7`?

ax3l (approved):
thx, all right!

I added some comments that can be addressed in a follow-up PR.

test/CoreTest.cpp
anokfireball: I don't see how this is sub-ideal. ```numAttributes``` counts the number of currently set Attributes. As this test happens in a very conttrolled environment, we should check precisely whether the defaults reflect the standard in that environment.

src/Iteration.cpp
anokfireball: Brain was not fully working at 7AM yet.
When writing the TODO, I was mislead into thinking the path gets created unconditionally in ```Series```, but it does 6 lines below.

test/CoreTest.cpp
ax3l: oh wait, it's a write test not a read test. you are right!

PR #67, opened 2018-02-14
Enable Parallel Test

ax3l:
Enable the parallel test again

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,7 +352,7 @@ set(openPMD_TEST_NAMES
     Core
     Auxiliary
     SerialIO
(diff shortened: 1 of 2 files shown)


ax3l:
still failing

ax3l:
@C0nsultant looks like there are still some issues in parallel I/O - either in the tests or the actual backends.

I could not reproduce the issues outside of travis, but I guess [valgrind-ing it](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/Debugging#parallel-debugging) could help.

anokfireball:
Fixed in #118.

Issue #70, opened 2018-02-15
Remove Boost Dependency


ax3l:
~~Please replace boost MPL lists with mp11 lists.
Requires Boost 1.66.0+~~

Update: oh wait, that is only boost variant you are controlling, right?

Anyway, careful with global `BOOST_MPL_` defines *in-code*! Better set them in CMake as `PUBLIC` target properties since downstream code might use `BOOST_MPL` as well! (and the compile-time impact of `BOOST_MPL_CFG_NO_PREPROCESSED_HEADERS` might be big).

There are also other C++11 libs of variant that we could consider to use in order to avoid Boost::MPL modifications.

Possible candidates for C++11 variants:

- https://github.com/mpark/variant (preferred)
- https://github.com/mapbox/variant

Possible replacements for C++11 filesystem (we require directory traversal, listing and creation):
- https://github.com/cginternals/cppfs
- https://github.com/wjakob/filesystem
- http://www.martinbroadhurst.com/list-the-files-in-a-directory-in-c.html
- https://github.com/cxong/tinydir
- https://github.com/LLNL/conduit/blob/master/src/libs/conduit/conduit_utils.cpp
- https://github.com/PollardBanknote/cppbackport

anokfireball:
Yes, those [```#defines```](https://github.com/openPMD/openPMD-api/blob/1eb2872cd33814bdfce145ed89a54bc0c2296b88/include/auxiliary/Variadic.hpp#L23) are there to increase the default limit of MPL lists.

Agreed, it is very hacky and unelegant. It it mainly in there since we did have the Boost dependency and I did not want to bother implementing a varaidic/any from scratch.

Agreed, this should be replaced to get rid of even more dependence on Boost in the future.

ax3l:
The last thing that is left is `boost::filesystem` (`boost/filesystem.hpp`).

The functionality we need from it is (@C0nsultant pls append if I missed something):
- `path` (object)
- `exists` (dir, file)
- `directory_iterator`
- `create_directories`
- `remove` (file)

(all of this: Linux/Unix, OSX, Windows)

anokfireball:
That should be it. To put it in terms less specific to Boost:
 - directory existence check
 - directory creation
 - directory content listing
 - file existence check
 - file deletion

ax3l:
[...]

PR #74, opened 2018-02-16
MPark.Variant

ax3l:
Draft for removing `boost::variant` for a `std::variant` in from [MPark.Variant](https://github.com/mpark/variant) #70


anokfireball (approved):
While the dependency is still not nice to have, it's worth it for getting Boost out of the system earlier.

ax3l:
rebased the draft to not use git submodules but the single-header version.

we can later on do exactly the same for replacing boost test with the single-header [catch2 library](https://github.com/catchorg/Catch2)

anokfireball:
@ax3l TIL [Quantstack](https://github.com/QuantStack/xtl/blob/master/include/xtl/xvariant.hpp) uses this mpark::variant as well.

ax3l:
Cool, didn't know that. Good sign :D

PR #77, opened 2018-02-17
MPark.Variant: Fix Include Dir

ax3l:
Use the same internal dir name for `MPark.Variant` as in the official install.

Allows switching to external install of it with ease.

Follow-up to #74


anokfireball (approved):

ax3l:
Feel free to merge :)

Issue #16, opened 2018-01-15
Structure: openPMD/ dir and namespace


ax3l:
All includes should be inside a `openPMD/` directory for grouping and additionally in an `openPMD::` namespace. A collective `openPMD/openPMD.hpp` include file would be nice as well.

ax3l:
I am currently working on this

PR #79, opened 2018-02-19
General Include Dir: `openPMD/`

ax3l:
A general include dir avoids naming collisions in headers after install. We mirror it in the source repo to keep the logic of includes simple.

First part for #16 (namespaces still missing).


test/SerialIOTest.cpp
```diff
@@ -1,11 +1,11 @@
 #define BOOST_TEST_MODULE libopenpmd_serial_io_test
 
-
-#include <boost/test/included/unit_test.hpp>
-
 /* make Writable::parent visible for hierarchy check */
 #define protected public
-#include "Series.hpp"
+#include "openPMD/openPMD.hpp"
+#undef protected
```
anokfireball, line 6: This little detail is very much appreciated.

ax3l: hurray :)

Issue #86, opened 2018-02-20
Travis: OSX


ax3l:
Add a single OSX build to the travis build matrix.

ax3l:
implemented :)
(throws some visibility warnings still)

PR #94, opened 2018-03-09
fix type conflict between Extent and std::vector< size_t >

amundson:
This patch is needed in order to get the code to compile with GCC 7 or Clang 5. (It is not needed for GCC 4.8.) Without this change, both newer compilers give errors like

```
 /Users/amundson/src/libopenPMD/src/Dataset.cpp:37:15: error: no viable overloaded '='
    chunkSize = cs;
 17 {
    ~~~~~~~~~ ^ ~~
/Library/Developer/CommandLineTools/usr/include/c++/v1/vector:545:13: note: candidate function not
      viable: no known conversion from 'const vector<size_t, allocator<unsigned long>>' to 'const
      vector<unsigned long long, allocator<unsigned long long>>' for 1st argument
    vector& operator=(const vector& __x);

```

### DIFF
--- a/include/openPMD/Dataset.hpp
+++ b/include/openPMD/Dataset.hpp
@@ -21,7 +21,7 @@ public:
     Dataset(Datatype, Extent);
 
     Dataset& extend(Extent newExtent);
-    Dataset& setChunkSize(std::vector< size_t > const&);
+    Dataset& setChunkSize(Extent const&);
     Dataset& setCompression(std::string const&, uint8_t const);
     Dataset& setCustomTransform(std::string const&);
 
--- a/src/Dataset.cpp
+++ b/src/Dataset.cpp
@@ -26,7 +26,7 @@ Dataset::extend(Extent newExtents)
 }
 
 Dataset&
-Dataset::setChunkSize(std::vector< size_t > const& cs)
+Dataset::setChunkSize(Extent const& cs)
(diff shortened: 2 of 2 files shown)


ax3l:
Thank you for the report and welcome to openPMD with your first contribution! :sparkles: :rocket: 

Wonderful you even directly propose a fix!

If everything works as expected, we already have Clang 5.0.0 in our (Ubuntu/Linux) test matrix and it seem [to pass](https://travis-ci.org/openPMD/openPMD-api/builds/348932160).  I tried to reproduce the GCC 7 issue in #96 and I can't get it to fail either.

Which version of Clang (GCC) are you using exactly and on which OS? Are you developing on OSX #86? Can you provide additional information on your issue so we can reproduce it?

ax3l (approved):
Look all right, the issue is rooted in #95.

We should also add OSX to the travis compile matrix #86 to make sure we catch this.

amundson:
Sorry for the late reply! I accidentally filtered the messages about this pull request in my email, so I didn't realize there were questions. Thanks for accepting the request. OpenPMD-api now builds cleanly and passes tests on my systems.

ax3l:
Wonderful, great to hear!

If you like, feel free to tell us what you will use it for and open issues if anything is unclear :-)

PR #101, opened 2018-03-12
[WIP] Travis: OSX

ax3l:
Try to add OSX testing (no-MPI): #86 

### To Do

Boost [RT bug](https://travis-ci.org/openPMD/openPMD-api/jobs/353383748):
- [ ] Probably a Boost.Test bug:
```
unknown location:0: fatal error: in "Test setup": boost::runtime::access_to_missing_argument: There is no argument provided for parameter color_output
```
Tried to debug with debug flags in boost but hard to grasp. Tried to run through with `lldb` but looks like the check (`breakpoint set --file /Users/Axel/lib/boost-1.66.0/include/boost/test/utils/runtime/argument.hpp --line 97`) is somehow triggert with a different param... as if something is writing in the stack... Valgrind is complaining about an invalid free from it. clang address sanitizier as well. Seen up to Boost 1.66.0. Also reproducible (besides travis' AppleClang 8) with AppleClang 7.0 on an older local mac.

See also: https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/issues/10#issuecomment-330061796

Maybe we just move to [Catch2](https://github.com/catchorg/Catch2) asap and get rid of it.

- [ ] linker warnings: @bstaletic recommends to hide and explicitly expose in the base lib as well; also positive for binary size and performance. https://gist.github.com/ax3l/ba17f4bb1edb5885a6bd01f58de4d542
issue:
```
ld: warning: direct access in function 'pybind11::detail::erase_all(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)' from file 'CMakeFiles/openPMD.py.dir/src/binding/python/openPMD.cpp.o' to global weak symbol 'std::__1::char_traits<char>::eq(char, char)' from file 'lib/libopenPMD.a(Series.cpp.o)' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ cache:
   directories:
     - $HOME/.cache/spack
     - $HOME/.cache/cmake-3.10.0
+    - $HOME/Library/Caches/Homebrew
   pip: true
 
 env:
@@ -104,6 +105,16 @@ jobs:
             - libopenmpi-dev
             - openmpi-bin
       before_install: *clang50_init
+      script: *script-cpp-unit
+    # Clang 8.1.0-apple + Python 2.7.10 @ OSX
+    - <<: *test-cpp-unit
+      os: osx
+      language: generic
+      env:
+        - USE_MPI=OFF USE_PYTHON=ON USE_HDF5=ON USE_ADIOS1=OFF USE_ADIOS2=OFF
+      compiler: clang
+      before_install:
+         - COMPILERSPEC="%clang@8.1.0"
       script: *script-cpp-unit
     # GCC 4.9.4
     - <<: *test-cpp-unit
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -366,6 +366,9 @@ foreach(testname ${openPMD_TEST_NAMES})
     if(TARGET Boost::unit_test_framework)
         target_link_libraries(${testname}Tests PRIVATE Boost::unit_test_framework)
     endif()
+    if(APPLE)
+        target_compile_definitions(${testname}Tests PRIVATE "-DBOOST_TEST_DYN_LINK")
```
ax3l: [wild guess](https://stackoverflow.com/questions/39171467/there-is-no-argument-provided-for-parameter-color-output-with-boost-test-and-cte), did not work.
remove

ax3l:
I will keep this open for reference and open a new OSX PR after we switched from Boost.Test to Catch2 #110

Sorry Boost.Test, I gave you my precious night-time to fix you and could not succeed. RIP :-)

CMakeLists.txt
mloskot: BTW, or, `target_compile_definitions(${testname}Tests PRIVATE $<$<CXX_COMPILER_ID:AppleClang>:BOOST_TEST_DYN_LINK>)`

ax3l: Thanks, yes that's more compact via a generator expression.
We migrated to Catch2 for testing now #110, and added OSX CI in #115 .

PR #108, opened 2018-03-26
Fix: MPI but no HDF5

ax3l:
Fixes the build if *MPI is found* but *HDF5 is not*.

There are in principle several ways to react on that in the code, either by calling the alternative constructor, returning a dummy and writing an error, throwing or by re-designing the `#if` in the HDF5 parallel backend.

@C0nsultant feel free to revise the strategy as it suits you :-)

### DIFF
--- a/src/IO/AbstractIOHandler.cpp
+++ b/src/IO/AbstractIOHandler.cpp
@@ -38,11 +38,18 @@ AbstractIOHandler::createIOHandler(std::string const& path,
     switch( f )
     {
         case Format::HDF5:
+#   if openPMD_HAVE_HDF5
             ret = std::make_shared< ParallelHDF5IOHandler >(path, at, comm);
+#   else
+            std::cerr << "Parallel HDF5 backend not found. "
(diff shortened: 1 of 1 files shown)


anokfireball:
There was a fail-_safe_ for HDF5+noMPI and noHDF5+MPI already: The construction of the handler would throw an [error](https://github.com/ax3l/libopenPMD/blob/5c5de627a5121f8aed1305110baa129c6ebdef20/src/IO/HDF5/ParallelHDF5IOHandler.cpp#L90). 
... Yes, I used ```throw``` inside a constructor. And yes, I know that is horrible.

ax3l:
Yes, I saw that :D

But on top of that was a *compile error* when MPI was found but *no* HDF5 because the interfaces of the fallback itself differed :)

anokfireball:
So the fallback does not actually work for both combinations. Thanks for pointing this one out.

This PR is a fix for the moment, but I will think about a solution that handles this more generally.

PR #110, opened 2018-04-04
Replace Boost.Test with Catch2

ax3l:
Replace [Boost.Test](https://www.boost.org/doc/libs/1_66_0/libs/test/doc/html/index.html) with [Catch2](https://github.com/catchorg/Catch2).

Step toward #70 and #86

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,7 @@ openpmd_option(ADIOS2 "Enable ADIOS2 support"  OFF)
 openpmd_option(PYTHON "Enable Python bindings" AUTO)
 
 option(openPMD_USE_INTERNAL_VARIANT "Use internally shipped MPark.Variant" ON)
+option(openPMD_USE_INTERNAL_CATCH   "Use internally shipped Catch2" ON)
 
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -563,6 +579,7 @@ message("")
 message("  Additionally installed third party libraries:")
 invertBoolean(openPMD_USE_INTERNAL_VARIANT openPMD_HAVE_EXTERNAL_VARIANT)
 message("    MPark.Variant: ${openPMD_HAVE_EXTERNAL_VARIANT}")
+message("           Catch2: OFF")
```
anokfireball, line 588: Should this do the same inversion as the lines above with ```openPMD_USE_INTERNAL_CATCH```?

anokfireball (approved):
Currently only a change in dependencies, but Catch2 will allow for more readable and maintainable tests.

CMakeLists.txt
ax3l: Good question! No, Catch2 will not be installed on `make install` simply because it is only needed at build-time.

anokfireball: It took me a while to get the intent of the messages.
That is a really elegant and clever solution and makes the Variant output really useful.

ax3l: Haha, thanks.

PR #114, opened 2018-04-05
Wpessimizing-move on return

ax3l:
Fixes clang warning:
```
moving a temporary object prevents copy elision [-Wpessimizing-move]
```

Returning a `std::move` seems to be an anti-pattern.


anokfireball (approved):
As these are merely ```unique_ptr```s, manually moving them would not have benefited much. This will furthermore enable RVO.

PR #126, opened 2018-04-11
ADIOS1 backend

anokfireball:
This adds both a serial and parallel ADIOS1 backend.

One potential drawback of this backend is that it can only mutually exclusively build a non-MPI/parallel version. This might induce contraints on other backends as well, depending on how they are built.

The implementation in this PR does the bare minimum to use ADIOS1 effectively:
 - It considers the typical ADIOS workflow of defining variables before opening a file and tries to reduce the number of times a file has to be (re-)opened. To do this, file opening is deferred until the first write must happen. Two IO queues are used in the backend to discern the two phases of operations (before / after file open). One condition under which re-opening a file can not be avoided is when a manual ```Series::flush()``` is called before new a variable is defined (there may be more complex ones that we have not recognized yet).
 - It bunches reads for every file to schedule a combined read of multiple chunks.
 - It keeps file handles open as long as is possible. It also reopenes them when required.
 - It allows to write dynamically sized chunks per dataset. Typically, ADIOS recommends to define a unique variable for every different-sized/-localized chunk for every dataset. We solve this by creating scalar varaibles for every every dimension of the extent and offset (in "/tmp", so it does not conflict with the openPMD standard). After being defined once, these variables are re-used as the extent and offset for any chunk of a dataset. To do this, the scalar variables just have to be overwritten with the desired extent and offset values before performing the chunk write.


CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ endfunction()
 
 openpmd_option(MPI    "Enable MPI support"     AUTO)
 openpmd_option(HDF5   "Enable HDF5 support"    AUTO)
-openpmd_option(ADIOS1 "Enable ADIOS1 support"  OFF)
+openpmd_option(ADIOS1 "Enable ADIOS1 support"  AUTO)
```
ax3l, line 44: also occurs in `README.md`, `docs/source/dev/buildoptions.rst` and `docs/source/dev/dependencies.rst`

CMakeLists.txt
```diff
@@ -158,11 +158,15 @@ else()
...
 endif()
 
-# TODO: check!
-#if(openPMD_HAVE_MPI AND openPMD_HAVE_ADIOS AND ADIOS_HAVE_SEQUENTIAL)
-#    message(FATAL_ERROR "Found MPI but requested ADIOS1 is serial."
-#                        "Set openPMD_USE_MPI=OFF to disable MPI.")
-#endif()
+# TODO we could support more combinations than MPI+pADIOS and noMPI+sADIOS
+if(openPMD_HAVE_MPI AND openPMD_HAVE_ADIOS1 AND ADIOS_HAVE_SEQUENTIAL)
+    message(FATAL_ERROR "Found MPI but requested ADIOS1 is serial. "
```
ax3l: are you sure a MPI-parallel ADIOS not not still ship a "sequential" component "as well"?

cmake/FindADIOS.cmake
```diff
@@ -178,6 +179,9 @@ if(ADIOS_FOUND)
     endforeach()
     # we could append ${CMAKE_PREFIX_PATH} now but that is not really necessary
 
+    # determine whether found library is serial only
```
ax3l: I think we can only determine:
- is serial *and* parallel
- is serial only

CMakeLists.txt
anokfireball: The library itself might provide both, but with the used ```FindAdios.cmake``` only mutually exclusive versions get linked. Since ADIOS has a unified interface (with [dummy MPI](https://github.com/ornladios/ADIOS/blob/master/src/public/adios_mpi.h) if serial), I doubt you can link both simultaneously.

cmake/FindADIOS.cmake
anokfireball: Alright, the vocabulary in this comment is not sufficient. This determines if the library *links*  as serial only with the specified components.
The library itself might be serial + parallel.

ax3l: ok cool.

want I wanted to say is just: if you build serial, the `lib/` dir will contain a regular and a `_nompi` version, as far as I remember.

If you check that only the noMPI is linked this is fine, but you might want to prefer it someway during search and validate it with two adios installations (/w and /wo MPI), printing all vars to make sure. Also includes (and their `#defines`) should match.

.travis/spack/packages.yaml
```diff
@@ -16,6 +16,9 @@ packages:
       openmpi@1.6.5%gcc@7.2.0 arch=linux-ubuntu14-x86_64: /usr
       openmpi@1.6.5%clang@5.0.0 arch=linux-ubuntu14-x86_64: /usr
     buildable: False
+packages:
```
ax3l: oh oh, pls remove this single line

ax3l:
I think there is more to do for the CI than we can do in the next few days.
Just leave it off for now (remove the last three commits) and we will add travis for ADIOS in a follow-up together end of the week.

anokfireball:
This is probably not a problem with our build process. The errors are caused by timeouts during download of our dependencies (e.g. https://zlib.net was down yesterday). Now, Boost & HDF5 are horribly slow, but making progress.

ax3l:
Yes, also that but I restarted them later on and saw some other issues, also spack-dependency-wise, that I need to investigate properly.

include/openPMD/auxiliary/Memory.hpp
```diff
@@ -83,6 +99,9 @@ allocatePtr(Datatype dtype, size_t numPoints)
             data = new bool[numPoints];
             del = [](void* p){ delete[] static_cast< bool* >(p); p = nullptr; };
             break;
+        case DT::STRING:
+            /* user assings c_str pointer */
```
ax3l: typo: assigns

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -37,31 +95,527 @@ ADIOS1IOHandler::flush()
...
+
+    int64_t id;
+    id = adios_define_var(m_group,
+                          name.c_str(),
+                          "",
+                          datatype,
+                          dims.c_str(),
+                          global_dims.c_str(),
+                          local_offsets.c_str());
+    ASSERT(id >= 0 /* ??? */, "Internal error: Failed to define ADIOS variable during Dataset writing");
```
ax3l: good candidate for a `VERIFY` macro - seen this a lot in the last on startup/env issues.

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -37,31 +95,527 @@ ADIOS1IOHandler::flush()
...
+        case DT::ARR_DBL_7:
+        {
+            auto ptr = reinterpret_cast< double* >(values.get());
+            auto const& arr = att.get< std::array< double, 7> >();
+            for( size_t i = 0; i < 7; ++i )
+                ptr[i] = arr[i];
+            break;
+        }
+        case DT::BOOL:
+            throw std::runtime_error("No workaround for ADIOS1 bool implemented");
```
ax3l: Current PIConGPU representation: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1756

also just as ints, unfortunately.

src/IO/ADIOS/ADIOS1IOHandler.cpp
anokfireball: Very much agreed. We should not exclude these assertions in production code, unless the user *really* wants to.

include/openPMD/auxiliary/StringManip.hpp
```diff
@@ -126,8 +126,9 @@ join(std::vector< std::string > const& vs, std::string const& delimiter)
         default:
             std::ostringstream ss;
             std::copy(vs.begin(),
-                      vs.end(),
+                      vs.end() - 1,
```
ax3l: o.0

this looks dirty and unsave for zero-lengths

ax3l: o.0

this changes looks a bit unsave, e.g. for zero-lengths :)

ax3l: offline discussion: add comment `// do not append separator after last element`

anokfireball: The empty and size 1 vector cases are handled above ;)

ax3l: yep, saw this too late :)

ax3l:
@C0nsultant can you pls add a little to-do list in the PR description what is still missing in this PR? :) Just curious :)

anokfireball:
I added the major two things that have me refrain from merging yet.
Nothing major, at least the testing can be done in a few hours (should have  time for that tomorrow).

ax3l:
sounds good.
leave the CI out for now, it looks still buggy and I had no time to progress on it.

src/IO/ADIOS/ADIOS1IOHandler.cpp
anokfireball: Currently tinkering with this. Adapting the behaviour of using unsigned bytes for bools means we can store them but not read them back typesafe.
Having every uint8_t of value 0 be reinterpreted to a bool false and every uint8_t of value 1 be reinterpreted to a bool true is undesirable when reading a file.

ax3l:
> can only mutually exclusively build a serial/parallel version

@C0nsultant but if I build the parallel version and satisfy the MPI dependency e.g. when running in mpiexec - can I still create a serial `Series` from a single rank? This should work, no?

CMakeLists.txt
```diff
@@ -301,6 +303,10 @@ if(openPMD_HAVE_ADIOS1)
     target_link_libraries(openPMD PUBLIC ${ADIOS_LIBRARIES})
     target_include_directories(openPMD SYSTEM PUBLIC ${ADIOS_INCLUDE_DIRS})
     target_compile_definitions(openPMD PUBLIC "-DopenPMD_HAVE_ADIOS1=1")
+    if(ADIOS_HAVE_SEQUENTIAL)
+        # TODO might be smarter to get ALL definitions from adios-config -s & parse it
```
ax3l: Yes, ideally in `FindADIOS.cmake` we should populate `ADIOS_DEFINITIONS`. We then use that var here.

CMakeLists.txt
```diff
@@ -399,17 +405,20 @@ if(BUILD_TESTING)
     endforeach()
 endif()
 
-foreach(examplename ${openPMD_EXAMPLE_NAMES})
-    if(${examplename} MATCHES ".+parallel$")
-        if(openPMD_HAVE_MPI)
+# Current examples all use HDF5, elaborate if other backends are used
+if(openPMD_HAVE_HDF5)
```
ax3l: don't exclude the build, just the test below is fine.
the examples should still be build and only throw errors at runtime if they try to use a backend that does not exist.

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
```diff
@@ -0,0 +1,150 @@
...
+        case DT::STRING:
+            return adios_string;
+        case DT::VEC_STRING:
+            return adios_string_array;
+        case DT::DATATYPE:
+            throw std::runtime_error("Meta-Datatype leaked into IO");
+        case DT::UNDEFINED:
+            throw std::runtime_error("Unknown Attribute datatype");
+        default:
+            throw std::runtime_error("Datatype not implemented in HDF5 IO");
```
ax3l: ADIOS :)

include/openPMD/Series.hpp
```diff
@@ -279,4 +280,12 @@ Format determineFormat(std::string const& filename);
  * @return  String containing the filename without filename extension.
  */
 std::string cleanFilename(std::string const& filename, Format f);
+
+/** Create a functor to determine if a file can be of a format given the filename on disk.
+ *
+ * @param   name        String containing desired filename without filename extension.
+ * @param   f           File format to check plausibility for.
```
ax3l: Maybe write "backend applicability" instead of "plausibility"?

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -53,15 +240,1228 @@ ADIOS1IOHandler::flush()
...
+            break;
+        case DT::VEC_STRING:
+            nelems = att.get< std::vector< std::string > >().size();
+            break;
+        case DT::ARR_DBL_7:
+            nelems = 7;
+            break;
+        case DT::UNDEFINED:
+        case DT::DATATYPE:
+            throw std::runtime_error("Unknown Attribute datatype");
```
ax3l, line 756: I am wondering if one has still to `break` after a throw...?
probably, no?

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -53,15 +240,1228 @@ ADIOS1IOHandler::flush()
...
+            case adios_string_array:
+            {
+                dtype = DT::VEC_STRING;
+                auto c = reinterpret_cast< char** >(data);
+                std::vector< std::string > vs;
+                vs.resize(size);
+                for( int i = 0; i < size; ++i )
+                {
+                    vs[i] = auxiliary::strip(std::string(c[i], std::strlen(c[i])), {'\0'});
+                    /* TODO pointer should be freed, but this causes memory curruption */
```
ax3l: make this a doxygen todo please: `/** @todo ...`

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -20,36 +20,243 @@
...
+{
+    std::stringstream params;
+    params << "num_aggregators=" << getEnvNum("OPENPMD_ADIOS_NUM_AGGREGATORS", "1")
+           << ";num_ost=" << getEnvNum("OPENPMD_ADIOS_NUM_OST", "0")
+           << ";have_metadata_file=" << getEnvNum("OPENPMD_ADIOS_HAVE_METADATA_FILE", "0")
+           << ";verbose=2";
+    char const* c = params.str().c_str();
+
+    int status;
+    /* TODO ADIOS_READ_METHOD_BP_AGGREGATE */
```
ax3l: `//! @todo ...`

test/SerialIOTest.cpp
```diff
@@ -1087,10 +1089,407 @@ TEST_CASE( "no_serial_hdf5", "[serial][hdf5]" )
...
+    REQUIRE(s.getAttribute("vecUchar").get< std::vector< char > >() == std::vector< char >({'u', 'c', 'h', 'a', 'r'}));
+    REQUIRE(s.getAttribute("vecUint16").get< std::vector< uint16_t > >() == std::vector< uint16_t >({65534u, 65535u}));
+    REQUIRE(s.getAttribute("vecUint32").get< std::vector< uint32_t > >() == std::vector< uint32_t >({4294967294u, 4294967295u}));
+    REQUIRE(s.getAttribute("vecUint64").get< std::vector< uint64_t > >() == std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
+    REQUIRE(s.getAttribute("vecFloat").get< std::vector< float > >() == std::vector< float >({0.f, 3.40282e+38f}));
+    REQUIRE(s.getAttribute("vecDouble").get< std::vector< double > >() == std::vector< double >({0., 1.79769e+308}));
+    REQUIRE(s.getAttribute("vecLongdouble").get< std::vector< long double > >() == std::vector< long double >({0.L, 1.18973e+4932L}));
+    REQUIRE(s.getAttribute("vecString").get< std::vector< std::string > >() == std::vector< std::string >({"vector", "of", "strings"}));
+    REQUIRE(s.getAttribute("bool").get< uint8_t >() == static_cast< uint8_t >(true));
```
ax3l, line 1163: can you add a comment on bool reads please?
maybe we also want to write this in the manual in a section on "backends: features & limitations" or so.

test/SerialIOTest.cpp
```diff
@@ -1087,10 +1089,407 @@ TEST_CASE( "no_serial_hdf5", "[serial][hdf5]" )
...
+        e_3["positionOffset"]["x"].storeChunk({i}, {1}, positionOffset_local_3);
+        o.flush();
+    }
+
+    o.flush();
+}
+
+TEST_CASE( "hzdr_adios1_sample_content_test", "[serial][adios1]" )
+{
+    // since this file might not be publicly available, gracefully handle errors
```
ax3l, line 1338: please add `/** @todo add bp example files to https://github.com/openPMD/openPMD-example-datasets */`

anokfireball:
> but if I build the parallel version and satisfy the MPI dependency e.g. when running in mpiexec - can I still create a serial Series from a single rank? This should work, no?

Yes. Yes, that always works with single-size MPI communicators.
Let's rephrase that to non-MPI and parallel versions.

CMakeLists.txt
anokfireball: I included the examples as tests in CI. With this approach, CI tests without HDF5 will always fail.

src/IO/ADIOS/ADIOS1IOHandler.cpp
anokfireball: Execution at that point aborts until you reach the next mathcing ```catch``` (if it reaches the top level, the process aborts). And even inside that catch, you can not possilbly recover to the point of failure (unless you use ```goto```-fuckery).

So no, there is no ```break``` required after a ```throw``` just like after a ```return```.

ax3l: ok, I was more wondering if the `default` can be reached without the `break`.

CMakeLists.txt
ax3l: no, the `add_tests` are in the block below and properly guarded. Those are the runtimeand require HDF5 files and backend. but the compile should still be done.

ax3l:
Funny, I just saw this: https://github.com/ornladios/ADIOS/blob/v1.13.1/KNOWN_BUGS

That means writing a var for each chunk's dimensions is indeed the right way to go in ADIOS1 ;)

PR #131, opened 2018-04-15
Enclose all exposed classes in namespaces

### LINKED ISSUE #130: Namespace for Tests
Tests should also be placed in C++ namespaces.

This way we can keep them included in doxygen builds but they do not pollute the root scope.

Please see http://www.openpmd.org/openPMD-api/annotated.html for reference :)

anokfireball:
Put the classes inside test cases into namespaces as they are exposed
to the public.

Closes #130.

### DIFF
--- a/test/AuxiliaryTest.cpp
+++ b/test/AuxiliaryTest.cpp
@@ -56,7 +56,7 @@ TEST_CASE( "string_test", "[auxiliary]" )
 
 namespace openPMD
 {
-namespace Test
+namespace test
 {
 struct S : public Writable
 {
@@ -77,7 +77,7 @@ TEST_CASE( "container_default_test", "[auxiliary]")
 
 namespace openPMD
 {
-namespace Test
+namespace test
 {
 class structure : public Attributable
 {
@@ -147,7 +147,7 @@ TEST_CASE( "container_retrieve_test", "[auxiliary]" )
 
 namespace openPMD
 {
-namespace Test
+namespace test
 {
 struct Widget : public Writable
 {
@@ -191,7 +191,7 @@ TEST_CASE( "attributable_default_test", "[auxiliary]" )
 
 namespace openPMD
 {
-namespace Test
+namespace test
 {
 class AttributedWidget : public Attributable
 {
@@ -240,7 +240,7 @@ TEST_CASE( "attributable_access_test", "[auxiliary]" )
 
 namespace openPMD
 {
-namespace Test
+namespace test
 {
 class Dotty : public Attributable
 {
--- a/test/AuxiliaryTest.cpp
+++ b/test/AuxiliaryTest.cpp
@@ -63,7 +63,7 @@ struct S : public Writable
     int val;
     bool written;
 };
-} // Test
+} // test
 } // openPMD
 
 TEST_CASE( "container_default_test", "[auxiliary]")
@@ -89,7 +89,7 @@ public:
     std::string text() const { return variantSrc::get< std::string >(getAttribute("text").getResource()); }
     structure& setText(std::string text) { setAttribute("text", text); return *this; }
 };
-} // Test
+} // test
 } // openPMD
 
 TEST_CASE( "container_retrieve_test", "[auxiliary]" )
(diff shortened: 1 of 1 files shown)


test/AuxiliaryTest.cpp
```diff
@@ -54,20 +54,31 @@ TEST_CASE( "string_test", "[auxiliary]" )
     REQUIRE(expected3 == split("/path/to/relevant/data/", "/"));
 }
 
+namespace openPMD
+{
+namespace Test
```
ax3l: let us rename `Test` -> `test` for sanity :)

test/AuxiliaryTest.cpp
```diff
@@ -54,20 +54,31 @@ TEST_CASE( "string_test", "[auxiliary]" )
...
+{
+struct S : public Writable
```
ax3l: maybe intend by one level here.

test/AuxiliaryTest.cpp
```diff
@@ -54,20 +54,31 @@ TEST_CASE( "string_test", "[auxiliary]" )
...
+    Container< openPMD::Test::S > c = Container< openPMD::Test::S >();
     c.IOHandler = AbstractIOHandler::createIOHandler("", AccessType::CREATE, Format::DUMMY);
 
     REQUIRE(c.size() == 0);
     REQUIRE(c.erase("nonExistentKey") == false);
 }
 
+namespace openPMD
+{
+namespace Test
```
ax3l: `test`

ax3l:
@C0nsultant can you please apply the suggested changes? :)

test/AuxiliaryTest.cpp
anokfireball: Indenting the ```struct``` would make this one class inconsistent with all the rest of the API. I'm for leaving this one unchanged.

ax3l: ok I see, then we can make such changes later on - at once and consistently.
Here is how we usually style nested namespaces: https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#6-namespaces

A namespace only causes intend on nested non-namespaces :)

anokfireball: I do not dislike the style, but that will require a change to every single C++ source file in this project.

test/AuxiliaryTest.cpp
```diff
@@ -78,9 +89,12 @@ class structure : public Attributable
     std::string text() const { return variantSrc::get< std::string >(getAttribute("text").getResource()); }
     structure& setText(std::string text) { setAttribute("text", text); return *this; }
 };
+} // Test
```
ax3l: // test

test/AuxiliaryTest.cpp
```diff
@@ -54,20 +54,31 @@ TEST_CASE( "string_test", "[auxiliary]" )
...
+namespace openPMD
+{
+namespace test
+{
+struct S : public Writable
+{
+    int val;
+    bool written;
+};
+} // Test
```
ax3l: // test

test/AuxiliaryTest.cpp
```diff
@@ -139,9 +157,12 @@ struct Widget : public Writable
     Widget(int)
     { }
 };
+} // Test
```
ax3l: // test

test/AuxiliaryTest.cpp
```diff
@@ -180,9 +205,12 @@ class AttributedWidget : public Attributable
         return getAttribute(key).getResource();
     }
 };
+} // Test
```
ax3l: // test

test/AuxiliaryTest.cpp
```diff
@@ -227,10 +259,12 @@ class Dotty : public Attributable
     Dotty& setAtt2(double d) { setAttribute("att2", d); return *this; }
     Dotty& setAtt3(std::string s) { setAttribute("att3", s); return *this; }
 };
+} // Test
```
ax3l: // test

ax3l (approved):

PR #133, opened 2018-04-16
README: Updates

ax3l:
C++ syntax is now implemented as shown.

Catch2 switch missing for external.


anokfireball (approved):
I bumped the requried ADIOS1 version to ```1.13.0``` in #126, will update the Readme there accordingly.

PR #142, opened 2018-04-23
Write Example: Simplify

ax3l:
Simplify the serial write example a little. Use more standard functionality, remove pointer magic and add comments. Remove "disk" speak.

### DIFF
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2017-2018 Fabian Koller
+/* Copyright 2017-2018 Fabian Koller, Axel Huebl
  *
  * This file is part of openPMD-api.
  *
@@ -22,6 +22,7 @@
 
 #include <iostream>
 #include <memory>
+#include <numeric>
 
 
 using std::cout;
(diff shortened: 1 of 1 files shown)


examples/3_write_serial.cpp
```diff
@@ -22,24 +22,26 @@
...
 {
-    // allocate a data set to write
+    // user input: size of matrix to write, default 3x3
     size_t size = (argc == 2 ? atoi(argv[1]) : 3);
+
+    // matrix dataset to write with values 0...size*size-1
     std::shared_ptr< double > global_data{
         new double[size*size],
-        [](double *p){ delete[] p; p = nullptr; }
+        std::default_delete< double[] >()
```
anokfireball, line 39: Was not aware this worked for C++11 and shared_ptr (shared_ptr does not have an array version until C++17, unique_ptr does from C++11).
It's beautiful.

anokfireball (approved):

examples/3_write_serial.cpp
ax3l: yes, that's weirdly solved in C++11.
shared_ptr always does `delete` and can change the deleter for arrays until the new interfaces in C++17 are there.
Interestingly, unique_ptr can already handle arrays well in C++11 :)
http://en.cppreference.com/w/cpp/memory/default_delete#Example
http://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared

PR #146, opened 2018-04-25
Rewrite ParticlePatches

anokfireball:
Use syntax very similar to Records to store and load single patch
record values.


ax3l:
note: the idea of the "affects latest release" badge is to mark bugs that are in a stable release (and not only in `dev`). We don't have such a release yet :)

ax3l:
@C0nsultant shall I fix the python bindings affected by this for you or are you already working on updating those?

ax3l:
Why don't we add this to `chunks`?

Chunks are (particle) patches. I plan to make them also available for meshes in the future, so that's consistent information that one can add to a stored chunk.

anokfireball:
The Pyhton bindings are failing as not all functions are implemented yet (the culprit is [```ParticlePatches::numPatches()```](https://github.com/openPMD/openPMD-api/pull/146/files#diff-f4a05ee6cfc27ea1cf02968dd626b020R36), which causes link errors). Once that is implemented, at least the build issue should go away.

Anyway, since this PR changes the public C++ API, the Python binding needs to be updated as well. I will take care of this wherever possible.


> Why don't we add this to ```chunks```?

I do not quite understand what you are asking here. Are you suggesting we somehow make the particlePatches a side-effect of ```RecordComponent::storeChunk```?

PR #154, opened 2018-05-03
RTD Build Pygments

ax3l:
Syntax highlighting with pygments in online docs missing.
Looks like pygments is outdated. Add it to requirements.


ax3l:
hm, still did not fix it.
but I also see no errors in the RTD build log... weird, locally it is colored.

PR #156, opened 2018-05-03
Wrap Writables

### LINKED ISSUE #143: Series Factories: Lifetime Control
Currently the `create` and `read` factories of the Series only return an instance.

Nevertheless, the Series is not cleanly movable atm, which makes it cumbersome to control its object lifetime outside of the stack or in member variables, e.g. as smart pointer instances. Also, it makes binding to Python hard since where I would either pass a unique pointer or move it fully into the python ownership.

Should we either return a smart pointer, e.g. shared, or make the class and all classes & members, containers, ... it depends on fully movable? Alternatively, a simple shared ptr as return type might be sufficient, there is probably no reason that these object would need to be unique and can not be shared.

### LINKED ISSUE #144: Reference Return is Dangerous
From the first `1_structure.cpp` example:
```C++
// ...
ParticleSpecies &electrons = series.iterations[1].particles["electrons"];

Record          &mass        = electrons["mass"];
RecordComponent &mass_scalar = electrons["mass"][RecordComponent::SCALAR];
// ...
```

Such constructs can be quite dangerous if users forget the `&`. Are for all those objects the
- copy constructors
- assignment operators
- move constructors?

properly protected?

Otherwise users will copy objects and operate on them without ever modifying the actual ones they aimed for.

Possible, verbose alternative: some kind of handles as wrappers.

cc @psychocoderHPC

anokfireball:
Closes #144 and closes #143.

- [x] Wrap writables so front-end objects are merely handles. This will probably also improve semantic understanding.
- [x] Remove Series factories:
Offer plain old constructors as the factories provided very little benefit for the effort required in python bindings.


ax3l (commented):
In the Python examples, change the imports from
```py
from openPMD import Series
```
to
```py
from openPMD import Series, Access_Type
```

I also forgot to expose the `AccessType` module in `src/binding/python/openPMD.cpp`, here is the full diff:
```diff
diff --git a/examples/2_read_serial.py b/examples/2_read_serial.py
index f7164d6..38256f1 100755
--- a/examples/2_read_serial.py
+++ b/examples/2_read_serial.py
@@ -6,11 +6,11 @@ Copyright 2018 openPMD contributors
 Authors: Axel Huebl
 License: LGPLv3+
 """
-from openPMD import Series
+import openPMD
 
 
 if __name__ == "__main__":
-    series = Series("../samples/git-sample/data%T.h5", Access_Type.read_only)
+    series = openPMD.Series("../samples/git-sample/data%T.h5", openPMD.Access_Type.read_only)
     print("Read a Series with openPMD standard version %s" %
           series.openPMD)
 
diff --git a/src/binding/python/openPMD.cpp b/src/binding/python/openPMD.cpp
index 5c2d9dc..8b4c00c 100644
--- a/src/binding/python/openPMD.cpp
+++ b/src/binding/python/openPMD.cpp
@@ -9,6 +9,7 @@ namespace py = pybind11;
 
 
 // forward declarations of exposed classes
+void init_AccessType(py::module &);
 void init_BaseRecord(py::module &);
 void init_BaseRecordComponent(py::module &);
 void init_Container(py::module &);
@@ -29,6 +30,7 @@ PYBIND11_MODULE(openPMD, m) {
     // m.doc() = ...;
 
     // note: order from parent to child classes
+    init_AccessType(m);
     init_Container(m);
     init_BaseRecord(m);
     init_Dataset(m);
```

We could also just do the classic `'r'`, `'w'`, `'a'` strings here instead of enums. I can do that as a refactoring later on.

include/openPMD/Series.hpp
```diff
@@ -50,20 +50,12 @@ class Series : public Attributable
...
-#endif
-    static Series create(std::string const& filepath,
-                         AccessType at = AccessType::CREATE);
-
-#if openPMD_HAVE_MPI
-    static Series read(std::string const& filepath,
-                       MPI_Comm comm,
-                       AccessType at = AccessType::READ_ONLY);
+    Series(std::string const& filepath,
+           AccessType at,
```
ax3l, line 54: did you intentionally not put the `AccessType` last? Having it at the last position will allow us to put a good default there later on (or are you planning to default to `MPI_COMM_WORLD` as communicator? probably that would be too implicit.)

ax3l:
@C0nsultant can you enable ["allow edits from maintainers"](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) in this pull request? I can then push you the commit that fixes your build.

anokfireball:
> can you enable "allow edits from maintainers"

I have never explicitly disabled the feature. It is currently allowed, so you should be good to go as you have write access to this repository.

include/openPMD/Series.hpp
anokfireball: The idea of a default ```AccessType``` was intuitive when the factories were in place, but IMO there is no clear, sane default for the constructor now.
We could think about assuming ```READ_ONLY``` if a file exists on disk and ```CREATE``` if it doesn't. Personally, I'd prefer to explicitly state it in any case. Users would have to look up the assumed behaviour rather than see it. Also, the constructor will probably not be called often.

Defaulting to ```MPI_COMM_WORLD``` would actually be a very bad idea: The implicit parallel constructor would then be equivalent to the explicit serial constructor.

ax3l:
It's about the branch of a PR which sits not in the main repo :)
Not it works, can push :)

include/openPMD/Series.hpp
ax3l: I agree to both, still a (to be determined) default for `AccessType` would be (if any) the only one we might want to put a default on so we could keep it last.

anokfireball:
We should also offer means of copying resources (i.e. let the user explicitly do what this PR removes):

```cpp
Series s = Series("./out.h5", AccessType::CREATE);

Iteration i1 = s.iterations[1];
/* ... modify i1 significantly ... */

/* does not acutally do what is intuitive (but reason for this PR):
 * merely creates a view into i1
 * every modification to i1 is reflected in i2 (and vice versa)
 * this might even be desirable in some cases */
s.iterations[2] = i1;

/* have an independent copy of the iteration that is not merely a view
 * any modification to i1 will not be reflected in i2 (and vice versa) */
s.iterations[3] = i1.deepCopy(); // or free function, whatever we decide on
```

ax3l:
Great idea. I think as in our `.deepCopy()` example is a good way to go. Numpy does the same, it simply calls it [`.copy()`](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.copy.html) and otherwise passes references/handles around as we do.

```C++
Series s = Series("./out.h5", AccessType::CREATE);

// handle
Iteration i1 = s.iterations[1];

// decoupled object
Iteration i2 = s.iterations[1].copy();
// changing i2 will not change i1 anymore
```

ax3l (approved):
@C0nsultant Thank you for the update!

In case you don't want to switch the access-order of the MPI-Series constructor, feel free to merge.

ax3l:
@C0nsultant is this merge-able or still `[WIP]`?

anokfireball:
We're good with respect to fixing #143 and #144 . I'll create a seperate PR for the ```copy()``` behaviour as this is a blocker.

PR #164, opened 2018-05-08
ADIOS filesystem: path shadowed by var

ax3l:
I sporadically get compile errors on Linux because the `path` type from boost filesystem is shadowed by a locally defined variable `path` (line 609) in one of the methods of the ADIOS1 backend.

This fixes it.


src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -622,12 +622,12 @@ ADIOS1IOHandlerImpl::deleteFile(Writable* writable,
         if( !auxiliary::ends_with(name, ".bp") )
             name += ".bp";
 
-        using namespace boost::filesystem;
-        path file(name);
```
ax3l, line 626: `path` is shadowed from a locally defined var in line 609 above

PR #165, opened 2018-05-08
[WIP] Symbol Visibility

ax3l:
For supported compilers, hide all non-public API symbols by default.

Defines `OPENPMD_PUBLIC` as a macro for all public classes, functions, etc. See details in [this gist](https://gist.github.com/ax3l/ba17f4bb1edb5885a6bd01f58de4d542).

This is a requirement for Windows support, which hides symbols by default in its dlls unless explicitly exposed.

Size of `libopenPMD.so`:
- before hiding: 2,2 MB
- after hiding: 2,1 MB

## To Do

- [x] maybe rename macro to shorter `OPENPMD_PUBLIC`
- [ ] Surprisingly, this does not yet remote the [linker warnings](https://travis-ci.org/openPMD/openPMD-api/jobs/376415481) we see on OSX. I probably have to set the same properties on the created executables.
- [x] lol, on windows we have to switch the headers from export to import as soon as we use the dlls: https://msdn.microsoft.com/en-us/library/8fskxacy.aspx
- [x] lol, publicly passing stl objects, e.g. in `Dataset` as `std::string` members or our usage of `std::vector< uint64_t >` need to be dllimported as well: https://social.msdn.microsoft.com/Forums/vstudio/en-US/df99b712-c00b-4af0-82fd-3764c8b6cbec/exporting-stdstring-from-a-dll-does-not-export-stdstringnpos?forum=vcgeneral - on the other hand, exposing STL objects seems to be considered a very bad idea in this thread and they are big time into C-objects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ openpmd_option(ADIOS1 "Enable ADIOS1 support"  AUTO)
 openpmd_option(ADIOS2 "Enable ADIOS2 support"  OFF)
 # openpmd_option(JSON "Enable JSON support" AUTO)
 openpmd_option(PYTHON "Enable Python bindings" AUTO)
+openpmd_option(SYMBOLHIDING "Enable symbol hiding in libraries" AUTO)
 
 option(openPMD_USE_INTERNAL_VARIANT "Use internally shipped MPark.Variant" ON)
 option(openPMD_USE_INTERNAL_CATCH   "Use internally shipped Catch2" ON)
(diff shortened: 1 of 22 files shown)


ax3l:
@C0nsultant I can keep this open and rebase it after you have your changes in #156 implemented and merged

CMakeLists.txt
```diff
@@ -238,6 +238,8 @@ set_target_properties(openPMD PROPERTIES
     CXX_EXTENSIONS OFF
     CXX_STANDARD_REQUIRED ON
     POSITION_INDEPENDENT_CODE ON
+    CXX_VISIBILITY_PRESET hidden
```
ax3l: for consistency with the `Visibility.hpp`  header we should check here:
- is this compiler MSVC or
- does this compiler have support for `__attribute__((visibility("default")))`

then hide, otherwise don't.

include/openPMD/auxiliary/Visibility.hpp
```diff
@@ -0,0 +1,38 @@
...
+ * Added to the symbols of the public API.
+ * Needs to be used as qualifier for all publicly exposed, compiled
+ * functionality.
+ */
+#if defined(__GNUC__) || defined(__clang__) || defined(__ICL) || defined(__xlC__)
+#   define OPENPMDAPI_PUBLIC __attribute__((visibility("default")))
+#elif defined(_MSC_VER)
+#   define OPENPMDAPI_PUBLIC __declspec( dllexport )
+#else
+#   define OPENPMDAPI_PUBLIC
```
psychocoderHPC: maybe throw also a warning that "maybe" all symbols are exported because a program developed against the lib with all exposed symbols can be incompatible with a library version compiled with hidden symbols if non public member functions are used.

CMakeLists.txt
ax3l: Maybe we can check via `$<COMPILE_FEATURES:cxx_attributes>` but checking specific attributes inside those is not yet test-able in CMake.

https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES

Will a compile test via `check_cxx_source_compiles` properly fail for unsupported compilers or will some just throw a warning or silently remove unsupported attributes?
https://stackoverflow.com/questions/50157727/how-to-check-for-compiler-support-of-certain-keywords-attributes

include/openPMD/auxiliary/Visibility.hpp
ax3l: Ok, but this would mean I will throw a warning just because e.g. PGI us used (without symbold hiding support) - even if only public API is properly used.

ax3l:
Hurray, we don't need this. CMake to the rescue! <3
https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

PR #168, opened 2018-05-14
Tests: Add Std Includes

ax3l:
Add missing stdlib includes to tests.

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -9,7 +9,17 @@ using namespace openPMD;
 #include <catch/catch.hpp>
 #if openPMD_HAVE_MPI
 #   include <mpi.h>
+#endif
 
+#include <iostream>
+#include <algorithm>
+#include <string>
+#include <vector>
+#include <array>
+#include <memory>
+
+
+#if openPMD_HAVE_MPI
 
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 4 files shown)


test/ParallelIOTest.cpp
```diff
@@ -9,8 +9,18 @@ using namespace openPMD;
 #include <catch/catch.hpp>
 #if openPMD_HAVE_MPI
 #   include <mpi.h>
+#endif
```
anokfireball, line 12: Except for catch, there's no need for _any_ headers when there's no MPI at the end of the day.

ax3l: ;-)

The all-serial use case will be the windows use case xD

ax3l: Since the tests have no headers and use these stdlib containers and algos directly, it's save to include them. Otherwise they are only pulled indirectly and this might change on unknown implementation details.

Or am I missing something?

anokfireball: Maybe we are concerned about different things. I absolutely agree with the inclusion of the headers when the stdlib is used.
But this is the _parallel_ test suite. So if MPI is not present, no tests are currently executed, therefore stdlib functionality and headers are not required outside the ```#ifdef```s.

ax3l: Ah I see, yes that's true. I just included them in a general scope of the file because of laziness. Didn't check the exact requirements of the fallbacks.

anokfireball: It should be irrelevant as long as we do not try to minimize compile time at all cost.

PR #169, opened 2018-05-14
Container Particle Species: Fix Overload

ax3l:
The overload of the `Container::operator[]` can not be done by re-defining the symbol again a second time. This surprisingly works with the Linux/OSX linkers but is hacky.

Instead, we can define a trait to pass additional modifications after creation/emplacement into a container.

As a nice side-effect, this also removes code-duplication.

Fixes a linker issue on MSVC as well.


anokfireball (approved):
Ah yes, the bauty of cross-platform compiler compliance. Thanks for taking the time to get MSVC to work (and fixing my non-standard-compliance).

ax3l:
rebased against the latest changes.

no problem, I am always happy when the effort unveils at least one or two little bugs one did not expect. Otherwise it would be only the pain of getting a new env up.

Issue #63, opened 2018-02-13
CI: Appveyor


ax3l:
Add appveyor to CI for testing Windows platforms.

Serial testing and python support is probably sufficient.

ax3l:
For this to be implemented, one first needs to properly implement [symbol visibility](https://gist.github.com/ax3l/ba17f4bb1edb5885a6bd01f58de4d542) throughout all public APIs. (Windows libraries have all symbols hidden by default.)

Related drafts:
- appveyor.yaml and windows CMake fixes: https://github.com/ax3l/openPMD-api/tree/topic-appveyor
- ~~symbol visibility: #165~~ we can use [a clever CMake flag](https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/) instead

PR #172, opened 2018-05-14
Container: Read Only Throw Missing

ax3l:
Missing keys in read-only containers should throw.

## To Do

- [x] fix existing tests ~@C0nsultant feel free to go on, I am a bit stuck why some throw~
- [x] add: C++ tests that throw
- [x] Py: throws a clean `IndexError` that can be catched :)

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -75,6 +75,21 @@ class Attributable
     friend class Iteration;
     friend class Series;
 
+    template< typename T_Key >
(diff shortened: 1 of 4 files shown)


anokfireball:
The problem here was subtle, but fatal:
```Series``` opened as ```READ_ONLY``` could not acutally be read in. When creating the first Iteration that was not yet in the container, an error was thrown since the ```AccessType``` was ```READ_ONLY```.

ax3l:
Smart, nice!

ax3l:
lol, I just broke MSVC #63 again with this merge xD

PR #173, opened 2018-05-15
Windows Support and AppVeyor CI

ax3l:
Adds support for MSVC builds on Windows.

Due to various constrains in dependencies, this is
- Python 3 only
- HDF5 only
- Serial only (this is fine)

ADIOS1 on MSVC is very unlikely but ADIOS2 in the future could be a possibility, also for conda-forge.

Related to #63

## To Do

- [x] clean up verbose lines from experimenting on Windows
- [x] finalize PowerShell download script
- [x] README: Add Badge
- [x] tests: does not link with Catch2 enabled
- [x] tests: python bindings build, but do they run?
- [x] install? looks ok-ish, might need to add a link to the main `openPMD.dll` later on as well
- [x] enable AppVeyor in main repo as well


test/SerialIOTest.cpp
```diff
@@ -815,7 +815,11 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
         s.setAttribute("vecUint64", std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
         s.setAttribute("vecFloat", std::vector< float >({0.f, 3.40282e+38f}));
         s.setAttribute("vecDouble", std::vector< double >({0., 1.79769e+308}));
+#ifdef _MSC_VER
+        s.setAttribute("vecLongdouble", std::vector< long double >({0.L, 1.49769e+308L}));
```
anokfireball: All of these calls could be replaced with [```std::numeric_limits::max```](http://en.cppreference.com/w/cpp/types/numeric_limits/max). This would make it compatible across all compilers (_that conform to the C++ standard_).

ax3l: Yep, just a quick hack so far :)

ax3l:
@C0nsultant I think the tests are not compiling because of the redefinition of `private/protected` to `public` which causes binary incompatibilities in MSVC: https://github.com/catchorg/Catch2/issues/258#issuecomment-195479675

Can you maybe control these "very internal unit tests" via a define so I can skip them in MSVC?

anokfireball:
> Can you maybe control these "very internal unit tests" via a define so I can skip them in MSVC?

Just so I get this right - you have something like ```-DopenPMD_NONINVASIVE_TESTS=ON```  in mind, which then skip all tests that rely on ```#define public private```?
I'd rather have some modifier like [this](https://github.com/nlohmann/json/issues/913#issuecomment-357508906) that allows us to still test the data.

ax3l:
No, really the first one.

The second solution means we would ship different access modifiers on Windows which is a hack. Instead, let us just test *observable* behavior on Windows - that's ok.

anokfireball:
> let us just test observable behavior on Windows

Sighh...
The platform is not out main target audience anyway, so that should be fine. Core tests then can be designated for unix and assumed to represent all other platforms. Not elegant, but it does the job.

CMakeLists.txt
```diff
@@ -240,7 +249,23 @@ set_target_properties(openPMD PROPERTIES
...
+    if(BUILD_SHARED_LIBS)
+        target_compile_options(openPMD PUBLIC "/MD")
+    else()
+        target_compile_options(openPMD PUBLIC "/MT")
+    endif()
+endif()
+if(WIN32)
+    # generate a the openpmd_export.h header with export defines we can use for
+    # static members and other global data
+    include(GenerateExportHeader)
```
ax3l: Comment out if we don't need this currently.

CMakeLists.txt
```diff
@@ -240,7 +249,23 @@ set_target_properties(openPMD PROPERTIES
     CXX_EXTENSIONS OFF
     CXX_STANDARD_REQUIRED ON
     POSITION_INDEPENDENT_CODE ON
+    WINDOWS_EXPORT_ALL_SYMBOLS ON
 )
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+    target_compile_options(openPMD PUBLIC "/bigobj")
+    if(BUILD_SHARED_LIBS)
+        target_compile_options(openPMD PUBLIC "/MD")
```
ax3l: check if we still need this block

.appveyor.yml
```diff
@@ -0,0 +1,140 @@
...
+  - cmd: mkdir build
+  - cmd: cd build
+
+  # Download example files
+  - ps: ..\.travis\download_samples.ps1
+
+  # Compiler & Generator Selection
+  - cmd: if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" set OPENPMD_CMAKE_GENERATOR=Visual Studio 14 2015
+  - cmd: if "%TARGET_ARCH%"=="x64" set OPENPMD_CMAKE_GENERATOR=%OPENPMD_CMAKE_GENERATOR% Win64
+#  - cmd: if "%TARGET_ARCH%"=="x86" "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64_x86
```
ax3l, line 127: note: I keep this for docs, this line is needed if we want to build everything from a `amd64` platform

test/SerialIOTest.cpp
```diff
@@ -861,7 +864,9 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
     REQUIRE(s.getAttribute("vecUint64").get< std::vector< uint64_t > >() == std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
     REQUIRE(s.getAttribute("vecFloat").get< std::vector< float > >() == std::vector< float >({0.f, 3.40282e+38f}));
     REQUIRE(s.getAttribute("vecDouble").get< std::vector< double > >() == std::vector< double >({0., 1.79769e+308}));
-    REQUIRE(s.getAttribute("vecLongdouble").get< std::vector< long double > >() == std::vector< long double >({0.L, 1.18973e+4932L}));
+#if !defined(_MSC_VER)
```
anokfireball, line 867: #179 is really this bad? Sad to see as ```long double``` is part of the C++ standard.

test/SerialIOTest.cpp
```diff
@@ -1171,7 +1176,9 @@ TEST_CASE( "adios1_dtype_test", "[serial][adios1]" )
     REQUIRE(s.getAttribute("uint64").get< uint64_t >() == 64u);
     REQUIRE(s.getAttribute("float").get< float >() == 16.e10f);
     REQUIRE(s.getAttribute("double").get< double >() == 1.e64);
+#if !defined(_MSC_VER)
     REQUIRE(s.getAttribute("longdouble").get< long double >() == 1.e80L);
```
anokfireball, line 1180: Why not use ```std::numeric_limits<long double>::max()``` here as well? Is it the initializer list causing the problems?

anokfireball (approved):

test/SerialIOTest.cpp
ax3l: The problem I caused by the `long double` template argument during read #179 
I will exclude it for now and open a PR that removes the exclusion to demonstrate.

test/SerialIOTest.cpp
ax3l: only during read :)
I think for some reason the variant takes the `double` alternative and then tries to read as `long double` or similar

PR #180, opened 2018-05-17
Module Feature Variants: As Map

ax3l:
Exposing them as a single attribute map is probably slicker than #178.

### DIFF
--- a/src/binding/python/openPMD.cpp
+++ b/src/binding/python/openPMD.cpp
@@ -1,8 +1,10 @@
 #include <pybind11/pybind11.h>
-//include <pybind11/stl.h>
+#include <pybind11/stl.h>
 
 #include "openPMD/version.hpp"
 
+#include <string>
+#include <map>
 #include <sstream>
 
 namespace py = pybind11;
@@ -55,11 +57,12 @@ PYBIND11_MODULE(openPMD, m) {
         openPMDapi << "-" << OPENPMDAPI_VERSION_LABEL;
     m.attr("__version__") = openPMDapi.str();
 
-    // variants
-    m.attr("have_mpi") = bool(openPMD_HAVE_MPI);
-    m.attr("have_hdf5") = bool(openPMD_HAVE_HDF5);
-    m.attr("have_adios1") = bool(openPMD_HAVE_ADIOS1);
-    m.attr("have_adios2") = bool(openPMD_HAVE_ADIOS2);
-    // m.attr("have_json") = bool(openPMD_HAVE_JSON);
+    // feature variants
+    m.attr("variants") = std::map<std::string, bool>{
+        {"mpi", bool(openPMD_HAVE_MPI)},
+        {"hdf5", bool(openPMD_HAVE_HDF5)},
+        {"adios1", bool(openPMD_HAVE_ADIOS1)},
+        {"adios2", bool(openPMD_HAVE_ADIOS2)}
+    };
 }
 


anokfireball:
For reference, [this](https://github.com/yt-project/yt/blob/master/yt/__init__.py#L86) is how yt exposes their [frontends](https://github.com/yt-project/yt/blob/master/yt/frontends/api.py#L56) (granted, they are modules and not just boolean values as in our case).

ax3l:
@C0nsultant Thanks for the link. Do you think a list and "present == enabled" is better? In reality, one would need to check for various upcoming releases (tm) "present and if yes, value" anyway, so why not.

On the other hand, I admit I also kind of like the explicitly of the bool. What do you think?

anokfireball:
Our feature set would not be significant in size, so explicitly storing all of them should be fine.
Having a list would certainly be better for digestability, but then the user would not be certain what to look for (e.g. ```camelCase```, ```under_scores``` or ```noneofthetwo```).

Boolean values for feature presence are a good way to go. They also should be all we need to expose to the user. Interest in backend internals is likely rare enough to not warrant frontend modifications.

Personally, I do not see much difference between a map and the current approach. The downside of the map is the same problem I outlined above (no IDE would offer autocompletion). [```namedtuple```](https://docs.python.org/3.6/library/collections.html#collections.namedtuple) would be a more usable solution as it provides named members (for autocompletion) and is immutable (contrary to the dict).

ax3l:
I think the main use case is for integrators in apps and other libs that want to know what they can offer in file support and serial/parallel write.

With a list they would do
```python
bp_support = 'adios1' in openPMD.variants
```

with a map they would do
```python
json_support = openPMD.variants.get('json', False)
```

and with a member one would first have to check if the member function exists at all which is a bit verbose.

src/binding/python/openPMD.cpp
```diff
@@ -55,11 +57,12 @@ PYBIND11_MODULE(openPMD, m) {
...
     m.attr("__version__") = openPMDapi.str();
 
-    // variants
-    m.attr("have_mpi") = bool(openPMD_HAVE_MPI);
-    m.attr("have_hdf5") = bool(openPMD_HAVE_HDF5);
-    m.attr("have_adios1") = bool(openPMD_HAVE_ADIOS1);
-    m.attr("have_adios2") = bool(openPMD_HAVE_ADIOS2);
-    // m.attr("have_json") = bool(openPMD_HAVE_JSON);
+    // feature variants
+    m.attr("feature") = std::map<std::string, bool>{
```
ax3l: @C0nsultant last question: plural *features*? :)
variant*s*?

anokfireball: To be consistent with the other containers in the API, plural is be preferred.

ax3l: and variant is used in the build step, so `variants` it is.

anokfireball (approved):
LGTM.
I am too unfamiliar with pybind, so this is just a quick idea: Maybe make the values ```bool const``` so they can not be overwritten. *If* ```varaiants```handles like a typical C++ map (i.e. default value creation for non-existent keys/non-present features), this also eradicates the need for 
```python
json_support = openPMD.variants.get('json', False)
```
and could be reduced to
```python
json_support = openPMD.variants['json']
```

ax3l:
No problem, here is the explanation:

The map is translated to a python dictionary, so one can access it indeed as:
```py
json_support = openPMD.variants['json']
```

but if it does not exist, one can wish for a default for non-existing keys (python dictionary interface):
```py
json_support = openPMD.variants.get('json', False)
```

Const-question: huh, the attribute is indeed not const right now...

ax3l:
Turns out you can even overwrite the `.__version__` in `numpy.__version__` so let's do this o.0

PR #182, opened 2018-05-18
MSVC: decltype function Work-Around

ax3l:
Adds a work-around for MSVC looking up `decltype(function(var))`. Due to missing two-phase lookup, MSVC always determines a type (instead of no-type) which prevents our SFINAE.

So of course, we implement a SFINAE for trait `HasToString` to allow using this still in a SFINAE.

Introduced in #172


ax3l:
Dang, this only worked on VS 2017 as a work-around but not on VS 2015.
Will use just the generic exception message now on MSVC.

PR #183, opened 2018-05-19
Key Error on Read-Only

ax3l:
Refactor the key error message for read-only containers.

Implement the non-MSVC SFINAE from #172 properly again. `std::to_string` is only defined for ints and floats. Also factor out duplicate code.

### DIFF
--- a/include/openPMD/auxiliary/OutOfRangeMsg.hpp
+++ b/include/openPMD/auxiliary/OutOfRangeMsg.hpp
@@ -1,0 +1,51 @@
+#pragma once
+
+#include <string>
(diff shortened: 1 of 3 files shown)


ax3l:
@C0nsultant we need to check how the `no_such_attribute_error` looks in python. We could also just use the `std::out_of_range` error here.

If we want to define our own exception, we have to expose it to Python (`std::` exceptions are exposed already) so it does not crash the interpreter (but can be `try:` - `except:` catched).

Update: works as it is already, just passed as `RuntimeError` by pybind11.

PR #184, opened 2018-05-19
SerialTest: MSVC long double read

### LINKED ISSUE #179: MSVC: get<long double> bad_variant_access
Currently the following test fails with MSVC 19 (VS 14, 2015), probably due to it's `double` == `long double` implementation:

```
3: C:\projects\openpmd-api\test\SerialIOTest.cpp(845): FAILED:
3:   REQUIRE( s.getAttribute("longdouble").get< long double >() == 1.e80L )
3: due to unexpected exception with message:
3:   bad_variant_access
3: 
3: Datatype::UNDEFINED caught during dataset creation (serial HDF5)
3: ===============================================================================
3: test cases:  13 |  12 passed | 1 failed
3: assertions: 213 | 212 passed | 1 failed
3: 
 3/10 Test  #3: Serial.SerialIO ..................***Failed    0.41 sec
test 4
      Start  4: Serial.1_structure
```

Example in https://ci.appveyor.com/project/ax3l/openpmd-api/build/1.0.140/job/jo3nh9xgb03poikp

I think we should be able to emulate that behavior with gcc with the `-mlong-double-64` switch (but from a first test this does not trigger it).

ax3l:
So far this does only enable the read of `long double` attributes on MSVC again. ~~Demonstrates~~ Fix #179.

~~Let's see if we can figure out the reason behind it.~~ Use `getCast` to read back a `long double` on Windows which is identical to a `double`. Since both are unique types, we have to check if they are identical which is exactly what `getCast` does, see #337.

In the future, we should integrate `getCast` into `auxiliary::Variant::get<>()`.

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -953,7 +953,7 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
     REQUIRE(getCast< uint64_t >(s.getAttribute("uint64")) == 64u);
     REQUIRE(s.getAttribute("float").get< float >() == 16.e10f);
     REQUIRE(s.getAttribute("double").get< double >() == 1.e64);
-    REQUIRE(s.getAttribute("longdouble").get< long double >() == 1.e80L);
+    REQUIRE(getCast< long double >(s.getAttribute("longdouble")) == 1.e80L);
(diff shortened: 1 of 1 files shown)


anokfireball (approved):
LGTM

PR #190, opened 2018-05-19
Travis-CI: ADIOS1 Tests

ax3l:
Try to test the ADIOS1 backend on Travis-CI.
Builds in the matrix one entry with the oldes supported ADIOS1 release and in others the latest stable release.

Don't be confused, ADIOS weirdly has a python build dependency. And up to 1.13.1 this needs to be python 2.7... https://github.com/ornladios/ADIOS/pull/178

Update: seems to trigger https://github.com/spack/spack/issues/8082

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ jobs:
       stage: 'C++ Unit Tests'
       language: cpp
       env:
-        - USE_MPI=OFF USE_PYTHON=OFF USE_HDF5=ON USE_ADIOS1=ON USE_ADIOS2=OFF USE_SAMPLES=ON
+        - USE_MPI=OFF USE_PYTHON=OFF USE_HDF5=ON USE_ADIOS1=OFF USE_ADIOS2=OFF USE_SAMPLES=ON
       compiler: clang
       addons:
         apt:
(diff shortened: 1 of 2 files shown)


ax3l:
@C0nsultant [OSX test shows](https://travis-ci.org/openPMD/openPMD-api/jobs/392149408): `ERROR: Attribute element /iterationFormat has invalid value attribute - SerialIOTests(6857,0x7fffafa293c0) malloc: *** error for object 0x100000005: pointer being freed was not allocated` in SerialIOTest. Could this be a bug in our ADIOS attribute implementation?

anokfireball:
Without having the ability to actually debug on OSX:
The statement that causes this is likely 
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/src/IO/ADIOS/ADIOS1IOHandler.cpp#L1338
Where ```void* data``` is filled by ADIOS:
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/src/IO/ADIOS/ADIOS1IOHandler.cpp#L1061-L1065
ADIOS [doc](https://users.nccs.gov/~pnorbert/ADIOS-UsersManual-1.13.1.pdf) says 
> 9.4.20
...
The memory for the data is allocated within the library.
You can use free() to free the memory after use.

ax3l:
OSX problem was probably just an artifact, but I will keep an eye on it.

ax3l:
Uh, that might work as well soon. The only thing that should fail now is the `SerialIOTest` #252 with `openPMD_USE_MPI=ON` && `openPMD_USE_ADIOS1=ON` which is addressed in #254 

But since you silently NOPed the serial ADIOS backend when MPI is on, this PR might actually pass already.

ax3l:
Interesting, the OSX double-free returned :scream: 
Maybe if we valgrind the SerialIOTests with serial ADIOS on Linux we find it already.

ax3l:
Ok, valgrinding the serial backends in `SerialIOTests` needs some love :)
I will not check ADIOS on OSX for now and merge all other tests. Will open a PR enabling OSX ADIOS as follow-up.

PR #191, opened 2018-05-21
Container: ipython completions

ax3l:
Add ipython completions for the `__getitem__` method in the `Container` class.

Mainly helps with all string-keys since ints seem not really to be completed, even in regular `dict`s.

### Example

```python
In [1]: import openPMD

In [2]: s = openPMD.Series("openPMD-example-datasets/example-2d/hdf5/data00000100.h5", openPMD.Access_Type.read_only)
Series constructor called with explicit iteration suggests loading a single file with groupBased iteration encoding. Loaded file is fileBased.

In [3]: it = s.iterations[100]

In [4]: list(it.particles)
Out[4]: ['Hydrogen1+', 'electrons']

In [5]: it.particles['Hy<TAB>
In [5]: it.particles['Hydrogen+
```

### Reference

https://ipython.readthedocs.io/en/5.x/config/integrating.html

### DIFF
--- a/src/binding/python/Container.cpp
+++ b/src/binding/python/Container.cpp
@@ -178,6 +178,16 @@ namespace detail
             &Map::size
         );
 
+        cl.def(
+            "_ipython_key_completions_",
+            []( Map & m ) {
+                auto l = py::list();
+                for( const auto &myPair : m )
+                    l.append( myPair.first );
+                return l;
+            }
+        );
+
         return cl;
     }
 } // namespace detail


ax3l:
@C0nsultant pls feel free to review & merge :)

anokfireball:
Oh the day when this is possible with Cling in the C++ frontend...

> Series constructor called with explicit iteration suggests loading a single file with groupBased iteration encoding. Loaded file is fileBased.

Do you think we could/should [wrap](http://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html) the error/warning messages into something more [pythonic](https://docs.python.org/3/library/logging.html)?

anokfireball (approved):

ax3l:
Good point with the prints! I think this will also be relevant for the notebook because otherwise we will write to the running kernel's log instead of the frontend. Can you open this as a new issue?

PR #194, opened 2018-05-22
Fix all Clang and GCC warnings

### LINKED ISSUE #98: Fix Clang Warnings
We should fix all the clang (5.0) warnings that are currently raised in the code.

We should try to aim to be warning free with the following compiler flags:
```bash
export CXXFLAGS="-Wall -Wextra -Woverloaded-virtual -Wshadow"
```

anokfireball:
-Wall -Wextra -Wpedantic -Wshadow

Travis: -Werror

Closes #98.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,10 +75,9 @@ include(CTest)
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address,memory,undefined")
     #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Weverything")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Wshadow")
     string(CONCAT CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} "
-                  "-Wno-weak-vtables -Wno-padded -Wno-switch-enum "
-                  "-Wno-undefined-func-template -Wno-c++98-compat-pedantic")
+            "-Wno-unknown-pragmas")
     # silence HDF5 (FIXME: -isystem should be enough)
     string(CONCAT CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} "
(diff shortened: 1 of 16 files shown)


CMakeLists.txt
```diff
@@ -75,24 +75,27 @@ include(CTest)
...
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wpedantic -Wshadow")
+    string(CONCAT CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} "
+            "-Wno-unknown-pragmas")
     # silence BOOST filesystem
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-cpp")
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     # Warning C4503: "decorated name length exceeded, name was truncated"
     # Symbols longer than 4096 chars are truncated
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd4503")
+else ()
```
ax3l: Usually we would not place warnings in cmakelists anyways since they are not required for a build. We just have to for now since we have not written cmake Toolchain files for Travis yet.
Anyway, don't warn please ;)

ax3l (commented):
Just a note on cmake

CMakeLists.txt
anokfireball: So would you prefer a plain old ```message()``` or no ```else``` at all?

ax3l: I would be ok with a status type message but no else is probably what I prefer ;)

anokfireball:
Turns out ```-Wshadow``` makes pybind11 so incredibly verbose that the resulting log exceeds travis's max size. With that, the build is assumed as error'ed.
We might have to do a little hacking (```set_target_properties(openPMD.py PROPERTIES COMPILE_FLAGS -Wno-shadow```).

ax3l:
I tried to hack it into the property of the target but this does not silence it. We could now regex replace it ... or just disable the `-Wshadow` check as soon as we build the python tests. We also have pure C++ builds in the travis matrix, so that is ok.

ax3l (approved):

anokfireball (commented):
After a first wave of repulsion by the verbose variable names, I see their point. Good thinking!

anokfireball:
> just disable the ```-Wshadow``` check as soon as we build the python tests

The python binding *build* is not of huge interest. ROI for anything more complex is probably too low.
If the C++ base builds without warnings, we should assume the binding to be fine as well.

ax3l:
At least for the shadow warning this will be sufficient, I hope.

PR #197, opened 2018-05-23
Windows: Test Python Import

ax3l:
Test if the dlls are all in good shape and place.

I am now just installing into the currently active miniconda location which already imports paths to binary directories (where `.dll`s live and are searched as well).


CMakeLists.txt
```diff
@@ -17,13 +17,11 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 # install directories
-set(CMAKE_INSTALL_BINDIR bin)
-set(CMAKE_INSTALL_INCLUDEDIR include)
+include(GNUInstallDirs)
```
ax3l, line 20: now these vars properly sit in cache and can be modified from the outside

PR #200, opened 2018-05-24
Docs: All Static Builds

ax3l:
Restructure the build options and give hints on all-static builds.

### DIFF
--- a/docs/source/dev/buildoptions.rst
+++ b/docs/source/dev/buildoptions.rst
@@ -5,6 +5,9 @@ Build Options
 
 .. sectionauthor:: Axel Huebl
 
+Variants
+--------
+
 The following options can be added to the ``cmake`` call to control features.
 CMake controls options with prefixed ``-D``, e.g. ``-DopenPMD_USE_MPI=OFF``:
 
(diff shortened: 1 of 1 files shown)


docs/source/dev/buildoptions.rst
```diff
@@ -21,7 +24,41 @@ CMake Option                   Values          Description
...
 :sup:`1` *not yet implemented*
-:sup:`2` e.g. C++ keywords, currently disabled only for MSVC*
+
+:sup:`2` e.g. C++ keywords, currently disabled only for MSVC
+
+
+Shared or Static
+----------------
+
+By default, we will build as a static library (``libopenPMD.a`` or ``openPMD.lib``) and installs also its headers.
```
anokfireball: Typo: installs -> install.

docs/source/dev/buildoptions.rst
```diff
@@ -21,7 +24,41 @@ CMake Option                   Values          Description
...
-:sup:`2` e.g. C++ keywords, currently disabled only for MSVC*
+
+:sup:`2` e.g. C++ keywords, currently disabled only for MSVC
+
+
+Shared or Static
+----------------
+
+By default, we will build as a static library (``libopenPMD.a`` or ``openPMD.lib``) and install also its headers.
+You can only build a static or a shared library (``libopenPMD.so`` or ``openPMD.dll``) at a time.
```
anokfireball: I had to read this a few times to understand what you mean. There are really two sentences here - one about either static or shared - and one about the names of the shared library.

ax3l: I can put the brackets of the first sentence in the second one.

anokfireball: Would make it easier to understand IMO.

Issue #140, opened 2018-04-21
storeChunk: shared_ptr only?


ax3l:
Quick question regarding the write `storeChunk` API.
Why does it only accept `shared_ptr`s as parameters?

I was playing around with use cases where user data is allocated in containers such as `std::array` and `std::vector` in which the data layout is guaranteed to be [contiguous and access is well-defined](http://en.cppreference.com/w/cpp/container/vector) ([ContiguousContainer](http://en.cppreference.com/w/cpp/concept/ContiguousContainer)s), e.g. via `vector<t>.data()`. How do we get such containers' data written? :-)

```C++
// user code
std::vector<int> v = {1, 1, 3, 4, 7, 11};

// write me, I am famous
```

Do we need a [`ContiguousIterator`](http://en.cppreference.com/w/cpp/concept/ContiguousIterator) interface as well? (Dang, that concept exists in `std::vector` since C++03 but the `ContiguousIterator` is C++17... Anyway, there are just three std lib types where this is relevant for now so we could overload or define traits. Fun fact: `vector<bool>` is special and not guaranteed to be contiguous.)

Anyway, the data could come from any user-data as well which might be badly designed C pointer. We can't enforce this backwards on the user to change their containers/data storage/etc. just to write. How do we handle such data? If we just add a `unique_ptr<T>` interface this would mean we would have to move this ptr out again when we are done writing it. Verbose and cumbersome...

Saying that, another question regarding the existing `shared_ptr` interface: when should the shared ptr reference be released by the API again? Imho, first then it is save to use the ptr at user-side again, so after `.close()` and `.flush()`.

anokfireball:
The main reason for employing ```shared_ptr```s was to make the shared ownership during unfinished IO explicit. Raw C-style pointers were also considered, but discarded as they give no insight into what happens to them over time.

ContiguousContainers store data in exactly the way ```storeChunk``` expects - as one unstrided contiguous slab of values, one after the other. So to get the values of a vector
```c++
std::vector< int > x = /* ... */;
```
you would just put the ```std::vector::data()```-pointer into a ```shared_ptr```. To not have double frees from both the vector and the```shared_ptr```, the deleter needs to be empty. You end up with
```c++
std::shared_ptr< int > x_chunk{
    x.data(),
    [](int*){  }
};
```
which you can use to write your data as was intended
```c++
Offset x_offset = /* ... */;
Extent x_extent = { x.size() };
// x_chunk.use_count() == 1
[...]

PR #204, opened 2018-05-24
Helper: Raw Pointer store/load

ax3l:
Quick & dirty draft to share raw pointers for read & write.

Related to #140

### To Do

- [x] add docstrings
- [x] add tests
- [x] loadChunk: also a `shared_ptr` data interface, separate out allocation API
- [x] move helpers into own header
- [x] ~unit tests for reference counts of `shared_ptr`s~ deferred to #211

### DIFF
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -60,8 +60,7 @@ int main()
 
     Offset chunk_offset = {1, 1, 1};
     Extent chunk_extent = {2, 2, 1};
-    std::shared_ptr< double > chunk_data;
-    E_x.loadChunk(chunk_offset, chunk_extent, chunk_data);
+    auto chunk_data = E_x.loadChunk<double>(chunk_offset, chunk_extent);
     cout << "Queued the loading of a single chunk from disk, "
             "ready to execute\n";
     series.flush();
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -60,7 +60,7 @@ int main()
 
     Offset chunk_offset = {1, 1, 1};
     Extent chunk_extent = {2, 2, 1};
-    std::unique_ptr< double[] > chunk_data;
+    std::shared_ptr< double > chunk_data;
     E_x.loadChunk(chunk_offset, chunk_extent, chunk_data);
     cout << "Queued the loading of a single chunk from disk, "
             "ready to execute\n";
@@ -72,7 +72,7 @@ int main()
         for( size_t col = 0; col < chunk_extent[1]; ++col )
             cout << "\t"
                  << '(' << row + chunk_offset[0] << '|' << col + chunk_offset[1] << '|' << 1 << ")\t"
-                 << chunk_data[row*chunk_extent[1]+col];
+                 << chunk_data.get()[row*chunk_extent[1]+col];
         cout << '\n';
     }
 
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -103,10 +103,10 @@ public:
     template< typename T >
     RecordComponent& makeConstant(T);
 
-    template< typename T >
+    template< typename T, typename D >
     void loadChunk(Offset const&,
                    Extent const&,
-                   std::unique_ptr< T[] >&,
+                   std::unique_ptr< T[], D >&,
                    Allocation = Allocation::AUTO,
                    double targetUnitSI = std::numeric_limits< double >::quiet_NaN() );
     template< typename T >
@@ -140,9 +140,9 @@ RecordComponent::makeConstant(T value)
     return *this;
 }
 
-template< typename T >
+template< typename T, typename D >
 inline void
-RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[] >& data, Allocation alloc, double targetUnitSI)
+RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[], D >& data, Allocation alloc, double targetUnitSI)
 {
     if( !std::isnan(targetUnitSI) )
         throw std::runtime_error("unitSI scaling during chunk loading not yet implemented");
@@ -170,7 +170,7 @@ RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[
         numPoints *= dimensionSize;
 
     if( (Allocation::AUTO == alloc && !data) || Allocation::API == alloc )
-        data = std::unique_ptr< T[] >(new T[numPoints]);
+        data = std::unique_ptr< T[], D >(new T[numPoints]);
     T* raw_ptr = data.get();
 
     if( *m_isConstant )
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -21,6 +21,7 @@
 #pragma once
 
 #include "openPMD/backend/BaseRecordComponent.hpp"
+#include "openPMD/auxiliary/ShareRaw.hpp"
 #include "openPMD/Dataset.hpp"
 
 #include <cmath>
@@ -29,47 +30,10 @@
 #include <queue>
 #include <string>
 #include <stdexcept>
-#include <vector>
-#include <array>
 
 
 namespace openPMD
 {
-//! @{
-/** Share ownership with a raw pointer
- *
- * Helper function to share write ownership
- * unprotected and without reference counting with a
- * raw pointer or stdlib container.
- *
- * @warning this is a helper function to bypass the shared-pointer
- *          API for storing data behind raw pointers. Using it puts
- *          the resposibility of buffer-consistency between stores
- *          and flushes to the users side without an indication via
- *          reference counting.
- */
-template< typename T >
-std::shared_ptr< T >
-storeRaw( T* x )
-{
-    return std::shared_ptr< T >( x, [](T*){} );
-}
-
-template< typename T >
-std::shared_ptr< T >
-storeRaw( std::vector< T > & v )
-{
-    return std::shared_ptr< T >( v.data(), [](T*){} );
-}
-
-template< typename T, std::size_t T_size >
-std::shared_ptr< T >
-storeRaw( std::array< T, T_size > & a )
-{
-    return std::shared_ptr< T >( a.data(), [](T*){} );
-}
-//! @}
-
 class RecordComponent : public BaseRecordComponent
 {
     template<
@@ -103,10 +67,10 @@ public:
     template< typename T >
     RecordComponent& makeConstant(T);
 
-    template< typename T, typename D >
+    template< typename T >
     void loadChunk(Offset const&,
                    Extent const&,
-                   std::unique_ptr< T[], D >&,
+                   std::shared_ptr< T >,
                    Allocation = Allocation::AUTO,
                    double targetUnitSI = std::numeric_limits< double >::quiet_NaN() );
     template< typename T >
@@ -140,13 +104,13 @@ RecordComponent::makeConstant(T value)
     return *this;
 }
 
-template< typename T, typename D >
+template< typename T >
 inline void
-RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[], D >& data, Allocation alloc, double targetUnitSI)
+RecordComponent::loadChunk(Offset const& o, Extent const& e, std::shared_ptr< T > data, Allocation alloc, double targetUnitSI)
 {
     if( !std::isnan(targetUnitSI) )
         throw std::runtime_error("unitSI scaling during chunk loading not yet implemented");
-    Datatype dtype = determineDatatype(std::shared_ptr< T >());
+    Datatype dtype = determineDatatype(data);
     if( dtype != getDatatype() )
         throw std::runtime_error("Type conversion during chunk loading not yet implemented");
 
@@ -170,7 +134,10 @@ RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[
         numPoints *= dimensionSize;
 
     if( (Allocation::AUTO == alloc && !data) || Allocation::API == alloc )
-        data = std::unique_ptr< T[], D >(new T[numPoints]);
+    {
+        auto newData = std::shared_ptr<T>( new T[numPoints], []( T *p ){ delete [] p; } );
+        data.swap(newData);
+    }
     T* raw_ptr = data.get();
 
     if( *m_isConstant )
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -21,7 +21,6 @@
 #pragma once
 
 #include "openPMD/backend/BaseRecordComponent.hpp"
-#include "openPMD/auxiliary/ShareRaw.hpp"
 #include "openPMD/Dataset.hpp"
 
 #include <cmath>
@@ -68,10 +67,14 @@ public:
     RecordComponent& makeConstant(T);
 
     template< typename T >
+    std::shared_ptr< T > loadChunk(Offset const&,
+                   Extent const&,
+                   double targetUnitSI = std::numeric_limits< double >::quiet_NaN() );
+
+    template< typename T >
     void loadChunk(Offset const&,
                    Extent const&,
                    std::shared_ptr< T >,
-                   Allocation = Allocation::AUTO,
                    double targetUnitSI = std::numeric_limits< double >::quiet_NaN() );
     template< typename T >
     void storeChunk(Offset, Extent, std::shared_ptr< T >);
@@ -105,8 +108,21 @@ RecordComponent::makeConstant(T value)
 }
 
 template< typename T >
+inline std::shared_ptr< T >
+RecordComponent::loadChunk(Offset const& o, Extent const& e, double targetUnitSI)
+{
+    uint64_t numPoints = 1u;
+    for( auto const& dimensionSize : e )
+        numPoints *= dimensionSize;
+
+    auto newData = std::shared_ptr<T>(new T[numPoints], []( T *p ){ delete [] p; });
+    loadChunk(o, e, newData, targetUnitSI);
+    return newData;
+}
+
+template< typename T >
 inline void
-RecordComponent::loadChunk(Offset const& o, Extent const& e, std::shared_ptr< T > data, Allocation alloc, double targetUnitSI)
+RecordComponent::loadChunk(Offset const& o, Extent const& e, std::shared_ptr< T > data, double targetUnitSI)
 {
     if( !std::isnan(targetUnitSI) )
         throw std::runtime_error("unitSI scaling during chunk loading not yet implemented");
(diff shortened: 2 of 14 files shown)


include/openPMD/RecordComponent.hpp
```diff
@@ -29,10 +29,41 @@
...
+#include <vector>
+#include <array>
 
 
 namespace openPMD
 {
+//! @{
+/** Share ownership with a raw pointer
+ *
+ * Helper function to share read/write ownership
```
ax3l: write

include/openPMD/RecordComponent.hpp
```diff
@@ -29,10 +29,41 @@
...
+//! @{
+/** Share ownership with a raw pointer
+ *
+ * Helper function to share write ownership
+ * unprotected and without reference counting with a
+ * raw pointer or stdlib container.
+ */
+template< typename T >
+std::shared_ptr< T >
+shareRaw( T* x )
```
ax3l: possible naming: `storeRaw`

since we will need another wrapper for loads.

on the other hand, we can also use `shareRaw`/`storeRaw` in `determineDatatype` arguments.

test/CoreTest.cpp
```diff
@@ -452,6 +452,22 @@ TEST_CASE( "structure_test", "[core]" )
...
 
+struct nop
+{
+    template <typename T>
+    void operator() (T const &) const noexcept { }
+};
+
+template< typename T, std::size_t T_size >
+std::unique_ptr< T[], nop >
+wrapRaw( std::array< T, T_size > & a )
```
ax3l: `loadRaw`?

test/CoreTest.cpp
```diff
@@ -503,11 +519,13 @@ TEST_CASE( "wrapper_test", "[core]" )
     mrc2.loadChunk({0}, {1}, loadData);
     o.flush();
     REQUIRE(loadData[0] == value);
-    value = 43;
+    value = 43.;
     mrc2.makeConstant(value);
-    o.iterations[4].meshes["E"]["y"].loadChunk({0}, {1}, loadData);
+    std::array< double, 1 > moreData = { 112233. };
+    auto pppp = wrapRaw(moreData);
```
ax3l: we need an `lvalue`

include/openPMD/RecordComponent.hpp
```diff
@@ -29,10 +29,41 @@
...
 
 namespace openPMD
 {
+//! @{
+/** Share ownership with a raw pointer
+ *
+ * Helper function to share write ownership
+ * unprotected and without reference counting with a
+ * raw pointer or stdlib container.
+ */
```
ax3l: add `@note` or `@warning` about user-responsibility of memory pre-allocation, buffer-consistency, etc.

include/openPMD/RecordComponent.hpp
anokfireball: Adapting from future C++ lingo, we might want to call this a [```span```](http://en.cppreference.com/w/cpp/container/span) (I'd still prefer the name [```array_view```](http://open-std.org/Jtc1/sc22/wg21/docs/papers/2017/p0122r4.pdf), but well :man_shrugging:).
One could think about ```shared_span```/```unique_span``` and ```readonly_span```/```readwrite_span```

anokfireball: Sidenote - ```span``` is available in [GSL](https://github.com/Microsoft/GSL/blob/master/include/gsl/span) should we want to use it.
It does pretty much the same thing as our use of ```std::unique_ptr``` and ```shared_ptr```, conveys the message that it is about a view into a range of values but does not indicate anything about ownership and lifetime.

ax3l: Span name: Fits quite good as well, similar to the `ContiguousIterator` concept discussed earlier.

GSL: not sure, maybe as an overload. But could be too heavy for now (C++11), also dependency constrain wise. (uhh, [GSL-lite](https://github.com/martinmoene/gsl-lite) ;-) )

Also, I think if I want to express "safe" numpy arrays in the Python API I will need the lifetime control as well there (changing numpy.array.flags).

test/CoreTest.cpp
```diff
@@ -503,11 +519,13 @@ TEST_CASE( "wrapper_test", "[core]" )
     mrc2.loadChunk({0}, {1}, loadData);
     o.flush();
     REQUIRE(loadData[0] == value);
-    value = 43;
+    value = 43.;
     mrc2.makeConstant(value);
-    o.iterations[4].meshes["E"]["y"].loadChunk({0}, {1}, loadData);
+    std::array< double, 1 > moreData = { 112233. };
+    auto pppp = loadRaw(moreData);
```
ax3l: note: we need an `lvalue` since `loadChunk` needs a `ref`

we could also provide a pure `*` interface that does this wrapping.

test/CoreTest.cpp
```diff
@@ -503,11 +519,13 @@ TEST_CASE( "wrapper_test", "[core]" )
     mrc2.loadChunk({0}, {1}, loadData);
     o.flush();
     REQUIRE(loadData[0] == value);
-    value = 43;
+    value = 43.;
     mrc2.makeConstant(value);
-    o.iterations[4].meshes["E"]["y"].loadChunk({0}, {1}, loadData);
+    std::array< double, 1 > moreData = {{ 112233. }};
+    auto pppp = loadRaw(moreData);
```
ax3l: note: we need an `lvalue` since `loadChunk` needs a `ref`

we could also provide a pure `*` interface that does this wrapping.

anokfireball: As we introduce more user-friendly ways of passing raw data here, you absolutely have my blessing if you want to change the interface of ```loadChunk``` and ```storeChunk```.

ax3l: I will probably just add two overloads for raw pointers then and will do the wrapping therein.
Same for `determineDatatype<T*>`.

anokfireball: Just so the intention behind the ```std::shared_ptr``` does not get lost in the ether:
It is supposed to signal that the data behind it must not be changed until the reference count is back to the initial value (i.e. the backend does not hold a reference any more):

```cpp
std::shared_ptr< double > data = /* ... */
Offset offset = /* ... */
Extent extent = /* ... */
long initial = data.use_count();

auto record_component = /* ... */
record_component.storeChunk(offset, extent, global_data);
std::future< void > future = s.flush();

while( data.use_count() > initial )
{
    /* wörk that DOES NOT modify *data */
}
/* no more such wörk */
future.wait();
/* wörk that DOES modify *data */
```

ax3l: Can you elaborate why it's a `shared` ptr for `stores` but a `unique` ptr for loads? (I understand the in-out meaning of `unique_ptr<T> &` and wonder why we use it only in load)

anokfireball: Legacy code and laziness to refactor (or even *delete*) it, in all honesty:
https://github.com/openPMD/openPMD-api/blob/62b854a0714e4f6516bc5515f0a4382e673a6b56/include/openPMD/RecordComponent.hpp#L162-L164
The previous interface induced a ```flush()``` and *returned* the pointer, eliminating the need for shared ownership. As the ```unique_ptr``` is more lightweight, it was chosen for that purpose. *This might not have been premature optimization, but now is overly specific.*

ax3l: So does it make sense to put both `unique_ptr<T>&` or both `shared_ptr<T>`?

anokfireball: [Semantic obviousness](https://herbsutter.com/2013/06/05/gotw-91-solution-smart-pointer-parameters/) for the user would dictate a ```shared_ptr<T>```. Note that this is also used internally when writing datasets (to take advantage of reference-counting):
https://github.com/openPMD/openPMD-api/blob/62b854a0714e4f6516bc5515f0a4382e673a6b56/include/openPMD/IO/IOTask.hpp#L162-L169
The downside with ```unique_ptr<T>&``` is that the original ```unique_ptr<T>``` might go out of scope for the user, destroy and deallocate what it points to. The reference would be dangling, and we cannot guarantee the IOTask has been completed by that point:
```cpp
auto record_component = /* ... */
{
    std::unique_ptr< double > data;
    Offset offset = /* ... */
    Extent extent = /* ... */
    record_component.loadChunk(offset, extent, data);
}
s.flush();
```
With a ```shared_ptr<T>```, the operation would be wasteful as the results are discarded. But at least the program will not explode in your face, thanks to reference-counting the runtime won't deallocate the pointed-to-data.

**NOTE - This is what the behaviour should be, but currently is not:**
For some reason (maybe the above?) I implemented ```loadChunk``` to always [```.flush()```](https://github.com/openPMD/openPMD-api/blob/069a9e2280c3e501007b5b164c27f5dcc375f29f/include/openPMD/RecordComponent.hpp#L195), but ```storeChunk``` to defer flushes with the help of a [queue](https://github.com/openPMD/openPMD-api/blob/069a9e2280c3e501007b5b164c27f5dcc375f29f/include/openPMD/RecordComponent.hpp#L277). Ideally, we would put both into the queue.

**MINOR NOTE** With similar logic, there should not be any need to [```.flush()```](https://github.com/openPMD/openPMD-api/blob/d16f1762fcb1be7e7441755a3209abdba7994173/include/openPMD/RecordComponent.hpp#L158) for every IOTask generated by ```storeChunk```.

ax3l: Nice, I agree with your points and the ownership arguments in the reference.
I think all in all it makes sense then to make `loadChunk` use a `shared_ptr` too.

ax3l: Should we then overload both methods with a "lazy"/"unaware" raw-pointer overload or just expose the `shareRaw` wrapper to the user?

anokfireball: What are use cases you're thinking of for raw pointers?
Even with such applications, I would rather keep the contract (i.e. shared ownership) in the interface (i.e. offer ```loadChunk```/```storeChunk``` with ```shared_ptr``` only) and offer a helper for raw pointers:
```cpp
template< typename T >
std::shared_ptr< T >
storeRaw( T* ptr )
{
    return std::shared_ptr< T >( ptr, [](T*){} );
}
```

anokfireball: This helper mechanism does not offer the safety-guarantees I referred to in my [comment](https://github.com/openPMD/openPMD-api/pull/204#discussion_r190864021), btw:
```cpp
auto record_component = /* ... */
{
    std::vector< double > data = /* ... */
    Offset offset = /* ... */
    Extent extent{data.size()};
    record_component.storeChunk(offset, extent, storeRaw(data));
}
//pointer behind `data` does not exist any more,
//but flush might not have happened
s.flush();
```

ax3l: > What are use cases you're thinking of for raw pointers?

Mainly contiguous containers (STL and third-party libs), that don't offer a shared view or unique-ptr into their data.

> I would rather keep the contract (i.e. shared ownership) in the interface

Agreed, so we will expose the helpers/wrappers and do not overload.

ax3l: While refactoring the `loadChunk` interface to `shared_ptr` I realized the following: For proper support of the "we allocate for you" API I would need to pass as `shared_ptr< T[] >&` which I find a little odd. Also, the implementation allocated immediately which does not offer too much convenience.

What do you think about if I:
- remove the allocation policy argument
- use a signature (1) that returns a `shared_ptr<T>` (we allocate)
- use a signature (2) that accepts a pre-allocated `shared_ptr<T>` (user allocates)

I think that makes things clearer.

anokfireball: Overall I'll have to agree. Didn't think far enough to realize that you can't allocate the ```shared_ptr<T>``` and have it visible to the caller. The proposal with the two signatures also makes sense.
Do you have any catchy names for semantic differentiation (i.e. ```loadChunkAllocated```)? 

Quick idea that does not entail two functions:
```cpp
template< typename T, typename D >
std::shared_ptr< T, typename D >
loadChunk(Offset const&,
          Extent const&,
          std::shared_ptr< T, D > = std::shared_ptr< T, D >()); // assume allocated if not default
          //double targetUnitSI = std::numeric_limits< double >::quiet_NaN() );
```

> use a signature (1) that returns a shared_ptr<T> (we allocate)

Good thing I was lazy enough to leave that code in there.

ax3l: I don't think we will need different names, the calling parameters are different :-)

> Quick idea that does not entail two functions.

Tried that, does not really work great also because `shared_ptr< T[] >` machinery was kind of forgotten in C++11... I can push you the draft in 6dc5ab956ba3fcec7016a4e17ac1b45f5e0800fe that RT crashes. But maybe I screwed up the deleter.

Anyway, I have the feeling that member function has too many parameters and does too little (just allocate immediately) that can not be separated otherwise cleanly, which makes it easier to grasp by being less complex.

We can implement this by the existing signature and the second just allocates, calls it and returns the new memory.

ax3l: Btw, I think [this example](https://github.com/openPMD/openPMD-api/pull/204#discussion_r190829625) you posted would make a great page in the manual! :)

anokfireball: > ```shared_ptr< T[] >``` machinery was kind of forgotten in C++11

How this is only coming in C++17 is beyond me...

include/openPMD/auxiliary/ShareRaw.hpp
```diff
@@ -0,0 +1,65 @@
...
+    shareRaw( T* x )
+    {
+        return std::shared_ptr< T >( x, [](T*){} );
+    }
+
+    template< typename T >
+    std::shared_ptr< T >
+    shareRaw( std::vector< T > & v )
+    {
+        return std::shared_ptr< T >( v.data(), [](T*){} );
```
anokfireball: Maybe we should SFINAE and do this for anything that offers ```T* TYPE::data()```?

include/openPMD/RecordComponent.hpp
```diff
@@ -170,7 +134,10 @@ RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[
         numPoints *= dimensionSize;
 
     if( (Allocation::AUTO == alloc && !data) || Allocation::API == alloc )
-        data = std::unique_ptr< T[], D >(new T[numPoints]);
+    {
+        auto newData = std::shared_ptr<T>( new T[numPoints], []( T *p ){ delete [] p; } );
+        data.swap(newData);
```
anokfireball: Are you doing this for safety guarantee when the allocation would fail?

examples/2_read_serial.cpp
```diff
@@ -60,8 +60,7 @@ int main()
 
     Offset chunk_offset = {1, 1, 1};
     Extent chunk_extent = {2, 2, 1};
-    std::unique_ptr< double[] > chunk_data;
-    E_x.loadChunk(chunk_offset, chunk_extent, chunk_data);
+    auto chunk_data = E_x.loadChunk<double>(chunk_offset, chunk_extent);
```
ax3l, line 63: I really wish the syntax `shared_ptr< auto > = someFunc< double >();` would exist...

include/openPMD/auxiliary/ShareRaw.hpp
ax3l: That's a good idea, but probably offering a trait is better.

`*mumble*` Although the concept of `SFINAE`s and `enable_if` are slick, they are quite compiler-dependent to work well in the corner-cases (msvc #182, nvcc) and are [super-slow in terms of compile time](https://www.reddit.com/r/cpp/comments/6gur2x/the_rule_of_chiel_aka_compiletime_cost_of/).

Well, actually: just because something has a `.data()` member does not mean it is contiguous and save to use. let us not do that by default as long as we can't check it via traits.

include/openPMD/RecordComponent.hpp
ax3l: I was trying to exchange the incoming shared_ptr but things did not work as expected and I started to go for the two-signature option for allocations...

include/openPMD/auxiliary/ShareRaw.hpp
anokfireball: > just because something has a .data() member does not mean it is contiguous and save to use

My interpretation of [```std::data()```](http://en.cppreference.com/w/cpp/iterator/data) is just that:
> ... a pointer to the block of memory containing the elements ...

[xtensor](http://xtensor.readthedocs.io/en/latest/api/xcontainer.html#_CPPv3N2xt10xcontainer4dataEv) does it as well, btw.
But that is C++17, so well have to deal with it manually, anyway.

ax3l: Interesting, didn't know `std::data()` yet! I stand corrected.

ax3l:
@C0nsultant Besides possible overloads for additional `.data()` providing containers, what do you think of the pushed changes?

include/openPMD/RecordComponent.hpp
```diff
@@ -123,17 +141,13 @@ RecordComponent::loadChunk(Offset const& o, Extent const& e, std::unique_ptr< T[
...
                                      + " - Chunk: " + std::to_string(o[i] + e[i])
                                      + ")");
-    if( Allocation::API == alloc && data )
-        throw std::runtime_error("Preallocated pointer passed with signaled API-allocation during chunk loading.");
-    else if( Allocation::USER == alloc && !data )
-        throw std::runtime_error("Unallocated pointer passed with signaled user-allocation during chunk loading.");
+    if( !data )
+        throw std::runtime_error("Unallocated pointer passed during chunk loading.");
 
     uint64_t numPoints = 1u;
```
anokfireball: This could now be pulled into a lower scope inside the ```if( *m_isConstant )``` branch.

anokfireball (approved):
Love the new syntax!
```cpp
auto chunk_data = E_x.loadChunk<double>(chunk_offset, chunk_extent);
```
is what the initial draft was and *now it's back*!

include/openPMD/RecordComponent.hpp
C0nsultant: Lines https://github.com/ax3l/openPMD-api/blob/fd7b2b1b14e8b574c3b35d5ef00bb50f59441926/include/openPMD/RecordComponent.hpp#L176-L226 can now safely be removed as the functionality is provided now.

anokfireball:
Lines https://github.com/ax3l/openPMD-api/blob/fd7b2b1b14e8b574c3b35d5ef00bb50f59441926/include/openPMD/RecordComponent.hpp#L176-L226 can now safely be removed as the functionality is provided now.

ax3l:
found a little design flaw. `writeChunk` should only take `shared_ptr<T const>` args. will fix on the way :)
(discovered while trying to write a `double const`)

update: I was overly optimistic, can only fix the internal `Parameter` type.

examples/2_read_serial.cpp
ax3l: should I write `shared_ptr<double>` here so users know what is going on?

anokfireball: As this is only the second piece of code a user would see, this is a good point in time to show them the way we handle raw data.

ax3l:
@C0nsultant this should now contain all changes :)

include/openPMD/RecordComponent.hpp
```diff
@@ -124,24 +140,16 @@ RecordComponent::loadChunk(Offset const& o, Extent const& e, std::shared_ptr< T
...
-        data.swap(newData);
-    }
     T* raw_ptr = data.get();
 
     if( *m_isConstant )
     {
+        uint64_t numPoints = 1u;
+        for( auto const& dimensionSize : e )
+            numPoints *= dimensionSize;
         /*
```
anokfireball: This block of code can also be removed. The constant value is cached within the ```RecordComponent``` itself now.

anokfireball:
By making ```loadChunk``` defer actual loading, we also need to introduce proper reference counting for the ```shared_ptr```, just like with ```storeChunk```.

This *probably* (haven't thought about it for long enough) means we have to make [this](
https://github.com/ax3l/openPMD-api/blob/23e95403e6bb212e83e2929b969eabf10dd7cef8/include/openPMD/IO/IOTask.hpp#L177) raw pointer into a ```shared_ptr< void >``` as well. Otherwise the same out-of-scope/deallocation problem could hit us during loading.

**EDIT**
Yes, I am fairly certain this needs to happen:
```cpp
auto record_component = /* ... */
{
    std::shared_ptr< double > data = record_component.loadChunk(offset, extent);
    //data.ref_count() == 1
}
//the IOTask in the backend still holds the raw pointer to the now deallocated memory behind data
s.flush();
```

Do you want me to do that, @ax3l? I'll push it to this PR. Just making sure we're not messing with the same branch simulateneously.

ax3l:
good point, feel free to go ahead :)
(maybe also add this ref count as an `REQUIRE` in a test? I missed that we should check that indeed)

ax3l:
@C0nsultant what I am quite surprised is that we are just doing

```C++
auto data = record_component.loadChunk(offset, extent);
// use data
```

without the need to call `flush()` first. Is this intended? Shouldn't the data first be available *after* `flush()`?

Side note: what happened to the CI in this PR? Somehow it's doubled since yesterday evening.

anokfireball:
> Shouldn't the data first be available after ```flush()```?

You are entirely correct. *In theory...* This is not the current behaviour, but will be introduced with #205.

> what happened to the CI in this PR?

Not sure, this happened after I pushed to your branch (following the steps GitHub's documentation lists).

ax3l:
All right then, let's wrap it :-)
Thanks for the push!

PR #206, opened 2018-05-26
Docs: Fix CMake Static ADIOS

ax3l:
Fix option name in static ADIOS1 libs (#200).


docs/source/dev/buildoptions.rst
```diff
@@ -42,7 +42,7 @@ CMake Option                   Values          Description
 ``BUILD_SHARED_LIBS``          ON/**OFF**      Build the C++ API as shared library
 ``Boost_USE_STATIC_LIBS``      ON/**OFF**      Require static Boost library
 ``HDF5_USE_STATIC_LIBRARIES``  ON/**OFF**      Require static HDF5 library
```
anokfireball, line 44: Presumably we'd want to do the same for HDF5?

anokfireball (approved):

docs/source/dev/buildoptions.rst
ax3l: No, these vars are `FindPackagename.cmake` specific (legacy) and not fully unified in the CMake community :)
HDF5 really uses the `_LIBRARIES` suffix xD

PR #207, opened 2018-05-27
Replace boost.filesystem

### LINKED ISSUE #70: Remove Boost Dependency

anokfireball:
Create a minimally slim replacement for boost.filesystem for WIN32 and
POSIX compliant systems.

Closes #70.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,10 +69,6 @@ include(CTest)
 
 # Dependencies ################################################################
 #
-# external library: Boost (mandatory)
-find_package(Boost 1.62.0 REQUIRED
-    COMPONENTS system filesystem)
-
 # external library: MPI (optional)
 if(openPMD_USE_MPI STREQUAL AUTO)
     find_package(MPI)
@@ -256,15 +252,6 @@ if(BUILD_TESTING)
         find_package(Catch2 2.2.1 CONFIG REQUIRED)
         message(STATUS "Catch2: Found version ${Catch2_VERSION}")
     endif()
-endif()
-
-if(TARGET Boost::filesystem)
-    target_link_libraries(openPMD PUBLIC
-        Boost::boost Boost::system Boost::filesystem)
-else()
-    target_link_libraries(openPMD PUBLIC
-        ${Boost_LIBRARIES})
-    target_include_directories(openPMD SYSTEM PUBLIC ${Boost_INCLUDE_DIRS})
 endif()
 
 if(openPMD_HAVE_MPI)
@@ -457,8 +444,6 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     endif()
     string(CONCAT CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} "
             "-Wno-unknown-pragmas")
-    # silence BOOST filesystem
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-cpp")
 elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
     # Warning C4503: "decorated name length exceeded, name was truncated"
     # Symbols longer than 4096 chars are truncated
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -210,24 +210,6 @@ set(IO_SOURCE
         src/IO/HDF5/HDF5IOHandler.cpp
         src/IO/HDF5/ParallelHDF5IOHandler.cpp)
 
-if(WIN32)
-    set(TARGET_PLATFORM "WIN32")
-    set_property(
-            SOURCE src/auxiliary/Filesystem.cpp
-            APPEND
-            PROPERTY COMPILE_DEFINITIONS
-            WIN32=1
-    )
-else()
-    set(TARGET_PLATFORM "UNIX")
-    set_property(
-            SOURCE src/auxiliary/Filesystem.cpp
-            APPEND
-            PROPERTY COMPILE_DEFINITIONS
-            UNIX=1
-    )
-endif()
-
 # library
 add_library(openPMD ${CORE_SOURCE} ${IO_SOURCE})
 
@@ -406,8 +388,6 @@ endif()
 if(BUILD_TESTING)
     foreach(testname ${openPMD_TEST_NAMES})
         add_executable(${testname}Tests test/${testname}Test.cpp)
-
-        target_compile_definitions(${testname}Tests PUBLIC "-D${TARGET_PLATFORM}=1")
 
         if(openPMD_HAVE_INVASIVE_TESTS)
             target_compile_definitions(${testname}Tests PUBLIC "-DopenPMD_HAVE_INVASIVE_TESTS=1")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -194,6 +194,7 @@ set(CORE_SOURCE
         src/Record.cpp
         src/RecordComponent.cpp
         src/Series.cpp
+        src/auxiliary/Filesystem.cpp
         src/backend/Attributable.cpp
         src/backend/BaseRecordComponent.cpp
         src/backend/MeshRecordComponent.cpp
@@ -208,6 +209,24 @@ set(IO_SOURCE
         src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
         src/IO/HDF5/HDF5IOHandler.cpp
         src/IO/HDF5/ParallelHDF5IOHandler.cpp)
+
+if(WIN32)
+    set(TARGET_PLATFORM "WIN32")
+    set_property(
+            SOURCE src/auxiliary/Filesystem.cpp
+            APPEND
+            PROPERTY COMPILE_DEFINITIONS
+            WIN32=1
+    )
+else()
+    set(TARGET_PLATFORM "UNIX")
+    set_property(
+            SOURCE src/auxiliary/Filesystem.cpp
+            APPEND
+            PROPERTY COMPILE_DEFINITIONS
+            UNIX=1
+    )
+endif()
 
 # library
 add_library(openPMD ${CORE_SOURCE} ${IO_SOURCE})
@@ -387,9 +406,13 @@ endif()
 if(BUILD_TESTING)
     foreach(testname ${openPMD_TEST_NAMES})
         add_executable(${testname}Tests test/${testname}Test.cpp)
+
+        target_compile_definitions(${testname}Tests PUBLIC "-D${TARGET_PLATFORM}=1")
+
         if(openPMD_HAVE_INVASIVE_TESTS)
             target_compile_definitions(${testname}Tests PUBLIC "-DopenPMD_HAVE_INVASIVE_TESTS=1")
         endif()
+
         if(openPMD_HAVE_MPI)
             target_compile_definitions(${testname}Tests PUBLIC "-DopenPMD_HAVE_MPI=1")
         else()
--- a/include/openPMD/auxiliary/Filesystem.hpp
+++ b/include/openPMD/auxiliary/Filesystem.hpp
@@ -1,0 +1,45 @@
+/* Copyright 2018 Fabian Koller
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * openPMD-api is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include <string>
+#include <vector>
+
+namespace openPMD
+{
+namespace auxiliary
+{
+bool
+path_exists( std::string const& path );
+
+bool
+file_exists( std::string const& path );
+
+std::vector< std::string >
+list_directory(std::string const& path );
+
+void
+create_directories( std::string const& path );
+
+void
+remove_file( std::string const& path );
+} // auxiliary
+} // openPMD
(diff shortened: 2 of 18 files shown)


CMakeLists.txt
```diff
@@ -209,6 +210,24 @@ set(IO_SOURCE
         src/IO/HDF5/HDF5IOHandler.cpp
         src/IO/HDF5/ParallelHDF5IOHandler.cpp)
 
+if(WIN32)
```
ax3l: I think we can just do `#if defined(_WIN32)` in-code and assume all others are POSIX.
Ref: https://stackoverflow.com/questions/5919996/how-to-detect-reliably-mac-os-x-ios-linux-windows-in-c-preprocessor

include/openPMD/auxiliary/Filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+path_exists( std::string const& path );
+
+bool
+file_exists( std::string const& path );
+
+std::vector< std::string >
+list_directory(std::string const& path );
+
+void
+create_directories( std::string const& path );
```
ax3l: docstring: this is `mkdir -p`, right?

include/openPMD/auxiliary/Filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+void
+remove_file( std::string const& path );
```
ax3l: docstring: this is equivalent to `rm` with `-r` or `-f` or both or none?

include/openPMD/auxiliary/Filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+namespace auxiliary
+{
+bool
+path_exists( std::string const& path );
+
+bool
+file_exists( std::string const& path );
+
+std::vector< std::string >
+list_directory(std::string const& path );
```
ax3l: docstring: non-recursive, files-only?

include/openPMD/auxiliary/Filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include <string>
+#include <vector>
+
+namespace openPMD
+{
+namespace auxiliary
+{
+bool
+path_exists( std::string const& path );
```
ax3l: is this `directory_exists`?

test/AuxiliaryTest.cpp
```diff
@@ -337,3 +338,34 @@ TEST_CASE( "dot_test", "[auxiliary]" )
...
+    REQUIRE(contains(dir_entries, "root"));
+    REQUIRE(!contains(dir_entries, "nonexistent_folder_in_root_directory"));
+
+    REQUIRE(path_exists("/"));
+    REQUIRE(path_exists("/boot"));
+    REQUIRE(path_exists("/etc"));
+    REQUIRE(path_exists("/home"));
+    REQUIRE(!path_exists("/nonexistent_folder_in_root_directory"));
+#endif
+}
```
ax3l: missing end-of-file newline

include/openPMD/auxiliary/Filesystem.hpp
anokfireball: That was the intention, to save the user from creating loops or copy-pasting.

include/openPMD/auxiliary/Filesystem.hpp
anokfireball: Non-recursive, every entry (i.e. directories and files).

include/openPMD/auxiliary/Filesystem.hpp
anokfireball: As this will be implemented with [```remove()```](http://pubs.opengroup.org/onlinepubs/9699919799/functions/remove.html), at least ```-r```.
```-f``` should probably not be the default with ```void``` - the user should be notified if the target does not exist. We could borrow from [```std::filesystem::remove()```](http://en.cppreference.com/w/cpp/filesystem/remove), return ```bool``` and then make ```-f``` the default.

File: include/openPMD/auxiliary/Filesystem.hpp (hunk 4 above)
Comment (ax3l) [reply include/openPMD/auxiliary/Filesystem.hpp]: Makes sense. Best to just put this info in the `doxygen` doc string I guess.

File: include/openPMD/auxiliary/Filesystem.hpp (hunk 5 above)
Comment (anokfireball) [reply include/openPMD/auxiliary/Filesystem.hpp]: Yes, it actually checks for a directory. This should be made more explicit.

File: CMakeLists.txt (hunk 1 above)
Comment (ax3l) [reply CMakeLists.txt]: Also, I think it's really unsave to set those vars globally and publicly, e.g. in the tests below ;-)

Comment (ax3l) [reply CMakeLists.txt]: I will push a commit to fix this...

**Comment by ax3l** [comment]
Nice, `FindHDF5.cmake` seems to forget to expose it needs to link against `zlib`. Fun.

**Comment by anokfireball** [comment]
Do not be mistaken:

> Unknown compression format zlib. This might mean that compression will not be enabled.

is a message by our own code:
https://github.com/openPMD/openPMD-api/blob/b8e8de9187e7cc179589c974d042de9cf3ae92b4/src/Dataset.cpp#L68-L70

The current failures are probably because ```mkdir -p``` is not yet working, but ```mkdir``` is (there's only a test for the latter).

**Comment by ax3l** [comment]
Ah ok, I thought it's triggered the other way around. Alright, you can pull again and work on it.

**Comment by ax3l** [comment]
@C0nsultant I think you should be able to traverse your Unix paths for nested creation of dirs like this:
```C++
#include <iostream>
#include <sstream>

void iterEm(std::string input)
{
    std::istringstream ss(input);
    std::string token;

    std::string iterPath;
    if( input.find("/") == 0 )
        iterPath += "/";
    while( std::getline( ss, token, '/' ) ) {
        if( token.size() > 0 )
            iterPath += token + "/";
        std::cout << iterPath << std::endl;
    }
}

int main()
{
    iterEm( "/abc/def/ghi" );
    iterEm( "abc/def/ghi" );
    iterEm( "abc/def/ghi/" );

}
```
output:
```
/
/abc/
/abc/def/
/abc/def/ghi/

abc/
abc/def/
abc/def/ghi/

abc/
abc/def/
abc/def/ghi/
```

ax3l (commented):
I am confused about deletes in cmake dirs ;)

test/AuxiliaryTest.cpp
```diff
@@ -337,3 +338,129 @@ TEST_CASE( "dot_test", "[auxiliary]" )
...
+    REQUIRE(!directory_exists(new_directory));
+    REQUIRE(!remove_directory(new_directory));
+
+    REQUIRE(!remove_file(".\\nonexistent_file_in_cmake_bin_directory"));
+    dir_entries = list_directory("..\\CMakeFiles");
+    if( !dir_entries.empty() )
+    {
+        std::string del = "..\\CMakeFiles\\" + *dir_entries.begin();
+        while( !file_exists(del) && directory_exists(del) )
+        {
```
ax3l: :-o what are you removing here? xD

test/AuxiliaryTest.cpp
```diff
@@ -337,3 +338,129 @@ TEST_CASE( "dot_test", "[auxiliary]" )
...
+    dir_entries = list_directory("../CMakeFiles");
+    if( !dir_entries.empty() )
+    {
+        std::string del = "../CMakeFiles/" + *dir_entries.begin();
+        while( !file_exists(del) && directory_exists(del) )
+        {
+            dir_entries = list_directory(del);
+            del += "/" + *dir_entries.begin();
+        }
+        if( file_exists(del) )
```
ax3l: This looks scary and probably harmful, e.g. for make install.

anokfireball:
The idea was to find a file to test deletion on realiably. Maybe we could delete some ```.cpp``` files instead?

ax3l:
Nooo, don't let us mess with the cake build tree our even worse the source tree :-o

Just use c++ standard file IO to write a new text file for deletion.

ax3l:
I will push an update on this in the next hour.

src/auxiliary/Filesystem.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+bool
+create_directories( std::string const& path )
+{
+    if( directory_exists(path) )
+        return true;
+
+    bool success = true;
+#ifdef _WIN32
+    success = CreateDirectory(path.c_str(), NULL);
```
ax3l: recursive?
https://stackoverflow.com/a/1530815/2719194

(or same as the solution below for POSIX)

src/auxiliary/Filesystem.cpp
```diff
@@ -0,0 +1,167 @@
...
+        } while (FindNextFile(hFind, &data) != 0);
+        FindClose(hFind);
+    }
+#else
+    auto directory = opendir(path.c_str());
+    if( !directory )
+        throw std::system_error(std::error_code(errno, std::system_category()));
+    dirent* entry;
+    while ((entry = readdir(directory)) != nullptr)
+        if( strncmp(entry->d_name, ".", 1) && strncmp(entry->d_name, "..", 2) )
```
anokfireball: This probably skips hidden folders starting with a ```.``` we could need checks for both ```strncmp``` if the ```strlen``` is larger than 1 or 2 respectively.

src/auxiliary/Filesystem.cpp
anokfireball: That or
https://msdn.microsoft.com/en-us/library/bb762130(v=vs.85).aspx
> This function creates a file system folder whose fully qualified path is given by pszPath. If one or more of the intermediate folders do not exist, it creates them.
To set security attributes on a new folder, use SHCreateDirectoryEx.

src/auxiliary/Filesystem.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+    if( directory_exists(path) )
+        return true;
+
+#ifdef _WIN32
+    char seperator = '\\';
+    auto mk = [](std::string const& p) -> bool { return CreateDirectory(p.c_str(), nullptr); };
+#else
+    char seperator = '/';
+    auto mk = [](std::string const& p) -> bool { return (0 == mkdir(p.c_str(), 0777));};
```
ax3l: Maybe let us take the default, environment umask instead of 0777?
https://stackoverflow.com/questions/10147990/how-to-create-directory-with-right-permissons-using-c-on-posix

test/AuxiliaryTest.cpp
```diff
@@ -337,3 +340,120 @@ TEST_CASE( "dot_test", "[auxiliary]" )
...
+#ifdef _WIN32
+    REQUIRE(directory_exists("C:\\"));
+    REQUIRE(directory_exists("C:\\Program Files"));
+    REQUIRE(directory_exists("C:\\Windows"));
+    REQUIRE(!directory_exists("C:\\nonexistent_folder_in_C_drive"));
+
+    auto entries = list_directory(".");
+    for( auto const& entry : entries )
+        std::cout << entry << std::endl;
+    //REQUIRE(file_exists("AuxiliaryTests.exe"));
```
ax3l: should these be enabled or removed?

should the output above be on or is it a debug leftover?

anokfireball: These should be enabled, they just failed because I could not figure out the cwd in appveyor.
Also, the above ```std::cout``` loop should be removed again.

src/auxiliary/Filesystem.cpp
anokfireball: This would actually make Windows and POSIX behaviour equivalent. Agreed!

test/AuxiliaryTest.cpp
anokfireball: For appveyor, this should then be 
```cpp
REQUIRE(file_exists("Release\\AuxiliaryTests.exe"));
```

ax3l: Yes, depends if it's a Debug or Release build ("Configuration") in multi-config build systems such as VS that is used as Generator. Maybe just do not check for this exe and check for another file?

anokfireball: Any file whose (relative) path we can reliably determine is fine, really.
Going from appveyor [logs](https://ci.appveyor.com/project/ax3l/openpmd-api/build/1.0.109/job/h0opcijm5hg0agc5?fullLog=true#L1514), the cwd is ```C:\projects\openpmd-api\build\bin```.
How about
```cpp
REQUIRE(file_exists("..\\..\\CMakeLists.txt"));
```

ax3l: Yep, this is just by the `WORKING_DIRECTORY` I add to the `add_test` calls in `CMakeLists.txt`. I am surprised it is not `bin\Release\` ...

Update: Actually `REQUIRE(file_exists("AuxiliaryTests.exe"));` should work as is - can you try again?

Checking the `CMakeLists.txt` this way is not possible, since it depends on the user-chosen location of the build directory. Let me see which file we can use...

anokfireball: If Python bindings are built, it [appears](https://ci.appveyor.com/project/ax3l/openpmd-api/build/1.0.109/job/h0opcijm5hg0agc5?fullLog=true#L1514) we could check for
```cpp
REQUIRE(file_exists(".\\2_read_serial.py"));

anokfireball: ~Thinking about it, why not just do the same empty file workaround from https://github.com/openPMD/openPMD-api/pull/207/commits/0b873348df53939e788e9c98be50e78ab0af9502?~
Actually, I think we are good with just relying on https://github.com/openPMD/openPMD-api/pull/207/commits/0b873348df53939e788e9c98be50e78ab0af9502, which does a ```file_exists()``` check on a relative path.
This **can** be removed, after all.

ax3l: perfect!

anokfireball:
Once both of these minor issues are addressed, we should be ready to merge. :fireworks:

ax3l:
Yay, nice! :rocket: Do you want to push it or shall I do it?

anokfireball:
@ax3l Pull the trigger.

ax3l (approved):

ax3l:
We are free, freeee!

PR #210, opened 2018-05-28
CI: GCC 4.8

ax3l:
I wonder if we run well with GCC 4.8.

Add an entry in the test matrix of Travis that compiles: gcc 4.8 + Python 3.5.
Update the GCC 6 minor version.

The reason behind this is the following: unfortunately, Python (binary) package and delivery platforms such as conda-forge and PyPI only support GCC 4.8 as of today. Details are e.g. defined in https://github.com/conda-forge/staged-recipes/issues/872 https://github.com/conda-forge/conda-forge.github.io/issues/460 (conda-forge channel) and [PEP571](https://www.python.org/dev/peps/pep-0571/) (PyPI manylinux binary wheels).

Our dependencies (catch2, mpark.variant, pybind11) do support GCC 4.8+, so that's a good start.

- [x] rebase after #224 is merged
- [x] lower supported version for gcc to 4.8

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -207,6 +207,25 @@ jobs:
             - *gcc64_deps
       before_install: *gcc64_init
       script: *script-cpp-unit
+    # GCC 4.8.5 + Python 3.5.5
+    - <<: *test-cpp-unit
+      language: python
+      python: "3.5"
+      env:
+        - USE_MPI=OFF USE_PYTHON=ON USE_HDF5=ON USE_ADIOS1=OFF USE_ADIOS2=OFF
+      compiler: gcc
+      addons:
+        apt:
+          <<: *apt_common_sources
+          packages:
+            - environment-modules
+            - gfortran
+            - g++
+            - libopenmpi-dev
+            - openmpi-bin
+      before_install: &gcc48_init
+         - CXX=g++-4.8 && CC=gcc-4.8 && COMPILERSPEC="%gcc@4.8.5"
+      script: *script-cpp-unit
 
 install:
   # spack install
--- a/.travis/spack/compilers.yaml
+++ b/.travis/spack/compilers.yaml
@@ -32,6 +32,19 @@ compilers:
     modules: []
     operating_system: ubuntu14.04
     paths:
+      cc: /usr/bin/gcc-4.8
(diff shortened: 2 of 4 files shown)


ax3l:
Looks so far like only the known-to-be-broken `<regex>` stdlib in GCC 4.8 is causing RT issues in `/src/Series.cpp` but otherwise we are good.

ax3l:
@C0nsultant I know it's a pain, but do you think we can implement the `matcher` in `Series.cpp` either
- without `<regex>` or
- with [slre](https://github.com/cesanta/slre) (or something, that was just the first one I found)

Alternatively, I could add a statically linked boost::regex fallback for GCC 4.8 for shipping with conda-forge & PyPI wheels...

anokfireball:
It should not be that big of an effort to replace ```<regex>``` (even if the reasons for doing so are not all that convincing to me). All that matcher does is check if a filename matches 
```regex
.*nameprefix\d+namesuffix\.ext$
```
If ```slre```  can match urls, it should be able to handle our case as well.

ax3l:
Btw, shouldn't this be moved into the individual IO backends?

anokfireball:
So you would propse a 
```cpp
static /* ??? */
std::function< bool(std::string const&) >
ConcreteIOHandler::matcher(std::string const& name);
```

I don't have a huge bias either way, but to me this feels like logic the backends should not be concerned with. It's pretty openPMD-specific, yet the backends know virtually nothing about openPMD.

ax3l:
> even if the reasons for doing so are not all that convincing to me

@C0nsultant I updated the PR description now to be more precise. Does it make more sense to you now? Sorry for the initial sketchy text, I just wanted to trigger the CI for a test.

> ConcreteIOHandler::matcher

Ah don't worry, the current location is fine as well.

ax3l:
> ConcreteIOHandler::matcher

Ah don't worry, the current location is fine for that.

anokfireball:
Your reasoning for getting gcc 4.8 to work is great. For most people, getting started with this library *could* then be even easier.

I guess that's the grudge of being a library author and maintainer - working around all the broken stuff the ecosystem throws at you...

anokfireball (approved):

PR #212, opened 2018-05-29
Minimize number of flushes in backend

### LINKED ISSUE #205: Improve utilization of IO-queue for loadChunk/storeChunk
It has come up in https://github.com/openPMD/openPMD-api/pull/204#discussion_r190864021 that there are unnecessarily many calls to ```.flush()``` during dataset chunk transfers.

anokfireball:
Closes #205.


src/backend/PatchRecordComponent.cpp
```diff
@@ -62,27 +62,39 @@ PatchRecordComponent::PatchRecordComponent()
...
-        IOHandler->flush();
-    }
+        while( !m_chunks->empty() )
+        {
+            IOHandler->enqueue(m_chunks->front());
+            m_chunks->pop();
+        }
 
-    while( !m_chunks->empty() )
+        IOHandler->flush();
```
anokfireball: This ```flush()``` should not be necessary. For ```AccessType::READ_ONLY```, it can be pulled up into ```Series::flush()```.

src/RecordComponent.cpp
```diff
@@ -67,45 +67,57 @@ RecordComponent::getExtent() const
...
-            dCreate.transform = m_dataset->transform;
-            IOHandler->enqueue(IOTask(this, dCreate));
+            IOHandler->enqueue(m_chunks->front());
+            m_chunks->pop();
         }
-        IOHandler->flush();
-    }
 
-    while( !m_chunks->empty() )
+        IOHandler->flush();
```
anokfireball: This ```flush()``` should not be necessary. For ```AccessType::READ_ONLY```, it can be pulled up into ```Series::flush()```.

anokfireball:
More generally, there should be no explicit ```IOHandler->flush()``` for ```IOTasks``` **without** input parameters (i.e. ```std::shared_ptr```s). ```IOTasks``` **with** input parameters typically will require ```IOHandler->flush()```. (There are exceptions when we need to manually re-link the hierarchy, e.g. for scalar records.)

All pending ```IOTasks``` without input parameters can be executed with a ```IOHandler->flush()``` at the end of ```Series::flush()```, unconditionally.

ax3l (approved):
Really nice!

PR #215, opened 2018-05-31
CI: Test HDF5 Minimal Version

ax3l:
Add a entry in the matrix that tests the minimal supported HDF5 version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -198,7 +198,7 @@ jobs:
       language: python
       python: "3.6"
       env:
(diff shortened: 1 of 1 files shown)


anokfireball:
The problem should not be as severe for ADIOS, but we might want to do the same at some point.

anokfireball (approved):

ax3l:
yep, with #190 as soon as the upstream spack bug with external dependencies is fixed :)

Issue #216, opened 2018-05-31
Avoid <regex> (for now)


ax3l:
As discussed in #210, we will try to also support GCC 4.8.

For this to work, the only thing we will need to change is the use of `<regex>` inside the library, as it is only implemented as a stub in GCC 4.8.

Possible alternatives are:
- own, simple string matcher for file names
- use another light-weight regex library, e.g. [slre](https://github.com/cesanta/slre)

For reference, [this](https://github.com/openPMD/openPMD-api/pull/217#issuecomment-393532139) is what needs to be solved.

ax3l:
@C0nsultant if you have an idea how to implement this efficiently, feel free to say how or go ahead. I just tried to quickly hack it but needs more thinking.

anokfireball:
You weren't far off. The only thing you didn't consider are the troubles caused by having both a prefix and a postfix.
I'll look into it.

ax3l:
Since it's a GCC 4.8 only issue, we could also use the POSIX [`<regex.h>` header](http://man7.org/linux/man-pages/man3/regex.3.html) or the [regex lib](https://developer.gnome.org/glib/stable/glib-Perl-compatible-regular-expressions.html) from `glibc` here as a fallback.

Resources:
- https://stackoverflow.com/questions/31628536/what-is-difference-between-c-and-c-regex
- http://lh3lh3.users.sourceforge.net/reb.shtml

ax3l:
Ok, I have a new draft based on the POSIX C `<regex.h>`

PR #217, opened 2018-05-31
Series: Replace File Matching Regex

ax3l:
Just check by file ending. A group based encoded series might not have a iteration in its (single-file) name anyway.

Draft for #216

### DIFF
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -24,7 +24,6 @@
 #include "openPMD/Series.hpp"
 
 #include <iostream>
-#include <regex>
 
 
 namespace openPMD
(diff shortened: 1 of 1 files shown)


anokfireball (changes requested):
This will match ```any.h5``` for series ```other%T.h5```.

ax3l:
But what if my series is group based inside a single file?
updated to only match `any*.h5` for series `any`.

anokfireball:
That is not what this regex is used for.
```groupBased``` openPMD files are read in by the concrete name provided by the user:
https://github.com/openPMD/openPMD-api/blob/2847752aecfb1c32710f2ecdcb0300eaab35ae68/src/Series.cpp#L590-L592
where ```m_name``` is the string provided in the ```Series``` constructor (with the leading path stripped away):
https://github.com/openPMD/openPMD-api/blob/2847752aecfb1c32710f2ecdcb0300eaab35ae68/src/Series.cpp#L153
No regex used there.

What it is actaully used for is iterating through a directory. Take our ```samples``` directory, for example:
You'll have a bunch of ```.h5```s, some of them belonging to a ```fileBased``` output, others to different ```groupBased``` outputs. There may even be more than one ```fileBased``` Series. Possibly even other file formats.
What this regex solves is this: Find every file that matches 
```regex
^prefix\d+posfix\.extension$
```
in
```
[N] prefixposfix.extension
[N] prefixNOTANUMBERposfix.extension
[N] prefix1posfix
[N] prefix1posfix.otherextension
[N] prefix1posfix.extension.dir
[N] 1posfix.extension
[N] prefix1.extension
[N] otherprefix1posfix.extension
[N] prefix1otherposfix.extension
[Y] prefix1posfix.extension
```

src/Series.cpp
```diff
@@ -770,20 +769,28 @@ cleanFilename(std::string const& filename, Format f)
...
 {
     switch( f )
     {
         case Format::HDF5:
         {
-            std::regex pattern(auxiliary::replace_last(name + ".h5$", "%T", "[[:digit:]]+"));
-            return [pattern](std::string const& filename) -> bool { return std::regex_search(filename, pattern); };
+            return [seriesName](std::string const& filename) -> bool
+            {
+                bool const matchEnd( filename.rfind(".h5") == filename.size() - 3u );
```
anokfireball, line 780: We have a whole bunch of [string manipulation](https://github.com/openPMD/openPMD-api/blob/dev/include/openPMD/auxiliary/StringManip.hpp) coded up and tested already, if you want to save some time.

ax3l:
ok well, then this won't even work as a quick-hack xD

PR #218, opened 2018-06-04
Compiler Docs: VS 2015

ax3l:
Document that Visual Studio 2015 (and potentially newer) are supported.

Compiler names [are weird](https://stackoverflow.com/a/31885449/2719194) in VS...


anokfireball (approved):
Strictly speaking: These are not compiler versions, but IDE versions. As you typically only get vc++ with VS, this should be fine, though.

PR #219, opened 2018-06-04
First Python Numpy Bindings

ax3l:
Add numpy bindings for loading and storing actual data.

Related to #32

## References

- pybind11:
  - https://github.com/pybind/pybind11/blob/master/tests/test_numpy_array.cpp
  - https://github.com/pybind/pybind11/blob/master/tests/test_numpy_dtypes.cpp
  - https://github.com/pybind/pybind11/issues/1299
  - https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html
- numpy: https://docs.scipy.org/doc/numpy/reference/arrays.dtypes.html

### DIFF
--- a/examples/2_read_serial.py
+++ b/examples/2_read_serial.py
@@ -37,14 +37,17 @@ if __name__ == "__main__":
     print("Field E.x has shape {0} and datatype {1}".format(
           shape, E_x.dtype))
 
+    offset = openPMD.Extent([1, 1, 1])
+    extent = openPMD.Extent([2, 2, 1])
     # TODO buffer protocol / numpy bindings
     # chunk_data = E_x[1:3, 1:3, 1:2]
+    chunk_data = E_x.load_chunk(offset, extent)
     # print("Queued the loading of a single chunk from disk, "
     #       "ready to execute")
     series.flush()
-    # print("Chunk has been read from disk\n"
-    #       "Read chunk contains:")
-    # print(chunk_data)
+    print("Chunk has been read from disk\n"
+          "Read chunk contains:")
+    print(chunk_data)
     # for row in range(2):
     #     for col in range(2):
     #         print("\t({0}|{1}|{2})\t{3}".format(
--- a/examples/2_read_serial.py
+++ b/examples/2_read_serial.py
@@ -41,9 +41,10 @@ if __name__ == "__main__":
     # chunk_data = E_x[1:3, 1:3, 1:2]
     # print("Queued the loading of a single chunk from disk, "
     #       "ready to execute")
-    # series.flush()
+    series.flush()
     # print("Chunk has been read from disk\n"
     #       "Read chunk contains:")
+    # print(chunk_data)
     # for row in range(2):
     #     for col in range(2):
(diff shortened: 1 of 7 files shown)


ax3l:
@C0nsultant I will make the next release `0.2.0-alpha` to have at least some indication of the new API and we decide on making any later version `0.X.0-beta`.

examples/3_write_serial.py
```diff
@@ -7,22 +7,22 @@
...
-    # global_data = np.range(size*size, dtype=np.double)
+    global_data = np.arange(size*size, dtype=np.double)
 
     print("Set up a 2D square array ({0}x{1}) that will be written".format(
         size, size))
 
     # open file for writing
     series = openPMD.Series(
-        "../samples/3_write_serial.h5",
+        "../samples/3_write_serial_py.h5",
```
anokfireball, line 25: As this is supposed to be run on all platforms, we probably should use [```os.path.join```](https://docs.python.org/3.6/library/os.path.html#os.path.join). This has already come up as a problem in #229.

anokfireball:
This addition is significant enough to warrant a version bump from ```0.1.X``` to ```0.2.0```.  Feel free to do so.

examples/3_write_serial.py
ax3l: Good catch!
Huh, I wonder why it passes the AppVeyor tests...

anokfireball: HDF5 triggers a [warning](https://ci.appveyor.com/project/ax3l/openpmd-api/build/1.0.211/job/og7t29wucynkm1r7?fullLog=true#L2831), but does not abort.
That is because the ```ASSERT``` in the backend that would catch it is a NOOP in ```Release``` builds (will be fixed in #229).

ax3l:
@C0nsultant fixed it :)

examples/2_read_serial.py
```diff
@@ -7,11 +7,14 @@
...
 """
 import openPMD
+import os
 
 
 if __name__ == "__main__":
-    series = openPMD.Series("../samples/git-sample/data%T.h5",
-                            openPMD.Access_Type.read_only)
+    series = openPMD.Series(
+        os.path.join("..", "samples", "git-sample", "data%T.h5"),
```
ax3l: ugh, on Windows:
```
RuntimeError: No matching iterations found: ..\samples\git-sample\data%T
```

anokfireball: This is very likely related to the ```cwd``` as discussed in https://github.com/openPMD/openPMD-api/pull/207#discussion_r192012409.

ax3l: shouldn't be, it's the same in `2_read_serial.cpp`, no?

anokfireball: Indeed that should be the same directory. All the more confusing that [2_read_serial.exe](https://ci.appveyor.com/project/ax3l/openpmd-api/build/1.0.214/job/aqgimr72ysdvrvvx?fullLog=true#L1675) passes, as that one uses UNIX-style paths.

ax3l: I will keep it as before for now and we can investigate later in #229

PR #224, opened 2018-06-05
GCC 4.8: <regex> to <regex.h>

### LINKED ISSUE #216: Avoid <regex> (for now)

ax3l:
Close #216 Avoid the missing `<regex>` library in GCC 4.8 by using the POSIX C library equivalent.

### DIFF
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -24,7 +24,18 @@
 #include "openPMD/Series.hpp"
 
 #include <iostream>
-#include <regex>
+
+#if defined(__GNUC__)
+#   if (__GNUC__ == 4 && __GNUC_MINOR__ < 9)
+#       define IS_GCC_48 1
+#   endif
+#endif
+
+#if defined(IS_GCC_48)
+#   include <regex.h>
+#else
+#   include <regex>
+#endif
 
 
 namespace openPMD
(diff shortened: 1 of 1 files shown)


src/Series.cpp
```diff
@@ -24,7 +24,18 @@
 #include "openPMD/Series.hpp"
 
 #include <iostream>
-#include <regex>
+
+#if defined(__GNUC__)
+#   if (__GNUC__ == 4 && __GNUC_MINOR__ < 9)
```
ax3l, line 29: I have tested this locally by replacing this test with a single `#   if __GNUC__ < 9`

ax3l:
@C0nsultant ping :)

anokfireball (approved):
Recompiling the regex on every call to the functor seems excessive. This should have no significant impact compared to the IO that follows, though.

ax3l:
yep. as a side note: that is probably what the `<regex>` `pattern` init is doing as well

Issue #87, opened 2018-02-20
Versioning Scheme


ax3l:
Versioning of the lib can be tricky since the `openPMD` [base standard](https://github.com/openPMD/openPMD-standard) also has a version and we want to avoid confusion.

The idea I came up with yet was to use always the **same major version** of `openPMD-api` as the base standard and try to support all minor versions of this major (but not previous major versions; see [updater](https://github.com/openPMD/openPMD-updater)).

The only problem comes in when we want to do incompatible API changes *within* a major version, which would break the usual [semantic versioning](https://semver.org/) of `openPMD-api`.

In the README (badges & test) and in `openPMD/version.hpp` defines we then define additional macros (C++: done already; Py: Needs Exposure) that expose which min and max of the openPMD standard are supported.

Of course we could also just do a simple, independent [semantic versioning](https://semver.org/) of `openPMD-api` (with the docs as mentioned above) but I would bet this will confuse people a lot...

ax3l:
@C0nsultant as soon as we have #207 and #216 merged I would like to tag a first alpha release, e.g. `1.0.0-alpha` or `1.0.0-alpha.1`, so I can try to set up a `conda-forge` feedstock.

After that I would say we [fix and finish leftovers](https://github.com/openPMD/openPMD-api/projects/2) until `1.0.0-beta.N` releases and invite users to test it.

anokfireball:
We should be feature-ready for non-critical deployment by this point.
#70 makes us fairly light-weight with respect to dependencies, so this is indeed a suitable candidate for stepping up from ```dev```.

When following your proposed scheme, we would tag ```1.1.0-alpha```. How about we start with ```1.1.0-alpha.0```?

ax3l:
As you like, I am just trying to be compatible with [semver.org](https://semver.org) and they suggest `1.0.0-alpha.1` as far as I can see.

ax3l:
Hm, I usually name all our projects without the last dot. Is `1.0.0-alpha1`, `1.0.0-alpha2`, `1.0.0-beta1`, `1.0.0-rc1`, `1.0.0-rc2`, `1.0.0`, ... ok with you?

ax3l:
Hm, CMake does not support suffixes in `project(VERSION ...)` and crops them in `VERSION_...` comparisons: https://cmake.org/pipermail/cmake-developers/2017-March/029928.html

ax3l:
[...]

PR #226, opened 2018-06-06
Version: 0.1.0-alpha

ax3l:
Prepare the first developer preview, `0.1.0-alpha`.

I will create a GPG signed tag after that and draw a release, including a DOI on [rodare](https://rodare.hzdr.de/).

Later on, we will do a `stable` and `release-newVersion` branches but for now we just bump versions happily inside the `dev`.

@C0nsultant do you have an [ORCiD](https://orcid.org/) that I can use to identify you?

Related to #87


anokfireball:
> do you have an ORCiD that I can use to identify you?

I have now: https://orcid.org/0000-0001-8704-1769

ax3l:
Made the version unique when suffix is dropped:
```
# non-semantic initial pre-releases
0.1.0-alpha
0.1.1-alpha
0.1.2-alpha

(there will not be a 0.1.Y release without -alpha suffix)

...
0.X.0-beta
0.X.1-beta

(there will not be a 0.X.Y release without -beta suffix)
...

# semantic releases, following `standardMAJOR.apiMAJOR.apiMINOR`
1.0.0 (stable)
```

anokfireball (approved):

ax3l:
just as a note: there is an issue with rodare and webhooks from GitHub. Will add the DOI to the release another time.

PR #227, opened 2018-06-06
SerialIOTest: loadChunk template ADIOS1

ax3l:
Forgot to change this access to `loadChunk` in #204.
Seems to cause only problems on GCC 4.8.**2** & ADIOS1

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,23 @@
 Changelog
 =========
 
+0.1.1-alpha
+-----------
+**Date:** TBD
+
+[Title TBA]
+
+[Short Summary]
+
+Changes to "0.1.0-alpha"
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Bug Fixes
+"""""""""
+
+- SerialIOTest: ``loadChunk`` template missing for ADIOS1 #227
+
+
 0.1.0-alpha
 -----------
 **Date:** 2018-06-06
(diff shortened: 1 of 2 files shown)


anokfireball:
How did [this](https://travis-ci.org/openPMD/openPMD-api/jobs/384911753#L916) even compile? :confused: 

**EDIT** Oh well, it is ADIOS related which we never built before.

ax3l:
Dunno, weird. Also, I am compiling it successfully with GCC 4.8.**5** in #210

update: ah, ADIOS coverage. ok. #190

anokfireball (approved):

ax3l:
It's still weird, since I am compiling successfully with adios (+mpi) locally (with GCC 6.3)

ax3l:
Oh, I know why. The serial tests are accidentally excluded if a parallel ADIOS is found. Will enable them.

ax3l:
Ah, adios is lovely. When using the serial API it still requires to be started with `mpirun`... Waaaaaaaa

ax3l:
`SerialIOTest`: When using the serial API it still requires to be started with `mpirun` if we link the parallel ADIOS lib...?

Are you relying on ADIOS' MPI moc library for the serial API? You must not call `MPI_Comm_dup` in serial IO (or any `MPI_` command). (We currently just assume it's in an MPI context anyway and we can use `MPI_COMM_SELF`. This is fine for a lot of applications, but subideal for usability. Anyway, this could be a current limitation caused by the design of ADIOS1...)

ax3l:
Found a solution: refactored us and added a fix to adios

ax3l:
I will split the ADIOS fix in a second PR since it is unrelated.

PR #228, opened 2018-06-06
Serial ADIOS: Refactor MPI Comm

ax3l:
Instead of relying on the ADIOS serial MPI moc library, we can also pass `MPI_COMM_NULL` as communicator to avoid triggering MPI functionality and requiring mpi startup.

Needs fix in

  https://github.com/ornladios/ADIOS/pull/182

in order to work without `mpiexec` and linked against a parallel ADIOS library (but will work with any serial ADIOS library as before).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Bug Fixes
 """""""""
 
 - SerialIOTest: ``loadChunk`` template missing for ADIOS1 #227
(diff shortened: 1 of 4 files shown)


anokfireball (approved):

ax3l:
hm, there is much more calling to MPI functions inside ADIOS' bp functionality which is not mocked when compiled in parallel... argh.

PR #240, opened 2018-06-11
setup.py for sdist

ax3l:
Developed a `setup.py` for python-level install *from source* (actually compiling!).

I am not sure it it does more harm than good to expose this to the user or if I should hide a perfectly tailored, `PyPI` distribution building only, `setup.py` deep in the source tree. Or I just keep this not as a documented (aka recommended) install method.

~~I will keep this here for now for documentation purposes.~~

Done:
- `sdist`

Intentionally skipepd:
- `bdist_wheel`: fully static, non-MPI
  - linux: [manylinux2010](https://www.python.org/dev/peps/pep-0571/) PEP571
  - win: ~~mingw~~ (no `std::future` support as of 7.0) -> VC 2015? PEP?
  - osx: apple clang? PEP?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,154 @@
+import os
+import re
+import sys
+import platform
+import subprocess
+
+from setuptools import setup, Extension
+from setuptools.command.build_ext import build_ext
+from distutils.version import LooseVersion
+
+
+class CMakeExtension(Extension):
+    def __init__(self, name, sourcedir=''):
+        Extension.__init__(self, name, sources=[])
+        self.sourcedir = os.path.abspath(sourcedir)
+
+
+class CMakeBuild(build_ext):
+    def run(self):
+        try:
+            out = subprocess.check_output(['cmake', '--version'])
+        except OSError:
+            raise RuntimeError(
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -0,0 +1,149 @@
...
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Natural Language :: English',
+        'Environment :: Console',
+        'Intended Audience :: Science/Research',
+        'Operating System :: OS Independent',
+        'Topic :: Scientific/Engineering',
+        'Topic :: Database :: Front-Ends',
+        'Programming Language :: C++',
+        'Programming Language :: Python',
```
ax3l: this line probably needs to go to make sure it's really just matching Python 3 on PyPI and a `pip2 install openpmd-api` can fail early.

ax3l: uh, it's [`python_requires`](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) for setuptools on PyPI.
([ref](https://github.com/pypa/pip/issues/3182#issuecomment-289073930))

PR #244, opened 2018-06-11
[WIP] Fix RODARE: Title/Descr

ax3l:
I am not allowed to change/set again the title/description
in RODARE since it must match the title in ROBIS...

update: hm, even seen with manual release. could be a RODARE bug... will get feedback by next week.


ax3l:
works now, RODARE issue fixed.

PR #229, opened 2018-06-06
Replace ASSERT with VERIFY

anokfireball:
Closes #151.

PR #247, opened 2018-06-12
Check validity of internally used pointers

### LINKED ISSUE #209: Assert non-null pointer during storeChunk
There is currenlty no check during chunk storage whether the passed in pointer is valid:

https://github.com/openPMD/openPMD-api/blob/d16f1762fcb1be7e7441755a3209abdba7994173/include/openPMD/RecordComponent.hpp#L238-L240

We should do a simple non-null check before the pointer cast.

### LINKED ISSUE #211: Verify reference counts of shared_ptrs in IO queue
To reduce the risk of reading from and writing to invalid pointers resulting from deallocated ```shared_ptr```s, we should unit-test the ```.ref_count()```s of all used smart pointers in the fronend, IO queue and backend.

Came up in https://github.com/openPMD/openPMD-api/pull/204#issuecomment-392599715.

anokfireball:
Closes #209 and closes #211.

### DIFF
--- a/test/CoreTest.cpp
+++ b/test/CoreTest.cpp
@@ -557,3 +557,24 @@ TEST_CASE( "wrapper_test", "[core]" )
     REQUIRE(pp["numParticles"][RecordComponent::SCALAR].m_chunks->empty());
 #endif
 }
+
+TEST_CASE( "use_count_test", "[core]" )
+{
+    Series o = Series("./new_openpmd_output", AccessType::CREATE);
+
+    MeshRecordComponent mrc = o.iterations[1].meshes["E"]["x"];
+    mrc.resetDataset(Dataset(Datatype::UINT16, {42}));
+    std::shared_ptr< uint16_t > storeData = std::make_shared< uint16_t >(44);
+    REQUIRE(storeData.use_count() == 1);
+    mrc.storeChunk({0}, {1}, storeData);
+    REQUIRE(storeData.use_count() == 2);
+    o.flush();
+    REQUIRE(storeData.use_count() == 1);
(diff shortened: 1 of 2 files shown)


test/CoreTest.cpp
```diff
@@ -557,3 +557,24 @@ TEST_CASE( "wrapper_test", "[core]" )
     REQUIRE(pp["numParticles"][RecordComponent::SCALAR].m_chunks->empty());
 #endif
 }
+
+TEST_CASE( "use_count_test", "[core]" )
+{
+    Series o = Series("./new_openpmd_output", AccessType::CREATE);
```
ax3l, line 563: no file ending? :-o how does it know its backend? :-o
what is this witchcraft? :-o

ax3l (approved):

test/CoreTest.cpp
anokfireball: If no known filename extension can be matched, we can't determine a suitable backend and [throw](https://github.com/C0nsultant/libopenPMD/blob/0b97c08c0094bc52f461d15d3d905e43b7ac44a6/src/Series.cpp#L798) (after #229 is merged). With the only exception being *no file extension at all*, in which case we do [NOOPs](https://github.com/openPMD/openPMD-api/blob/63dc8818d12ba868466e950b5c252afec71536af/src/IO/AbstractIOHandler.cpp#L106) with help of [```Format::DUMMY```](https://github.com/openPMD/openPMD-api/blob/e05b136b2b6a4d6247be573fc44922f943d49ce7/src/Series.cpp#L770).

PR #249, opened 2018-06-12
unittests for python API

CFGrote:
Related to #32 

This PR introduces unittesting of the python wrappers. Only new dependency is the [unittest library](https://docs.python.org/3/library/unittest.html) which comes with the standard python libraries. Some tests use the sample files that have to be pulled with the .travis/download_samples.sh script. 

To run the tests, do `python test/python/unittest/Test.py -v`. To add more tests, use `tests/python/unittest/API/APITest.py` as a template and don't forget to import new test modules `Test.py` and to add the corresponding test suite. I'll be happy to assist.


ax3l:
@CFGrote hurray, thank you so much for the PR, Carsten!

Can you please add a PR description? Especially, can you highlight what dependencies are needed for the tests, e.g. are they part of the python standard libraries or external? How to call the tests manually, how to add more tests (e.g. in separate files?), etc.?

include/openPMD/Mesh.hpp
```diff
@@ -174,6 +174,10 @@ class Mesh : public BaseRecord< MeshRecordComponent >
     template< typename T >
     Mesh& setTimeOffset(T timeOffset);
 
+    /** Access to records.
+     */
+    Container< MeshRecordComponent > records;
```
ax3l: Please revert the addition in this file since it's a bit outdated.
Records and record components are accessed like this instead:
https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/examples/2_read_serial.py#L34-L44

src/binding/python/Mesh.cpp
```diff
@@ -63,6 +66,7 @@ void init_Mesh(py::module &m) {
         .def_property_readonly("time_offset", &Mesh::timeOffset<double>)
         .def_property_readonly("time_offset", &Mesh::timeOffset<long double>)
 
+        .def_readwrite("records", &Mesh::records)
```
ax3l: same here, this should not be needed to be exposed

src/binding/python/Mesh.cpp
```diff
@@ -29,6 +29,9 @@
 namespace py = pybind11;
 using namespace openPMD;
 
+using PyRecordsContainer = Container< MeshRecordComponent >;
```
ax3l: same here, this should not be needed

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,326 @@
...
+        obj = openPMD.Particle_Container()
+
+    def testRecord(self):
+        """ Test openPMD.Record. """
+        # Has only copy constructor.
+        self.assertRaises(TypeError, openPMD.Record)
+
+        ### FIXME
+        ## Get a record (fails)
+        #record = self.__series.iterations[100].particles['electrons'].properties['positions'].components['x']
```
ax3l: Should be accessible via:
```python
record = self.__series.iterations[100].particles['electrons']['positions']['x']
```

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,326 @@
...
+        Ex = E["x"]
+
+        print (type(Ex))
+        self.assertIsInstance(Ex, openPMD.Mesh_Record_Component)
+
+
+
+# TODO: add __getitem__ to openPMD.Mesh and openPMD.Particle object for
+#       non-scalar records: return a record component
+# E_x = i.meshes["E"]["x"]
```
ax3l: you can do that now, see the example here :-)
https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/examples/2_read_serial.py#L34-L44

ax3l (dismissed):
Very nice draft!
I added some inline comments on already implemented parts that can be used.

CFGrote:
@ax3l can you add to .travis.yml that these test actrually run ? sth. like python test/unittest/Test.py -v should do

ax3l:
Just as a note, I am right now fixing a CI issue with HDF5. Please rebase your branch against `dev` as soon as we merged #250

Yes, I can push on your branch to address CMake and CI related features in the end :-)

ax3l:
@CFGrote CI fixed, please fetch and rebase :-)
edit: never mind, my manual restart fixes this as well and merged onto the new dev.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,339 @@
...
+        obj = openPMD.Particle_Container()
+        del obj
+
+    def testRecord(self):
+        """ Test openPMD.Record. """
+        # Has only copy constructor.
+        self.assertRaises(TypeError, openPMD.Record)
+#        ### FIXME
+#        ## Get a record (fails)
+#        #record = self.__series.iterations[100].particles[
```
ax3l: can you try if

```python
record = self.__series.iterations[100].particles['electrons']['position']['x']
```
works?

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,339 @@
...
+
+        E = self.__series.iterations[100].meshes["E"]
+        Ex = E["x"]
+
+        print(type(Ex))
+        self.assertIsInstance(Ex, openPMD.Mesh_Record_Component)
+
+# TODO: add __getitem__ to openPMD.Mesh and openPMD.Particle object for
+#       non-scalar records: return a record component
+# E_x = i.meshes["E"]["x"]
```
ax3l: works now, please see example in https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/examples/2_read_serial.py#L34-L47

ax3l (changes requested):
Forgotten inline comments re-added :-)

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,339 @@
...
+
+        # Check type.
+        i = series.iterations[100]
+
+        self.assertTrue(hasattr(i.particles, "__getitem__"))
+        for ps in i.particles:
+            self.assertIsInstance(ps, str)
+            self.assertIsInstance(i.particles[ps], openPMD.ParticleSpecies)
+
+    def testAllocation(self):
```
anokfireball: This can be removed together with the C++ [code](https://github.com/openPMD/openPMD-api/blob/06ae64508f726d18eaedc60d7e73580e7f24a054/include/openPMD/RecordComponent.hpp#L52-L57) and [bindings](https://github.com/openPMD/openPMD-api/blob/b95e33c6627aa41afca80bad0d2265846451f5e4/src/binding/python/RecordComponent.cpp#L178-L182) behind it. We do not allow the user to specify allocation strategy any more.
If you include this, I'll remove it when we get rid of the ```enum``` with #253.

test/python/unittest/Test.py
```diff
@@ -0,0 +1,28 @@
+""" :module: Top level test suite  """
```
ax3l: please add the same copyright line as in the other file :)

test/python/unittest/TestUtilities/TestUtilities.py
```diff
@@ -0,0 +1,24 @@
+""" :module: Test Utilities """
+
```
ax3l: please here as well

test/python/unittest/API/APITest.py
CFGrote:
```
ERROR: testRecord (__main__.APITest)
Test openPMD.Record.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "APITest.py", line 293, in testRecord
    record = self.__series.iterations[100].particles['electrons']['positions']['x']
TypeError: 'openPMD.ParticleSpecies' object is not subscriptable

```

test/python/unittest/API/APITest.py
CFGrote: I'll leave it in but modify: now asserts AttributeError on openPMD.Allocation .

test/python/unittest/Test.py
CFGrote: done

test/python/unittest/TestUtilities/TestUtilities.py
CFGrote: done

test/python/unittest/API/APITest.py
CFGrote: ok, works. test added

ax3l:
@CFGrote I think the rebase includes a little too many commits now. Just rebase on mainline's `dev` (fetch first) and remove the ones not coming from you on top :-)

test/python/unittest/API/APITest.py
ax3l: Ah, alright. Not implemented yet, a `FIXME` note is great.

If you rebase against the latest `dev`, the particle access should work up to:
```python
self.__series.iterations[100].particles['electrons']['position']
```

and still lacks access to the *record component* `['x']`

ax3l:
@CFGrote can I help somehow to finish this PR? :-)

CFGrote:
i fear i f***ed up my repo. @ax3l, can you solve the conflict from your end?

ax3l (approved):

ax3l:
@CFGrote shall I fix the PEP issues and add CI?
(that means I will directly push on your PR branch, so don't forget to fetch & pull later on. Also, I think you forgot to branch before the PR ;-) )

ax3l (approved):

ax3l:
@C0nsultant feel free to review & merge this if all passes :)

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+    def setUp(self):
+        """ Setting up a test. """
+        self.__files_to_remove = []
+        self.__dirs_to_remove = []
+
+        path_to_field_data = generateTestFilePath(
+                "issue-sample/no_particles/data00000400.h5")
+        path_to_particle_data = generateTestFilePath(
+                "issue-sample/no_fields/data00000400.h5")
+        path_to_data = generateTestFilePath("git-sample/data00000100.h5")
```
anokfireball: To suppress all warnings
> Series constructor called with explicit iteration suggests loading a single file with groupBased iteration encoding. Loaded file is fileBased.

plase replace the explicit iteration numbers in these filenames (i.e. ```00000400``` and ```00000100```) with the [iterationFormat](https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#iterations-and-time-series) regex ```%T```.
This will still load your desired iterations (additionally to all other iterations in that Series).

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        del self.__series
+
+    def testFieldData(self):
+        """ Testing serial IO on a pure field dataset. """
+
+        # Get reference to series stored on test case.
+        series = self.__field_series
+
+        print("Read a Series with openPMD standard version %s" %
+              series.openPMD)
```
anokfireball: If the ```print()``` statements are used for file inspection, replace them with assertions where possible.
Reflecting the logic from the Python examples may be helpful during development of test cases, but only adds noise when running the tests.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+
+        print("The Series contains {0} iterations:".format(
+            len(series.iterations)))
+        for i in series.iterations:
```
anokfireball: This is a good candidate for a ```testContainer``` case where you check if 
```python
iter(series.iterations)
```
returns a sane result.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+            print("\t {0}".format(i))
+        print("")
+
+        self.assertEqual(len(series.iterations), 1)
+
+        i = series.iterations[400]
```
anokfireball: Before accessing with ```[]```, you should assert that Iteration 400 is actually contained. This API creates objects (with sane openPMD defaults) for missing keys instead of raising ```KeyError```.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        print("Iteration 100 contains {0} meshes:".format(len(i.meshes)))
```
anokfireball: Iteration number is hardcoded.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        E = i.meshes["E"]
+        self.assertIsInstance(E, openPMD.Mesh)
+
+        print("Iteration 100 contains {0} particle species:".format(
+            len(i.particles)))
+        for ps in i.particles:
+            print("\t {0}".format(ps))
+        print("")
+
+        # Get a mesh.
```
anokfireball:
```python
# Get a particle species.
```

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        extent = [2, 2, 1]
+
+        chunk_data = E_x.load_chunk(offset, extent)
+        series.flush()
+
+        print("Chunk has been read from disk\n"
+              "Read chunk contains:")
+        print(chunk_data)
+
+    def testLoadSeries(self):
```
anokfireball, line 168: This is reduant in the current form. The one contained assertion is included in the other three Series test cases.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        E_x = i.meshes["E"]["x"]
+        shape = E_x.shape
+
+        print("Field E.x has shape {0} and datatype {1}".format(
+              shape, E_x.dtype))
+
+        offset = [1, 1, 1]
+        extent = [2, 2, 1]
+
+        chunk_data = E_x.load_chunk(offset, extent)
```
anokfireball, line 161: This should be moved into ```testRecord_Component```.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        series.flush()
+
+        print("Chunk has been read from disk\n"
+              "Read chunk contains:")
+        print(chunk_data)
```
anokfireball, line 166: The shape, datatype and content of the read chunk should be verified.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        # Loop over iterations.
+        self.assertIsInstance(series.iterations, openPMD.Iteration_Container)
+        self.assertTrue(hasattr(series.iterations, "__getitem__"))
+        for i in series.iterations:
+            # self.assertIsInstance(i, int)
+            self.assertIsInstance(series.iterations[i], openPMD.Iteration)
+
+        # Check type.
+        i = series.iterations[100]
+        self.assertIsInstance(i, openPMD.Iteration)
```
anokfireball: This is now done in the loop above. Maybe we should check if this succeeds for any newly created Iteration
```python
i not in series.iterations
```

test/python/unittest/API/APITest.py
anokfireball: In that testcase, we should also check what happens when you access invalid keys.
```python
series.iterations['definitely not an int']
```
In C++, this is caught by the compiler. I'm not sure about the result in Python.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def testData(self):
+        """ Test IO on data containing particles and meshes."""
+
+        # Get series.
+        series = self.__series
+        self.assertIsInstance(series, openPMD.Series)
+
+        print("Read a Series with openPMD standard version %s" %
+              series.openPMD)
```
anokfireball: The datatype and content of contained attributes should be verified in all testcases that test concrete openPMD objects.

anokfireball: Getters and setters for attributes required by the standard should also be verified.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        obj = openPMD.Geometry(0)
+        del obj
+
+    def testIteration(self):
+        """ Test openPMD.Iteration. """
+        self.assertRaises(TypeError, openPMD.Iteration)
+
+        iteration = self.__particle_series.iterations[400]
+
+        copy_iteration = openPMD.Iteration(iteration)
```
anokfireball: @ax3l does this return deep copies or are these shallow copies as intended with #144?
Assuming it is a shallow copy, we should check if
 - the contained elements and attributes are the same and 
 - reflect modification in the other handle 

(as well as in all other objects with this behaviour).

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+
+        self.assertIsInstance(copy_iteration, openPMD.Iteration)
+
+    def testIteration_Container(self):
+        """ Test openPMD.Iteration_Container. """
+        obj = openPMD.Iteration_Container()
```
anokfireball: This is a mis-design in the C++ API. ```Container``` should not have a publicly available constructor. #280

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
...
+        # copy_record = openPMD.Record(record)
+
+        # Check.
+        # self.assertIsInstance(copy_record, openPMD.Record)
+
+    def testRecord_Component(self):
+        """ Test openPMD.Record_Component. """
+        self.assertRaises(TypeError, openPMD.Record_Component)
+
+    def testFieldRecord(self):
```
anokfireball, line 312: A ```Mesh``` is a ```Record``` by [design](https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#mesh-based-records). What this actually tests are ```Mesh_Record_Component```s.

test/python/unittest/TestUtilities/TestUtilities.py
```diff
@@ -0,0 +1,31 @@
...
+def generateTestFilePath(file_name):
+    """
+    Returns the absolute path to a test file located in ../TestFiles.
+
+    @param file_name : The name of the file in ../TestFiles.
+    @return : The absolute path to ../TestFiles/<file_name> .
+    """
+
+    test_files_dir = path.join('../samples/', file_name)
```
anokfireball: Hard-coded directory separators defeat the purpose of using ```os.path.join```.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,324 @@
+"""
+This file is part of the openPMD-api.
+
+Copyright 2018 openPMD contributors
+Authors: Axel Huebl, Carsten Fortmann-Grote
+License: LGPLv3+
+"""
+
+import openPMD
```
anokfireball, line 9: This should be moved to the third-party imports.

anokfireball (changes requested):
I noted a lot of stuff. Do not despair, you don't need to do all of it.
The annotations are there so any leftovers can be done in a follow-up.

The iterationFormat regex and the ```print()``` noise are a blocker, though.

test/python/unittest/API/APITest.py
anokfireball: This should be fine as the Series is opened with ```openPMD.Access_Type.read_only```.

test/python/unittest/API/APITest.py
ax3l: In python I throw on that as soon as the series is read_only. That's also the place where it originally triggered me that we need an asset in C++ as well :)

ax3l:
@C0nsultant I tried to address a whole bunch of review comments, but could not yet fix all.
`print()`s are removed, paths fixed and more verifies added. If you like, we could merge the current state and improve it ourselves further?

test/python/unittest/API/APITest.py
ax3l: I have to check, I hope it's a reference in the first access and a shallow-copy / alias in the second case.

ax3l: That's not an ideal place to check, here it uses a readonly series which means I can't modify the iteration for proper inspection :S

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,322 @@
...
+
+import os
+import shutil
+import unittest
+try:
+    import numpy as np
+    found_numpy = True
+except ImportError:
+    found_numpy = False
```
anokfireball, line 18: Ah yes, a simple runtime ```try```-```except``` to turn off features instead of CMake magic with ```#ifdef``` conditional compilation. :slightly_smiling_face:

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,322 @@
...
+    def setUp(self):
+        """ Setting up a test. """
+        self.__files_to_remove = []
+        self.__dirs_to_remove = []
+
+        path_to_field_data = generateTestFilePath(
+                "issue-sample/no_particles/data%T.h5")
+        path_to_particle_data = generateTestFilePath(
+                "issue-sample/no_fields/data%T.h5")
+        path_to_data = generateTestFilePath("git-sample/data%T.h5")
```
anokfireball, line 45: Ideally, these paths would be created with ```os.path.join``` as well.

test/python/unittest/API/APITest.py
```diff
@@ -0,0 +1,322 @@
...
+
+        # Check entries.
+        self.assertEqual(len(i.meshes), 2)
+        self.assertEqual(len(i.particles), 0)
+
+    def testParticleData(self):
+        """ Testing serial IO on a pure particle dataset. """
+
+        # Get reference to series stored on test case.
+        series = self.__field_series
```
anokfireball, line 91: This does not use the pure particle dataset referred to in the docstring.

anokfireball (approved):

anokfireball:
Thank you very much, @CFGrote!

ax3l:
What could possibly go wrong... ;)

test/python/unittest/API/APITest.py
ax3l: What could possibly go wrong... ;)

PR #250, opened 2018-06-12
Travis HDF5: 1.10.1

### LINKED ISSUE #248: Travis CI: MPI
I don't know what happened to the package installed trusty openmpi 1.6.5 since I last deleted the caches, but it does not look healthy... :confused: 

https://travis-ci.org/openPMD/openPMD-api/jobs/391288806

ax3l:
Stay with HDF5 1.10.1 instead of 1.10.2 on Travis.

Building parallel HDF5 itself fails with

```
H5Dmpio.c:3027:17: error: use of undeclared identifier 'MPI_Message'
     1064                    MPI_Message message;
```

in the latest HDF5 release.

Close #248

### DIFF
--- a/.travis/spack/packages.yaml
+++ b/.travis/spack/packages.yaml
@@ -16,6 +16,8 @@ packages:
       openmpi@1.6.5%gcc@7.2.0 arch=linux-ubuntu14-x86_64: /usr
       openmpi@1.6.5%clang@5.0.0 arch=linux-ubuntu14-x86_64: /usr
     buildable: False
+  hdf5:
+    version: [1.10.1, 1.8.13]
   python:
     version: [2.7.14, 2.7.10, 3.6.3]
     paths:


anokfireball:
Do we currently benefit in any way from 1.10.2?

anokfireball (approved):

ax3l:
Dunno. We didn't need it yet and the problem seems to be HDF5 deployment, so it's an upstream deployment issue that does not hit us. Of course, as nice people we should reach out...

Edit: [did my duty](https://twitter.com/axccl/status/1006644516072382464)

ax3l:
Yeah, bug. They use a [MPI-3 feature](https://twitter.com/science_dot/status/1006644951151607808) and don't announce they need it.

(MPI-3+ is cool, just OpenMPI 1.6.5 that we by coincidence use on trusty is old and provides some MPI 1.3? / 2.x standard.)

PR #251, opened 2018-06-12
Docs: MPI Version

ax3l:
We currently even test on OpenMPI 1.6.5 which only provides MPI 2.1

(Also, MPI 2.3 was a typo. There is only ... 1.3, 2.0, 2.1, 2.2, 3.0, 3.1. Hm, I initially learned the brand-new 1.3 standard back in the days.)

Of course we can require something more recent anytime we need it. This just means we have to compile it or use the mpich2 libs on trusty (or better: both, in order to continue testing the oldest claimed supported version as well).

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Optional I/O backends:
 * [ADIOS2](https://github.com/ornladios/ADIOS2) 2.1+ (*not yet implemented*)
 
 while those can be build either with or without:
-* MPI 2.3+, e.g. OpenMPI or MPICH2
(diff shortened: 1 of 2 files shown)


anokfireball (approved):
> This just means we have to compile it or use the mpich2 libs on trusty (or better: both, in order to continue testing the oldest claimed supported version as well)

In the spirit of #215, I would assume it would be sufficient if we test the minimum version we list (in the current case ```OpenMPI 1.6.5```) in one build and pull a default that satisfies the dependency for all other builds.

PR #254, opened 2018-06-13
ADIOS1: Link Serial & Parallel Separately

### LINKED ISSUE #252: ADIOS1: SerialIOTest needs MPI with Parallel Lib
Even with #228 there are way more calls to `MPI_` functions when linking a parallel ADIOS1 lib. That leads to the need to actually start serial programs, such as `SerialIOTest`, with `mpirun/exec`.

That's sad. Is the MPI-moc library the only way to call the serial ADIOS API? If so, a lot more refactoring inside ADIOS [`src/core/bp_utils.c`](https://github.com/ornladios/ADIOS/blob/v1.13.1/src/core/bp_utils.c) (incomplete below) and a little more inside `src/read/read_bp.c` would be needed:

```diff
diff --git a/src/core/bp_utils.c b/src/core/bp_utils.c
index 1ac8990a..94ef1532 100644
--- a/src/core/bp_utils.c
+++ b/src/core/bp_utils.c
@@ -70,6 +70,78 @@ inline BP_FILE * GET_BP_FILE (const ADIOS_FILE * fp)
 void * bp_read_data_from_buffer(struct adios_bp_buffer_struct_v1 *b, enum ADIOS_DATATYPES type, int nelems);
 int bp_parse_characteristics (struct adios_bp_buffer_struct_v1 * b, struct adios_index_var_struct_v1 ** root, uint64_t j);
 
+/* dummies: for building serial apps against a parallel ADIOS lib */
+#if defined(__APPLE__) || defined(__WIN32__) || defined(__CYGWIN__) 
+#    define lseek64 lseek
+#    define open64  open
+#endif
+
+static char mpierrmsg_dummy[MPI_MAX_ERROR_STRING];
+
+static int typesize_dummy(MPI_Datatype type)
+{
+    switch( type )
+    {
+      case MPI_INT : return sizeof( int );
+                     break;
+      case MPI_CHAR: return sizeof( char );
+                     break;
+      case MPI_REAL : return sizeof( float );
+                     break;
+      case MPI_DOUBLE : return sizeof( double );
+                     break;
+      case MPI_UNSIGNED_LONG_LONG : return sizeof( unsigned long long );
+                     break;
+      case MPI_UINT64_T : return sizeof( uint64_t );
+                     break;
+      default      : return 1 ;
+    }
+    return 1;
+}
+
+int MPI_File_open_dummy(MPI_Comm comm, char *filename, int amode, MPI_Info info, MPI_File *fh) 
+{
+    *fh = open64 (filename, amode);
+    if (*fh == -1) {
+        snprintf(mpierrmsg_dummy, MPI_MAX_ERROR_STRING, "File not found: %s", filename);
+        return -1;
+    }
+    return MPI_SUCCESS;
+}
+
+int MPI_File_close_dummy(MPI_File *fh) { return close(*fh); }
+
+int MPI_File_get_size_dummy(MPI_File fh, MPI_Offset *size) {
+    uint64_t curpos = lseek64(fh, 0, SEEK_CUR); // get the current seek pos
+    uint64_t endpos = lseek64(fh, 0, SEEK_END); // go to end, returned is the size in bytes
+    lseek64(fh, curpos, SEEK_SET);             // go back where we were
+    *size = (MPI_Offset) endpos;
+    //printf("MPI_File_get_size: fh=%d, size=%lld\n", fh, *size);
+    return MPI_SUCCESS;
+}
+
+int MPI_File_read_dummy(MPI_File fh, void *buf, int count, MPI_Datatype datatype, MPI_Status *status)
+{
+    // FIXME: int count can read only 2GB (*datatype size) array at max
+    uint64_t bytes_to_read = count * typesize_dummy(datatype);
+    uint64_t bytes_read;
+    bytes_read = read (fh, buf, bytes_to_read);
+    if (bytes_read != bytes_to_read) {
+        snprintf(mpierrmsg_dummy, MPI_MAX_ERROR_STRING, "could not read %" PRIu64 " bytes. read only: %" PRIu64 "\n", bytes_to_read, bytes_read);
+        return -2;
+    }
+    *status = bytes_read;
+    //printf("MPI_File_read: fh=%d, count=%d, typesize=%d, bytes read=%lld\n", fh, count, datatype, *status);
+    return MPI_SUCCESS;
+}
+
+int MPI_File_seek_dummy(MPI_File fh, MPI_Offset offset, int whence)
+{
+    uint64_t off = (uint64_t) offset;
+    lseek64 (fh, off, whence);
+    //printf("MPI_File_seek: fh=%d, offset=%lld, whence=%d\n", fh, off, whence);
+    return MPI_SUCCESS;
+}
 
 
 void bp_alloc_aligned (struct adios_bp_buffer_struct_v1 * b, uint64_t size)
@@ -233,18 +305,23 @@ static int bp_read_open (const char * filename,
           MPI_Comm comm,
           BP_FILE * fh)
 {
-    int  err;
-    int  rank;
+    int  err = MPI_SUCCESS;
+    int  rank = 0;
 
-    MPI_Comm_rank (comm, &rank);
+    if (comm != MPI_COMM_NULL)
+        MPI_Comm_rank (comm, &rank);
 
     // variable definition
     MPI_Offset  file_size;
 
     // open a file by the multiple processors within the same
     // communicator
-    err = MPI_File_open (comm, (char *) filename, MPI_MODE_RDONLY,
-            (MPI_Info) MPI_INFO_NULL, &(fh->mpi_fh));
+    if (comm == MPI_COMM_NULL)
+        err = MPI_File_open_dummy (comm, (char *) filename, MPI_MODE_RDONLY,
+                (MPI_Info) MPI_INFO_NULL, &(fh->mpi_fh));
+    else
+        err = MPI_File_open (comm, (char *) filename, MPI_MODE_RDONLY,
+                (MPI_Info) MPI_INFO_NULL, &(fh->mpi_fh));
     if (err != MPI_SUCCESS) {
         char e [MPI_MAX_ERROR_STRING];
         int len = 0;
@@ -254,7 +331,10 @@ static int bp_read_open (const char * filename,
         return adios_flag_no;
     }
 
-    MPI_File_get_size (fh->mpi_fh, &file_size);
+    if (comm == MPI_COMM_NULL)
+        MPI_File_get_size_dummy (fh->mpi_fh, &file_size);
+    else
+        MPI_File_get_size (fh->mpi_fh, &file_size);
     fh->b->file_size = file_size;
     fh->mfooter.file_size = file_size;
 
@@ -265,26 +345,43 @@ static int bp_read_open_rootonly (const char * filename,
           MPI_Comm comm,
           BP_FILE * fh)
 {
-    int  err;
-    int  rank;
+    int  err = MPI_SUCCESS;
+    int  rank = 0;
 
-    MPI_Comm_rank (comm, &rank);
+    if (comm != MPI_COMM_NULL)
+        MPI_Comm_rank (comm, &rank);
 
     // variable definition
     MPI_Offset  file_size = 0;
 
     if (rank == 0)
     {
-        // open a file by one processor then broadcast the size and the success
-        err = MPI_File_open (MPI_COMM_SELF, (char *) filename, MPI_MODE_RDONLY,
-                (MPI_Info) MPI_INFO_NULL, &(fh->mpi_fh));
-        if (err == MPI_SUCCESS) {
-            MPI_File_get_size (fh->mpi_fh, &file_size);
-            err = 0;
+        if (comm == MPI_COMM_NULL)
+        {
+            // open a file by one processor then broadcast the size and the success
+            err = MPI_File_open_dummy (MPI_COMM_SELF, (char *) filename, MPI_MODE_RDONLY,
+                    (MPI_Info) MPI_INFO_NULL, &(fh->mpi_fh));
+            if (err == MPI_SUCCESS) {
+                MPI_File_get_size_dummy (fh->mpi_fh, &file_size);
+                err = 0;
+            }
+        }
+        else
+        {
+            // open a file by one processor then broadcast the size and the success
+            err = MPI_File_open (MPI_COMM_SELF, (char *) filename, MPI_MODE_RDONLY,
+                    (MPI_Info) MPI_INFO_NULL, &(fh->mpi_fh));
+            if (err == MPI_SUCCESS) {
+                MPI_File_get_size (fh->mpi_fh, &file_size);
+                err = 0;
+            }
         }
     }
-    MPI_Bcast (&err, 1, MPI_INT, 0, comm);
-    MPI_Bcast (&file_size, 1, MPI_UNSIGNED_LONG_LONG, 0, comm);
+    if (comm != MPI_COMM_NULL)
+    {
+        MPI_Bcast (&err, 1, MPI_INT, 0, comm);
+        MPI_Bcast (&file_size, 1, MPI_UNSIGNED_LONG_LONG, 0, comm);
+    }
     fh->b->file_size = file_size;
     fh->mfooter.file_size = file_size;
     if (err)
@@ -305,9 +402,10 @@ int bp_open (const char * fname,
              MPI_Comm comm,
              BP_FILE * fh)
 {
-    int rank;
+    int rank = 0;
 
-    MPI_Comm_rank (comm, &rank);
+    if (comm != MPI_COMM_NULL)
+        MPI_Comm_rank (comm, &rank);
 
     adios_buffer_struct_init (fh->b);
 
@@ -326,7 +424,8 @@ int bp_open (const char * fname,
     }
 
     /* Broadcast to all other processors */
-    MPI_Bcast (&fh->mfooter, sizeof (struct bp_minifooter), MPI_BYTE, 0, comm);
+    if (comm != MPI_COMM_NULL)
+        MPI_Bcast (&fh->mfooter, sizeof (struct bp_minifooter), MPI_BYTE, 0, comm);
 
     if (fh->mfooter.pgs_index_offset > 0)
     {
@@ -336,7 +435,12 @@ int bp_open (const char * fname,
          * secure for future if a metadata+data BP file will ever have subfiles.
          */
         if (rank == 0)
-            MPI_File_close(&fh->mpi_fh);
+        {
+            if (comm == MPI_COMM_NULL)
+                MPI_File_close_dummy(&fh->mpi_fh);
+            else
+                MPI_File_close(&fh->mpi_fh);
+        }
         if (bp_read_open (fname, comm, fh))
         {
             return -1;
@@ -357,7 +461,8 @@ int bp_open (const char * fname,
         }
     }
 
-    MPI_Barrier (comm);
+    if (comm != MPI_COMM_NULL)
+        MPI_Barrier (comm);
     // Broadcast the index which may be bigger than 2GB, so do it in chunks
     uint64_t bytes_sent = 0;
     int32_t to_send = 0;
@@ -373,8 +478,8 @@ int bp_open (const char * fname,
         {
             to_send = footer_size - bytes_sent;
         }
-
-        MPI_Bcast (fh->b->buff + bytes_sent, to_send, MPI_BYTE, 0, comm);
+        if (comm != MPI_COMM_NULL)
+            MPI_Bcast (fh->b->buff + bytes_sent, to_send, MPI_BYTE, 0, comm);
         bytes_sent += to_send;
     }
 
@@ -604,8 +709,13 @@ int bp_close (BP_FILE * fh)
     MPI_File mpi_fh = fh->mpi_fh;
 
     adios_errno = 0;
-    if (fh->mpi_fh != MPI_FILE_NULL)
-        MPI_File_close (&mpi_fh);
+    if (mpi_fh != MPI_FILE_NULL)
+    {
+        if (fh->comm == MPI_COMM_NULL)
+            MPI_File_close_dummy (&mpi_fh);
+        else
+            MPI_File_close (&mpi_fh);
+    }
 
     close_all_BP_subfiles (fh);
 
@@ -820,8 +930,16 @@ int bp_read_minifooter (BP_FILE * bp_struct)
         memset (b->buff, 0, MINIFOOTER_SIZE);
         b->offset = 0;
     }
-    MPI_File_seek (bp_struct->mpi_fh, (MPI_Offset) attrs_end, MPI_SEEK_SET);
-    MPI_File_read (bp_struct->mpi_fh, b->buff, MINIFOOTER_SIZE, MPI_BYTE, &status);
+    if (bp_struct->comm == MPI_FILE_NULL)
+    {
+        MPI_File_seek_dummy (bp_struct->mpi_fh, (MPI_Offset) attrs_end, MPI_SEEK_SET);
+        MPI_File_read_dummy (bp_struct->mpi_fh, b->buff, MINIFOOTER_SIZE, MPI_BYTE, &status);
+    }
+    else
+    {
+        MPI_File_seek (bp_struct->mpi_fh, (MPI_Offset) attrs_end, MPI_SEEK_SET);
+        MPI_File_read (bp_struct->mpi_fh, b->buff, MINIFOOTER_SIZE, MPI_BYTE, &status);
+    }
 
     /*memset (&mh->pgs_index_offset, 0, MINIFOOTER_SIZE);
     memcpy (&mh->pgs_index_offset, b->buff, MINIFOOTER_SIZE);*/
@@ -900,9 +1018,14 @@ int bp_read_minifooter (BP_FILE * bp_struct)
     /* It will be sent to all processes */
     uint64_t footer_size = mh->file_size - mh->pgs_index_offset;
     bp_realloc_aligned (b, footer_size);
-    MPI_File_seek (bp_struct->mpi_fh,
-            (MPI_Offset)  mh->pgs_index_offset,
-            MPI_SEEK_SET);
+    if (bp_struct->comm == MPI_FILE_NULL)
+        MPI_File_seek_dummy (bp_struct->mpi_fh,
+                (MPI_Offset)  mh->pgs_index_offset,
+                MPI_SEEK_SET);
+    else
+        MPI_File_seek (bp_struct->mpi_fh,
+                (MPI_Offset)  mh->pgs_index_offset,
+                MPI_SEEK_SET);
 
     // if we need to read > 2 GB, need to do it in parts
     // since count is limited to MAX_MPIWRITE_SIZE (signed 32-bit max).
@@ -921,7 +1044,10 @@ int bp_read_minifooter (BP_FILE * bp_struct)
             to_read = footer_size - bytes_read;
         }
 
-        err = MPI_File_read (bp_struct->mpi_fh, b->buff+bytes_read, to_read, MPI_BYTE, &status); 
+        if (bp_struct->comm == MPI_FILE_NULL)
+            err = MPI_File_read_dummy (bp_struct->mpi_fh, b->buff+bytes_read, to_read, MPI_BYTE, &status); 
+        else
+            err = MPI_File_read (bp_struct->mpi_fh, b->buff+bytes_read, to_read, MPI_BYTE, &status); 
         if (err != MPI_SUCCESS) 
         {
             char e [MPI_MAX_ERROR_STRING];
diff --git a/src/read/read_bp.c b/src/read/read_bp.c
index e3c9a42f..2b4a59f5 100644
--- a/src/read/read_bp.c
+++ b/src/read/read_bp.c
@@ -1721,14 +1721,15 @@ ADIOS_FILE * adios_read_bp_open (const char * fname, MPI_Comm comm, enum ADIOS_L
 
 ADIOS_FILE * adios_read_bp_open_file (const char * fname, MPI_Comm comm)
 {
-    int rank;
+    int rank = 0;
     BP_PROC * p;
     BP_FILE * fh;
     ADIOS_FILE * fp;
 
     log_debug ("adios_read_bp_open_file\n");
 
-    MPI_Comm_rank (comm, &rank);
+    if (comm != MPI_COMM_NULL)
+        MPI_Comm_rank (comm, &rank);
 
     fh = BP_FILE_alloc (fname, comm);
```

ax3l:
CMake draft to link the `_nompi` and MPI libs of ADIOS1 separately.

Close #252 

## Current Status 1 

It's ugly, it's verbose, it's hard to maintain, it excludes compilers that can not hide symbols such as pgc++ when building with ADIOS1, the system must have shared library support, it requires static ADIOS1 libraries (at least that's the default), ... but it works :-D

## Current Status 2

It's complex but doable.

## To Do

- [x] inline comments, clean up
- [ ] requires serial ADIOS library? (can't test: ADIOS 1.13.1 seems just to build to static...) -> reported in https://github.com/ornladios/ADIOS/issues/185
- [x] `FindADIOS.cmake` should honor `QUIET` in its `message()`s
- [x] ~~add CMake option for SHARED adios intermediate libs (default: on == MPI/Serial safe)?~~
- [x] clean up [serial interface](https://github.com/openPMD/openPMD-api/issues/252#issuecomment-396907630): @C0nsultant ~heeelp please :)~ #258 
- [x] very likely needs explicit symbol (un)hiding in ADIOS translation units
- [x] do the same symbol linker magic on OSX
- [x] Win has no ADIOS1 support
- [x] we can not inherit, because same symbols do not mean same implementation in the end
- [x] ~~[`objcopy --strip-symbols`](https://stackoverflow.com/a/42824356/2719194)?~~
- [x] ~~maybe we need to either call `MPI_Initialize()` (if MPI context missing) if ADIOS is build with MPI even in serial mode... or load ADIOS via `dlopen` to avoid symbol clashes/mismatches, etc.~~


CMakeLists.txt
```diff
@@ -119,7 +119,18 @@ if(openPMD_HAVE_HDF5 AND HDF5_IS_PARALLEL AND NOT openPMD_HAVE_MPI)
...
 
-# external library: ADIOS1 (optional)
+#   always search for a sequential lib first, so we can mock MPI
+find_package(ADIOS 1.13.1 COMPONENTS sequential QUIET)
+set(ADIOS_DEFINITIONS_SEQUENTIAL ${ADIOS_DEFINITIONS})
+set(ADIOS_LIBRARIES_SEQUENTIAL ${ADIOS_LIBRARIES})
+set(ADIOS_INCLUDE_DIRS_SEQUENTIAL ${ADIOS_INCLUDE_DIRS})
+unset(ADIOS_FOUND CACHE)
+unset(ADIOS_DEFINITIONS CACHE)
+unset(ADIOS_LIBRARIES CACHE)
```
ax3l: attention: still contains serial libs... not fully cleared!

CMakeLists.txt
```diff
@@ -277,10 +296,99 @@ else()
...
+    target_compile_features(openPMD.ADIOS1.Serial
+        PUBLIC cxx_std_11
+    )
+    set_target_properties(openPMD.ADIOS1.Serial PROPERTIES
+        CXX_EXTENSIONS OFF
+        CXX_STANDARD_REQUIRED ON
+        POSITION_INDEPENDENT_CODE ON
+        CXX_VISIBILITY_PRESET hidden
+        VISIBILITY_INLINES_HIDDEN ON
+        LINK_FLAGS "-Wl,--exclude-libs,libadios_nompi.a"
```
ax3l: or simply `-Wl,--exclude-libs,ALL`

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp
```diff
@@ -30,13 +30,19 @@
 #include <memory>
 #include <string>
 
+#if _MSC_VER
```
ax3l: define and included in `AbstractIOHandler.hpp` already

CMakeLists.txt
```diff
@@ -277,10 +296,99 @@ else()
 endif()
 
 if(openPMD_HAVE_ADIOS1)
-    target_link_libraries(openPMD PUBLIC ${ADIOS_LIBRARIES})
-    target_include_directories(openPMD SYSTEM PUBLIC ${ADIOS_INCLUDE_DIRS})
+    add_library(openPMD.ADIOS1.Serial SHARED ${IO_ADIOS1_SEQUENTIAL_SOURCE})
```
ax3l: ha, have to link this always xD

ax3l:
@C0nsultant lol, I made it. That's the worst hack to isolate an upstream issue I ever did. I feel ashamed and a bit used :face_with_head_bandage:

CMakeLists.txt
```diff
@@ -119,7 +119,15 @@ if(openPMD_HAVE_HDF5 AND HDF5_IS_PARALLEL AND NOT openPMD_HAVE_MPI)
         "to disable HDF5 or provide a serial install of HDF5.")
 endif()
 
-# external library: ADIOS1 (optional)
+#   always search for a sequential lib first, so we can mock MPI
+find_package(ADIOS 1.13.1 COMPONENTS sequential QUIET)
```
ax3l, line 136: todo: the QUIET should be honored in `FindADIOS.cmake`

CMakeLists.txt
```diff
@@ -119,7 +119,15 @@ if(openPMD_HAVE_HDF5 AND HDF5_IS_PARALLEL AND NOT openPMD_HAVE_MPI)
...
+set(ADIOS_DEFINITIONS_SEQUENTIAL ${ADIOS_DEFINITIONS})
+set(ADIOS_LIBRARIES_SEQUENTIAL ${ADIOS_LIBRARIES})
+set(ADIOS_INCLUDE_DIRS_SEQUENTIAL ${ADIOS_INCLUDE_DIRS})
+unset(ADIOS_FOUND CACHE)
+unset(ADIOS_VERSION CACHE)
```
ax3l, line 141: we could use this and look for the parallel component with `EXACT`.
but actually it's not two, totally independent backends anyway.
nevertheless, same version makes life easier.

anokfireball:
@ax3l You survived the war. Undoubtedly you have witnessed the bad and the worse, but you made it to the other side. People are going to love you for it.

Great work!

CMakeLists.txt
```diff
@@ -276,15 +289,95 @@ else()
     target_compile_definitions(openPMD PUBLIC "-DopenPMD_HAVE_HDF5=0")
 endif()
 
+add_library(openPMD.ADIOS1.Serial SHARED ${IO_ADIOS1_SEQUENTIAL_SOURCE})
+add_library(openPMD.ADIOS1.Parallel SHARED ${IO_ADIOS1_SOURCE})
```
anokfireball: Just for better understanding:
Are all 4 combinations
```
SER PAR
 N   N
 N   Y
 Y   N
 Y   Y
```
possible with this change?

ax3l: Yes indeed :dancer: 

Even though an ADIOS install of the form
```
SER PAR
 N   Y
```

is not build by ADIOS (it always ships at least a serial lib with its lovely mock magic).

anokfireball: Impressive!
... and that only requires building two seperate library files by compiling with different options and linking them with tinkered symbols.

> its lovely mock magic

:joy:

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp
```diff
@@ -22,39 +22,80 @@
...
+
+    virtual int64_t open_write(Writable *);
+    virtual ADIOS_FILE* open_read(std::string const& name);
+    void close(int64_t);
+    void close(ADIOS_FILE*);
+
+protected:
+    int64_t m_group;
+    std::string m_groupName;
+    ADIOS_READ_METHOD m_readMethod;
```
anokfireball: If we don't use that varaible for both backends anymore, we cna probably hard-code the ```ADIOS_READ_METHOD``` used in each one.

CMakeLists.txt
ax3l: Close: no symbol tinkering, just really aggressive symbols hiding for encapsulation :smiling_imp: 

And note the weird code sharing in the implementation files :-o

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -314,6 +316,11 @@ ParallelADIOS1IOHandlerImpl::open_read(std::string const& name)
 
     return f;
 }
+
+#define ADIOS1IOHandlerImpl ParallelADIOS1IOHandlerImpl
+#include "ADIOS1IOCommon.cpp"
+#undef ADIOS1IOHandlerImpl
```
anokfireball: When you *think* you've seen all the ways of sharing code between classes...

ax3l: yeah, that's my secret present for you.
did I mention it's the *worst hack* ever?

ax3l: Note that inheritance does not work since *identical symbols do not mean identical implementation* thanks to the MPI mock...

ax3l: I could inherit from yet another baseclass though, that we then hide again in symbols and compile twice... ahhhh... it must stop.

anokfireball: The reason for the duplicated code is not too diffcult to understand. Just maybe replace ```ADIOS1IOHandlerImpl``` with ```CommonImpl``` in ```"ADIOS1IOCommon.cpp"``` and then do the ```#define```s in both handlers?

ax3l: Yeah, thought so, too!

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp
ax3l: I don't fully get this but I guess we can do this in a follow-up as well.

anokfireball: The methods are initialized once per handler in ```init()```
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/src/IO/ADIOS/ADIOS1IOHandler.cpp#L210
https://github.com/openPMD/openPMD-api/blob/551ab6df4c3ad27b5cbee7cbf9db888fd503d37c/src/IO/ADIOS/ParallelADIOS1IOHandler.cpp#L221-L222
With the hard separation of serial and parallel backend, we do not really need the variable any more.

ax3l:
@C0nsultant for proper linking and header separation, I used your already excellent abstraction in the `AbstractIOHandler` and `Impl` files and separated it now rigorously into independent headers for the various translation units. (Otherwise we would have cyclic dependencies between the various IO backends.) Sorry, it get's really big by now ;)

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp
```diff
@@ -22,79 +22,65 @@
...
-class ParallelADIOS1IOHandler : public AbstractIOHandler
+namespace openPMD
 {
-    friend class ParallelADIOS1IOHandlerImpl;
+    class EXPORT ParallelADIOS1IOHandlerImpl;
 
-public:
-#if openPMD_HAVE_MPI
-    ParallelADIOS1IOHandler(std::string const& path, AccessType, MPI_Comm);
+#if openPMD_HAVE_ADIOS1 && openPMD_HAVE_MPI
```
anokfireball: This check is redundant with the ```# if openPMD_HAVE_MPI``` below.

anokfireball:
Separating the two is probably a good thing.
Using the ```Impl``` is only the recommended way to use backends, not a constraint. I placed them in the same files up to this point since all backends implemented their own ```Impl```s.

ax3l:
@C0nsultant I think we could merge this as well now. It's also hard to rebase if we keep this open for too long.

CMakeLists.txt
ax3l: done

CMakeLists.txt
```diff
@@ -218,11 +252,22 @@ set(CORE_SOURCE
...
         src/IO/IOTask.cpp
-        src/IO/ADIOS/ADIOS1IOHandler.cpp
-        src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
         src/IO/HDF5/HDF5IOHandler.cpp
         src/IO/HDF5/ParallelHDF5IOHandler.cpp)
+set(IO_ADIOS1_SEQUENTIAL_SOURCE
+        src/IO/AbstractIOHandler.cpp
+        src/IO/AbstractIOHandlerImpl.cpp
+        src/auxiliary/Filesystem.cpp
+#        src/IO/IOTask.cpp
```
anokfireball: Remove if not required to build this library.

CMakeLists.txt
```diff
@@ -218,11 +252,22 @@ set(CORE_SOURCE
...
+        src/IO/ADIOS/ADIOS1IOHandler.cpp)
+set(IO_ADIOS1_SOURCE
+        src/IO/AbstractIOHandler.cpp
+        src/IO/AbstractIOHandlerImpl.cpp
+        src/auxiliary/Filesystem.cpp
+#        src/IO/IOTask.cpp
```
anokfireball: Remove if not required to build this library.

CMakeLists.txt
```diff
@@ -292,22 +338,113 @@ else()
...
+    target_compile_definitions(openPMD.ADIOS1.Serial PRIVATE "-DopenPMD_HAVE_ADIOS1=1")
+    target_compile_definitions(openPMD.ADIOS1.Serial PRIVATE "-DopenPMD_HAVE_MPI=0")
+    target_compile_definitions(openPMD.ADIOS1.Serial PRIVATE "-D_NOMPI=1")
+
+    if(openPMD_HAVE_MPI)
+        set_target_properties(openPMD.ADIOS1.Parallel PROPERTIES
+            CXX_EXTENSIONS OFF
+            CXX_STANDARD_REQUIRED ON
+            POSITION_INDEPENDENT_CODE ON
+            CXX_VISIBILITY_PRESET "hidden"
```
anokfireball: Is there a reasong why a string is used for ```CXX_VISIBILITY_PRESET``` here and a raw literal in ```openPMD.ADIOS1.Serial```?

share/openPMD/cmake/FindADIOS.cmake
```diff
@@ -107,7 +107,9 @@ if(ADIOS_FIND_COMPONENTS)
         elseif(comp STREQUAL "sequential")
             set(OPTLIST "${OPTLIST}s")
         else()
-            message("ADIOS component ${COMP} is not supported. Please use fortran, readonly, or sequential")
+            if(NOT ADIOS_FIND_QUIETLY)
+                message("ADIOS component ${COMP} is not supported. Please use fortran, readonly, or sequential")
```
anokfireball, line 111: This is potentially a case where we could signal ```FATAL_ERROR```.
The user is requesting a component we will never be able to provide.

anokfireball (approved):
LGTM.

CMakeLists.txt
ax3l: That's unfortunately syntax in CMake to set `-fvisibility-inlines-hidden` (or equivalent)

share/openPMD/cmake/FindADIOS.cmake
ax3l: Oh, quite the contrary! We are inside a `Find*.cmake` script. The worst thing it is allowed to do is *not find* what is requested.

Fatal errors are controlled by setting `REQUIRED` in `find_package` or by handling the `Package_FOUND` variable manually.

So we are all set :-)

CMakeLists.txt
ax3l: Ah, now I see. yes, we can skip the quotes. Updated.

Issue #255, opened 2018-06-13
Setters/Getters of Attributable


ax3l:
It's cool that a `Container` itself is *attributable*, but why are attributes not simply implemented as `Container` themselves as well?

I think similar to `h5py`'s [`AttributeManager`](http://docs.h5py.org/en/latest/high/dataset.html#creating-and-reading-empty-or-null-datasets-and-attributes), using our container both for data sets and attributes would make it very natural to interact with them. Or am I missing something?

anokfireball:
I frankly do not understand your question or the motivation behind it.

**DISCLAIMER** - Attributes were one of the first features. Please excuse design flaws and antipatterns.
Let me elaborate a bit on the choices concerning attributes:
There's three major types of abstract objects in this API:
  1. Datasets (i.e. ```RecordComponent```)
  2. Groups with arbitrary child-nodes (i.e. ```Mesh```), all implementing the [```Container```](https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/include/openPMD/backend/Container.hpp#L52-L65) interface
  3. Groups with well-defined child-nodes (i.e. ```Iteration```)

All of these can be attributed with arbitrary key-value pairs (with [```Attributable::setAttribute()```](https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/include/openPMD/backend/Attributable.hpp#L90-L101)).
Some of them additionally carry pairs that are restricted by openPMD (e.g. [```Series::setOpenPMD()```](https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/include/openPMD/Series.hpp#L65-L70)).

To have management of attributes centralized, the [```Attributable```](https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/include/openPMD/backend/Attributable.hpp#L57-L64) mixin was created. Essentially it boils down to a map that stores key-value pairs with ```std::string``` keys and [```Attribute```](https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/include/openPMD/backend/Attribute.hpp#L39-L44) values. The attributes may be of any [```Datatype```](https://github.com/openPMD/openPMD-api/blob/0.2.0-alpha/include/openPMD/Datatype.hpp#L33-L61) mentioned in openPMD and supported by major backends.

[...]

PR #256, opened 2018-06-13
Python: Attribute Keys

ax3l:
Expose attributes of containers to Python.

Realized #255 while doing so.

### DIFF
--- a/src/binding/python/Container.cpp
+++ b/src/binding/python/Container.cpp
@@ -32,11 +32,13 @@
 #include "openPMD/Iteration.hpp"
 #include "openPMD/Mesh.hpp"
 #include "openPMD/ParticleSpecies.hpp"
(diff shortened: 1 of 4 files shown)


ax3l:
@C0nsultant ping :)

src/binding/python/Container.cpp
```diff
@@ -141,6 +143,16 @@ namespace detail
             >()
         );
 
+        cl.def_property_readonly(
```
anokfireball, line 146: Is this your inspiration for #255?

anokfireball (approved):

src/binding/python/Container.cpp
ax3l: yep, can expose it in python easily as property behind the scenes later on, but in C++ #255 we could make this already pretty

PR #257, opened 2018-06-13
Write Examples: Scalar Record Name

ax3l:
Rename the scalar mesh records written in the examples. Use "rho" which is used for scalar observables such as densities instead of "E" which is used for vector fields.

### DIFF
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -47,10 +47,10 @@ int main(int argc, char *argv[])
     );
     cout << "Created an empty " << series.iterationEncoding() << " Series\n";
 
-    MeshRecordComponent E =
+    MeshRecordComponent rho =
       series
           .iterations[1]
-          .meshes["E"][MeshRecordComponent::SCALAR];
+          .meshes["rho"][MeshRecordComponent::SCALAR];
     cout << "Created a scalar mesh Record with all required openPMD attributes\n";
 
     Datatype datatype = determineDatatype(shareRaw(global_data));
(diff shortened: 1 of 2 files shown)


examples/3_write_serial.cpp
```diff
@@ -59,14 +59,14 @@ int main(int argc, char *argv[])
     cout << "Created a Dataset of size " << dataset.extent[0] << 'x' << dataset.extent[1]
          << " and Datatype " << dataset.dtype << '\n';
 
-    E.resetDataset(dataset);
+    rho.resetDataset(dataset);
     cout << "Set the dataset properties for the scalar field E in iteration 1\n";
```
anokfireball: Output string should be adjusted accordingly.

examples/3_write_serial.py
```diff
@@ -39,7 +39,7 @@
     print("Created a Dataset of size {0}x{1} and Datatype {2}".format(
         dataset.extent[0], dataset.extent[1], dataset.dtype))
 
-    E.reset_dataset(dataset)
+    rho.reset_dataset(dataset)
     print("Set the dataset properties for the scalar field E in iteration 1")
```
anokfireball: This string as well.

anokfireball (changes requested):

examples/3_write_serial.py
```diff
@@ -29,7 +29,7 @@
     print("Created an empty {0} Series".format(series.iteration_encoding))
 
     print(len(series.iterations))
-    E = series.iterations[1].meshes["E"][openPMD.Mesh_Record_Component.SCALAR]
+    rho = series.iterations[1].meshes["rho"][openPMD.Mesh_Record_Component.SCALAR]
```
anokfireball: Line length now breaks PEP-8.

anokfireball (approved):

PR #258, opened 2018-06-13
Remove MPI from serial ADIOS interface

anokfireball:
To be able to compile both serial and parallel ADIOS1 backends, remove
all traces of MPI from the serial interface. MPI_COMM_NULL is used in
the implementation, but not exposed.

Suggested in https://github.com/openPMD/openPMD-api/issues/252#issuecomment-396907630, blocker for #254.

### DIFF
--- a/include/openPMD/IO/ADIOS/ADIOS1IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS1IOHandler.hpp
@@ -25,7 +25,6 @@
 #if openPMD_HAVE_ADIOS1
 #   include "openPMD/IO/AbstractIOHandlerImpl.hpp"
 #   include <adios.h>
-#   include <adios_mpi.h> /* includes a dummy version of mpi if -D_NOMPI */
 #   include <adios_read.h>
 #endif
 
@@ -46,7 +45,7 @@ class ADIOS1IOHandler;
 class ADIOS1IOHandlerImpl : public AbstractIOHandlerImpl
 {
 public:
-    ADIOS1IOHandlerImpl(AbstractIOHandler*, MPI_Comm = MPI_COMM_NULL);
+    ADIOS1IOHandlerImpl(AbstractIOHandler*);
     virtual ~ADIOS1IOHandlerImpl();
 
     virtual void init();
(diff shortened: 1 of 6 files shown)


test/SerialIOTest.cpp
```diff
@@ -1121,7 +1121,7 @@ TEST_CASE( "no_serial_hdf5", "[serial][hdf5]" )
     REQUIRE(true);
 }
 #endif
-#if openPMD_HAVE_ADIOS1
+#if openPMD_HAVE_ADIOS1 && !openPMD_HAVE_MPI
```
anokfireball, line 1124: You can revert this change when #254 is ready.

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -36,74 +37,59 @@ namespace openPMD
...
+    ASSERT(status == MPI_SUCCESS, "Internal error: Failed to duplicate MPI communicator");
 }
 
-std::string
-getEnvNum(std::string const& key, std::string const& defaultValue)
+ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
 {
-    char const* env = std::getenv(key.c_str());
-    if( env != nullptr )
+    /* create all files where ADIOS file creation has been deferred, but this has never been triggered
```
ax3l: the comment is not very clear: what is "this" what has never been triggered and what is "this" that happens in the line below?

ax3l (approved):

ax3l:
really nice, thank you!

include/openPMD/IO/ADIOS/ADIOS1IOHandler.hpp
```diff
@@ -25,7 +25,6 @@
 #if openPMD_HAVE_ADIOS1
 #   include "openPMD/IO/AbstractIOHandlerImpl.hpp"
 #   include <adios.h>
```
ax3l: hm, `adios.h` also includes `adios_mpi.h`... argh.. but this calls the dummy... of have to check my setters more.

ax3l: I think you need `adios.h` and `mpi.h` in `ParallelADIOS1...hpp` only.

In the serial header, use `adios_read.h` and in the serial impl. `mpidummy.h` as you do.

test/SerialIOTest.cpp
ax3l: uff, I rebased on your branch and applied the mentioned changes.
I won't get a running solution with our new approach unless I do explicit symbol hiding/exposing in the two transalation units for adios, since otherwise symbols like `adios_open` and `common_adios_open` will be mismatched at link and runtime.

Can do, but this will exclude pgic++ at least since not all compilers know symbol hiding. harsh world...

Or so some symbol renaming with `--prefix-symbols` via `objcpy`... gosh!

ax3l: We could maybe go for `dlopen()` for one ADIOS part xD

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
anokfireball: True, the sentence really is a bit vague.
First "this" means actual file creation. The operation (file creation) has been deferred, but because no IO op that triggers execution of the deferred operation (file creation) has been scheduled across the whole Series, we have to do it manually.
Seconds "this" stands for the situation described above (i.e. the scenario in which no appropiate IO op has been scheduled).

ax3l (approved):

PR #261, opened 2018-06-14
Fix recursive create_directories

ax3l:
`mk()` on an existing directory returns a non-zero exit status.


anokfireball (approved):
Just curious: Windows or UNIX?

ax3l:
Saw it in `SerialIOTest` on Linux with first access to `samples` dir: `../samples/dtype_test.h5` crashed in `../`

We did not see this in CI since we always download samples. Will remove in one example.

PR #266, opened 2018-06-14
Python: Expose Attributes Fully

ax3l:
Looks still very C++-ish #255 but exposes all needed functionality for now.

Maybe we could read all attributes at once into an intermediate class (`AttributeCollection?` h5py: `AttributeManager`) and return that for read/write access? In both C++ and Python we would need to couple/link this class again safely into the parent so it can not get invalidated and un-flushed without it knowing it.

- [x] rebase after #268

### DIFF
--- a/examples/2_read_serial.py
+++ b/examples/2_read_serial.py
@@ -37,8 +37,8 @@ if __name__ == "__main__":
     print("Field E.x has shape {0} and datatype {1}".format(
           shape, E_x.dtype))
 
-    offset = openPMD.Extent([1, 1, 1])
-    extent = openPMD.Extent([2, 2, 1])
+    offset = [1, 1, 1]
+    extent = [2, 2, 1]
     # TODO buffer protocol / numpy bindings
     # chunk_data = E_x[1:3, 1:3, 1:2]
     chunk_data = E_x.load_chunk(offset, extent)
--- a/src/binding/python/Attributable.cpp
+++ b/src/binding/python/Attributable.cpp
@@ -1,0 +1,173 @@
+/* Copyright 2018 Axel Huebl
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * openPMD-api is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <pybind11/pybind11.h>
+#include <pybind11/stl_bind.h>
+#include <pybind11/stl.h>
+
+#include "openPMD/backend/Attributable.hpp"
+#include "openPMD/backend/Attribute.hpp"
+#include "openPMD/auxiliary/Variant.hpp"
+
+#include <string>
+#include <vector>
+#include <array>
+
+
+// std::variant
+//   https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html
+namespace pybind11 {
+namespace detail {
+    template< typename... Ts >
+    struct type_caster< variantSrc::variant< Ts... > > :
+        variant_caster< variantSrc::variant< Ts... > >
+    {};
+
+} // namespace detail
+} // namespace pybind11
+
+namespace py = pybind11;
+using namespace openPMD;
+
+using PyAttributeKeys = std::vector< std::string >;
+//PYBIND11_MAKE_OPAQUE(PyAttributeKeys)
+
+void init_Attributable(py::module &m) {
+    py::class_<Attributable>(m, "Attributable")
+        .def(py::init<>())
+        .def(py::init<Attributable const &>())
+
+        .def("__repr__",
+            [](Attributable const & attr) {
+                return "<openPMD.Attributable with'" + std::to_string(attr.numAttributes()) + "' attributes>";
+            }
+        )
+
+        .def_property_readonly(
+            "attributes",
+            []( Attributable & attr )
+            {
+                return attr.attributes();
+            },
+            // ref + keepalive
+            py::return_value_policy::reference_internal
+        )
+
+        // C++ pass-through API
+        //   note: since the value is expected as T&& we have to write a lambda here
+        .def("set_attribute", []( Attributable & attr, std::string const& key, char const& v ) {
+            attr.setAttribute(key, v);
+        })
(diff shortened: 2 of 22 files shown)


src/binding/python/Attributable.cpp
```diff
@@ -0,0 +1,173 @@
...
+                return attr.attributes();
+            },
+            // ref + keepalive
+            py::return_value_policy::reference_internal
+        )
+
+        // C++ pass-through API
+        //   note: since the value is expected as T&& we have to write a lambda here
+        .def("set_attribute", []( Attributable & attr, std::string const& key, char const& v ) {
+            attr.setAttribute(key, v);
```
ax3l: @C0nsultant do you think the `T&& value` signature is necessary for `setAttribute`? Or would a `T&` do as well?

examples/2_read_serial.py
```diff
@@ -37,8 +37,8 @@
     print("Field E.x has shape {0} and datatype {1}".format(
           shape, E_x.dtype))
 
-    offset = openPMD.Extent([1, 1, 1])
-    extent = openPMD.Extent([2, 2, 1])
+    offset = [1, 1, 1]
```
anokfireball, line 40: Will any Python object that is not cleanly castable to a ```std::vector< uint64_t >``` raise?
What about 
```python
offset = [-1, -1, -1]
```
or
```python
offset = 1
```
or
```python
offset = np.ones(3)
```
?

src/binding/python/Attributable.cpp
anokfireball: In restrospect ```T&&``` was indeed premature optimization. It improves performance when the user passes literals to the setters, but otherwise is not really useful.

I would not mourn after ```T&&``` if we make it ```T const&```.

src/binding/python/Attributable.cpp
```diff
@@ -0,0 +1,173 @@
...
+        .def("set_attribute", []( Attributable & attr, std::string const& key, std::vector<double> const& v ) {
+            attr.setAttribute(key, v);
+        })
+        .def("set_attribute", []( Attributable & attr, std::string const& key, std::vector<long double> const& v ) {
+            attr.setAttribute(key, v);
+        })
+        .def("set_attribute", []( Attributable & attr, std::string const& key, std::vector<std::string> const& v ) {
+            attr.setAttribute(key, v);
+        })
+        .def("set_attribute", []( Attributable & attr, std::string const& key, std::array< double, 7 > const& v ) {
```
anokfireball: Will Python ever store ```std::array< double, 7 >```?
I would assume any list with doubles of length 7 would be caught as a ```std::vector<double>``` first.

anokfireball: To expand on this a little:
```python
record.set_unit_dimension([...])
```
will probably work (i.e. store a ```std::array< double, 7 >```), but
```python
udim = [...]
record.set_attribute('unitDimension', udim)
```
will probably not.

anokfireball: This is precicely my concern with #255, btw. It is bad enough if we have this problem in Python, having it in our core C++ API will only be worse.

src/binding/python/Attributable.cpp
ax3l: Ok, that allows me to write a on-liner per specialization here. Cool.

src/binding/python/Attributable.cpp
ax3l: Why does it even matter if we pass the unitDimension attributes as `std::vector<double>` or `array< double, 7>`? In the end it must be and is the same representation in the files (intentionally). And only in `set_unit_dimension` we have to control the input length.

I will still try the examples you posted and report where it specializes to.

examples/2_read_serial.py
ax3l: I will just try it and report back :)

src/binding/python/Attributable.cpp
anokfireball: > In the end it must be and is the same representation in the files (intentionally).

True, but only for the backends we currently employ. You could think of a *highly theoretical* serialization backend (streaming IO?) where this does not apply.

I was worried about retrieval, but we are covered:
https://github.com/openPMD/openPMD-api/blob/4832b11a91a70a9a6b7d91e119222fb02602b5ea/include/openPMD/backend/BaseRecord.hpp#L190-L206

ax3l: I think that's fine. We just say for unitDimensions: "it's an array as usual, but of size 7"

examples/2_read_serial.py
ax3l: The error is mean, but yes it throws as it should:
```python
offset = [-1, -1, -1]
E_x.load_chunk(offset, extent)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-31a35ce5d5ec> in <module>()
----> 1 E_x.load_chunk(offset, extent)

TypeError: load_chunk(): incompatible function arguments. The following argument types are supported:
    1. (self: openPMD.Record_Component, arg0: List[int], arg1: List[int]) -> array

Invoked with: <openPMD.Mesh_Record_Component of dimensionality '3'>, [-1, -1, -1], [2, 2, 1]
```

We just need do add a better docstring (and docstrings at all) to the Python APIs (args).

## Wrong Dimension

```python
offset = [1, 1]

In [15]: E_x.load_chunk(offset, extent)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-15-31a35ce5d5ec> in <module>()
----> 1 E_x.load_chunk(offset, extent)

RuntimeError: Dimensionality of chunk and dataset do not match.

In [20]: offset = 1

In [21]: E_x.load_chunk(offset, extent)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-31a35ce5d5ec> in <module>()
----> 1 E_x.load_chunk(offset, extent)

TypeError: load_chunk(): incompatible function arguments. The following argument types are supported:
    1. (self: openPMD.Record_Component, arg0: List[int], arg1: List[int]) -> array

Invoked with: <openPMD.Mesh_Record_Component of dimensionality '3'>, 1, [2, 2, 1]
```

### Numpy Indices

I could allow that with an interface change, but question is if that's a good idea. Maybe...
```python
import numpy as np

# indices
x = np.ones(2, dtype=np.int)
y = x.astype(np.float)

# data
a = [[1.,2.],[3.,4.]]
b = np.array(a)

b[x]
# array([[3., 4.],
#      [3., 4.]])

a[x]
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-17-8fab1da9f470> in <module>()
----> 1 a[x]

TypeError: only integer scalar arrays can be converted to a scalar index

b[y]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-6-a124d409158d> in <module>()
----> 1 b[y]

IndexError: arrays used as indices must be of integer (or boolean) type
```

```python
offset = np.ones(3)

In [19]: E_x.load_chunk(offset, extent)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-31a35ce5d5ec> in <module>()
----> 1 E_x.load_chunk(offset, extent)

TypeError: load_chunk(): incompatible function arguments. The following argument types are supported:
    1. (self: openPMD.Record_Component, arg0: List[int], arg1: List[int]) -> array

Invoked with: <openPMD.Mesh_Record_Component of dimensionality '3'>, array([1., 1., 1.]), [2, 2, 1]
```

anokfireball (approved):

PR #267, opened 2018-06-14
Fix ADIOS Warnings

ax3l:
Fix warnings and uninitialized vars in implementation.

### DIFF
--- a/src/IO/ADIOS/ADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS1IOHandler.cpp
@@ -279,7 +279,7 @@ ADIOS1IOHandlerImpl::open_write(Writable* writable)
         m_existsOnDisk[res->second] = true;
     }
 
-    int64_t fd;
+    int64_t fd = -1;
 #if !openPMD_HAVE_MPI
     int status;
     status = adios_open(&fd,
@@ -294,9 +294,14 @@ ADIOS1IOHandlerImpl::open_write(Writable* writable)
 }
 
 ADIOS_FILE*
-ADIOS1IOHandlerImpl::open_read(std::string const& name)
+ADIOS1IOHandlerImpl::open_read(
+    std::string const&
+#if !openPMD_HAVE_MPI
+    name
+#endif
+)
 {
-    ADIOS_FILE *f;
+    ADIOS_FILE *f = nullptr;
 #if !openPMD_HAVE_MPI
     f = adios_read_open_file(name.c_str(),
                              m_readMethod,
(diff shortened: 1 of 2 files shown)


src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -294,9 +294,14 @@ ADIOS1IOHandlerImpl::open_write(Writable* writable)
...
+ADIOS1IOHandlerImpl::open_read(
+    std::string const&
+#if !openPMD_HAVE_MPI
+    name
+#endif
+)
 {
-    ADIOS_FILE *f;
+    ADIOS_FILE *f = nullptr;
 #if !openPMD_HAVE_MPI
```
ax3l, line 305: @C0nsultant can you tell me why we snip out these region? If MPI is enabled, where does the serial implementation come from?

anokfireball: Because of the ADIOS MPI debacle, I got rid of all mock traces if we build without ```_NOMPI```. As ```ParallelADIOSIOHandlerImpl``` inherits from ```ADIOSIOHandlerImpl```, we need to compile both. To make it easier not to run into MPI problems, I hid them when the serial ADIOS would not work as intended anyway. In that case, it can't be used as a backend, but merely provides functionality for the parallel backend.

The destructor has to be snipped out regardless because the logic is copied in the parallel version. Calling it twice with already finalized ADIOS will make something explode.

anokfireball (approved):

src/IO/ADIOS/ADIOS1IOHandler.cpp
ax3l: But in the current version, ADIOS will include a proper MPI as soon as we compile parallel which will work, besides the need to be started in an MPI context, also serial.

Does that mean you NOP-ed the serial backend if the parallel backend can be build? It's a bit tricky xD

anokfireball: > But in the current version, ADIOS will include a proper MPI

Which is part of my reasoning behind this. Having that compiled into the serial backend seems like the wrong thing to do. #258 seperated serial and parallel backends to the point where we do not asumme a working MPI mock employed in the parallel code (for the serial backend).
Depending on where ADIOS goes with respect to https://github.com/ornladios/ADIOS/issues/183, we can change this.

> Does that mean you NOP-ed the serial backend if the parallel backend can be build?

When it *will* be built, not just if it *could*.
But yes, it neutures a big part of the serial backend in that case.

ax3l: > #258 seperated serial and parallel backends to the point where we do not asumme a working MPI mock employed in the parallel code (for the serial backend).

Makes absolute sense, that was the idea.

> When it will be built, not just if it could.

Ok, I see.

I don't think that is fully necessary to clip it since it will still work as long as an MPI context is around. Maybe I get #254 to work, otherwise I will un-neuture the serial part again and document the RT MPI context detail with enabled parallel ADIOS backend using serial mode.

PR #269, opened 2018-06-16
Fix ADIOS1 String Attributes

ax3l:
Fix broken string attribute writes in ADIOS1: Used invalid pointer to string array before and leaked memory as well.

Also enable serial ADIOS1 on OSX for CI testing.

## Background

Valgrind says there are a few memory leaks in `std::string` attribute writing.

Also, there seems to be a `SerialIOTest` crash in ADIOS1 attribute reading, probably some un-init bytes? Before crash it reads:
```
ERROR: Attribute element /iterationFormat has invalid value attribute
ERROR: Attribute element /iterationFormat has invalid value attribute
```

which looks like a string Datatype issue in our backend.

Message origin in ADIOS: https://github.com/ornladios/ADIOS/blob/v1.13.1/src/core/adios_internals.c#L1146-L1161

During valgrind runs, don't get confused by some noise added by https://github.com/ornladios/ADIOS/issues/184

### DIFF
--- a/src/IO/ADIOS/ADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS1IOHandler.cpp
@@ -718,7 +718,7 @@ ADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
 
     Attribute att = Attribute(parameters.resource);
 
-    int nelems;
+    int nelems = 0;
     switch( parameters.dtype )
     {
         using DT = Datatype;
@@ -840,9 +840,9 @@ ADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
         }
         case DT::STRING:
         {
-            using uptr = std::unique_ptr< void, std::function< void(void*) > >;
-            values = uptr(const_cast< char* >(att.get< std::string >().c_str()),
-                          [](void*){ });
+            auto const & v = att.get< std::string >();
+            values = auxiliary::allocatePtr(Datatype::CHAR, v.length() + 1u);
+            strcpy((char*)values.get(), v.c_str());
             break;
         }
         case DT::VEC_CHAR:
@@ -977,7 +977,7 @@ ADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
                                             "",
                                             getBP1DataType(parameters.dtype),
                                             nelems,
-                                            values.get());
+                                            values.get()); // Invalid read of size 1 (could be in ADIOS itself?)
     VERIFY(status == err_no_error, "Internal error: Failed to define ADIOS attribute by value");
 
     if( parameters.dtype == Datatype::VEC_STRING )
@@ -1053,9 +1053,9 @@ ADIOS1IOHandlerImpl::readAttribute(Writable* writable,
         attrname += "/";
     attrname += parameters.name;
 
-    ADIOS_DATATYPES datatype;
-    int size;
-    void* data;
+    ADIOS_DATATYPES datatype = adios_unknown;
+    int size = 0;
+    void* data = nullptr;
 
     int status;
     status = adios_get_attr(f,
--- a/.travis.yml
+++ b/.travis.yml
@@ -116,7 +116,7 @@ jobs:
       os: osx
       language: generic
       env:
-        - USE_MPI=OFF USE_PYTHON=ON USE_HDF5=ON USE_ADIOS1=OFF USE_ADIOS2=OFF USE_SAMPLES=ON
+        - USE_MPI=OFF USE_PYTHON=ON USE_HDF5=ON USE_ADIOS1=ON USE_ADIOS2=OFF USE_SAMPLES=ON
       compiler: clang
       before_install:
          - COMPILERSPEC="%clang@8.1.0"
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,6 +37,7 @@ Bug Fixes
 - recursive directory creation with existing base #261
 - ``FindADIOS.cmake``: reset on multiple calls #263
 - ``SerialIOTest``: remove variable shadowing #262
+- ADIOS1: memory violation in string attribute writes #269
 
 Other
 """""


src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -1335,7 +1335,8 @@ ADIOS1IOHandlerImpl::readAttribute(Writable* writable,
         }
     }
 
-    free(data);
+    if(data)
```
ax3l: does not change the OSX issue

ax3l:
@C0nsultant from the valgrind issues (annotated inline) and the error message, I think we might actually have a little bug on our ADIOS1 IO backend side, at least for the representation of `std::string`s.

anokfireball:
Can you narrow down the conditions under which this triggers?
Does this only happen for files that we write? Does it also happen for files written by other codes (e.g. PIC)?

If the former is the case, this might actually be a bug in ADIOS attribute writing, rather than reading
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp#L118-L119
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/src/IO/ADIOS/ADIOS1IOHandler.cpp#L767-L768
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/src/IO/ADIOS/ADIOS1IOHandler.cpp#L841-L847

Relevant ADIOS [manual](https://users.nccs.gov/~pnorbert/ADIOS-UsersManual-1.13.1.pdf) part:
```
4.1.8
...
Simple string attributes should be defined 
with the type adios_string,
nelems=1 and
values should be the char* type pointer.
```

anokfireball:
I may have spotted the culprit (see #271).
The ```Variant::get()``` returns a copy.
https://github.com/openPMD/openPMD-api/blob/0ccf00b5ecc7661a0c26d4db4c8c6770d772c4e3/include/openPMD/auxiliary/Variant.hpp#L62-L72
When used to assign the uptr, all is fine.
https://github.com/openPMD/openPMD-api/blob/265828b7f79428e2cf3caa3e9252b03c0f29f1c9/src/IO/ADIOS/ADIOS1IOHandler.cpp#L841-L847
But as soon as the assignement completes, the temporary ```std::string``` gets destroyed.
The ```.c_str()``` pointer immediately becomes invalid.

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -977,7 +977,7 @@ ADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
                                             "",
                                             getBP1DataType(parameters.dtype),
                                             nelems,
-                                            values.get());
+                                            values.get()); // Invalid read of size 1 (could be in ADIOS itself?)
```
ax3l, line 980: in `strlen()` in `adios_get_type_size (adios_bp_v1.c:2415)` in `adios_common_define_attribute_byvalue (adios_internals.c:1109)` in `adios_define_attribute_byvalue (adios.c:442)` in `openPMD::ADIOS1IOHandlerImpl::writeAttribute` on string value: `serial_fileBased_write%T`

ax3l: we already pass the invalid c-string.

`strlen((const char*)values.get())` is an invalid read of size 1

anokfireball: Very likely because of https://github.com/openPMD/openPMD-api/pull/269#issuecomment-398033949.

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -842,7 +842,7 @@ ADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
         {
             using uptr = std::unique_ptr< void, std::function< void(void*) > >;
             values = uptr(const_cast< char* >(att.get< std::string >().c_str()),
-                          [](void*){ });
+                          [](void*){ }); // leaks memory
```
ax3l: found it, will replace this part

src/IO/ADIOS/ADIOS1IOHandler.cpp
ax3l: ah dang, just found it as well :)
yes, we can not defer/forward the pointer we get there :)

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -841,8 +841,10 @@ ADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
         case DT::STRING:
         {
             using uptr = std::unique_ptr< void, std::function< void(void*) > >;
-            values = uptr(const_cast< char* >(att.get< std::string >().c_str()),
-                          [](void*){ });
+            auto const & v = att.get< std::string >();
+            values = uptr(new char[v.length() + 1u],
+                          [](void* ca){ delete [] (char*)ca; });
```
anokfireball: There's [```auxiliary::allocatePtr()```](https://github.com/openPMD/openPMD-api/blob/d16f1762fcb1be7e7441755a3209abdba7994173/include/openPMD/auxiliary/Memory.hpp#L51) ;)

anokfireball (approved):
With the inline comments removed, this is good to go.

ax3l:
Fun fact: the `long double` valgrind reports we get in ADIOS1 are actually a valgrind bug:
https://bugs.kde.org/show_bug.cgi?id=397313

anokfireball:
You know that ```long double``` is uncommon when even a tool dealing with such wizardry as _memory debugging_ has not cared for it in the last [16+](https://sourceware.org/git/?p=valgrind.git;a=commit;h=f4da2db6b6d58e7eab2e2fd85cc9d0f1895a3a4a) years.

PR #270, opened 2018-06-17
Pad iteration for fileBased output

anokfireball:
Closes #265.

PR #272, opened 2018-06-18
Version: 0.3.0-alpha

ax3l:
Changelog for the 0.3.0-alpha release.

I just want to publish a new, updated version since people want to / shall play with it. So let's continue to release out alphas often. I decided to bump to `0.3.0` instead of `0.2.1` since we added also quite a lot of features again.


anokfireball:
Should we delay this until #270 is ready? It'll be ready soon when tests are added.

ax3l:
We could but since we are in alpha, let's just go fast, release often and break things :)

There is no reason why there should not be another release a day after if need be.

anokfireball (approved):

PR #277, opened 2018-06-20
CMake: Add openPMD::openPMD ALIAS

ax3l:
Adding an alias with the same name as the installed, exported target allows to include the full source-code of openPMD-api in a project similar to header-only libraries.

Instead of starting with `find_package(openPMD)`
```cmake
cmake_minimum_required(VERSION 3.10)

# supports:                       COMPONENTS MPI NOMPI HDF5 ADIOS1 ADIOS2
find_package(openPMD 0.1.0 CONFIG COMPONENTS REQUIRED)

add_executable(YourTarget 3_write_serial.cpp)

if(openPMD_FOUND) # well, it will be if REQUIRED is set
    target_link_libraries(YourTarget PUBLIC openPMD::openPMD)
endif()
```

just do:
```cmake
cmake_minimum_required(VERSION 3.10)

add_subdirectory(share/project/openPMD-api)

target_link_libraries(myProject PRIVATE openPMD::openPMD)
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Changes to "0.3.0-alpha"
 Features
 """"""""
 
+- CMake:
+
+  - add ``openPMD::openPMD`` alias for full-source inclusion #...
+
 Bug Fixes
 """""""""
 
--- a/README.md
+++ b/README.md
@@ -219,3 +219,10 @@ if(openPMD_FOUND)
     target_link_libraries(YourTarget PRIVATE openPMD::openPMD)
 endif()
 ```
+
+*Alternatively*, add the openPMD-api repository directly to your project and add it via:
+```cmake
+add_subdirectory("path/to/source/of/openPMD-api")
+
+target_link_libraries(YourTarget PRIVATE openPMD::openPMD)
+```
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,6 +211,7 @@ set(IO_SOURCE
 
 # library
 add_library(openPMD ${CORE_SOURCE} ${IO_SOURCE})
+add_library(openPMD::openPMD ALIAS openPMD)
 
 # properties
 target_compile_features(openPMD


README.md
```diff
@@ -219,3 +219,10 @@ if(openPMD_FOUND)
     target_link_libraries(YourTarget PRIVATE openPMD::openPMD)
 endif()
 ```
+
+*Alternatively*, add the openPMD-api repository directly to your project and add it via:
+```cmake
+add_subdirectory("path/to/source/of/openPMD-api")
+
+target_link_libraries(YourTarget PRIVATE openPMD::openPMD)
```
anokfireball, line 227: Just to see if I understand CMake here:
Assuming ```add_subdirectory``` does only add the path without scanning anything inside, ```target_link_libraries``` will do (among other things) the scan of all subdirectories (if no library ```openPMD::openPMD``` is available yet), find and scan our ```CMakeLists.txt```, find the desired ```openPMD::openPMD``` library, build it and then link it?

Damn, CMake can do _a lot_.

EDIT [```add_subdirectory```](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) does the heavy lifting.

anokfireball (approved):

README.md
ax3l: `add_subdirectory` will try to access the `CMakeLists.txt` inside its argument.

From there on, a *separate*, *isolated* build for openPMD-api is done (which does not pollute the project's CMake vars that does the `add_subdirectory`). Often such projects then even expose their dependencies as `IMPORTED` targets again, e.g. in `myProject::thirdparty::openPMD`.

C0nsultant: Ah well, [```add_subdirectory```](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) does the heavy lifting.
Still impressive.

ax3l: Yep, modern CMake is awesome :-)

CHANGELOG.rst
```diff
@@ -17,6 +17,10 @@ Changes to "0.3.0-alpha"
 Features
 """"""""
 
+- CMake:
+
+  - add ``openPMD::openPMD`` alias for full-source inclusion #...
```
ax3l, line 22: ups, forgot to add #277

PR #278, opened 2018-06-20
Spack: Python3

ax3l:
Mainline spack sadly defaults to python2. So let's explicitly tell users how to use python3 by default.

We could also increase our requirement in the package, but we do indeed (unofficially) even run with python2. Nevertheless, our official support is for Python 3, so let's install it by default as such.

Spack usage reference: https://spack.readthedocs.io/en/latest/features.html#customize-dependencies

Thx to @ritiek for the report!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Features
 
 - CMake:
 
-  - add ``openPMD::openPMD`` alias for full-source inclusion #...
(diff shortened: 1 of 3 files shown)


anokfireball (approved):
The [king](https://docs.scipy.org/doc/numpy/neps/dropping-python2.7-proposal.html) is [dead](https://github.com/pandas-dev/pandas/blob/master/doc/source/install.rst) anyway.

ax3l:
I could not agree more.

Issue #213, opened 2018-05-30
minimum version requirement of HDF5


qiminglu:
The document says that it needs the 1.8.6+ version of HDF5. However in HDF5IOHandler.cpp it has called the H5free_memory() method which is only available from HDF5 1.8.13+, please see the following link

https://support.hdfgroup.org/HDF5/doc/RM/RM_H5.html#Library-FreeMemory

With HDF5 v1.8.13 it compiles and the examples run fine for me.  So maybe should update the document for the minimum HDF5 version requirements.

anokfireball:
Thank you very much for this report, @qiminglu!
This is a prime example of what an issue should contain - a clear description, a reason for the behaviour and even a tested fix.

@ax3l do you think we should consider adding one CI step that builds with the minimal required dependency versions so we catch this by ourself in the future?

ax3l:
@qiminglu thank you for the report! :sparkles: 

@C0nsultant no CI needed for that, we just bump the version inf`find_package()` in CMake and the docs accordingly. Or do you mean we should build the *oldest supported* HDF5 version in CI? Good idea!

anokfireball:
>  Or do you mean we should build the oldest supported HDF5 version in CI?

Yeah, that's where I was going. Currently, we just pull the default versions of [HDF5](https://github.com/openPMD/openPMD-api/blob/15cda606d55ecc9d3748000182a56babbc21a0ea/.travis.yml#L279) and [ADIOS](https://github.com/openPMD/openPMD-api/blob/15cda606d55ecc9d3748000182a56babbc21a0ea/.travis.yml#L286).
Maybe we should have one build in travis that builds against the minimum requirements (i.e. oldest supported version).

ax3l:
Added in #215 for HDF5.

PR #281, opened 2018-06-20
 Ship pybind11 v2.2.3

ax3l:
Add an internal version of pybind11, version 2.2.3.

Modifications from the release:
- add fix https://github.com/pybind/pybind11/pull/1434
- remove test dirs
- remove wheel packaging (setup.py/cfg)
- remove github templates

Increase requirement from pybind 2.2.1+ to 2.2.3+ with latest fixes.


CHANGELOG.rst
```diff
@@ -20,6 +20,7 @@ Features
 - CMake:
 
   - add ``openPMD::openPMD`` alias for full-source inclusion #277
+  - include internally shipped pybind11 v2.2.3 #281
```
anokfireball, line 24: We should probably note somewehere that the (currently) included version includes modifications w.r.t. the actual ```pybind11 v2.2.3```. This would be a good place.

CMakeLists.txt
```diff
@@ -166,15 +167,25 @@ endif()
...
 # external library: pybind11 (optional)
 if(openPMD_USE_PYTHON STREQUAL AUTO)
-    find_package(pybind11 2.2.1 CONFIG)  # 2.3.0
-    if(pybind11_FOUND)
+    if(openPMD_USE_INTERNAL_PYBIND11)
+        add_subdirectory("${openPMD_SOURCE_DIR}/share/openPMD/thirdParty/pybind11")
         set(openPMD_HAVE_PYTHON TRUE)
     else()
-        set(openPMD_HAVE_PYTHON FALSE)
+        find_package(pybind11 2.2.3 CONFIG)  # 2.3.0
```
anokfireball: Do we require the changes from pybind/pybind11#1434?
If so, this is actually a lie.

CHANGELOG.rst
ax3l: in principle, absolutely yes.

the modifications I did are only relevant for direct inclusion via `add_subdirectories` (and to save some space: ~~tests~~). if one switches to `openPMD_USE_INTERNAL_PYBIND11=OFF` with the exact same release, they are not needed.

Therefore, I think the note in the commit message is good enough if one wants to see what was added when.

CMakeLists.txt
ax3l: no, we don't need those, every public release in the allowed version range will work :-)
that's the beauty of it, see above :)

anokfireball (approved):

.travis.yml
```diff
@@ -301,9 +301,9 @@ install:
     fi
   - if [ $USE_PYTHON == "ON" ]; then
       travis_wait spack install
-        py-pybind11@2.2.1 ^python@$TRAVIS_PYTHON_VERSION
+        python@$TRAVIS_PYTHON_VERSION
         $COMPILERSPEC &&
-      spack load --dependencies py-pybind11@2.2.1 ^python@$TRAVIS_PYTHON_VERSION $COMPILERSPEC;
+      spack load python@$TRAVIS_PYTHON_VERSION $COMPILERSPEC;
```
anokfireball, line 312: > We don't need to "install" pybind11 beforehand now.

True, but for a reasons similar to #213, we should consider _abusing_ one entry in the build matrix where we do.

ax3l: maybe we should abuse one to install the *latest* pybind11 version (since we ship the oldest supported version)?

Still, here it's a bit less severe, since we ship a supported version and the ones getting into trouble are packagers (well, and spack users if we keep the upper version range open).

ax3l: Have to find a creative idea how to add extra cmake flags without being super-verbose.

ax3l: @C0nsultant fixed :)

.travis.yml
```diff
@@ -304,6 +310,12 @@ install:
         python@$TRAVIS_PYTHON_VERSION
         $COMPILERSPEC &&
       spack load python@$TRAVIS_PYTHON_VERSION $COMPILERSPEC;
+      if [ "$USE_INTERNAL_PYBIND11" == "OFF" ]; then
+        travis_wait spack install
+          py-pybind11@2.2.1 ^python@$TRAVIS_PYTHON_VERSION
```
anokfireball: In the heat of the moment, you probably forgot that you bumped the minimum.
```py-pybind11@2.2.3```

File: .travis.yml [hunk 5]
```diff
@@ -304,6 +310,12 @@ install:
...
+          $COMPILERSPEC &&
+        spack load py-pybind11@2.2.1 ^python@$TRAVIS_PYTHON_VERSION $COMPILERSPEC;
```
anokfireball: Same here.

ax3l: heat of the... day!

.travis.yml
ax3l: copy-pasted from the diff when re-adding it xD

ax3l:
@C0nsultant ok to be merged? :) update: ah.. changelog.. argh.

PR #282, opened 2018-06-20
Make Container constructor protected

### LINKED ISSUE #280: Container constructor is public
Having the [```Container::Container()```](https://github.com/openPMD/openPMD-api/blob/0.3.0-alpha/include/openPMD/backend/Container.hpp#L88-L89) be publicly available is actually a design flaw. It should be ```private``` or ```protected``` at the very least.

It breaks our rule of not allowing users to create dangling objects.

anokfireball:
Avoid users creating dangling objects from a user-constructed Container.

Closes #280.

### DIFF
--- a/src/binding/python/Container.cpp
+++ b/src/binding/python/Container.cpp
@@ -90,7 +90,7 @@ namespace detail
             std::forward< Args >( args ) ...
         );
 
-        cl.def( py::init<>() );
+        cl.def( py::init<Map const &>() );
 
         // Register stream insertion operator (if possible)
         py::detail::map_if_insertion_operator<
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,7 @@ Bug Fixes
 """""""""
 
 - improved handling of ``fileBased`` Series and ``READ_WRITE`` access
+- expose ``Container`` constructor as ``protected`` rather than ``public`` #282
 
 Other
 """""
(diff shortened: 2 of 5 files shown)


src/binding/python/Container.cpp
```diff
@@ -90,7 +90,7 @@ namespace detail
             std::forward< Args >( args ) ...
         );
 
-        cl.def( py::init<>() );
+        cl.def( py::init<Map const &>() );
```
anokfireball, line 93: @ax3l Is this the correct way to stop exposing the (non-copy-)constructor in the Python interface?

ax3l: yes, you simply just define the copy constructor (but no other constructor) here.

ax3l:
@C0nsultant pls rebase for the changelog :)

ax3l (approved):

ax3l:
Wow, I used the fancy online editor xD
feel free to rebase and remove my ugly merge commit :)

anokfireball:
It sure is nice for people that need to merge something fast and can't wait for the PR author to fix it. But then again, there's ```Allow edits from maintainers```.

ax3l:
I guess so, yes.
Also, as a person that touched a windows PC last like 10 years ago, I kind of like clicking at new buttons.

ax3l:
Python unittest probably needs an update now :)

anokfireball:
> Python unittest probably needs an update now :)

That should be due to #282 as far as I can see.
I will fix the tests as soon as possible.

PR #284, opened 2018-06-22
Python: Particle Record Components

ax3l:
Add particle record component support (store/load) to the python API.

## To Do

- [x] rebase against #286

### DIFF
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -162,11 +162,18 @@ class APITest(unittest.TestCase):
         if found_numpy:
             self.assertEqual(y_data.dtype, np.float64)
             self.assertEqual(w_data.dtype, np.float64)
-        # TODO data test
-        # pytest.approx(value)
-        # self.assertSequenceEqual(y_data,
-        #     [-9.60001131e-06, -8.80004967e-06, -8.00007455e-06, ...])
-        # self.assertSequenceEqual(w_data, [1600000., 1600000., ...])
+
+            np.testing.assert_allclose(
+                y_data,
+                [-9.60001131e-06, -8.80004967e-06, -8.00007455e-06,
+                 -7.20008487e-06, -6.40007232e-06, -5.60002710e-06,
+                 -4.79993871e-06, -3.99980648e-06, -3.19964406e-06,
+                 -2.39947455e-06]
+            )
+            np.testing.assert_allclose(
+                w_data,
+                np.ones((10,)) * 1600000.
+            )
 
         E_x = E["x"]
         shape = E_x.shape
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -119,8 +119,8 @@ class APITest(unittest.TestCase):
             self.assertTrue(i in [100, 200, 300, 400, 500])
 
         # Check type.
-        self.assertTrue(100 in series.iterations)
-        i = series.iterations[100]
+        self.assertTrue(400 in series.iterations)
+        i = series.iterations[400]
         self.assertIsInstance(i, openPMD.Iteration)
         with self.assertRaises(TypeError):
             series.iterations[-1]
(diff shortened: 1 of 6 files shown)


test/python/unittest/API/APITest.py
```diff
@@ -146,25 +146,36 @@ def testData(self):
...
+        self.assertSequenceEqual(w.shape, [270625, ])
+        if found_numpy:
+            self.assertEqual(pos_x.dtype, np.float64)
+            self.assertEqual(w.dtype, np.float64)
+        #x_data =
+        pos_x.load_chunk([200000, ], [100, ])
+        # w_data =
+        w.load_chunk([200000, ], [100, ])
+        series.flush()
+        # assertSequenceEqual(x_data, [floats...])
```
ax3l: looks to me like the data is all-zero (or uninitialized)

test/python/unittest/API/APITest.py
```diff
@@ -146,25 +146,36 @@ def testData(self):
         # Get a particle species.
         electrons = i.particles["electrons"]
         self.assertIsInstance(electrons, openPMD.ParticleSpecies)
+        pos_x = electrons["momentum"]["x"]
+        w = electrons["weighting"][openPMD.Mesh_Record_Component.SCALAR]
```
ax3l: I wonder if we could realize this as well was `w = electrons["weighting"]` somehow

test/python/unittest/API/APITest.py
```diff
@@ -146,25 +146,36 @@ def testData(self):
         # Get a particle species.
         electrons = i.particles["electrons"]
         self.assertIsInstance(electrons, openPMD.ParticleSpecies)
+        pos_x = electrons["momentum"]["x"]
+        w = electrons["weighting"][openPMD.Record_Component.SCALAR]
```
ax3l: I wonder if we could realize this as well was `w = electrons["weighting"]` somehow or expose the constant magic attribute directly in `openPMD.SCALAR` as well

anokfireball: ~```w = electrons["weighting"]``` will very likely not be possible. On first access, there is ambiguity whether you want to create a ```Record``` or a scalar ```RecordComponent```.
While it would be reasonable to assume a scalar and then later discard that scalar if a named ```RecordComponent``` is added, this will result in runtime failures (not all backends support node deletion). *Note that the opposite is currently assumed.*~ #287
```w = electrons["weighting"]``` to create a scalar is not possible because the user never resolves the ambiguity.

I'd rather keep a constant. [Explicit is better than implicit.](https://www.python.org/dev/peps/pep-0020/)

Comment (ax3l) [reply test/python/unittest/API/APITest.py]: ok, then let's probably just define a shortcut alias as mentioned above

**Comment by ax3l** [comment]
@C0nsultant can you review the PR please? :-)

File: test/python/unittest/API/APITest.py [hunk 4]
```diff
@@ -146,12 +146,31 @@ def testData(self):
...
+        self.assertSequenceEqual(y_data.shape, [10, ])
+        self.assertSequenceEqual(w_data.shape, [10, ])
+        if found_numpy:
+            self.assertEqual(y_data.dtype, np.float64)
+            self.assertEqual(w_data.dtype, np.float64)
+        # TODO data test
+        # pytest.approx(value)
+        # self.assertSequenceEqual(y_data,
+        #     [-9.60001131e-06, -8.80004967e-06, -8.00007455e-06, ...])
+        # self.assertSequenceEqual(w_data, [1600000., 1600000., ...])
```
anokfireball: Is data-transfer for non-mesh ```RecordComponent```s not implemented yet?
If this is commented out because of floating-point comparisons, you could use the same ```np.testing.assert_allclose``` as below.

anokfireball (approved):

test/python/unittest/API/APITest.py
ax3l: true, I found that later!

PR #285, opened 2018-06-22
Minor fix to README source compilation process

StanczakDominik:
The directory was off for the .travis/download_samples.sh script, and given that I'm on a relatively new laptop, I actually didn't have unzip installed and that crashed the download process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,8 @@ git clone https://github.com/openPMD/openPMD-api.git
 mkdir -p openPMD-api-build
 cd openPMD-api-build
 
-# optional: for full tests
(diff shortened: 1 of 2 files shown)


ax3l (commented):
Thank you very much for the fix!
Dang, I just updated it recently and missed the base again.

Can you please also update `docs/source/install/install.rst` which contains the same typo?

StanczakDominik:
Welp, forgot to `ag` for that path! Will do that in the evening.

ax3l (approved):
Thanks! :+1:

PR #286, opened 2018-06-25
Python Load Data

ax3l:
Properly load data in python interface.

### To Do

- [x] verify values in test
- [x] more data types (as in `store_chunk`)
- [x] check array memory is not leaked

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -108,9 +108,92 @@ void init_RecordComponent(py::module &m) {
 
         // deprecated: pass-through C++ API
         .def("load_chunk", [](RecordComponent & r, Offset const & offset, Extent const & extent) {
-            std::vector<ptrdiff_t> c(extent.size());
-            std::copy(std::begin(extent), std::end(extent), std::begin(c));
-            return py::array( c, r.loadChunk<double>(offset, extent).get() );
+            std::vector<ptrdiff_t> shape(extent.size());
+            std::copy(std::begin(extent), std::end(extent), std::begin(shape));
+
+            auto dtype = py::dtype("void");
+
+            if( r.getDatatype() == Datatype::CHAR ) dtype = py::dtype("b");
+            else if( r.getDatatype() == Datatype::UCHAR ) dtype = py::dtype("B");
+            else if( r.getDatatype() == Datatype::INT16 ) dtype = py::dtype("int16");
+            else if( r.getDatatype() == Datatype::INT32 ) dtype = py::dtype("int32");
+            else if( r.getDatatype() == Datatype::INT64 ) dtype = py::dtype("int64");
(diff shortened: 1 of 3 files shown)


ax3l:
@C0nsultant ready for review (in case you have internet somewhere) ;-)

src/binding/python/RecordComponent.cpp
```diff
@@ -108,9 +108,92 @@ void init_RecordComponent(py::module &m) {
...
+            else if( r.getDatatype() == Datatype::INT32 ) dtype = py::dtype("int32");
+            else if( r.getDatatype() == Datatype::INT64 ) dtype = py::dtype("int64");
+            else if( r.getDatatype() == Datatype::UINT16 ) dtype = py::dtype("uint16");
+            else if( r.getDatatype() == Datatype::UINT32 ) dtype = py::dtype("uint32");
+            else if( r.getDatatype() == Datatype::UINT64 ) dtype = py::dtype("uint64");
+            else if( r.getDatatype() == Datatype::LONG_DOUBLE ) dtype = py::dtype("longdouble");
+            else if( r.getDatatype() == Datatype::DOUBLE ) dtype = py::dtype("double");
+            else if( r.getDatatype() == Datatype::FLOAT ) dtype = py::dtype("float");
+
+            /* @todo
```
anokfireball, line 128: All datatypes in this block are (currently) invalid arguments for chunk transfers.

src/binding/python/RecordComponent.cpp
```diff
@@ -108,9 +108,92 @@ void init_RecordComponent(py::module &m) {
...
+            .value("VEC_UINT64", Datatype::VEC_UINT64)
+            .value("VEC_FLOAT", Datatype::VEC_FLOAT)
+            .value("VEC_DOUBLE", Datatype::VEC_DOUBLE)
+            .value("VEC_LONG_DOUBLE", Datatype::VEC_LONG_DOUBLE)
+            .value("VEC_STRING", Datatype::VEC_STRING)
+            .value("ARR_DBL_7", Datatype::ARR_DBL_7)
+            */
+
+            else if( r.getDatatype() == Datatype::BOOL )
+                r.loadChunk<bool>(offset, extent, shareRaw((bool*) a.mutable_data()));
```
anokfireball, line 192: This will trigger not-yet-implemented type conversion errors when using the ADIOS backend.
When reading ```bool``` from ADIOS files, the data will be transferred as ```uint8_t``` with values ```0``` and ```1```.
https://github.com/openPMD/openPMD-api/blob/875d1ffc82cf5acff44996574ed5e8798e9afa4f/include/openPMD/RecordComponent.hpp#L129-L131

anokfireball (approved):

src/binding/python/RecordComponent.cpp
ax3l: Ok good to know.
That's fine, since errors will propagate cleanly.

Issue #135, opened 2018-04-17
Test cases with 0-sized reads & writes


anokfireball:
To verify the desired parallel behaviour, we should add tests where multiple participants from an MPI-communicator contribute 0-sized chunks to a collective IO operation.

A minimal example would consist of 3 ranks with multiple tested combinations of empty/non-empty chunks.

ax3l:
In case `storeChunk`/`loadChunk` are (not) always collective: a test where the zero-writers do/do *not* participate should also be added.

(I fear this is probably backed-dependent...)

ax3l:
To do: read-back written data in tests https://github.com/openPMD/openPMD-api/pull/309#discussion_r212986804

PR #297, opened 2018-07-23
Fail fast during flush of empty Records

### LINKED ISSUE #287: Record creation implies non-scalar RecordComponents
When creating a ```Record``` without supplying any ```RecordComponents```, ```Record::m_containsScalar``` is [default](https://github.com/openPMD/openPMD-api/blob/4832b11a91a70a9a6b7d91e119222fb02602b5ea/include/openPMD/backend/BaseRecord.hpp#L86)-initialized to ```false```. Under this assumption, an [empty group](https://github.com/openPMD/openPMD-api/blob/aca17f8563e53841b3bbb44a05a2a9e7cb3afb7d/src/Record.cpp#L60-L76) is created for such ```Record```s on ```Series::flush()```.

With major backends (e.g. [ADIOS](https://github.com/openPMD/openPMD-api/blob/dev/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L355-L367)) lacking node deletion entirely, this assumption is a design-flaw: The user can not at a later point resolve the ambiguity by specifying the ```RecordComponent``` as ```SCALAR```. This would lead to a [```std::runtime_error```](https://github.com/openPMD/openPMD-api/blob/dev/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L359) as the empty group can not reliably be replaced with a dataset.

To solve this, we should not make assumptions about (non-)scalar components. This means that we must not flush ```Record```s that contain no ```RecordComponents```. There are two options for reporting this to the user:
 - (preferred) ```throw``` when a ```Series::flush()``` reaches an empty ```Record```
 - (not preferred) skip the ```Record```; only create a dataset or group when the ambiguity is removed. When there is no such ```RecordComponent``` on file closing, this will silently discard the ```Record```.

anokfireball:
Do not write write records to disk unless scalar/non-scalar ambiguity is resolved. This is done when the user creates a record component that is either scalar/nonscalar and possibly constant.

Also fixes major bugs in the ADIOS1 backend concerning ```fileBased``` output.

Closes #287.

### DIFF
--- a/examples/1_structure.cpp
+++ b/examples/1_structure.cpp
@@ -47,5 +47,10 @@ int main()
     Dataset dataset = Dataset(Datatype::DOUBLE, Extent{1});
     mass_scalar.resetDataset(dataset);
 
+    /* Required Records and RecordComponents are created automatically.
+     * Initialization has to be done explicitly by the user. */
+    electrons["position"][RecordComponent::SCALAR].resetDataset(dataset);
+    electrons["positionOffset"][RecordComponent::SCALAR].resetDataset(dataset);
+
     return 0;
 }
--- a/include/openPMD/ParticleSpecies.hpp
+++ b/include/openPMD/ParticleSpecies.hpp
@@ -62,11 +62,11 @@ namespace traits
 
             auto& np = ret.particlePatches["numParticles"];
             auto& npc = np[RecordComponent::SCALAR];
-            npc.resetDataset(Dataset(Datatype::UINT64, {0}));
+            npc.resetDataset(Dataset(Datatype::UINT64, {1}));
             npc.parent = np.parent;
             auto& npo = ret.particlePatches["numParticlesOffset"];
             auto& npoc = npo[RecordComponent::SCALAR];
-            npoc.resetDataset(Dataset(Datatype::UINT64, {0}));
+            npoc.resetDataset(Dataset(Datatype::UINT64, {1}));
             npoc.parent = npo.parent;
         }
     };
--- a/src/IO/ADIOS/ADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS1IOHandler.cpp
@@ -37,7 +37,7 @@ namespace openPMD
 {
 #if openPMD_HAVE_ADIOS1
 #   if openPMD_USE_VERIFY
-#       define VERIFY(CONDITION, TEXT) { if(!(CONDITION)) throw std::runtime_error(std::string((TEXT))); }
+#       define VERIFY(CONDITION, TEXT) { if(!(CONDITION)) throw std::runtime_error((TEXT)); }
 #   else
 #       define VERIFY(CONDITION, TEXT) do{ (void)sizeof(CONDITION); } while( 0 )
 #   endif
--- a/src/IO/ADIOS/ADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS1IOHandler.cpp
@@ -43,15 +43,14 @@ namespace openPMD
 #   endif
 
 ADIOS1IOHandlerImpl::ADIOS1IOHandlerImpl(AbstractIOHandler* handler)
-        : AbstractIOHandlerImpl(handler),
-          m_groupName{"data"}
+        : AbstractIOHandlerImpl(handler)
 { }
 
 ADIOS1IOHandlerImpl::~ADIOS1IOHandlerImpl()
 {
-  /* create all files where ADIOS file creation has been deferred,
-   * but execution of the deferred operation has never been triggered
-   * (happens when no Operation::WRITE_DATASET is performed) */
+    /* create all files where ADIOS file creation has been deferred,
+     * but execution of the deferred operation has never been triggered
+     * (happens when no Operation::WRITE_DATASET is performed) */
     for( auto& f : m_existsOnDisk )
     {
         if( !f.second )
@@ -59,7 +58,7 @@ ADIOS1IOHandlerImpl::~ADIOS1IOHandlerImpl()
             int64_t fd;
             int status;
             status = adios_open(&fd,
-                                m_groupName.c_str(),
+                                f.first->c_str(),
                                 f.first->c_str(),
                                 "w",
                                 MPI_COMM_NULL);
@@ -209,11 +208,6 @@ ADIOS1IOHandlerImpl::init()
     status = adios_read_init_method(m_readMethod, MPI_COMM_NULL, "");
     VERIFY(status == err_no_error, "Internal error: Failed to initialize ADIOS reading method");
 
-    ADIOS_STATISTICS_FLAG noStatistics = adios_stat_no;
-    status = adios_declare_group(&m_group, m_groupName.c_str(), "", noStatistics);
-    VERIFY(status == err_no_error, "Internal error: Failed to declare ADIOS group");
-    status = adios_select_method(m_group, "POSIX", "", "");
-    VERIFY(status == err_no_error, "Internal error: Failed to select ADIOS method");
 }
 #endif
 
@@ -279,7 +273,7 @@ ADIOS1IOHandlerImpl::open_write(Writable* writable)
     int64_t fd = -1;
     int status;
     status = adios_open(&fd,
-                        m_groupName.c_str(),
+                        res->second->c_str(),
                         res->second->c_str(),
                         mode.c_str(),
                         MPI_COMM_NULL);
@@ -299,6 +293,19 @@ ADIOS1IOHandlerImpl::open_read( std::string const& name )
     VERIFY(f != nullptr, "Internal error: Failed to open_read ADIOS file");
 
     return f;
+}
+
+int64_t
+ADIOS1IOHandlerImpl::initialize_group(std::string const &name)
+{
+    int status;
+    int64_t group;
+    ADIOS_STATISTICS_FLAG noStatistics = adios_stat_no;
+    status = adios_declare_group(&group, name.c_str(), "", noStatistics);
+    VERIFY(status == err_no_error, "Internal error: Failed to declare ADIOS group");
+    status = adios_select_method(group, "POSIX", "", "");
+    VERIFY(status == err_no_error, "Internal error: Failed to select ADIOS method");
+    return group;
 }
 
 #define CommonADIOS1IOHandlerImpl ADIOS1IOHandlerImpl
--- a/src/IO/ADIOS/ADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS1IOHandler.cpp
@@ -48,32 +48,30 @@ ADIOS1IOHandlerImpl::ADIOS1IOHandlerImpl(AbstractIOHandler* handler)
 
 ADIOS1IOHandlerImpl::~ADIOS1IOHandlerImpl()
 {
-    /* create all files where ADIOS file creation has been deferred,
-     * but execution of the deferred operation has never been triggered
-     * (happens when no Operation::WRITE_DATASET is performed) */
-    for( auto& f : m_existsOnDisk )
-    {
-        if( !f.second )
-        {
-            int64_t fd;
-            int status;
-            status = adios_open(&fd,
-                                f.first->c_str(),
-                                f.first->c_str(),
-                                "w",
-                                MPI_COMM_NULL);
-            if( status != err_no_error )
-                std::cerr << "Internal error: Failed to open_flush ADIOS file\n";
-
-            m_openWriteFileHandles[f.first] = fd;
-        }
-    }
-
     for( auto& f : m_openReadFileHandles )
         close(f.second);
+    m_openReadFileHandles.clear();
 
-    for( auto& f : m_openWriteFileHandles )
-        close(f.second);
+    if( this->m_handler->accessType != AccessType::READ_ONLY )
+    {
+        for( auto& f : m_openWriteFileHandles )
+            close(f.second);
+        m_openWriteFileHandles.clear();
+
+        for( auto& group : m_attributeWrites )
+            for( auto& att : group.second )
+                flush_attribute(group.first, att.first, att.second);
+
+        /* create all files, even if ADIOS file creation has been deferred,
+         * but execution of the deferred operation has never been triggered
+         * (happens when no Operation::WRITE_DATASET is performed) */
+        for( auto& f : m_filePaths )
+            if( m_openWriteFileHandles.find(f.second) == m_openWriteFileHandles.end() )
+                m_openWriteFileHandles[f.second] = open_write(f.first);
+
+        for( auto& f : m_openWriteFileHandles )
+            close(f.second);
+    }
 
     int status;
     status = adios_read_finalize_method(m_readMethod);
--- a/src/RecordComponent.cpp
+++ b/src/RecordComponent.cpp
@@ -18,6 +18,7 @@
  * and the GNU Lesser General Public License along with openPMD-api.
  * If not, see <http://www.gnu.org/licenses/>.
  */
+#include "openPMD/auxiliary/Memory.hpp"
 #include "openPMD/RecordComponent.hpp"
 #include "openPMD/Dataset.hpp"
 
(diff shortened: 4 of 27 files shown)


src/backend/MeshRecordComponent.cpp
```diff
@@ -44,10 +44,16 @@ MeshRecordComponent::read()
     Attribute a = Attribute(*aRead.resource);
     if( *aRead.dtype == DT::VEC_FLOAT )
         setPosition(a.get< std::vector< float > >());
+    else if( *aRead.dtype == DT::FLOAT )
+        setPosition(std::vector< float >(a.get< float >()));
```
anokfireball: This actually initializes a vector with
```cpp
static_cast< std::vector< float >::size_type >(a.get< float >())
```
default-initalized elements.
The intended behaviour is realized with
```cpp
setPosition(std::vector< float >({a.get< float >()}));
```

ax3l: @C0nsultant do you want to update the branch with it?

src/backend/MeshRecordComponent.cpp
```diff
@@ -44,10 +44,16 @@ MeshRecordComponent::read()
...
     else if( *aRead.dtype == DT::VEC_DOUBLE )
         setPosition(a.get< std::vector< double > >());
+    else if( *aRead.dtype == DT::DOUBLE )
+        setPosition(std::vector< double >(a.get< double >()));
```
ax3l: same here

src/backend/MeshRecordComponent.cpp
```diff
@@ -44,10 +44,16 @@ MeshRecordComponent::read()
...
     else if( *aRead.dtype == DT::VEC_LONG_DOUBLE )
         setPosition(a.get< std::vector< long double > >());
+    else if( *aRead.dtype == DT::LONG_DOUBLE )
+        setPosition(std::vector< long double >(a.get< long double >()));
```
ax3l: same here

test/SerialIOTest.cpp
```diff
@@ -1451,9 +1443,11 @@ TEST_CASE( "adios1_fileBased_write_test", "[serial][adios1]" )
     {
         *positionOffset_local_1 = positionOffset_global[i];
         e_1["positionOffset"]["x"].storeChunk({i}, {1}, positionOffset_local_1);
-        o.flush();
     }
 
+    //TODO breaks openWriteFile logic when removed
+    //o.flush();
```
ax3l: works now?

examples/1_structure.cpp
```diff
@@ -47,5 +47,10 @@ int main()
     Dataset dataset = Dataset(Datatype::DOUBLE, Extent{1});
     mass_scalar.resetDataset(dataset);
 
+    /* Required Records and RecordComponents are created automatically.
+     * Initialization has to be done explicitly by the user. */
+    electrons["position"][RecordComponent::SCALAR].resetDataset(dataset);
```
ax3l, line 52: I just realized `resetDataset` in `(Patch)RecordComponent` lacks doxygen docs. What does it do, overwrite a dataset?

test/SerialIOTest.cpp
anokfireball: No, unfortunately. I figured out where it all goes haywire, but not what causes the behavior.
It has to do with ADIOS's distinction between read/write files. The scenario under which I can reproduce this is when creating a ```fileBased``` Series and *not* flushing between creation of two Iterations (i.e. files). The problem then becomes that the file with the earliest creation time will have an on-disk size of 0Bytes and only the last file will be created as expected.
I doubt this is a problem with ADIOS itself, but the way the backend is using it:
File creation (i.e. opening a write file) is deferred until required (i.e. encountering a Dataset write operation). This might be a problem if no Dataset write occurs. Once this file is opened, it will not be closed until new defines or reads happen.
... and this is exactly where the problem happens. One of these calls happens at the wrong time (for ADIOS), so the file on disk will be empty. Unfortunately, ADIOS doesn't fail (only warns) when opening a 0Byte file (```adios_errno``` is not set, either). The ```SIGSEGV``` in the test happens when that faulty file is closed.

src/backend/MeshRecordComponent.cpp
anokfireball: Will do before merging.

examples/1_structure.cpp
anokfireball: It overwrites the default Dataset (single ```char```) with a user-defined dataset, including datatype, extent, chunking, compression, etc.
The Dataset is used when flushing a RecordComponent to disk, see [```RecordComponent::flush()```](https://github.com/openPMD/openPMD-api/blob/7501e3435913216812f3efe8707b30892b3aa915/src/RecordComponent.cpp#L97-L106) for example.

anokfireball: At this point I am not sure if a default *valid* dataset is the best solution. It surely is helpful for prototyping, but having a ```flush()``` fail with an uninitialized, default dataset sends a clear message.

test/SerialIOTest.cpp
anokfireball: (Likely) found the culprit:
In ```fileBased``` Series, *one* ADIOS group was used for *all* files. This works when only one file is open at a time, but that is not a constraint enforced in the ADIOS backends.
A solution is to create a group for every file handle. The commit alleviates the CI failure, but does not reliably use the group across the backend yet.

anokfireball:
@ax3l OSX builds on Travis are failing after an upgrade to High Sierra. In particular, spack install of a compiler [fails](https://travis-ci.org/openPMD/openPMD-api/jobs/410981935#L949):
```
=> Error: No compilers found for operating system highsierra and target x86_64.
If previous installations have succeeded, the operating system may have been updated.
Compilers are defined for the following operating systems and targets:
	sierra-x86_64
	ubuntu14.04-x86_64
Run 'spack compiler find' to add compilers.
```
Can you look into that?

ax3l:
> OSX builds on Travis are failing after an upgrade to High Sierra

ugh, sorry about that. yes will update: #299

ref: https://blog.travis-ci.com/2018-07-19-xcode9-4-default-announce

anokfireball:
There is a pretty major blocker in the current ADIOS backend:
Attributes may only be defined *once* when using ```adios_define_attribute_byvalue()```. Otherwise, they will occur multiple times under multiple IDs.
The problem is that (borrowing from HDF5 behaviour), the backend was implemented under the assumption that successive definitions overwrite previous values.

```cpp
#define _NOMPI
#include <adios.h>


int main()
{
  adios_init_noxml(MPI_COMM_NULL);

  int64_t m_adios_group;
  int64_t m_adios_file;

  adios_declare_group(&m_adios_group, "att_test.bp", "", adios_stat_default);
  adios_select_method(m_adios_group, "POSIX", "", "");

  /* ############################################################################################################ */

  int i = 0;
  adios_define_attribute_byvalue(m_adios_group,
                                 "attribute", "/foo", adios_integer, 1, &i);

  i = 1;
  adios_define_attribute_byvalue(m_adios_group,
                                 "attribute", "/foo", adios_integer, 1, &i);

  /* ############################################################################################################ */

  adios_open(&m_adios_file, "att_test.bp", "att_test.bp", "w", MPI_COMM_NULL);
  adios_close(m_adios_file);

  adios_finalize(0);
  return 0;
}
```

leads to

```bash
 $ bpdump att_test.bp
[...]
 Attributes Count: 5
                Attribute Name (ID): attribute (1)
                Attribute Path: /foo
                Datatype: integer
                # of elements:   1
                Length in bytes: 4
                Value: 0

                Attribute Name (ID): attribute (2)
                Attribute Path: /foo
                Datatype: integer
                # of elements:   1
                Length in bytes: 4
                Value: 1
[...]
```

There are two options to explore for solving this problem:
- Deferring *all* attribute writes until the IOHandler is terminated. At that point we can determine which write happens last and discard all overwritten values.
- Using ```adios_define_attribute``` and write the attribute value to a variable. This is similar to what the backend already does for dynamic chunk sizes. Note ADIOS [doc](https://users.nccs.gov/~pnorbert/ADIOS-UsersManual-1.13.1.pdf):
> It is not appropriate to define a foating-point type attribute with high-precision, and cannot be used to define an array attribute. For those cases, see the next function below, ```adios_define_attribute_byvalue```.

ax3l:
The first approach looks more forward-thinking, e.g. for other backends that will have similar limitations.

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -48,32 +48,30 @@ ADIOS1IOHandlerImpl::ADIOS1IOHandlerImpl(AbstractIOHandler* handler)
...
+        for( auto& group : m_attributeWrites )
+            for( auto& att : group.second )
+                flush_attribute(group.first, att.first, att.second);
+
+        /* create all files, even if ADIOS file creation has been deferred,
+         * but execution of the deferred operation has never been triggered
+         * (happens when no Operation::WRITE_DATASET is performed) */
+        for( auto& f : m_filePaths )
+            if( m_openWriteFileHandles.find(f.second) == m_openWriteFileHandles.end() )
+                m_openWriteFileHandles[f.second] = open_write(f.first);
```
anokfireball, line 70: ~For this to work with ```AccessType::READ_WRITE```, [```openFile()```](https://github.com/openPMD/openPMD-api/blob/2cc1cef79f9b75d181296956611dc6acd2aaed3f/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L453-L487) needs to set ```m_existsOnDisk``` so [```open_write()```](https://github.com/openPMD/openPMD-api/blob/2cc1cef79f9b75d181296956611dc6acd2aaed3f/src/IO/ADIOS/ADIOS1IOHandler.cpp#L247-L281) does not overwrite, but append.~
Done.

anokfireball:
@ax3l Should be ready for review. I'll add a few more tests and a changelog for the recent PRs today.

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -43,38 +43,35 @@ namespace openPMD
...
+
+        for( auto& f : m_openWriteFileHandles )
+            close(f.second);
+    }
```
anokfireball, line 74: ~This needs to be done in ```ParallelADIOS1IOHandler.cpp``` as well.~
Done.

include/openPMD/ParticleSpecies.hpp
```diff
@@ -62,11 +62,11 @@ namespace traits
 
             auto& np = ret.particlePatches["numParticles"];
             auto& npc = np[RecordComponent::SCALAR];
-            npc.resetDataset(Dataset(Datatype::UINT64, {0}));
+            npc.resetDataset(Dataset(Datatype::UINT64, {1}));
```
ax3l, line 65: Why do we set an extend of `1` here already? Doesn't this imply a first entry already? What if someone does not write data / writes exactly one entry later on? Just wondering about corner cases.

src/IO/ADIOS/ADIOS1IOHandler.cpp
ax3l: But does `openFile()` set this now?

src/IO/ADIOS/ADIOS1IOHandler.cpp
ax3l: done as far as I can see :)

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -301,6 +293,19 @@ ADIOS1IOHandlerImpl::open_read( std::string const& name )
     return f;
 }
 
+int64_t
+ADIOS1IOHandlerImpl::initialize_group(std::string const &name)
+{
+    int status;
+    int64_t group;
+    ADIOS_STATISTICS_FLAG noStatistics = adios_stat_no;
```
ax3l, line 301: just a note: these statistics are quite interesting, but I never evaluated the RT costs in our use cases, so we turned them off by default.

if we could control from the outside we could evaluate them.

maybe good for an issue.

src/RecordComponent.cpp
```diff
@@ -103,6 +104,21 @@ RecordComponent::flush(std::string const& name)
                 dCreate.compression = m_dataset->compression;
                 dCreate.transform = m_dataset->transform;
                 IOHandler->enqueue(IOTask(this, dCreate));
+                if( m_chunks->empty() )
+                {
+                    /* Ensure at least one WRITE_DATASET per dataset occurs
+                     * ADIOS1 backend only creates a dataset after at least once cell has been written */
```
ax3l, line 110: do we or ADIOS define the term cell?

test/CoreTest.cpp
```diff
@@ -102,8 +102,11 @@ TEST_CASE( "output_constructor_test", "[core]" )
 
     o.setMeshesPath("customMeshesPath").setParticlesPath("customParticlesPath");
 
-    o.iterations[1].meshes["foo"];
-    o.iterations[1].particles["bar"];
+    o.iterations[1].meshes["foo"]["baz"].resetDataset(Dataset(Datatype::DOUBLE, {1}));
```
ax3l, line 105: can you add a (global) extend-zero write to the tests and it's defined behaviour? Basically `empty_record_test` but where the user sets the extend explicitly to `{0}`?

Do we already have a test that does a parallel write with extend > 0 (1D and ND) in which some of the MPI ranks contribute *locally* zero extends? This is quite often the case for e.g. particle data.

ax3l (commented):
Thank you so much, these are fantastic updates :sparkles: 

I have added some inline questions, probably most of them already addressed.

src/IO/ADIOS/ADIOS1IOHandler.cpp
anokfireball: Yes it does, right [here](https://github.com/openPMD/openPMD-api/blob/414f2ddbdd1e0542d6f82aa32885e3adbfe7eef4/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L490).

include/openPMD/ParticleSpecies.hpp
anokfireball: ADIOS does not handle datasets without extent well. The default here is just used to enforce standard compliance and create deterministic behaviour across backends.
To address your concern about corner cases: [```PatchRecordComponent::resetDataset()```](https://github.com/openPMD/openPMD-api/blob/0.3.1-alpha/src/backend/PatchRecordComponent.cpp#L33-L42) should be used by users that do want to write actual data. It ~will~ should be replaced when writing production data.
If the user does *not* write dataset values ever, the created dataset will contain one potentially [default](https://github.com/openPMD/openPMD-api/blob/414f2ddbdd1e0542d6f82aa32885e3adbfe7eef4/src/RecordComponent.cpp#L107-L121)-initalized value.

src/RecordComponent.cpp
anokfireball: Interesting. ADIOS *does* use the same lingo for what I called cell. (See e.g. [```adios_define_mesh_unstructured```](https://users.nccs.gov/~pnorbert/ADIOS-UsersManual-1.13.1.pdf) on page 52).
I used it independently here, though.

test/CoreTest.cpp
anokfireball: You're probably misreading what is happening here.
```resetDataset()``` is not actually a data write-through to disk. It *defines* the *global* datatype, extent, compression (...) of each dataset. This is a collective call, every rank will have to call this with the same global data.
If the extent passed to ```resetDataset()``` is zero in any dimension (and is later flushed to disk), undefined behaviour in certain backends ensures.

Where you would actually pass the zero-extent in the scenario you're describing is in [```RecordComponent::storeChunk()```](https://github.com/openPMD/openPMD-api/blob/0.3.1-alpha/include/openPMD/RecordComponent.hpp#L167-L196), which triggers a ```Operation::WRITE_DATASET``` in the backends (see [HDF5](https://github.com/openPMD/openPMD-api/blob/0.3.1-alpha/src/IO/HDF5/HDF5IOHandler.cpp#L653-L735) and [ADIOS](https://github.com/openPMD/openPMD-api/blob/0.3.1-alpha/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L376-L416)).

#135 is for implementing such a local zero-extent write into a global dataset. Currently, there is no such test.

ax3l: Ok, great. Just one thing:

> If the extent passed to resetDataset() is zero in any dimension (and is later flushed to disk), undefined behaviour in certain backends ensures.

Instead of undef behaviour, depending on backend, can we just implement defined similar behaviour for all backends? E.g., if global zero extent is given we throw (directly or in flush) or if zero is given we skip the write (and document that), etc.?

ax3l (approved):
Very nice! We can define a consistent behavior for global-zero extends in a follow-up.

test/CoreTest.cpp
anokfireball: > if global zero extent is given we throw (directly or in flush)

That's the behaviour I currently implement in #309. Skipping the undefined write (however verbose) does not appeal to me, as the dataset itself does not really make sense.

PR #298, opened 2018-07-30
Spack: load -r

ax3l:
`spack load -r` is shorter and easier to type than `--dependencies`.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -106,12 +106,11 @@ install:
 
   # Add our channels.
   - cmd: conda config --set show_channel_urls true
-  - cmd: conda config --remove channels defaults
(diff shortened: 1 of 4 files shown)


anokfireball (approved):

ax3l:
Something seems wrong in the AppVeyor MKL-FFT install on 32 bit windows inside numpy 1.15.0 with Python 3.5 ...

ax3l:
I will keep the numpy 1.14.5 for testing on appveyor pinned for now, but the originating issue is that I install it's dependency `mkl_fft` from `conda-forge` and mix channels in an unhealthy way. (Something that conda dependency resolution is supposed to keep track of...)

PR #300, opened 2018-08-05
IWYU: Public Headers

ax3l:
## `openPMD.hpp`: Export all public headers

Build a proper facade header that includes all public API headers.

> `IWYU pragma`: export to tell IWYU that one header serves as the
> provider for all symbols in another, included header (e.g. facade
> headers). Use IWYU pragma: begin_exports/end_exports for a whole
> group of included headers

Ref:
  https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md

## Variant

Export wrapped underlying variant types.

## Private Headers

~~Experiment with private headers and `friend` declarations for internal includes and tests.~~

-> this will probably not help much/work well since `friend` regex would have to include tests and all internal `.cpp` files and something general such as an `openPMD/` directory could also be used in user source code.

### DIFF
--- a/include/openPMD/openPMD.hpp
+++ b/include/openPMD/openPMD.hpp
@@ -20,6 +20,33 @@
  */
 #pragma once
 
+// IWYU pragma: begin_exports
+#include "openPMD/Dataset.hpp"
+#include "openPMD/Datatype.hpp"
+#include "openPMD/IterationEncoding.hpp"
+#include "openPMD/Iteration.hpp"
+#include "openPMD/Mesh.hpp"
+#include "openPMD/ParticlePatches.hpp"
+#include "openPMD/ParticleSpecies.hpp"
+#include "openPMD/RecordComponent.hpp"
+#include "openPMD/Record.hpp"
 #include "openPMD/Series.hpp"
+
+#include "openPMD/backend/Attributable.hpp"
+#include "openPMD/backend/Attribute.hpp"
+#include "openPMD/backend/BaseRecordComponent.hpp"
+#include "openPMD/backend/BaseRecord.hpp"
+#include "openPMD/backend/Container.hpp"
+#include "openPMD/backend/MeshRecordComponent.hpp"
+#include "openPMD/backend/PatchRecord.hpp"
+#include "openPMD/backend/PatchRecordComponent.hpp"
+#include "openPMD/backend/Writable.hpp"
+
+#include "openPMD/IO/AccessType.hpp"
+
+#include "openPMD/auxiliary/OutOfRangeMsg.hpp"
 #include "openPMD/auxiliary/ShareRaw.hpp"
+#include "openPMD/auxiliary/Variant.hpp"
+
 #include "openPMD/version.hpp"
+// IWYU pragma: end_exports
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -22,6 +22,7 @@
 
 #include <iostream>
 #include <memory>
+#include <cstddef>
 
 
 using std::cout;
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -23,6 +23,7 @@
 #include <iostream>
 #include <memory>
 #include <numeric>
+#include <cstdlib>
 
 
 using std::cout;
--- a/examples/4_read_parallel.cpp
+++ b/examples/4_read_parallel.cpp
@@ -24,6 +24,7 @@
 
 #include <iostream>
 #include <memory>
+#include <cstddef>
 
 
 using std::cout;
--- a/include/openPMD/ParticlePatches.hpp
+++ b/include/openPMD/ParticlePatches.hpp
@@ -24,6 +24,7 @@
 #include "openPMD/backend/PatchRecord.hpp"
 
 #include <vector>
+#include <cstddef>
 
 
 namespace openPMD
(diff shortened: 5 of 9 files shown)


include/openPMD/Series.hpp
```diff
@@ -20,6 +20,10 @@
  */
 #pragma once
 
+// IWYU pragma: private, include <openPMD/openPMD.hpp>
+// IWYU pragma: friend ".*openPMD/.*"
```
ax3l: that's not enough, e.g. everything in `src/` such as `Iteration.cpp` would not be matched

include/openPMD/Series.hpp
```diff
@@ -20,6 +20,10 @@
...
+// IWYU pragma: friend ".*Test\.cpp"
```
ax3l: we could also let out tests include `openPMD/openPMD.hpp` altogether.

anokfireball: At first it seemed strange to me to include facade headers in tests rather than specifying the components under test.
It probably is not a bad idea, as it would catch regressions caused by implicit and/or transitive imports, just as they would occur in production-code.

Note: Not all headers in ```include/``` are publicly (transitively) included by means of ```openPMD.hpp```.

ax3l: > Note: Not all headers in include/ are publicly (transitively) included by means of openPMD.hpp.

Could be that I included a few too many now...
Can you list which ones are purely `// IWYU pragma: private` for internal `.cpp` files?

include/openPMD/openPMD.hpp
```diff
@@ -20,6 +20,67 @@
...
+#include "openPMD/ParticleSpecies.hpp"
+#include "openPMD/RecordComponent.hpp"
+#include "openPMD/Record.hpp"
 #include "openPMD/Series.hpp"
+
+#include "openPMD/IO/AbstractIOHandler.hpp"
+#include "openPMD/IO/AbstractIOHandlerHelper.hpp"
+#include "openPMD/IO/AccessType.hpp"
+#include "openPMD/IO/Format.hpp"
+#include "openPMD/IO/IOTask.hpp"
```
anokfireball: All of the headers in this block are included multiple times.

include/openPMD/openPMD.hpp
```diff
@@ -20,6 +20,67 @@
...
+#include "openPMD/IO/HDF5/HDF5IOHandler.hpp"
+#include "openPMD/IO/HDF5/HDF5IOHandlerImpl.hpp"
+#include "openPMD/IO/HDF5/ParallelHDF5IOHandler.hpp"
+#include "openPMD/IO/HDF5/ParallelHDF5IOHandlerImpl.hpp"
+#include "openPMD/IO/IOTask.hpp"
+
+#include "openPMD/backend/Attributable.hpp"
+#include "openPMD/backend/Attribute.hpp"
+#include "openPMD/backend/Container.hpp"
+#include "openPMD/backend/Writable.hpp"
```
anokfireball: Same here.

include/openPMD/Series.hpp
anokfireball: ~Without any verification, just the best judgement I can give right now:~
```cpp
/* internal only */

#include "openPMD/IO/AbstractFilePosition.hpp"
#include "openPMD/IO/AbstractIOHandlerImpl.hpp"

#include "openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp"
#include "openPMD/IO/ADIOS/ADIOS1FilePosition.hpp"
#include "openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp"
#include "openPMD/IO/ADIOS/ADIOS2FilePosition.hpp"
#include "openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp"

#include "openPMD/IO/HDF5/HDF5Auxiliary.hpp"
#include "openPMD/IO/HDF5/HDF5FilePosition.hpp"
#include "openPMD/IO/HDF5/HDF5IOHandlerImpl.hpp"
#include "openPMD/IO/HDF5/ParallelHDF5IOHandlerImpl.hpp"

#include "openPMD/auxiliary/Filesystem.hpp"
#include "openPMD/auxiliary/Memory.hpp"
#include "openPMD/auxiliary/StringManip.hpp"
```

```cpp
/* unnecessarily publicly exposed */

#include "openPMD/IO/AbstractIOHandler.hpp" //Series, Attributable
#include "openPMD/IO/IOTask.hpp" //Series->AbstractIOHandler

#include "openPMD/IO/Format.hpp" //Series, Series->AbstractIOHandler
```

- [X] Verify all the claimed private includes

ax3l: That also means we should not *install* those headers :)

ax3l: @C0nsultant were you able to verify the private-ness of those headers? (how? manually?)

anokfireball: Will do so today. And yes, that means going through them manually.

anokfireball: @ax3l Done. There are a few headers that are currently publicly exposed without any real need to.

ax3l:
@C0nsultant with the new public header annotations, I think we can already merge this.

It's still weird it does not recognize `<openPMD/openPMD.hpp>` in examples as the proper include and requests the `""` version...

I would say other include suggestions can be addressed in follow-up PRs to keep things small.

anokfireball (approved):
This only fixes external library use, not yet internal use of our own headers.

PR #302, opened 2018-08-06
Fix ADIOS2 File Position Include

ax3l:
non-existent include.

was not used yet since the backend is a stub.


anokfireball (approved):

anokfireball:
There is probably more faulty code in the ADIOS2 backend. It was largely a dupe of old ADIOS1 code with minor modifications.

PR #303, opened 2018-08-06
Travis: Fix Overlapping Caches

ax3l:
Fix overlapping caches in Travis-CI.
Make "env" unique per build matrix entry:

https://docs.travis-ci.com/user/caching/#caches-and-build-matrices

This causes corruptions.


ax3l:
@C0nsultant yay!

PR #304, opened 2018-08-07
Default Constructors/Destructors

ax3l:
Address clang-tidy [`hicpp-use-equals-default`](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-equals-default.html).
Use `= default` for trivial constructors and destructors.

### DIFF
--- a/src/IO/ADIOS/ADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS1IOHandler.cpp
@@ -225,8 +225,7 @@ ADIOS1IOHandler::ADIOS1IOHandler(std::string const& path, AccessType at)
     m_impl->init();
 }
 
-ADIOS1IOHandler::~ADIOS1IOHandler()
-{ }
+ADIOS1IOHandler::~ADIOS1IOHandler() = default;
 
 std::future< void >
 ADIOS1IOHandler::flush()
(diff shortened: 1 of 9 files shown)


anokfireball (approved):

ax3l:
Is it possible that `HDF5IOHandlerImpl::~HDF5IOHandlerImpl()` needs to be declared as `~HDF5IOHandlerImpl() override;` instead of `virtual ~HDF5IOHandlerImpl();`?

anokfireball:
Yes. Yes, it should just like all other ```IOHandlerImpl```s that inherit from ```AbstractIOHandlerImpl```.
... But you already did that. :hourglass:

ax3l:
Nope, did not change their declarations yet. They don't declare `override` yet.

include/openPMD/IO/ADIOS/ADIOS1IOHandler.hpp
```diff
@@ -47,7 +47,7 @@ namespace openPMD
 
     public:
         ADIOS1IOHandler(std::string const& path, AccessType);
-        ~ADIOS1IOHandler() override;
+        ~ADIOS1IOHandler() override = default;
```
ax3l: Hm, can't move those up since `ADIOS1IOHandlerImpl` is still incomplete here.

anokfireball:
> Nope, did not change their declarations yet. They don't declare override yet.

My bad. You did the Handlers, not the HandlerImpls.

> ADIOS1IOHandlerImpl is still incomplete here.

Generally speaking: Why do you want to put the ```= default``` specifications inside the headers? They can be put in the ```.cpp``` sources just as well. This does not clutter the interfaces and does not enforce as many constraints.

Header ```test.hpp```
```cpp
class Test
{
public:
    Test();
    ~Test();
};
```

Source ```test.cpp```
```cpp
#include "test.hpp"

Test::Test() = default;
Test::~Test() = default;
```

ax3l:
Yeah, `coveralls`... *relative* coverage decreased because I removed LOC ;)

ax3l:
@C0nsultant works! :)

PR #305, opened 2018-08-07
Path Arguments: Pass-by-Value

ax3l:
Address clang-tidy [`modernize-pass-by-value`](https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html).

Pass `std::string`s by value and `std::move` them where needed, e.g. in constructors.

At worst, same copy cost as before for member init. At best, add rvalue movability.

### DIFF
--- a/include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
@@ -75,7 +75,7 @@ namespace openPMD
         void listAttributes(Writable*, Parameter< Operation::LIST_ATTS > &) override;
 
         virtual int64_t open_write(Writable *);
-        virtual ADIOS_FILE* open_read(std::string const& name);
+        virtual ADIOS_FILE* open_read(std::string const & name);
         void close(int64_t);
         void close(ADIOS_FILE*);
 
--- a/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
+++ b/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
@@ -76,7 +76,7 @@ namespace openPMD
         void listAttributes(Writable*, Parameter< Operation::LIST_ATTS > &) override;
 
         virtual int64_t open_write(Writable *);
-        virtual ADIOS_FILE* open_read(std::string const& name);
+        virtual ADIOS_FILE* open_read(std::string const & name);
         void close(int64_t);
         void close(ADIOS_FILE*);
 
--- a/include/openPMD/IO/AbstractIOHandlerHelper.hpp
+++ b/include/openPMD/IO/AbstractIOHandlerHelper.hpp
@@ -37,7 +37,7 @@ namespace openPMD
      */
     std::shared_ptr< AbstractIOHandler >
     createIOHandler(
-        std::string const& path,
+        std::string path,
         AccessType accessType,
         Format format,
         MPI_Comm comm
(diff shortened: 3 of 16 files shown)


include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
```diff
@@ -75,7 +75,7 @@ namespace openPMD
         void listAttributes(Writable*, Parameter< Operation::LIST_ATTS > &) override;
 
         virtual int64_t open_write(Writable *);
-        virtual ADIOS_FILE* open_read(std::string const& name);
+        virtual ADIOS_FILE* open_read(std::string name);
```
ax3l: not sure if it's clever to change this as well

include/openPMD/IO/AbstractIOHandlerHelper.hpp
```diff
@@ -37,7 +37,7 @@ namespace openPMD
      */
     std::shared_ptr< AbstractIOHandler >
     createIOHandler(
-        std::string const& path,
+        std::string path,
```
ax3l, line 40: makes sense to change this as well, no?

ax3l:
@C0nsultant ready for review :)

include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
anokfireball: Internally, this just passes the ```.c_str()``` into ```adios_read_open_file```. There won't be any benefit from moving or creating a temporary as the string is not sunk. If anything, this could be changed to
```cpp
virtual ADIOS_FILE* open_read(char * const name);
```

include/openPMD/IO/AbstractIOHandlerHelper.hpp
anokfireball: Yes, this then can forward the string to the constructor of the appropriate IOHandler (which also take their path-strings by value).

src/IO/ADIOS/ADIOS1IOHandler.cpp
```diff
@@ -218,8 +219,8 @@ ADIOS1IOHandlerImpl::init()
 #endif
 
 #if openPMD_HAVE_ADIOS1
-ADIOS1IOHandler::ADIOS1IOHandler(std::string const& path, AccessType at)
-        : AbstractIOHandler(path, at),
+ADIOS1IOHandler::ADIOS1IOHandler(std::string path, AccessType at)
+        : AbstractIOHandler(std::move(path), at),
```
anokfireball, line 223: You'd hope today's compilers were all smart enough to do this on their own, but it's good to state the intention nonetheless.

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
```diff
@@ -76,7 +76,7 @@ namespace openPMD
         void listAttributes(Writable*, Parameter< Operation::LIST_ATTS > &) override;
 
         virtual int64_t open_write(Writable *);
-        virtual ADIOS_FILE* open_read(std::string const& name);
+        virtual ADIOS_FILE* open_read(std::string name);
```
anokfireball: This also just takes the ```.c_str()``` without sinking the string itself.

anokfireball (approved):

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
ax3l: yep, reverted this one

Issue #312, opened 2018-08-27
Python: Read the Write Example


ax3l:
The 0.3.1-alpha python write example HDF5 files lack the `geometry` attribute. Reading them in `fileBased` again fails.

Python: adding
```python
    del series


    s = openPMD.Series("../samples/3_write_serial_py.h5",
        openPMD.Access_Type.read_only)
```
to `3_write_serial.py` fails on missing `geometry` attribute.

**Expected behavior**
Do not fail :-)

**Software Environment:**
 - version of openPMD-api: 0.3.1-alpha
 - installed openPMD-api via: conda-forge
 - operating system: OSX
 - machine: [Are you running on a public cluster? It's likely we compute on it as well!]
 - name and version of Python implementation: [e.g. CPython 3.6]
 - version of HDF5: [e.g. 1.10.1]
 - version of ADIOS1: [e.g. 1.13.1]
 - name and version of MPI: [e.g. OpenMPI 3.0.0]

**Additional context**

This bug is already fixed in `dev`, tested with 1e9c2200069a4767058e2ca4aedfa3f31e3cdde3.

anokfireball:
The symptom of the bug report is not the actual cause. The behaviour was a side-effect of https://github.com/openPMD/openPMD-api/pull/156. File-based handles were incorrectly over-written and files were thus only partially complete.
~(I am unsure which commit this was fixed in.)~ See https://github.com/openPMD/openPMD-api/pull/193.

ax3l:
But aren't both linked PRs already in `0.3.1-alpha`? I would guess that actually #297 might have fixed it?

anokfireball:
Both are involved, actually. #193 restores the hierarchy-relationship that was broken by #156 ([here](https://github.com/C0nsultant/libopenPMD/blob/7a33bee89f5da63b846151fb6e98f461210a3a60/src/Mesh.cpp#L190)), and #297 writes it to disk at the correct point in time ([here](https://github.com/C0nsultant/libopenPMD/blob/414f2ddbdd1e0542d6f82aa32885e3adbfe7eef4/src/Mesh.cpp#L199)).

PR #309, opened 2018-08-16
Handle zero-extent dataset creation and dataset write

anokfireball:
Adds tests for #135.

PR #313, opened 2018-08-27
Version: 0.4.0-alpha

ax3l:
Finalize the 0.4.0 release: hardened output.

I would like to release this after #309 is merged since the latest updates increased stability a lot and we have a user in-house that suffers from #312


anokfireball (approved):

ax3l:
Will merge - tag - release - update `dev` version when I am at home, later this evening.

Issue #171, opened 2018-05-14
Provide copying mechanism for frontend handles


anokfireball:
As a continuation of #156, a mechanism to allow detatched copies of frontend objects (i.e. ```Attributable```s) should be provided. While it is possible to create two objects with the same parameters explicitly, it is cumbersome e.g. when assigning multiple iterations in a series with the same openPMD data and differing record data.

(Shamelessly stealing numpy's lingo with [```.copy()```](https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.copy.html) to have familiar syntax and semantics.)

```cpp
Series s = Series("./out.h5", AccessType::CREATE);

Iteration i1 = s.iterations[1];

/* merely creates a view into i1
 * every modification to i1 is reflected in i2 (and vice versa) */
s.iterations[2] = i1;

/* have an independent copy of the iteration that is not merely a view
 * any modification to i1 will not be reflected in i2 (and vice versa) */
s.iterations[3] = i1.copy();
```

anokfireball:
Going off of https://github.com/openPMD/openPMD-api/issues/290#issuecomment-407335778, I doubt we can make ```Attributable::copy()``` happen.
The user should not be able to create dangling objects at all, especially not this easily.
This problem could be avoided by copying values into an existing (non-dangling) object inside a ```openPMD::Series``` structure, e.g. with ```Iteration::copyTo(Iteration&)```.

ax3l:
Ok, but we should be able to implement `it.copy()`, etc.?

PR #314, opened 2018-08-27
Python: Extended Write Example

ax3l:
Add the `7_extended_write_serial` example in python.

### DIFF
--- a/examples/7_extended_write_serial.py
+++ b/examples/7_extended_write_serial.py
@@ -169,16 +169,20 @@ if __name__ == "__main__":
         electrons["positionOffset"]["x"].store_chunk(o, e, partial_particleOff)
 
         electrons.particle_patches["numParticles"][SCALAR].store(
-            i, np.uint64(numParticles))
+            i, np.array([numParticles], dtype=np.uint64))
         electrons.particle_patches["numParticlesOffset"][SCALAR].store(
-            i, np.uint64(numParticlesOffset))
+            i, np.array([numParticlesOffset], dtype=np.uint64))
 
         electrons.particle_patches["offset"]["x"].store(
-            i, particle_position[numParticlesOffset])
+            i,
+            np.array([particle_position[numParticlesOffset]],
+                     dtype=np.float32))
         electrons.particle_patches["extent"]["x"].store(
             i,
-            particle_position[numParticlesOffset + numParticles - 1] -
-            particle_position[numParticlesOffset])
+            np.array([
+                particle_position[numParticlesOffset + numParticles - 1] -
+                particle_position[numParticlesOffset]
+            ], dtype=np.float32))
 
     mesh["y"].reset_dataset(d)
     mesh["y"].set_unit_SI(4)
--- a/examples/7_extended_write_serial.py
+++ b/examples/7_extended_write_serial.py
@@ -1,0 +1,194 @@
+#!/usr/bin/env python
+"""
+This file is part of the openPMD-api.
+
+Copyright 2018 openPMD contributors
+Authors: Axel Huebl, Fabian Koller
+License: LGPLv3+
+"""
+from openPMD import Series, Access_Type, Dataset, Mesh_Record_Component, \
+    Unit_Dimension
+import numpy as np
+
+
+SCALAR = Mesh_Record_Component.SCALAR
+
+
+if __name__ == "__main__":
+    # open file for writing
+    f = Series(
+        "working/directory/2D_simData_py.h5",
+        Access_Type.create
+    )
+
+    # all required openPMD attributes will be set to reasonable default values
+    # (all ones, all zeros, empty strings,...)
+    # manually setting them enforces the openPMD standard
+    f.set_meshes_path("custom_meshes_path")
+    f.set_particles_path("long_and_very_custom_particles_path")
+
+    # it is possible to add and remove attributes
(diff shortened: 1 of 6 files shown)


src/binding/python/PatchRecordComponent.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+void init_PatchRecordComponent(py::module &m) {
+    py::class_<PatchRecordComponent, BaseRecordComponent >(m, "Patch_Record_Component")
+        .def("set_unit_SI", &PatchRecordComponent::setUnitSI)
+        .def("reset_dataset", &PatchRecordComponent::resetDataset)
+        .def_property_readonly("ndims", &PatchRecordComponent::getDimensionality)
+        .def_property_readonly("shape", &PatchRecordComponent::getExtent)
+
+        /*
+        template< typename T >
```
ax3l: todo: implement

examples/7_extended_write_serial.py
```diff
@@ -0,0 +1,195 @@
...
+    f.delete_attribute("custom_attribute_name")
+
+    # setting attributes can be chained in JS-like syntax for compact code
+    tmpItObj = f.iterations[1] \
+        .set_time(42.0) \
+        .set_dt(1.0) \
+        .set_time_unit_SI(1.39e-16)
+    # everything that is accessed with [] should be interpreted as permanent
+    # storage the objects sunk into these locations are deep copies
+    del tmpItObj
```
anokfireball: Is this `del` supposed to signal that only the *handle* and not the underlying sunk resource is deleted?
If so, a comment to explain what is going on should be added.

examples/7_extended_write_serial.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # this is a reference to an iteration
+    reference = f.iterations[1]
+    reference.set_comment("Modifications to a reference will always be visible"
+                          " in the output")
+    del reference
+
+    # alternatively, a copy may be created and later re-assigned to
+    # f.iterations[1]
+    copy = f.iterations[1]  # TODO .copy()
```
anokfireball, line 59: This is actually outdated in the C++ example. Please add the **TODO** there as well.

examples/7_extended_write_serial.py
```diff
@@ -0,0 +1,195 @@
...
+
+    cur_it = f.iterations[1]
+
+    # the underlying concept for numeric data is the openPMD Record
+    # https://github.com/openPMD/openPMD-standard/blob/upcoming-1.0.1/STANDARD.md#scalar-vector-and-tensor-records
+    # Meshes are specialized records
+    cur_it.meshes["generic_2D_field"].set_unit_dimension(
+        {Unit_Dimension.L: -3, Unit_Dimension.M: 1})
+
+    # as this is a reference, it modifies the original resource
```
anokfireball, line 75: Outdated in the C++ example as well.

examples/7_extended_write_serial.py
```diff
@@ -0,0 +1,195 @@
...
+    lowRez = cur_it.meshes["generic_2D_field"]
+    lowRez.set_grid_spacing([1, 1, 1, 1, 1, 1]) \
```
anokfireball: This `gridSpacing` would actually make it a 6D field.

anokfireball (approved):
Python example looks good, but C++ needs some love. Will do that when tackling #171.

examples/7_extended_write_serial.py
ax3l: Yes, I thought so. Let's update both.

examples/7_extended_write_serial.py
ax3l: Let's check this again when updating the C++ example

examples/7_extended_write_serial.py
ax3l: huh, don't we have the same in the C++ example? Should be three values in both examples, no?

anokfireball: The [C++ code](https://github.com/openPMD/openPMD-api/blob/0.6.0-alpha/examples/7_extended_write_serial.cpp#L98) for this reads
```cpp
/* construct a vector of [6, 1] */
lowRez.setGridSpacing(std::vector< double >{6, 1}).setGridGlobalOffset({0, 600});
```
Your interpretation was
```cpp
/* construct a vector of six 1's */
lowRez.setGridSpacing(std::vector< double >(6, 1)).setGridGlobalOffset({0, 600});
```

> Should be three values in both examples, no?

As this is a **2D** mesh, it should contain two values: `[6, 1]`.

https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#required-attributes-for-each-mesh-record

> `gridSpacing`
type: 1-dimensional array containing N *(floatX)* elements, where N is the **number of dimensions in the simulation**

ax3l: Oh wow, that's dangerous xD
I thought the initializer-list constructor should read `({6, 1})` but your code is indeed ok and does exactly what you say.

examples/7_extended_write_serial.py
ax3l: probably not needed anymore. I did not control the garbage collector properly before `0.6.0-alpha`

ax3l:
Ah cheese, I should remove Python 2 builds altogether.

PR #318, opened 2018-08-29
Numpy: float is double

ax3l:
As it turns out, `np.float` is an alias for `float64`.

Broke working with actual single precision floats in the Python bindings.

I found no way to underly the actual C type `float` yet, so we assume it's `float32` now... Gosh!
  https://docs.scipy.org/doc/numpy/user/basics.types.html

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,10 @@ Features
 Bug Fixes
 """""""""
 
+- Python
+
+  - single precision support: ``numpy.float`` is an alias for ``float64`` #318
+
 Other
 """""
 
--- a/src/binding/python/BaseRecordComponent.cpp
+++ b/src/binding/python/BaseRecordComponent.cpp
@@ -48,6 +48,7 @@ void init_BaseRecordComponent(py::module &m) {
         .def_property_readonly("dtype", [](BaseRecordComponent & brc) {
             using DT = Datatype;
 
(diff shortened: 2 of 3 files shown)


anokfireball:
Do you think the Python bindings are a point where datatype tests like in [```CoreTest```](https://github.com/openPMD/openPMD-api/blob/dev/test/CoreTest.cpp#L26-L80) and [```SerialIOTest```](https://github.com/openPMD/openPMD-api/blob/dev/test/SerialIOTest.cpp#L867-L942) are possible?

ax3l:
Sure, we should add the equivalents of those to the [python unit tests](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/test/python/unittest/API/APITest.py#L154-L156) as well

anokfireball:
The bulk of those tests are Attribute datatype tests. As attribute setting is not yet implemented in the Python [binding](https://github.com/openPMD/openPMD-api/blob/a12c330ae7d92721245533d229a07fb804e5b340/src/binding/python/Attributable.cpp), the only thing that can be tested are the dataset datatypes, which now all have explicit bit-width anyway.
The unit tests can be deferred to either #249 or #314.

anokfireball (approved):

ax3l:
Yep. The datatypes are a bit half-way implemented in the Python API so far. I already return proper numpy `.dtype`s but setting them is currently still done via our `enum`s instead of directly passing `numpy.dtype` types.

PR #320, opened 2018-08-29
Numpy: but C float is single

ax3l:
The proper naming for the C-type float in numpy is `single`.

```python
import numpy as np

np.dtype("float")
# dtype('float64')

np.dtype("single")
# dtype('float32')
```

Follow-up to #318 

Ref: https://github.com/numpy/numpy/issues/11833

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,7 +22,7 @@ Bug Fixes
 
 - Python
 
-  - single precision support: ``numpy.float`` is an alias for ``float64`` #318
+  - single precision support: ``numpy.float`` is an alias for ``float64`` #318 #320
   - ``Dataset`` method namings to underscores #319
 
 Other
--- a/src/binding/python/BaseRecordComponent.cpp
+++ b/src/binding/python/BaseRecordComponent.cpp
@@ -70,7 +70,7 @@ void init_BaseRecordComponent(py::module &m) {
             else if( brc.getDatatype() == DT::DOUBLE )
                 return py::dtype("double");
             else if( brc.getDatatype() == DT::FLOAT )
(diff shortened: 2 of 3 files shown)


anokfireball (approved):
A debatable decision on numpy's side, really. Their desire to establish the larger bit-width of the two as default is  understandable, but this breaks with all estabilshed languages so much that you have to ask their devs (NOT docs) what the correct datatype is.

ax3l:
Improved docs: https://github.com/numpy/numpy/pull/11837

Docs I missed that already mentioned `np.single`: https://docs.scipy.org/doc/numpy-1.14.0/reference/arrays.scalars.html

CHANGELOG.rst
```diff
@@ -22,7 +22,7 @@ Bug Fixes
 
 - Python
 
-  - single precision support: ``numpy.float`` is an alias for ``float64`` #318
+  - single precision support: ``numpy.float`` is an alias for ``float64`` #318 #320
```
eric-wieser, line 25: This is not true - `np.float` is an alias for `builtins.float`

ax3l: @eric-wieser thanks, forgot to update that in the fix!
Does "builtin" refer to the Python builtin float type? I found only [this](https://docs.python.org/3/tutorial/floatingpoint.html#representation-error) in the official docs:
> [on] almost all platforms map Python floats to IEEE-754 “double precision”.

and am now wondering: which (Python supported) platforms are *not* defaulting floats to double? :-)

eric-wieser: I mean that if you type `np.float is float` in a console (or `import builtins; builtins.float is np.float`), it returns `True`.

`np.float` should never be used - either use `np.float_` for the numpy type, or `float` for the pure-python type.

PR #322, opened 2018-08-30
Store: More Detailed Type Mismatch Error

ax3l:
Improve the error message for mismatched datatype and data passed to `store_chunk` by including both as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,13 +23,14 @@ Bug Fixes
 - avoid impact on unrelated classes in invasive tests #324
 - Python
 
-  - single precision support: ``numpy.float`` is an alias for ``float64`` #318 #320
+  - single precision support: ``numpy.float`` is an alias for ``builtins.float`` #318 #320
   - ``Dataset`` method namings to underscores #319
 
 Other
 """""
 
 - CMake: invasive tests not enabled by default #323
+- ``store_chunk``: more detailed type mismatch error #322
 
 
 0.4.0-alpha
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -28,6 +28,7 @@
 #include <limits>
 #include <queue>
 #include <string>
+#include <sstream>
 #include <stdexcept>
 
 // expose private and protected members for invasive testing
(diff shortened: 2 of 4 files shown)


include/openPMD/RecordComponent.hpp
```diff
@@ -174,7 +175,15 @@ RecordComponent::storeChunk(Offset o, Extent e, std::shared_ptr<T> data)
         throw std::runtime_error("Unallocated pointer passed during chunk store.");
     Datatype dtype = determineDatatype(data);
     if( dtype != getDatatype() )
-        throw std::runtime_error("Datatypes of chunk and dataset do not match.");
+    {
+        std::ostringstream oss;
+        oss << "Datatypes of chunk ("
+            << getDatatype()
```
ax3l: @C0nsultant did I mix up the two? is `dtype` of the incoming data the current `chunk` or the `dataset`? ;-)

ax3l: yep, I think I mixed it up. will also rename `of chunk` to `of chunk data`

anokfireball: [```getDatatype()```](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/backend/BaseRecordComponent.cpp#L47-L51) is equal to the datatype assigned to a dataset/```RecordComponent``` through [```resetDataset()```](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/RecordComponent.cpp#L46-L60).
So yes, it is actually the wrong way around.

ax3l: thx, already fixed locally. will push the rebase when #324 is in later today :)

CHANGELOG.rst
```diff
@@ -30,6 +30,7 @@ Other
 """""
 
 - CMake: invasive tests not enabled by default #323
+- ``store_chunk``: more detailed detailed type mismatch error #322
```
ax3l: double `detailed` -.-

include/openPMD/RecordComponent.hpp
ax3l: updated :)

test/CoreTest.cpp
```diff
@@ -567,6 +570,8 @@ TEST_CASE( "wrapper_test", "[core]" )
     REQUIRE(o.iterations[6].particles["electrons"].particlePatches["numParticles"][RecordComponent::SCALAR].m_chunks->size() == 1);
     REQUIRE(pp["numParticles"][RecordComponent::SCALAR].m_chunks->size() == 1);
 #endif
+    REQUIRE_THROWS_WITH(o.iterations[6].particles["electrons"].particlePatches["numParticles"][RecordComponent::SCALAR].store(idx+1, 42.),
+                        Catch::Equals("Datatypes of chunk data (DOUBLE) and dataset (UINT64) do not match."));
```
anokfireball, line 574: Kudos for also including unit tests!

ax3l: sneakily improving coverage, one corner case at a time :)

PR #325, opened 2018-08-30
File Error: Remove const char signature

ax3l:
Remove the C-like string constructor.
- c-strings are casted to c++ strings in the constructor anyway
- c-strings are not C++-ish
- coverage shows the constructor is unused

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Other
 
 - CMake: invasive tests not enabled by default #323
 - ``store_chunk``: more detailed type mismatch error #322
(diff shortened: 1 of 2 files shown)


anokfireball (approved):

anokfireball:
There's also the same pattern in [```Attributable.hpp```](https://coveralls.io/builds/18759862/source?filename=include/openPMD/backend/Attributable.hpp#L53).
I tried to push the changes, but as it seems `Allow edits from maintainers.` is not enabled?

ax3l:
Weird, I do have that allowed (as it is by default).
Did you push to the right repo?

PR #326, opened 2018-08-30
Dataset: Fix setCompression Detection

ax3l:
Fix the warning and error logic in `Dataset::setCompression`.

### DIFF
--- a/src/Dataset.cpp
+++ b/src/Dataset.cpp
@@ -62,9 +62,11 @@ Dataset::setChunkSize(Extent const& cs)
 Dataset&
 Dataset::setCompression(std::string const& format, uint8_t const level)
 {
-    if( (format == "zlib" || format == "gzip" || format == "deflate")
-        && level > 9 )
-        throw std::runtime_error("Compression level out of range for " + format);
+    if(format == "zlib" || format == "gzip" || format == "deflate")
+    {
+        if(level > 9)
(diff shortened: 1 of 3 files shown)


src/Dataset.cpp
```diff
@@ -62,9 +62,11 @@ Dataset::setChunkSize(Extent const& cs)
 Dataset&
 Dataset::setCompression(std::string const& format, uint8_t const level)
 {
-    if( (format == "zlib" || format == "gzip" || format == "deflate")
-        && level > 9 )
-        throw std::runtime_error("Compression level out of range for " + format);
+    if(format.compare("zlib") == 0 || format.compare("gzip") == 0 || format.compare("deflate") == 0)
```
anokfireball: What is your goal in using ```std::string::compare()``` over ```std::operator==(std::string const&, char* const)```?
Looking at the [C++ standard](https://timsong-cpp.github.io/cppwp/n4140/string::operator==#4), you're just replicating what is happening already.

ax3l: Uh, I just checked quickly in [cplusplus.com](http://www.cplusplus.com/reference/string/string/compare/) and did not see the `operator==`. Will keep using `==` instead, it's indeed there!
https://en.cppreference.com/w/cpp/string/basic_string

anokfireball (approved):

PR #330, opened 2018-09-03
set_attribute: broken numpy, list and string support

### LINKED ISSUE #329: How to name comment
I can make a comment using set_comment in Value. But the name is always "comment" in Name. How can I name it as I  want?

ax3l:
Fix `Attributable` write:

- [x] python `buildin.float` was down-casted in binding to single, now double
- [x] string writing was broken
- [x] add numpy scalar and array support
- [x] add python array support
- [x] add list support
- [x] N-D lists/arrays/np.arrays will be flattended to 1D
- [x] unit test added for HDF5 and ADIOS1 backend

Left for future:
- [ ] list & array of numpy scalars are cropped to native python type (ok-ish - use np.array instead!)
- [ ] `get_attribute`: return lists/arrays as np.arrays, not Python lists #348
- [ ] verify numpy version at RT via `auto numpy = py::module::import("numpy"); auto version = numpy.attr("__version__");`  #349

`set_attribute` requires numpy 1.15.0+ in order to handle scalars properly (`ndim == 0`).

Refs:
  https://github.com/numpy/numpy/issues/10265
  https://github.com/pybind/pybind11/issues/1224#issuecomment-354357392

Fix #329

### DIFF
--- a/src/binding/python/Attributable.cpp
+++ b/src/binding/python/Attributable.cpp
@@ -49,6 +49,166 @@ using namespace openPMD;
 using PyAttributeKeys = std::vector< std::string >;
 //PYBIND11_MAKE_OPAQUE(PyAttributeKeys)
 
+bool setAttributeFromBufferInfo(
+    Attributable & attr,
+    std::string const& key,
+    py::buffer_info const& buf
+) {
+    // Numpy: Handling of arrays and scalars
+    // work-around for https://github.com/pybind/pybind11/issues/1224
+    // -> passing numpy scalars as buffers needs numpy 1.15+
+    //    https://github.com/numpy/numpy/issues/10265
+    //    https://github.com/pybind/pybind11/issues/1224#issuecomment-354357392
+    // scalars, see PEP 3118
+    // requires Numpy 1.15+
+    if( buf.ndim == 0 ) {
+        // refs:
+        //   https://docs.scipy.org/doc/numpy-1.15.0/reference/arrays.interface.html
+        //   https://docs.python.org/3/library/struct.html#format-characters
+        // std::cout << "  scalar type '" << buf.format << "'" << std::endl;
+        // typestring: encoding + type + number of bytes
+        if( buf.format.find("?") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<bool*>(buf.ptr) );
+        else if( buf.format.find("h") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<short*>(buf.ptr) );
+        else if( buf.format.find("i") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<int*>(buf.ptr) );
+        else if( buf.format.find("l") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<long*>(buf.ptr) );
+        else if( buf.format.find("q") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<long long*>(buf.ptr) );
+        else if( buf.format.find("H") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<unsigned short*>(buf.ptr) );
+        else if( buf.format.find("I") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<unsigned int*>(buf.ptr) );
+        else if( buf.format.find("L") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<unsigned long*>(buf.ptr) );
+        else if( buf.format.find("Q") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<unsigned long long*>(buf.ptr) );
+        else if( buf.format.find("f") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<float*>(buf.ptr) );
+        else if( buf.format.find("d") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<double*>(buf.ptr) );
+        else if( buf.format.find("g") != std::string::npos )
+            return attr.setAttribute( key, *static_cast<long double*>(buf.ptr) );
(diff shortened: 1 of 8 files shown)


test/python/unittest/API/APITest.py
```diff
@@ -104,6 +106,188 @@ def testParticleData(self):
...
+            self.assertListEqual(series.get_attribute("l_int64"),
+                                 [np.int64(45), np.int64(48)])
+            self.assertListEqual(series.get_attribute("l_uint16"),
+                                 [np.uint16(23), np.uint16(26)])
+            self.assertListEqual(series.get_attribute("l_uint32"),
+                                 [np.uint32(34), np.uint32(37)])
+            self.assertListEqual(series.get_attribute("l_uint64"),
+                                 [np.uint64(45), np.uint64(48)])
+            # self.assertListEqual(series.get_attribute("l_single"),
+            #     [np.single(5.6), np.single(5.9)])
```
ax3l: bugged?

test/python/unittest/API/APITest.py
```diff
@@ -104,6 +106,188 @@ def testParticleData(self):
...
+            # list of ...
+            series.set_attribute("l_int16", [np.int16(23), np.int16(26)])
+            series.set_attribute("l_int32", [np.int32(34), np.int32(37)])
+            series.set_attribute("l_int64", [np.int64(45), np.int64(48)])
+            series.set_attribute("l_uint16", [np.uint16(23), np.uint16(26)])
+            series.set_attribute("l_uint32", [np.uint32(34), np.uint32(37)])
+            series.set_attribute("l_uint64", [np.uint64(45), np.uint64(48)])
+            series.set_attribute("l_single", [np.single(5.6), np.single(5.9)])
+            series.set_attribute("l_double", [np.double(6.7), np.double(7.1)])
+            series.set_attribute("l_longdouble",
```
ax3l, line 170: bug: stored as `native double` in HDF5
(not that this is clever to do. one should use a `np.array(..., dtype=np.longdouble)` instead!)

test/python/unittest/API/APITest.py
```diff
@@ -104,6 +106,188 @@ def testParticleData(self):
...
+            series.set_attribute("arr_double",
+                                 (np.double(6.7), np.double(7.1), ))
+            # list of ...
+            series.set_attribute("l_int16", [np.int16(23), np.int16(26)])
+            series.set_attribute("l_int32", [np.int32(34), np.int32(37)])
+            series.set_attribute("l_int64", [np.int64(45), np.int64(48)])
+            series.set_attribute("l_uint16", [np.uint16(23), np.uint16(26)])
+            series.set_attribute("l_uint32", [np.uint32(34), np.uint32(37)])
+            series.set_attribute("l_uint64", [np.uint64(45), np.uint64(48)])
+            series.set_attribute("l_single", [np.single(5.6), np.single(5.9)])
```
ax3l, line 168: bug: stored as `native int` in HDF5
(not that this is clever to do. one should use a `np.array(..., dtype=np.single)` instead!)

src/binding/python/Attributable.cpp
```diff
@@ -49,14 +51,151 @@ using namespace openPMD;
...
+    // scalars, see PEP 3118
+    // requires Numpy 1.15+
+    if( buf.ndim == 0 ) {
+        // refs:
+        //   https://docs.scipy.org/doc/numpy-1.15.0/reference/arrays.interface.html
+        //   https://docs.python.org/3/library/struct.html#format-characters
+        // std::cout << "  scalar type '" << buf.format << "'" << std::endl;
+        // typestring: encoding + type + number of bytes
+        if( buf.format.find("h") != std::string::npos )
+            attr.setAttribute( key, *static_cast<short*>(buf.ptr) );
```
ax3l: probably have to add `const_cast<...>(...)`s here for MSVC

src/binding/python/Attributable.cpp
```diff
@@ -49,14 +51,151 @@ using namespace openPMD;
...
+        //   https://docs.scipy.org/doc/numpy-1.15.0/reference/arrays.interface.html
+        //   https://docs.python.org/3/library/struct.html#format-characters
+        // std::cout << "  scalar type '" << buf.format << "'" << std::endl;
+        // typestring: encoding + type + number of bytes
+        if( buf.format.find("h") != std::string::npos )
+            attr.setAttribute( key, *const_cast<short*>(static_cast<short*>(buf.ptr)) );
+        else if( buf.format.find("i") != std::string::npos )
+            attr.setAttribute( key, *const_cast<int*>(static_cast<int*>(buf.ptr)) );
+        else if( buf.format.find("l") != std::string::npos )
+            attr.setAttribute( key, *const_cast<long*>(static_cast<long*>(buf.ptr)) );
```
ax3l: @C0nsultant hm, somehow this is still `const` xD Am I missing something?
https://travis-ci.org/openPMD/openPMD-api/jobs/424882144

Could make the `py::buffer_info const& buf` non-const...

ax3l: Huh, or I really need to write `Attribute(*const_cast<T*>(&value))` in the C++ backend...

ax3l: Hm, I really need to translate from the type aliases such as long to a bit-ed type. OMG...

ax3l: e.g. on clang OSX in Travis, `long` is `long long` but only one of them is in the variant list.
On MSVC, `long` and `int` are identical, etc.

Problem is, that I have to start with those type and translate them to fixed size types, simply because numpy/python only gives me the generic types in the dtype.

ax3l:
@C0nsultant we can merge this if you like

PR #333, opened 2018-09-06
[WIP] setAttribute with int/long on OSX & Win

ax3l:
Test that writing attributes with non-fixed size types works.

Seems to crash on some environments, such as:
- OSX in Travis, `long` is `long long`
- MSVC, `long` and `int` are identical

Problem is, that only one of them is in the variant list and it seems automatic detection that they are the same is not performed in `variant` on the two obscure OSes above. We should therefore write a trait, that converts types such as `int` to their current, platform-specific fixed-size type during access, e.g. in `Attributable` where `Attribute(value)` is called.

Or in simple words: on OSX and with MSVC on Windows this did not compile:
```c++
        long l = 32;
        s.setAttribute("long", l);
```

since `Attribute(value)` complains that our `Variant` does not contain `long` (but e.g. only `int` and `long long` although `int` is equivalent to `long`).

[First seen here.](https://github.com/openPMD/openPMD-api/pull/330#discussion_r215350563)

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -97,9 +97,15 @@ public:
      * @param   key     Key (i.e. name) to identify and store an Attributes value by.
      * @param   value   Value of Attribute stored with the provided key.
      * @return  true if key was already present, false otherwise
+     *
+     * @{
      */
     template< typename T >
     bool setAttribute(std::string const& key, T const& value);
+    bool setAttribute(std::string const& key, char const value[]);
+    /** @}
+     */
+
     /** Retrieve value of Attribute stored with provided key.
      *
      * @throw   no_such_attribute_error If no Attribute is currently stored with the provided key.
@@ -107,6 +113,7 @@ public:
      * @return  Stored Attribute in Variant form.
      */
     Attribute getAttribute(std::string const& key) const;
+
     /** Remove Attribute of provided value both logically and physically.
      *
      * @param   key Key (i.e. name) of the Attribute to remove.
@@ -230,6 +237,11 @@ Attributable::setAttribute(std::string const& key, T const& value)
                                    std::make_pair(key, Attribute(value)));
         return false;
     }
+}
+inline bool
+Attributable::setAttribute(std::string const& key, char const value[])
+{
+    return this->setAttribute(key, std::string(value));
 }
 
 extern template
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -906,8 +906,8 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
         s.setAttribute("float", f);
         double d = 1.e64;
         s.setAttribute("double", d);
-        long double l = 1.e80L;
-        s.setAttribute("longdouble", l);
+        long double ld = 1.e80L;
+        s.setAttribute("longdouble", ld);
         std::string str = "string";
         s.setAttribute("string", str);
         s.setAttribute("vecChar", std::vector< char >({'c', 'h', 'a', 'r'}));
@@ -923,6 +923,8 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
         s.setAttribute("vecLongdouble", std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
         s.setAttribute("vecString", std::vector< std::string >({"vector", "of", "strings"}));
         s.setAttribute("bool", true);
+
+        // non-fixed size integer types
         short ss = 16;
         s.setAttribute("short", ss);
         int si = 32;
@@ -939,6 +941,14 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
         s.setAttribute("ulong", ul);
         unsigned long long ull = 128u;
         s.setAttribute("ulonglong", ull);
+        s.setAttribute("vecShort", std::vector< short >({32766, 32767}));
+        s.setAttribute("vecInt", std::vector< int >({32766, 32767}));
+        s.setAttribute("vecLong", std::vector< long >({2147483646, 2147483647}));
+        s.setAttribute("vecLongLong", std::vector< long long >({2147483644, 2147483643}));
+        s.setAttribute("vecUShort", std::vector< unsigned short >({65534u, 65535u}));
+        s.setAttribute("vecUInt", std::vector< unsigned int >({65533u, 65531u}));
+        s.setAttribute("vecULong", std::vector< unsigned long >({65532u, 65530u}));
+        s.setAttribute("vecULongLong", std::vector< unsigned long long >({65531u, 65529u}));
     }
 
     Series s = Series("../samples/dtype_test.h5", AccessType::READ_ONLY);
@@ -983,6 +993,15 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
     REQUIRE(s.getAttribute("uint").dtype == Datatype::UINT);
     REQUIRE(s.getAttribute("ulong").dtype == Datatype::ULONG);
     REQUIRE(s.getAttribute("ulonglong").dtype == Datatype::ULONGLONG);
+
+    REQUIRE(s.getAttribute("vecShort").dtype == Datatype::VEC_SHORT);
+    REQUIRE(s.getAttribute("vecInt").dtype == Datatype::VEC_INT);
+    REQUIRE(s.getAttribute("vecLong").dtype == Datatype::VEC_LONG);
+    REQUIRE(s.getAttribute("vecLongLong").dtype == Datatype::VEC_LONGLONG);
+    REQUIRE(s.getAttribute("vecUShort").dtype == Datatype::VEC_USHORT);
+    REQUIRE(s.getAttribute("vecUInt").dtype == Datatype::VEC_UINT);
+    REQUIRE(s.getAttribute("vecULong").dtype == Datatype::VEC_ULONG);
+    REQUIRE(s.getAttribute("vecULongLong").dtype == Datatype::VEC_ULONGLONG);
 #endif
     REQUIRE(isSame(s.getAttribute("short").dtype, Datatype::SHORT));
     REQUIRE(isSame(s.getAttribute("int").dtype, Datatype::INT));
@@ -992,6 +1011,15 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
     REQUIRE(isSame(s.getAttribute("uint").dtype, Datatype::UINT));
     REQUIRE(isSame(s.getAttribute("ulong").dtype, Datatype::ULONG));
     REQUIRE(isSame(s.getAttribute("ulonglong").dtype, Datatype::ULONGLONG));
+
+    REQUIRE(isSame(s.getAttribute("vecShort").dtype, Datatype::VEC_SHORT));
+    REQUIRE(isSame(s.getAttribute("vecInt").dtype, Datatype::VEC_INT));
+    REQUIRE(isSame(s.getAttribute("vecLong").dtype, Datatype::VEC_LONG));
+    REQUIRE(isSame(s.getAttribute("vecLongLong").dtype, Datatype::VEC_LONGLONG));
+    REQUIRE(isSame(s.getAttribute("vecUShort").dtype, Datatype::VEC_USHORT));
+    REQUIRE(isSame(s.getAttribute("vecUInt").dtype, Datatype::VEC_UINT));
+    REQUIRE(isSame(s.getAttribute("vecULong").dtype, Datatype::VEC_ULONG));
+    REQUIRE(isSame(s.getAttribute("vecULongLong").dtype, Datatype::VEC_ULONGLONG));
 }
 
 TEST_CASE( "hdf5_write_test", "[serial][hdf5]" )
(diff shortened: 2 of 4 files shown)


anokfireball:
I'm not yet fully grapsing the problem.
My assumption was that datatype derivation in `Variant` worked similar to what I did manaully in [```determineDatatype()```](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/Datatype.hpp#L103-L140):
```cpp
using Attribute = Variant< T_dtype_enum, T_one, T_two, /* ... */ >;
T_two two = /* ... */;
Attribute(two).dtype == determineDatatype< T_two >();
```

Which it probably does, but as
```cpp
Attribute(T_one()).dtype == Attribute(T_two()).dtype
```
is apparently true for some platforms and combinations, this might break the logic of correlating datatypes by simply looking up the [enum position](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/auxiliary/Variant.hpp#L58).

Now to the problem at the pillar:
> We should therefore write a trait, that converts types such as int to their current, platform-specific fixed-size type during access

The core of this already exists as run-time behaviour in the form of [```determineDatatype()```](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/Datatype.hpp#L103-L140):
Instead of 
```cpp
template< class T_DTYPES, typename ... T >
class Variant
{
public:
    using resource = variantSrc::variant< T ... >;

    Variant(resource r)
            : dtype{static_cast<T_DTYPES>(r.index())},
              m_data{r}
    { }
} //Variant
```
we would have to explicitly cast the passed resource 
```cpp
template< class T_DTYPES, typename ... T >
class Variant
{
public:
    using resource = variantSrc::variant< T ... >;

    template< typename U >
    Variant(U&& u)
            : dtype{determineDatatype< U >()},
              m_data{resource(static_cast< YOUR_TRAIT< U >::type >(u))}
    { }
} //Variant
```

ax3l:
@C0nsultant wrote to you on Gitter, we have a solution

include/openPMD/backend/Attribute.hpp
```diff
@@ -61,4 +62,81 @@ using Attribute = auxiliary::Variant< Datatype,
...
+            U,
+            T
+        >::value;
+
+        // will throw below in case it is not T
+        using type = T;
+
+        static_assert(
+            wasFound || value,
+            "Type was not found!"
```
anokfireball: Is this the compilation error one will get when doing something like
```cpp
struct NotAnAttributeDType
{ };

Series s = /* ...  */;
auto attr = NotAnAttributeDType();
s.setAttribute("nonsense", attr);
```
?

include/openPMD/backend/Attribute.hpp
```diff
@@ -61,4 +62,81 @@ using Attribute = auxiliary::Variant< Datatype,
...
+        using type = typename FindTypeImpl<false, U, T_Available...>::type;
+    };
+} // namespace detail
+
+    /** Convert a fundamental type to its fixed size equivalent
+     *
+     * Takes a fundamental type in T_Input and converts it to its fixed size equivalent.
+     * T_Input is compared against all supported types in @see Attribute .
+     *
+     * Returns the original type of T_Input is already a fixed size type.
```
anokfireball: Typo? 
> the original type *if* T_Input

test/SerialIOTest.cpp
```diff
@@ -899,6 +899,26 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
...
+        unsigned int ui = 16u;
+        s.setAttribute("uint", ui);
+        unsigned long ul = 32u;
+        s.setAttribute("ulong", ul);
+        s.setAttribute("vecShort", std::vector< short >({32766, 32767}));
+        s.setAttribute("vecInt", std::vector< int >({32766, 32767}));
+        s.setAttribute("vecLong", std::vector< long >({2147483646, 2147483647}));
+        s.setAttribute("vecUShort", std::vector< unsigned short >({65534u, 65535u}));
+        s.setAttribute("vecUInt", std::vector< unsigned int >({65533u, 65531u}));
+        s.setAttribute("vecULong", std::vector< unsigned long >({65532u, 65530u}));
```
anokfireball: There already is a testcase in [```CoreTest.cpp```](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/test/CoreTest.cpp#L26-L80) that does similar checks. As this is (probably?) not supposed to check the workings in the backend, and rather in ```Attribute```, the tests are better suited there.

include/openPMD/backend/Attribute.hpp
ax3l: Yes, e.g. on GCC 6.3 you get (I removed the c-string handling for demonstration):
```
openPMD-api/include/openPMD/backend/Attribute.hpp:97:9:
error: static assertion failed: Type was not found!
         static_assert(
         ^~~~~~~~~~~~~
In file included from openPMD-api/include/openPMD/Iteration.hpp:23:0,
                 from openPMD-api/include/openPMD/openPMD.hpp:27,
                 from openPMD-api/test/SerialIOTest.cpp:8:
openPMD-api/include/openPMD/backend/Attributable.hpp: In instantiation of ‘bool openPMD::Attributable::setAttribute(const string&, const T&) [with T = char [59]; std::__cxx11::string = std::__cxx11::basic_string<char>]’:
openPMD-api/test/SerialIOTest.cpp:1290:80:   required from here
openPMD-api/include/openPMD/backend/Attributable.hpp:234:32: warning: the compiler can assume that the address of ‘value’ will always evaluate to ‘true’ [-Waddress]
         it->second = Attribute(static_cast<FixedSize const&>(value));
                                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
openPMD-api/include/openPMD/backend/Attributable.hpp:240:66: warning: the compiler can assume that the address of ‘value’ will always evaluate to ‘true’ [-Waddress]
                                    std::make_pair(key, Attribute(static_cast<FixedSize const&>(value))));
```

ax3l:
thx for the review, addressed all points. Will push the other commits after the test has failed, so I see if the issue is still visible as intended.

test/CoreTest.cpp
```diff
@@ -71,6 +71,56 @@ TEST_CASE( "attribute_dtype_test", "[core]" )
     REQUIRE(Datatype::ARR_DBL_7 == a.dtype);
     a = Attribute(static_cast< bool >(false));
     REQUIRE(Datatype::BOOL == a.dtype);
+
+    // translation from non-fixed size integer types
+    a = Attribute(static_cast< short >(1));
+    REQUIRE((
+        Datatype::INT16 == a.dtype || Datatype::INT32 == a.dtype || Datatype::INT64 == a.dtype
```
anokfireball: [```determineDatatype()```](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/Datatype.hpp#L103-L140) provides the datatype of explicit fixed width you are looking for. Would be a little easier to read and more explicit.
```cpp
a = Attribute(static_cast< short >(1));
REQUIRE(determineDatatype< short >() == a.dtype);
```

ax3l: Sounds legit, but wouldn't we then test `determineDatatype()` only against itself?

Anyway, the real test this adds is about the line above, if it compiles:
`a = Attribute(static_cast< short >(1));`

I just added the `REQUIRE` line as well for consistency with the existing checks.

anokfireball: > wouldn't we then test determineDatatype() only against itself?

No, because `Attribute.datatype` is *not* determined by `determineDatatype()`, but by (`std::`/`mpark::`)[`Variant`](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/auxiliary/Variant.hpp#L58) with a subsequent lookup in the `Datatype` enum.

ax3l:
@C0nsultant a note on your proposal [here](https://github.com/openPMD/openPMD-api/pull/333#issuecomment-419026604): I would like to move the trait into the `openPMD::Variant` constructor directly because it's more general. The problem is, that we don't have access to all "allowed" types at this point (aka `Attribute`) since they are set later.

Since this will be a larger refactoring, e.g. unifying `Attribute.hpp` and `Variant.hpp` I would leave this open as a possible refactoring later on. For now, I want to get #330 finished with this PR.

anokfireball (approved):

test/SerialIOTest.cpp
ax3l: Ah dang, no... I moved the test to `CoreTest` now where a (non-user typical) `Attribute()` is created directly. I did not fix that yet ([see reason below](https://github.com/openPMD/openPMD-api/pull/333#issuecomment-419088770)). I fixed the user-interface `setAttribute`.

ax3l:
@C0nsultant will have to push once more. The test needs to go into `SerialTest` again since there `setAttribute` is tested.

test/CoreTest.cpp
```diff
@@ -71,6 +71,36 @@ TEST_CASE( "attribute_dtype_test", "[core]" )
     REQUIRE(Datatype::ARR_DBL_7 == a.dtype);
     a = Attribute(static_cast< bool >(false));
     REQUIRE(Datatype::BOOL == a.dtype);
+
+    /* translation from non-fixed size integer types
+     * @todo see
```
ax3l: @C0nsultant I will leave this prepared test in here, for #333

include/openPMD/backend/Attribute.hpp
```diff
@@ -61,4 +62,81 @@ using Attribute = auxiliary::Variant< Datatype,
                             std::vector< std::string >,
                             std::array< double, 7 >,
                             bool >;
-} // openPMD
+namespace detail
+{
+    template<bool wasFound, typename U, typename T_Head, typename ...T>
+    struct FindTypeImpl
+    {
+        static constexpr bool value = std::is_same<
```
ax3l: hm, on the platforms with trouble, `is_same` will still return false although they are the same... Argh!

ax3l:
Regarding our underlying issue, this might be interesting reads:
https://github.com/MicrosoftDocs/cpp-docs/issues/411
https://en.cppreference.com/w/cpp/language/types#Integer_types

One could now probably either follow this and go for the fundamental types instead of fixed-size types in all relevant comparisons (e.g. all four `short`, `int`, `long`, `long long` instead of the three `int16_t`, `int32_t`, `int64_t`) - which is actually what we already do for floating point types - or we add our own `is_equivalent_fundamental` trait to the functionality inside this PR.

It's also interesting to see that our variadic templates in `Attribute` change the type in the parameter pack from the fixed size types (which are aliases) to their actual types (e.g. `short`, `int`, `long long` on OSX & Win). This also means there is no `cstdint` alias for `long` there.

include/openPMD/backend/Attribute.hpp
ax3l: turns out they are not the same. each of them are distinct types, just some happen to implement the same precision. `int32_t` and others are `typedef` aliases of the funnily named ones (`short`, `int`) but not vice versa and not across the latter.

test/SerialIOTest.cpp
```diff
@@ -1468,6 +1496,26 @@ TEST_CASE( "adios1_dtype_test", "[serial][adios1]" )
...
```
ax3l: read: `getAttribute` check missing

ax3l: `LongLong` missing

ax3l:
This PR is mainly solved with #337.

I will re-open what is left in here (increased coverage, maybe unique C-string signature for set-attribute, maybe the scope fix in one of the tests) into separate PRs.

include/openPMD/backend/Attributable.hpp
```diff
@@ -97,16 +97,23 @@ class Attributable
      * @param   key     Key (i.e. name) to identify and store an Attributes value by.
      * @param   value   Value of Attribute stored with the provided key.
      * @return  true if key was already present, false otherwise
+     *
+     * @{
```
anokfireball, line 101: Out of curiosity:
I am unfamiliar with this systax. Does this tell doxygen to generate only one doc for the both of these?

test/SerialIOTest.cpp
```diff
@@ -1467,7 +1495,33 @@ TEST_CASE( "adios1_dtype_test", "[serial][adios1]" )
         s.setAttribute("vecDouble", std::vector< double >({0., 1.79769e+308}));
         s.setAttribute("vecLongdouble", std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
         s.setAttribute("vecString", std::vector< std::string >({"vector", "of", "strings"}));
-        s.setAttribute("bool", static_cast< unsigned char >(true));
+        // s.setAttribute("bool", true);
```
anokfireball, line 1498: See https://github.com/openPMD/openPMD-api/pull/337#discussion_r217307170 in review comment https://github.com/openPMD/openPMD-api/pull/337#pullrequestreview-153581857 (not uncollapsed when using direct link).

test/SerialIOTest.cpp
```diff
@@ -1500,7 +1554,25 @@ TEST_CASE( "adios1_dtype_test", "[serial][adios1]" )
     REQUIRE(s.getAttribute("vecLongdouble").get< std::vector< long double > >() == std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
 #endif
     REQUIRE(s.getAttribute("vecString").get< std::vector< std::string > >() == std::vector< std::string >({"vector", "of", "strings"}));
-    REQUIRE(s.getAttribute("bool").get< unsigned char >() == static_cast< unsigned char >(true));
+    // REQUIRE(s.getAttribute("bool").get< bool >() == true);
```
anokfireball, line 1557: See https://github.com/openPMD/openPMD-api/pull/337#discussion_r217302014 in review comment https://github.com/openPMD/openPMD-api/pull/337#pullrequestreview-153581857 (not uncollapsed when using direct link).

anokfireball (approved):
LGTM

include/openPMD/backend/Attributable.hpp
ax3l: Yes, exactly. It's doxyen group-ing for exactly that use case :)

ax3l:
@C0nsultant I will rip this PR apart, since it's otherwise too confusing (https://github.com/openPMD/openPMD-api/pull/333#issuecomment-420705235)

PR #334, opened 2018-09-06
Travis: Skip clang-tidy & IWYU

ax3l:
Although both tools are very promising, we are currently not fixing their reported issues to 100%.

That would be no problem to fix them gradually, but since they do not succeed yes, travis is not caching our (dependency) build artifacts. Due to that, both take up 20min and block development time.

Proposal how to enable them: in case there are no false positives or we can add config files disabling those or actual unfixed defects, e.g. for `clang-format` we also need to [add a config file](https://github.com/ComputationalRadiationPhysics/contributing/pull/32), a PR should fix issues caused by them and enable them in the process (or later).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,105 +62,105 @@ jobs:
     ###########################################################################
     # Stage: Static Code Analysis                                             #
     ###########################################################################
-    - &static_code_cpp
-      stage: 'Static Code Analysis'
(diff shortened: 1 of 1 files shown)


anokfireball (approved):
Love Travis for what they provide, so let's not hammer them for results we don't use.

Issue #65, opened 2018-02-13
JSON


ax3l:
A (serial-only) JSON IO-backend would be cool.

ax3l:
A good backend provider for JSON is most likely the popular, header-only: https://github.com/nlohmann/json

ax3l:
With a JSON backend, validation via a [schema](https://schema.org/docs/gs.html) with existing tools / defintions such as [JSON-LD](https://en.wikipedia.org/wiki/JSON-LD) or [json-schema.org](https://json-schema.org/) could be way easier. (On a side-note, HDF5 seems to explore schema-support as well: https://2018esipsummermeeting.sched.com/event/Eypl)

ax3l:
Continued from https://github.com/openPMD/openPMD-api/pull/337#discussion_r216041160 on

> floating point datatypes already have the same issue and I ignored that so far

It's ok if a backend does not support (conserve) all data types or might even cast all floats to double or all ints to int64_t. 

In JSON, depending on the backend, we could signal different precision `floats` just by significant digits in the mantissa. (Ignore what languages make out of that, e.g. a Java app only knows one double type. In Python, we rely on Numpy to handle floats besides the one builtin.float type properly.)

With integers, we probably have to decide on a fixed type we cast all `int`s to. My idea is to avoid introducing additional annotations just for the sake of work-arounding JSON limitations, also because no standard JSON reader would understand such. But maybe you can convince me that this is useful at some points.

In the end, we will need the JSON backend in practice to (temporarily) transport int and double data with description of the meta data for small GUI-like applications. It's not a long term storage backend for primary data.

Of course, we need to make a section in the manual writing down limitations of specific backends (e.g. no MPI support or no support for certain types or cast of types during storage, etc.).

franzpoeschel:
Ok, that sounds sensible.
> or might even cast all floats to double or all ints to int64_t.

Not a huge issue in JSON anyway since values are pretty-printed rather than casted.

>we could signal different precision floats just by significant digits in the mantissa.

Would the number of bytes be an alternative? That would work for integers, too, and if different FP standards are actually used, we are loosing information anyway.

>With integers, we probably have to decide on a fixed type we cast all `ints` to.

After looking into things a bit more in detail, such a decision would actually be of relevance more for the frontend than the backend. With the current state of affairs, the user actively chooses the C++ datatype to be used:
```cpp
auto chunk_data = E_x.loadChunk<double>(chunk_offset, chunk_extent);
```
[...]

PR #337, opened 2018-09-07
Integer Type System: Refactor

ax3l:
Refactor the integer type system to support all fundamental C/C++ integer types: `short`, `int`, `long`, `long long` plus the unsigned versions of those.

Fixed size and "least size" ints are aliases of the above, leading to issues on MSVC since there is no fixed int alias for `long` in the old design.
Another detail is that `uint64_t` is an alias for `long` on 64bit Linux but an alias for `long long` on 64bit  OSX. On both platforms, `long` and `long long` are 64 bit.

## User Changes

Users will see no changes as long as they use helpers such as `determineDatatype<T>()`. Otherwise, they must be aware that `Datatype`s such as `DT::INT32` are now in one of `DT::SHORT`, `DT::INT`, etc. Types such as `int32_t` can still be used for attributes and data chunks, since they are simple aliases.

## Backend Changes

Instead of implementing types such as `INT16`, `INT32`, `INT64`, ... one now needs to implement the four types `short`, `int`, `long` and `long long`. Throw runtime errors on missing types, e.g. ADIOS 1.13.1 does not support `long long` yet.

The fileformat is responsible for portability, e.g. HDF5 stores automatically platform information with the stored types, making sure a stored `long` on one platform is an `int32_t` on one and and `int64_t` on another platform (and will not show up as `long` on both, potentially).

### Related

Related and depending issues are #330 and #333.


ax3l:
@franzpoeschel I am sorry, but this PR will affect your current json backend
@lge0303 I am verry sorry, but this PR will affect your current netcdf backend

please feel free to raise any questions in issues that might occur from this change!

anokfireball:
> ... your current json backend

Oh yes, I'm excited :dizzy_face:

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
```diff
@@ -87,23 +87,23 @@ getBP1DataType(Datatype dtype)
...
             return adios_unsigned_short;
-        case DT::UINT32:
-        case DT::VEC_UINT32:
+        case DT::UINT:
+        case DT::VEC_UINT:
             return adios_unsigned_integer;
-        case DT::UINT64:
-        case DT::VEC_UINT64:
+        case DT::ULONG:
+        case DT::VEC_ULONG:
```
anokfireball: We're very likely going to have problems with ADIOS in this regard.
From what I can see, they do **NOT** make the destinction that causes problems with OSX and MSVC. Rather, they treat the bit-widths as I did initally (see [```adios_types.h```](https://github.com/ornladios/ADIOS/blob/v1.13.1/src/public/adios_types.h#L15-L39), which explicitly lists sizes of the datatypes).
As e.g. `int` may now be either 16 or 32 bit, and ADIOS assumes a size of 32, this might explode in our face.

ax3l: Yep, I know but that's ADIOS' portability problem not ours: https://github.com/ornladios/ADIOS/issues/187

(Side note: ADIOS1 only compiles on Linux and OSX, not on MSVC.)

ax3l: Uh wait, you mean we will not be able to store a `(u)int64_t` since ADIOS has no `long long`... Yeah... Well.

anokfireball: > that's ADIOS' portability problem not ours

So the baseline here is to rely on the C standard and ignore potentially faulty behaviour in one of our backends. Not ideal, but probably the only thing we ~can~ should do.

> you mean we will not be able to store a (u)int64_t

We most likely can, unless we're on exotic platforms (*MSVC*, anyone?) where `LONG` and `ULONG` are 32 Bits wide... Which again comes back to the standard compliance stated above.

ax3l:
Note: The leftover HDF5 CI errors on OSX and MSVC are "just" the `shape` type for constant records.

Note2: we might have to fake types in the ADIOS1 backend on OSX in case it's really assuming fixed sizes behind it's `adios_int/long/...` types.

src/ParticlePatches.cpp
```diff
@@ -69,7 +69,7 @@ ParticlePatches::read()
         IOHandler->flush();
 
         using DT = Datatype;
-        if( DT::UINT64 != *dOpen.dtype )
+        if( DT::ULONG != *dOpen.dtype )
```
anokfireball: Correct me if I'm wrong, but this does not strictly enforce the standard, `ULONG` may be 32 or 64.
```cpp
if( determineDatatype< uint64_t >() != *dOpen.dtype )
```

src/Series.cpp
```diff
@@ -768,7 +768,7 @@ Series::readBase()
     aRead.name = "openPMDextension";
     IOHandler->enqueue(IOTask(this, aRead));
     IOHandler->flush();
-    if( *aRead.dtype == DT::UINT32 )
+    if( *aRead.dtype == DT::UINT )
```
anokfireball:
```cpp
if( *aRead.dtype == determineDatatype< uint32_t >() )
```

src/binding/python/BaseRecordComponent.cpp
```diff
@@ -53,17 +53,17 @@ void init_BaseRecordComponent(py::module &m) {
...
+            else if( brc.getDatatype() == DT::LONG )
                 return py::dtype("int64");
-            else if( brc.getDatatype() == DT::UINT16 )
+            else if( brc.getDatatype() == DT::USHORT )
                 return py::dtype("uint16");
-            else if( brc.getDatatype() == DT::UINT32 )
+            else if( brc.getDatatype() == DT::UINT )
                 return py::dtype("uint32");
-            else if( brc.getDatatype() == DT::UINT64 )
+            else if( brc.getDatatype() == DT::ULONG )
```
anokfireball: This is now probably not precise enough as `ULONG` may be 32 or 64.
Numpy provides a number of compatible C types, but not all of the applicable ones:
https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.scalars.html#built-in-scalar-types

anokfireball: I see you recorgnized the problem in `RecordComponent.cpp` already. :+1:

test/SerialIOTest.cpp
```diff
@@ -749,19 +749,19 @@ TEST_CASE( "hzdr_hdf5_sample_content_test", "[serial][hdf5]" )
 
         RecordComponent& e_positionOffset_x = e_positionOffset["x"];
         REQUIRE(e_positionOffset_x.unitSI() == 2.599999993753294e-07);
-        REQUIRE(e_positionOffset_x.getDatatype() == Datatype::INT32);
+        REQUIRE(e_positionOffset_x.getDatatype() == Datatype::INT);
```
anokfireball: Same *problem* for all datatypes read back from a file.
`INT` may be 16, 32 or 64.
```cpp
REQUIRE(e_positionOffset_x.getDatatype() == determineDatatype< int32_t >());
```

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
ax3l: What I try to say is: if `adios_long` is defined as 8 byte on *all* platforms, we will implement it as such and convert properly in the backend from whatever matches (`int` or `long` or `long long`).

src/ParticlePatches.cpp
ax3l: exactly, that's the leftover from my replace marathon and your suggestion is correct!

src/binding/python/BaseRecordComponent.cpp
ax3l: Yep, pls feel free to annotate any line you catch :)

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
franzpoeschel: >The fileformat is responsible for portability, e.g. HDF5 stores automatically platform information with the stored types

Hi, just so I get this right: When reading e.g. a `long` from some stored file, I will also have to take into account the platform that wrote the value and use the reading platform's datatype that corresponds with the actual length of the value stored, so possibly not a `long` but an `int` for example?
If so, it's good you mention this, because floating point datatypes already have the same issue and I ignored that so far.

ax3l: > I will also have to take into account the platform

My idea is to handle it exactly the other way around. Our backends, such as HDF5, are portable across platforms. If HDF5 stores for example a `long` (aka `int64_t`) on x86-64 Linux it does store in the back automatically the type of the platform. If you read this file back on x86-64 Windows, the HDF type presented to you will be `long long` (still aka `int64_t`) since `int` (and `long`) are `int32_t` there. The precision of the data is preserved.

> JSON

Let's migrate that discussion here: https://github.com/openPMD/openPMD-api/issues/65

test/SerialIOTest.cpp
```diff
@@ -907,10 +907,10 @@ TEST_CASE( "hdf5_dtype_test", "[serial][hdf5]" )
     REQUIRE(s.getAttribute("uchar").get< unsigned char >() == 'u');
     REQUIRE(s.getAttribute("int16").get< int16_t >() == 16);
     REQUIRE(s.getAttribute("int32").get< int32_t >() == 32);
-    REQUIRE(s.getAttribute("int64").get< int64_t >() == 64);
+    REQUIRE(getCast< int64_t >(s.getAttribute("int64")) == 64);
```
ax3l, line 950: As mentioned elsewhere, I would suggest we refactor `Variant` and `Attribute` to be closer / in one file.
With that, we could implement `getCast` alongside (or as) `get` directly in `openPMD::auxiliary::Variant`. This method will be useful for the user-facing `Attribute` code in many places.

ax3l:
@C0nsultant yay, it's done! Minus some convenience functions we should add (maybe in follow-ups) this is ready for review! :)

include/openPMD/Datatype.hpp
```diff
@@ -177,10 +189,338 @@ determineDatatype(std::shared_ptr< T >)
...
+
+/** Return number of bits representing a Datatype
+ *
+ * @param d Datatype
+ * @return number of bits
+ */
+inline size_t
+toBits( Datatype d )
+{
+    return toBytes( d ) * CHAR_BIT;
```
anokfireball, line 282: Just out of curiosity: Are you aware of any modern, usable, real-world platform where `CHAR_BIT != 8`?

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
```diff
@@ -87,24 +90,86 @@ getBP1DataType(Datatype dtype)
...
+        case DT::LONG:
+        case DT::VEC_LONG:
+            if( sizeof(long) == 2u )
+                return adios_short;
+            else if( sizeof(long) == 4u )
+                return adios_integer;
+            else if( sizeof(long) == 8u )
+                return adios_long;
+            else
+                throw unsupported_data_error("No native equivalent for Datatype::LONG found.");
```
anokfireball, line 122: `long` is at least 32 Bit, making the
```cpp
if( sizeof(long) == 2u )
```
branch dead code.
[C++ fundamental types](https://en.cppreference.com/w/cpp/language/types)

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
```diff
@@ -87,24 +90,86 @@ getBP1DataType(Datatype dtype)
...
+        case DT::LONGLONG:
+        case DT::VEC_LONGLONG:
+            if( sizeof(long long) == 2u )
+                return adios_short;
+            else if( sizeof(long long) == 4u )
+                return adios_integer;
+            else if( sizeof(long long) == 8u )
+                return adios_long;
+            else
+                throw unsupported_data_error("No native equivalent for Datatype::LONGLONG found.");
```
anokfireball, line 132: `long long` is at least 64 Bit, making the
```cpp
if( sizeof(long long) == 2u )
if( sizeof(long long) == 4u )
```
branches dead code.
[C++ fundamental types](https://en.cppreference.com/w/cpp/language/types)

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
```diff
@@ -87,24 +90,86 @@ getBP1DataType(Datatype dtype)
...
+        case DT::ULONG:
+        case DT::VEC_ULONG:
+            if( sizeof(unsigned long) == 2u )
+                return adios_unsigned_short;
+            else if( sizeof(unsigned long) == 4u )
+                return adios_unsigned_integer;
+            else if( sizeof(unsigned long) == 8u )
+                return adios_unsigned_long;
+            else
+                throw unsupported_data_error("No native equivalent for Datatype::ULONG found.");
```
anokfireball, line 162: Same for the vector types.

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
```diff
@@ -87,24 +90,86 @@ getBP1DataType(Datatype dtype)
...
+        case DT::ULONGLONG:
+        case DT::VEC_ULONGLONG:
+            if( sizeof(unsigned long long) == 2u )
+                return adios_unsigned_short;
+            else if( sizeof(unsigned long long) == 4u )
+                return adios_unsigned_integer;
+            else if( sizeof(unsigned long long) == 8u )
+                return adios_unsigned_long;
+            else
+                throw unsupported_data_error("No native equivalent for Datatype::ULONGLONG found.");
```
anokfireball, line 172: Same for the vector types.

include/openPMD/backend/Attribute.hpp
```diff
@@ -42,23 +46,154 @@ namespace openPMD
...
+
+/** Retrieve a stored specific Attribute and cast if possible.
+ *
+ * @throw   std::runtime_error if stored object is not static castable to U.
+ * @tparam  U   Type of the object to be casted to.
+ * @return  Copy of the retrieved object, casted to type U.
+ */
+template< typename U >
+inline U
+getCast( Attribute const a )
```
anokfireball, line 127: Knit-picking here. `DoConvert` returns copies of the cast values anyway, so is there any benefit in the additional copy of the `Attribute`?

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -46,26 +47,32 @@ CommonADIOS1IOHandlerImpl::flush_attribute(int64_t group, std::string const& nam
         case DT::VEC_CHAR:
             nelems = att.get< std::vector< char > >().size();
         break;
-        case DT::VEC_INT16:
-            nelems = att.get< std::vector< int16_t > >().size();
+        case DT::VEC_SHORT:
+            nelems = att.get< std::vector< short > >().size();
         break;
```
anokfireball: I introduced these `break`s as unindented... and it looks real ugly. Will clean up when I touch this the next time.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -276,8 +311,8 @@ CommonADIOS1IOHandlerImpl::flush_attribute(int64_t group, std::string const& nam
         }
         case DT::BOOL:
         {
-            auto ptr = reinterpret_cast< unsigned char* >(values.get());
-            *ptr = static_cast< unsigned char >(att.get< bool >());
+            auto ptr = reinterpret_cast< bool* >(values.get());
+            *ptr = static_cast< bool >(att.get< bool >());
```
anokfireball: I am unsure about this.
As far as I can remember, ADIOS does not offer support for native `bool` due to their C origins.
Which is why there was a [workaround](https://github.com/openPMD/openPMD-api/pull/126/commits/ccd9351709a6438583bb36a73140ff4d17f0477c) borrowed from [PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1756).

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -533,32 +568,89 @@ CommonADIOS1IOHandlerImpl::openDataset(Writable* writable,
...
+            if( sizeof(short) == 2u )
+                dtype = DT::SHORT;
+            else if( sizeof(int) == 2u )
+                dtype = DT::INT;
+            else if( sizeof(long) == 2u )
+                dtype = DT::LONG;
+            else if( sizeof(long long) == 2u )
+                dtype = DT::LONGLONG;
+            else
+                throw unsupported_data_error("No native equivalent for Datatype adios_short found.");
```
anokfireball, line 590: `long` is at least 32 Bit.
`long long` is at least 64 Bit.
Which maskes the 
```cpp
else if( sizeof(long) == 2u )
else if( sizeof(long long) == 2u )
```
branches dead code.
[C++ fundamental types](https://en.cppreference.com/w/cpp/language/types)

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -533,32 +568,89 @@ CommonADIOS1IOHandlerImpl::openDataset(Writable* writable,
...
+            if( sizeof(short) == 4u )
+                dtype = DT::SHORT;
+            else if( sizeof(int) == 4u )
+                dtype = DT::INT;
+            else if( sizeof(long) == 4u )
+                dtype = DT::LONG;
+            else if( sizeof(long long) == 4u )
+                dtype = DT::LONGLONG;                
+            else
+                throw unsupported_data_error("No native equivalent for Datatype adios_integer found.");
```
anokfireball, line 602: `long long` is at least 64 Bit.
Which maskes the 
```cpp
else if( sizeof(long long) == 4u )
```
branch dead code.
[C++ fundamental types](https://en.cppreference.com/w/cpp/language/types)

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -533,32 +568,89 @@ CommonADIOS1IOHandlerImpl::openDataset(Writable* writable,
...
+            if( sizeof(unsigned short) == 2u )
+                dtype = DT::USHORT;
+            else if( sizeof(unsigned int) == 2u )
+                dtype = DT::UINT;
+            else if( sizeof(unsigned long) == 2u )
+                dtype = DT::ULONG;
+            else if( sizeof(unsigned long long) == 2u )
+                dtype = DT::ULONGLONG;
+            else
+                throw unsupported_data_error("No native equivalent for Datatype adios_unsigned_short found.");
```
anokfireball, line 629: Same for the vector types.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -533,32 +568,89 @@ CommonADIOS1IOHandlerImpl::openDataset(Writable* writable,
...
+            if( sizeof(unsigned short) == 4u )
+                dtype = DT::USHORT;
+            else if( sizeof(unsigned int) == 4u )
+                dtype = DT::UINT;
+            else if( sizeof(unsigned long) == 4u )
+                dtype = DT::ULONG;
+            else if( sizeof(unsigned long long) == 4u )
+                dtype = DT::ULONGLONG;                
+            else
+                throw unsupported_data_error("No native equivalent for Datatype adios_unsigned_integer found.");
```
anokfireball, line 641: Same for the vector types.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -852,32 +948,152 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                    dtype = DT::LONG;
+                    a = Attribute(*reinterpret_cast< long* >(data));
+                }
+                else if( sizeof(long long) == 2u )
+                {
+                    dtype = DT::LONGLONG;
+                    a = Attribute(*reinterpret_cast< long long* >(data));
+                }
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_short found.");
```
anokfireball, line 972: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -852,32 +948,152 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                    dtype = DT::LONG;
+                    a = Attribute(*reinterpret_cast< long* >(data));
+                }
+                else if( sizeof(long long) == 4u )
+                {
+                    dtype = DT::LONGLONG;
+                    a = Attribute(*reinterpret_cast< long long* >(data));
+                }
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_integer found.");
```
anokfireball, line 996: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -852,32 +948,152 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                    dtype = DT::ULONG;
+                    a = Attribute(*reinterpret_cast< unsigned long* >(data));
+                }
+                else if( sizeof(unsigned long long) == 2u )
+                {
+                    dtype = DT::ULONGLONG;
+                    a = Attribute(*reinterpret_cast< unsigned long long* >(data));
+                }
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_unsigned_short found.");
```
anokfireball, line 1048: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -852,32 +948,152 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                    dtype = DT::ULONG;
+                    a = Attribute(*reinterpret_cast< unsigned long* >(data));
+                }
+                else if( sizeof(unsigned long long) == 4u )
+                {
+                    dtype = DT::ULONGLONG;
+                    a = Attribute(*reinterpret_cast< unsigned long long* >(data));
+                }
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_unsigned_integer found.");
```
anokfireball, line 1072: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -925,35 +1141,44 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                if( sizeof(short) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< short >(data, size), DT::VEC_SHORT);
+                else if( sizeof(int) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< int >(data, size), DT::VEC_INT);
+                else if( sizeof(long) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< long >(data, size), DT::VEC_LONG);
+                else if( sizeof(long long) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< long long >(data, size), DT::VEC_LONGLONG);
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_short found.");
```
anokfireball, line 1153: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -925,35 +1141,44 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                if( sizeof(short) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< short >(data, size), DT::VEC_SHORT);
+                else if( sizeof(int) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< int >(data, size), DT::VEC_INT);
+                else if( sizeof(long) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< long >(data, size), DT::VEC_LONG);
+                else if( sizeof(long long) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< long long >(data, size), DT::VEC_LONGLONG);
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_integer found.");
```
anokfireball, line 1167: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -969,35 +1194,44 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                if( sizeof(unsigned short) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned short >(data, size), DT::VEC_USHORT);
+                else if( sizeof(unsigned int) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned int >(data, size), DT::VEC_UINT);
+                else if( sizeof(unsigned long) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned long >(data, size), DT::VEC_ULONG);
+                else if( sizeof(unsigned long long) == 2u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned long long >(data, size), DT::VEC_ULONGLONG);
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_unsigned_short found.");
```
anokfireball, line 1206: Same here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -969,35 +1194,44 @@ CommonADIOS1IOHandlerImpl::readAttribute(Writable* writable,
...
+                if( sizeof(unsigned short) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned short >(data, size), DT::VEC_USHORT);
+                else if( sizeof(unsigned int) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned int >(data, size), DT::VEC_UINT);
+                else if( sizeof(unsigned long) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned long >(data, size), DT::VEC_ULONG);
+                else if( sizeof(unsigned long long) == 4u )
+                    std::tie(a, dtype) = std::make_tuple(readVectorAttributeInternal< unsigned long long >(data, size), DT::VEC_ULONGLONG);
+                else
+                    throw unsupported_data_error("No native equivalent for Datatype adios_unsigned_integer found.");
```
anokfireball, line 1220: Same here.

src/binding/python/BaseRecordComponent.cpp
```diff
@@ -49,22 +49,30 @@ void init_BaseRecordComponent(py::module &m) {
...
-                return py::dtype("uint32");
-            else if( brc.getDatatype() == DT::UINT64 )
-                return py::dtype("uint64");
+            else if( brc.getDatatype() == DT::SHORT )
+                return py::dtype("short");
+            else if( brc.getDatatype() == DT::INT )
+                return py::dtype("int");
+            // missing in numpy: covered by uint or ulonglong
+            // else if( brc.getDatatype() == DT::LONG )
+            //     return py::dtype("long");
```
anokfireball: Now this is weird.
https://docs.scipy.org/doc/numpy/user/basics.types.html
> `int_` | Default integer type (same as C `long`; normally either `int64` or `int32`)

https://docs.scipy.org/doc/numpy-1.14.2/reference/arrays.scalars.html
> `int_` | compatible: Python int | `'l'`

https://docs.python.org/3/c-api/long.html
(Note that there's two meanings of `long`here: (a) C `long` (b) Pythons's 'bignum' integer of arbitrary precision)
> PyLong_FromLong(long v)

Note the lacking `PyLong_FromInt(int v)`

So there seems to be an equivalent of `long` in `int_`, but no counterpart to `unsigned long`.

src/binding/python/RecordComponent.cpp
```diff
@@ -115,26 +117,32 @@ void init_RecordComponent(py::module &m) {
...
-            else if( r.getDatatype() == Datatype::INT16 ) dtype = py::dtype("int16");
-            else if( r.getDatatype() == Datatype::INT32 ) dtype = py::dtype("int32");
-            else if( r.getDatatype() == Datatype::INT64 ) dtype = py::dtype("int64");
-            else if( r.getDatatype() == Datatype::UINT16 ) dtype = py::dtype("uint16");
-            else if( r.getDatatype() == Datatype::UINT32 ) dtype = py::dtype("uint32");
-            else if( r.getDatatype() == Datatype::UINT64 ) dtype = py::dtype("uint64");
+            else if( r.getDatatype() == Datatype::SHORT ) dtype = py::dtype("short");
+            else if( r.getDatatype() == Datatype::INT ) dtype = py::dtype("int");
+            // missing in numpy: covered by int or longlong
+            // else if( r.getDatatype() == Datatype::LONG ) dtype = py::dtype("long");
```
anokfireball: See [comment](https://github.com/openPMD/openPMD-api/pull/337/files#r216158495).

src/binding/python/BaseRecordComponent.cpp
anokfireball: https://docs.scipy.org/doc/numpy/user/basics.types.html#array-types-and-conversions-between-types
> Additionally to `intc` the platform dependent C integer types `short`, `long`, `longlong` and their unsigned versions are defined.

Quick test:
```Python
In [1]: import numpy as np

In [2]: np.dtype('long')
Out[2]: dtype('int64')
```

Haven't figured out how `unsigned long` works yet.

anokfireball: https://github.com/numpy/numpy/issues/10678#issuecomment-369363551
> Docs are wrong. The C integer type `long` is called `np.int_`, and the integer type `unsigned long` is called `np.uint`

So there we have it:
C `int`/ `unsigned int` is Python `intc`/`uintc`
C `long`/`unsigned long` is Python `int_`/ `uint`

```Python
In [1]: import numpy as np

In [2]: np.dtype('short')
Out[2]: dtype('int16')

In [3]: np.dtype('intc')
Out[3]: dtype('int32')

In [4]: np.dtype('int_')
Out[4]: dtype('int64')

In [5]: np.dtype('longlong')
Out[5]: dtype('int64')
```

src/binding/python/RecordComponent.cpp
```diff
@@ -217,18 +231,24 @@ void init_RecordComponent(py::module &m) {
...
-            else if( a.dtype().is(py::dtype("uint32")) )
-                r.storeChunk( offset, extent, shareRaw( (uint32_t*)a.mutable_data() ) );
-            else if( a.dtype().is(py::dtype("uint64")) )
-                r.storeChunk( offset, extent, shareRaw( (uint64_t*)a.mutable_data() ) );
+            else if( a.dtype().is(py::dtype("short")) )
+                r.storeChunk( offset, extent, shareRaw( (short*)a.mutable_data() ) );
+            else if( a.dtype().is(py::dtype("int")) )
+                r.storeChunk( offset, extent, shareRaw( (int*)a.mutable_data() ) );
+            // missing in numpy: covered by int or longlong
+            // else if( a.dtype().is(py::dtype("long")) )
```
anokfireball: See [comment](https://github.com/openPMD/openPMD-api/pull/337/files#r216158495).

test/CoreTest.cpp
```diff
@@ -71,6 +81,37 @@ TEST_CASE( "attribute_dtype_test", "[core]" )
...
+        a = Attribute(static_cast< long long >(0));
+        REQUIRE(Datatype::LONG == a.dtype);
+    }
+    if( sizeof(int) == sizeof(long) )
+    {
+        a = Attribute(static_cast< long >(0));
+        REQUIRE(Datatype::INT == a.dtype);
+        a = Attribute(static_cast< int >(0));
+        REQUIRE(Datatype::LONG == a.dtype);
+    }
```
anokfireball: Scary thought:
There might be platforms where 
```cpp
sizeof(short) > 2u
```
making it impossible to write or read 16 Bit wide data (as `int16_t` does not alias to any of our provided integer types).

test/SerialIOTest.cpp
anokfireball: Agreed.
Is this `getCast` here for demonstration purposes? If not, why is it not used on the other explicit-width comparisons?

anokfireball (changes requested):
First of all, excuse the avalanche of comments.
Python dtypes need adjustment and a few questions need to be resolved.

test/SerialIOTest.cpp
ax3l: Yes, demonstration and necessity.

> If not, why is it not used on the other explicit-width comparisons?

I don't get this, can you elaborate? Maybe I missed an already existing way to do this.

include/openPMD/Datatype.hpp
ax3l: I just like being explicit. Of course, if I should ever get my hands on a C++11 capable compiler running on a UNIVAC ... <3 Besides that, some DSPs like TexasInstrucments are using 16 and 32bit Bytes.
Found this SO on it: https://stackoverflow.com/questions/5516044/system-where-1-byte-8-bit

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
ax3l: But a Byte is not necessarily 8 Bit ;-)
So if we have 4 bit bytes in our [Intel 4004](https://en.wikipedia.org/wiki/Intel_4004) ... :D

ax3l: I just want to mention that I can't see portability in ADIOS1 itself for bytes other than 8 bit, but maybe I am missing it: https://github.com/ornladios/ADIOS/issues/187

include/openPMD/backend/Attribute.hpp
ax3l: The only reason I don't put a reference `const &` here is that I can pass temporaries, such as `.dtype()`.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
ax3l: fixed :)

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
ax3l: oh yes, let me revert this. I thought about this as well and was reckless.

test/CoreTest.cpp
ax3l: Hm, that would be weird. But if a platform does not implement a 2 byte fundamental integer type, there will also be no alias `int16_t` for it.

src/binding/python/BaseRecordComponent.cpp
ax3l: Wow, as fancy as my float-confusion in Python xD Thanks a ton for checking!

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
ax3l: I would leave this in, just because it makes it easier to read and the compile time dead branches should be optimized out by a good compiler, no? (Or does it throw warnings somewhere?)

src/binding/python/BaseRecordComponent.cpp
ax3l: Ah yes, should be fixed now together with my float question in numpy's docs: https://github.com/numpy/numpy/pull/11837

test/SerialIOTest.cpp
anokfireball: Ok, more explicit this time.

Why is `getCast` used for `int64_t`, but not for `int32_t` or `int16_t`?
The only explaination I could come up with is the inconsistency between the `long` and `long long` aliasing, but this happens on a single platform where any stored HDF5 attribute should be read back in the `H5T_NATIVE_X`.

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
anokfireball: > I can't see portability in ADIOS1 itself for bytes other than 8 bit

No automatic way, they assume the user to do it by hand (please don't).

> I would leave this in

There's a point where software is generalized too much. This is not there yet, but close. Leave it in, I don't blame you for copy pasting and saving time. ;)

test/CoreTest.cpp
anokfireball: Just to throw out [this](https://stackoverflow.com/questions/5516044/system-where-1-byte-8-bit/5516161#5516161) SO answer again, this is present for `char` on certain (albeit rare) systems.

include/openPMD/backend/Attribute.hpp
anokfireball: Maybe I'm missing something here.
With `Attribute const&`, 
```cpp
auto value = getCast< uint64_t >(magicalTemporary());
```
would not lead to a dangling reference as the value is consumed (i.e. cast and copied) before the destruction of the temporary (as long as that itself is not a reference to an object on the stack).

Anyway, let's not waste too much time on this little detail.

anokfireball (approved):

test/SerialIOTest.cpp
ax3l: Got ya, the reason is as follows: first of all, it should be everywhere to be maximally portable. That's why I propose to unify `auxiliary:Variant` with `Attribute` so that `Variant::get` can know all supported types already and we can implement the external `getCast` *inside* the current `Variant::get` for maximum user comfort :)

(That's just too much of a structural change for this PR.)

I am only changing the `int64_t` in this PR, since on OSX 64bit, which has like Linux 64bit `long` and `long long` as `int64_t`, the `int64_t` alias is not on `long long` but on `long`, thus returning those as `long`.

ax3l: That's the same reason why #184 fails, so we can fix it the same way for `long double` on MSVC :)

test/CoreTest.cpp
ax3l: :D

ax3l:
(just pushing again since travis forgot to report its success state)

include/openPMD/IO/ADIOS/ADIOS1Auxiliary.hpp
anokfireball: > But a Byte is not necessarily 8 Bit ;-)

This logic then only works for 
```Python
CHAR_BIT in (4, 8, 16, 32)
```
btw ;)

test/SerialIOTest.cpp
anokfireball: I agree with your proposal as that will solve the problem in the general case.

I just don't grasp all the details of the problem, still. :/
So this one cast is (with our current platform exploration) mandatory only for `int64_t` as that is stored in the `Attibute` as `long long`, but the `Variant::get< int64_t >()` is equivalent to a `Variant::get< long >()`.

... Right?

ax3l: Yes, it is mandatory since on OSX a written `long long` is not a `int64_t` (typedef of `long`) and will round-trip as `long` during read.

Yet, as on Linux x86-64bit, `long` and `long long` as distinct types implementing the same precision: a 8 byte / 64 bit integer.

We always (consistently) choose the "smallest" generic type that fits, that means if a `long long` is also a `long` a written `long long` will be returned as `long`. To hide this nastiness and it's implications for `std::variant` (not an alias == bad access) from the user, we use a static cast for explicitly possible types.

We can decide if we allow precision loss in `get<>`. E.g., only exact same implementations vs. all possible `static_cast`s. I think even the latter is fine, a user can check the type at runtime.

anokfireball: There is already a similar casting and reporting mechanism in place for floating point attributes that the openPMD standard does not fix w.r.t. bid-width. This might become obsolete in the process of implementing this extended `Variant::get< T >()`.
See [here](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/backend/Attributable.cpp#L282-L290).

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
ax3l: I just realized I intentionally changed this while updating the depending PRs :) This is the read part, of course we want an actual `bool` back here. Will change it therein.

anokfireball: > This is the read part

Lines [314-315](https://github.com/ax3l/openPMD-api/blob/159d8f4f7effd0f17e5af4ee236e491c6cebb3ed/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L314-L315) are actually part of [```CommonADIOS1IOHandler::flushAttribute()```](https://github.com/ax3l/openPMD-api/blob/159d8f4f7effd0f17e5af4ee236e491c6cebb3ed/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L40-L42) (the deferred execution of [```CommonADIOS1IOHandler::writeAttribute()```](https://github.com/ax3l/openPMD-api/blob/159d8f4f7effd0f17e5af4ee236e491c6cebb3ed/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L787-L790)). So unless you count the transfer from in-memory to on-disk a _read_, then no. Then this is not the read-API. Changining this to `bool` will break ADIOS (for the stated reason).

The place where something like your idea can be implemented is in lines [1022-1025](https://github.com/ax3l/openPMD-api/blob/159d8f4f7effd0f17e5af4ee236e491c6cebb3ed/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L1022-L1025) of [```CommonADIOS1IOHandler::readAttribute()```](https://github.com/ax3l/openPMD-api/blob/159d8f4f7effd0f17e5af4ee236e491c6cebb3ed/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L865-L868). The problem here is different, though. The Attribute is stored as `uint8_t` and we can not determine whether it is a true `uint8_t` or a workaround for `bool` (casting for values `0` and `1` will lead to even more havoc).

This cast is better suited in your proposal #340 for unifying `Attribute` and `Variant`. This ADIOS1 corner-case would then be appropriately handled by `getCast`.

anokfireball: This change is not entirely invalid, though.
```cpp
case DT::BOOL:
{
    /* values is a void-pointer that is given to ADIOS
     * to support bool writing, this will have to be allocated as uint8_t */
    auto ptr = reinterpret_cast< unsigned char* >(values.get());
    /* the current workaround is heaved into the frontend
     * (i.e. the user has to pass uint8_t when storing bool in ADIOS)
     * the backend can do the appropriate widening from bool to uint8_t */
    *ptr = static_cast< unsigned char >(att.get< bool >());
    break;
}
```

ax3l: Thanks, that's important information I missed!

Just to make things worse: As I just remembered from the special case of `std::vector< bool >` in the stdlibs is that a bool is not necessarily the same size as an `uint8_t` ;)
https://stackoverflow.com/questions/4897844/is-sizeofbool-defined

Of course I don't know any implementation where this is not a Byte, but who knows ;-)

anokfireball: > a bool is not necessarily the same size as an uint8_t

That should not be a problem.
```cpp
/* frontend */
Series s = Series("bool_test.bp", AccessType::CREATE);
bool attribute_value = true;
s.setAttribute("bool", attribute_value); /* schedules a IOTask with a bool */
s.flush();

/* ADIOS1 backend steps in pseudo-code 
 * 1. schedule attribute write, not modifying any parameters yet
 * 2. in flushAttribute(), allocate a uint8_t behind a void* to give to ADIOS
 * 3. retrieve the bool as a bool (!) from the Attribute
 * 4. static_cast the bool to uint8_t, store it in the void*
 * 5. give  void* to ADIOS
 * 6. ???
 * 7. profit */
```

In this case, it won't even matter what `sizeof(bool)` is, as `0`/`1` can *always* be stored in `uint8_t`.

PR #338, opened 2018-09-10
JSON backend

franzpoeschel:
Implementation of a JSON backend for openPMD.

Todo/Issues:
* [x] user documentation
* [x] Some of the methods in `AbstractIOHandlerImpl.hpp` did not behave exactly as documented (modulo misunderstandings on my side). I will address this separately.
* [ ] see the discussion in #65 
* [x] integration with the build system: #384
* [x] remove commits from #337 and rebase onto master
* [x] put appropriate license, copyright and author headers in every file

### DIFF
--- a/docs/source/backends/json.rst
+++ b/docs/source/backends/json.rst
@@ -1,0 +1,83 @@
+.. _backends-json:
+
+JSON Backend
+============
+
+openPMD supports writing to and reading from JSON files.
+For this, the installed copy of openPMD must have been built with support for the JSON backend.
+To build openPMD with support for JSON, use the CMake option ``-DopenPMD_USE_JSON=ON``.
+For further information, check out the :ref:`installation guide <install>`,
+:ref:`build dependencies <development-dependencies>` and the :ref:`build options <development-buildoptions>`.
+
+
+JSON File Format
+----------------
+A JSON file uses the file ending ``.json``. The JSON backend is chosen by creating
+a ``Series`` object with a filename that has this file ending.
+
+The top-level JSON object is a group representing the openPMD root group ``"/"``.
+Any **openPMD group** is represented in JSON as a JSON object with two reserved keys:
+
+ * ``attributes``: Attributes associated with the group. This key may be null or not be present
+   at all, thus indicating a group without attributes.
+ * ``platform_byte_widths`` (root group only): Byte widths specific to the writing platform.
+   Will be overwritten every time that a JSON value is stored to disk, hence this information 
+   is only available about the last platform writing the JSON value.
+
+All datasets and subgroups contained in this group are represented as a further key of
+the group object. ``attributes`` and ``platform_byte_widths`` have
+hence the character of reserved keywords and cannot be used for group and dataset names
+when working with the JSON backend.
+Datasets and groups have the same namespace, meaning that there may not be a subgroup
+and a dataset with the same name contained in one group.
+
+Any **openPMD dataset** is a JSON object with three keys:
+
+ * ``attributes``: Attributes associated with the dataset. May be ``null`` or not present if no attributes are associated with the dataset.
+ * ``datatype``: A string describing the type of the stored data.
+ * ``data`` A nested array storing the actual data in row-major manner.
+   The data needs to be consistent with the fields ``datatype`` and ``extent``.
+   Checking whether this key points to an array can be (and is internally) used to distinguish groups from datasets.
+
+**Attributes** are stored as a JSON object with a key for each attribute.
+Every such attribute is itself a JSON object with two keys:
+
+ * ``datatype``: A string describing the type of the value.
+ * ``value``: The actual value of type ``datatype``.
+
+Restrictions
+------------
+For creation of JSON serializations (i.e. writing), the restrictions of the JSON backend are
+equivalent to those of the `JSON library by Niels Lohmann <https://github.com/nlohmann/json>`_
+used by the openPMD backend.
+
+Numerical values, integral as well as floating point, are supported up to a length of
+64 bits.
+Since JSON does not support special floating point values (i.e. NaN, Infinity, -Infinity),
+those values are rendered as ``null``.
+
+Instructing openPMD to write values of a datatype that is too wide for the JSON
+backend does *not* result in an error:
+ * If casting the value to the widest supported datatype of the same category (integer or floating point)
+   is possible without data loss, the cast is performed and the value is written.
+   As an example, on a platform with ``sizeof(double) == 8``, writing the value
+   ``static_cast<long double>(std::numeric_limits<double>::max())`` will work as expected
+   since it can be cast back to ``double``.
+ * Otherwise, a ``null`` value is written.
+
+Upon reading ``null`` when expecting a floating point number, a NaN value will be
+returned. Take notice that a NaN value returned from the deserialization process
+may have originally been +/-Infinity or beyond the supported value range.
+
+Upon reading ``null`` when expecting any other datatype, the JSON backend will
+propagate the exception thrown by Niels Lohmann's library.
+
+A parallel (i.e. MPI) implementation is *not* available.
+
+Example
+-------
+The example code in the :ref:`usage section <usage-serial>` will produce the following JSON serialization
+when picking the JSON backend:
+
+.. literalinclude:: json_example.json
+
--- a/docs/source/backends/json_example.json
+++ b/docs/source/backends/json_example.json
@@ -1,0 +1,142 @@
+{
+  "attributes": {
+    "basePath": {
+      "datatype": "STRING",
+      "value": "/data/%T/"
+    },
+    "iterationEncoding": {
+      "datatype": "STRING",
+      "value": "groupBased"
+    },
+    "iterationFormat": {
+      "datatype": "STRING",
+      "value": "/data/%T/"
+    },
+    "meshesPath": {
+      "datatype": "STRING",
+      "value": "meshes/"
+    },
+    "openPMD": {
+      "datatype": "STRING",
+      "value": "1.1.0"
+    },
+    "openPMDextension": {
+      "datatype": "UINT",
+      "value": 0
+    }
+  },
+  "data": {
+    "1": {
+      "attributes": {
+        "dt": {
+          "datatype": "DOUBLE",
+          "value": 1
+        },
+        "time": {
+          "datatype": "DOUBLE",
+          "value": 0
+        },
+        "timeUnitSI": {
+          "datatype": "DOUBLE",
+          "value": 1
+        }
+      },
+      "meshes": {
+        "rho": {
+          "attributes": {
+            "axisLabels": {
+              "datatype": "VEC_STRING",
+              "value": [
+                "x"
+              ]
+            },
+            "dataOrder": {
+              "datatype": "STRING",
+              "value": "C"
+            },
+            "geometry": {
+              "datatype": "STRING",
+              "value": "cartesian"
+            },
+            "gridGlobalOffset": {
+              "datatype": "VEC_DOUBLE",
+              "value": [
+                0
+              ]
+            },
+            "gridSpacing": {
+              "datatype": "VEC_DOUBLE",
+              "value": [
+                1
+              ]
+            },
+            "gridUnitSI": {
+              "datatype": "DOUBLE",
+              "value": 1
+            },
+            "position": {
+              "datatype": "VEC_DOUBLE",
+              "value": [
+                0
+              ]
+            },
+            "timeOffset": {
+              "datatype": "FLOAT",
+              "value": 0
+            },
+            "unitDimension": {
+              "datatype": "ARR_DBL_7",
+              "value": [
+                0,
+                0,
+                0,
+                0,
+                0,
+                0,
+                0
+              ]
+            },
+            "unitSI": {
+              "datatype": "DOUBLE",
+              "value": 1
+            }
+          },
+          "data": [
+            [
+              0,
+              1,
+              2
+            ],
+            [
+              3,
+              4,
+              5
+            ],
+            [
+              6,
+              7,
+              8
+            ]
+          ],
+          "datatype": "DOUBLE"
+        }
+      }
+    }
+  },
+  "platform_byte_widths": {
+    "BOOL": 1,
+    "CHAR": 1,
+    "DOUBLE": 8,
+    "FLOAT": 4,
+    "INT": 4,
+    "LONG": 8,
+    "LONGLONG": 8,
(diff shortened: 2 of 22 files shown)


ax3l:
@franzpoeschel I know the changes were quite heavy recently (big sorry again). Can you please try the following to update this branch? I would like to remove the number of commits and would like to have no merge commit in it.

```bash
# backup in case things go south
git checkout -b json-backend-bak
# switch back to json-backend 
git checkout -

# rebase: squash
git fetch --all
# I assume your "git remote -v" for this mainline repo is called "mainline"
git rebase -i mainline
# an editor will open
#   change "pick" to "s" ("squash") to all your commits *but the first entry*
#   remove the lines with "merge commits",
#   remove commits a7a7974 and 176d37d 
# save
# solve merge conflicts one

# if anything goes very wrong
# git rebase --abort

# in the end, update the PR via
git push -f
```

franzpoeschel:
Concerning user documentation, is there a place where the file formats are / should be documented?

@ax3l Done, thanks for the steps.

ax3l:
>Concerning user documentation, is there a place where the file formats
>are / should be documented?

Yes, it's even recommended :) You can update the documentation files (.rst) in the same PR that lands the feature :)

ax3l:
@franzpoeschel I added complete build system support for you in #384.
Is it ok if I rebase this branch and force-push update it against those changes?

ax3l:
Pushed the rebased branch :)
Please make sure that you don't have uncommited/unpushed *local* changes before continuing to work on this. Fetch the latest online version, revert the last local commit if they were identical before my rebase and pull the fresh commit.

```bash
git status -sb
# nothing uncommited?
# latest commits were pushed online?

# assuming yes

# fetch the updates I pushed on your branch and remove local "outdated" commit
git fetch --all
git reset --hard HEAD~
git pull --ff-only
```

After that, you can continue as usual.

If you *do* have local changes that were not yet pushed - commit them and rebase the branch against the online commit (also removing the one outdated commit but keeping all other changes on top).

```bash
git fetch --all

git rebase -i mainline/dev
# remove outdated commit
# keep all others
```

franzpoeschel:
Thanks for the help and the detailed instructions!

include/openPMD/auxiliary/Memory.hpp
```diff
@@ -24,8 +24,10 @@
 #include "openPMD/Datatype.hpp"
 
 #include <functional>
+#include<iostream>
```
ax3l: little formatting issue :)

include/openPMD/auxiliary/Filesystem.hpp
```diff
@@ -87,5 +87,27 @@ remove_directory(std::string const& path);
  */
 bool
 remove_file(std::string const& path);
+
+// /** Opens the file identified by the given path.
```
ax3l: can this  dead code be removed? :)

include/openPMD/IO/JSON/JSONFilePosition.hpp
```diff
@@ -0,0 +1,53 @@
...
+#endif
+
+namespace openPMD
+{
+
+
+    struct JSONFilePosition :
+        public AbstractFilePosition
+#if openPMD_HAVE_JSON
+    {
```
ax3l, line 39: can you please implement the bidy in a `.cpp` file? Avoids pulling in the public `nlohmann/json.hpp` header into our public API.

include/openPMD/Datatype.hpp
```diff
@@ -518,6 +519,184 @@ isSame( openPMD::Datatype const d, openPMD::Datatype const e )
...
+ * @tparam Action The functor's type.
+ * @tparam Args The functors argument types.
+ * @param dt The openPMD datatype.
+ * @param action The functor.
+ * @param args The functor's arguments.
+ * @return The return value of the functor, when calling its <operator()>() with
+ * the passed arguments and the template parameter type corresponding to the
+ * openPMD type.
+ */
+
```
ax3l: maybe no empy newline here. You can put the `OPENPMD_TEMPLATE_OPERATOR` define block above the doxygen block

docs/source/backends/json_example.json
```diff
@@ -0,0 +1,154 @@
+{
```
ax3l, line 1: When we are finished with the review (until then, pls update here):
Shall we rather put this file into https://github.com/openPMD/openPMD-example-datasets ?

docs/source/backends/json.rst
```diff
@@ -0,0 +1,82 @@
...
+propagate the exception thrown by Niels Lohmann's library.
+
+A parallel (i.e. MPI) implementation is *not* available.
+
+Example
+-------
+The example code in the :ref:`usage section <usage-serial>` will produce the following JSON serialization
+when picking the JSON backend:
+
+.. literalinclude:: json_example.json
```
ax3l, line 82: maybe link to https://github.com/openPMD/openPMD-example-datasets when it's included there?

docs/source/backends/json_example.json
```diff
@@ -0,0 +1,154 @@
...
+    "SHORT": 2,
+    "UCHAR": 1,
+    "UINT": 4,
+    "ULONG": 8,
+    "ULONGLONG": 8,
+    "USHORT": 2
+  },
+  "subgroups": {
+    "data": {
+      "subgroups": {
```
ax3l: why do you add a `subgroups` directory here and before `data`? Can we skip those maybe?

test/SerialIOTest.cpp
```diff
@@ -2296,3 +2296,475 @@ TEST_CASE( "no_serial_adios1", "[serial][adios]")
     REQUIRE(true);
 }
 #endif
+
+#if openPMD_HAVE_JSON
+
+TEST_CASE( "json_write_test", "[serial][json]")
```
ax3l: Do you think we can just generalize the `hdf5_`/`adios_`/`json_` write test into a helper function and loop with a `Series` object for each backend? Otherwise we copy so much code...

ax3l (commented):
Dang, I reviewed this a week ago and forgot to submit my comments.

docs/source/backends/json_example.json
franzpoeschel: Short answer: separating namespaces. If this indirection were dropped, the subgroups' and datasets' names will have the same namespace as all other JSON objects at that level, including the objects `attributes` and for the top-level group `platform_byte_widths` (meaning you can't have groups with those names). This seems like a nasty corner case for a user of the API to run into, hence I chose clarity at the expense of a certain informational overhead as is explained in [here](https://github.com/openPMD/openPMD-api/blob/ab2fb040060089a72acaa2c14a8c72a909df2b66/docs/source/backends/json.rst).
Do you have a better solution? I can imagine wrapping the subgroups into quotes or maybe reserving a prefix for "keywords" (such as `__attributes` maybe). Would this be preferrable?

include/openPMD/auxiliary/Filesystem.hpp
franzpoeschel: done ;)

include/openPMD/IO/JSON/JSONFilePosition.hpp
franzpoeschel: Done.... but I still need to include `nlohmann/json.hpp` in the header file?
This being said, `JSONFilePosition.hpp` is only ever included in `JSONIOHandlerImpl.hpp`, which is included in `JSONIOHandler.hpp` and that header file is only included in `*.cpp` files, so it should be fine?

include/openPMD/auxiliary/Memory.hpp
franzpoeschel: Ah, good you spotted that, those includes were left from when I added a function to that file which I later removed again. Removed the includes.

include/openPMD/Datatype.hpp
franzpoeschel: done

docs/source/backends/json_example.json
ax3l: Yes, there is a risk of collisions which in practice should not be too often the case. Nevertheless, rather adding a *dict of attributes* inside an `attributes` group would probably be best (no `__`). This keeps the collisions minimal and at the same time is very readable. But that's already the case, no?

The same is true for data sets and groups. Without the `subgroups` "namespace" one can not write the same group and data set name in the same level and neither can be named "attributes". But imho that's fine.

include/openPMD/IO/JSON/JSONFilePosition.hpp
ax3l: No, if no symbols/declarations are used from `nlohmann/json.hpp` in the header file, then do not include it there.

> This being said, [...]

That sounds good, I wasn't sure.

test/SerialIOTest.cpp
franzpoeschel: Pushed a commit that does this. This leaves only the tests using samples as backend-specific tests.

docs/source/backends/json_example.json
franzpoeschel: >But that's already the case, no?

yep

I changed it now, so attributes keep their separate namespace but datasets and subgroups are now directly linked into the containing group's object, see [here](https://github.com/franzpoeschel/openPMD-api/blob/json-backend/docs/source/backends/json_example.json).

include/openPMD/IO/JSON/JSONFilePosition.hpp
franzpoeschel: >if no symbols/declarations are used from nlohmann/json.hpp in the header file

They are. `JSONFilePosition.hpp` is implemented as a wrapper around `json::json_pointer`.

docs/source/backends/json_example.json
ax3l: Looks good!
Do you think you need the `"object_type": "dataset"` and `"object_type": "group"` keys? They don't hurt, but I wonder if one can already guess by the name (e.g. `data`) and content already.

franzpoeschel: I added those for the `listPaths` and `listDatasets` operations. I did not want to rely on checking the structure of JSON objects to distinguish groups from datasets, so I made the distinction explicit.
I mean, both layouts (i.e. of groups vs. datasets) should actually be computationally distinguishable, but should we go that length?

ax3l: I get the first part, but think it's probably ok to check if a `data` members is present.

Can please you elaborate what you mean with the last sentence (`computationally distinguishable`)?

franzpoeschel: >Can please you elaborate what you mean with the last sentence (computationally distinguishable)?

Pretty much what you said already, datasets and groups can be distinguished by inspecting their members.
>think it's probably ok to check if a data members is present

Probably not a good idea if you look at [line 28](https://github.com/franzpoeschel/openPMD-api/blob/json-backend/docs/source/backends/json_example.json) of the example JSON file ;)
The safest bet would probably be to check whether the `extent` key exists *and* whether it points to a JSON array. That should work, but it *feels* like a hack. Should I do it this way?

ax3l: Why do we need `extent`? Can't we determine this from a list (of lists (of lists)) by querying their lengths, even in ND-cases?

Would checking for "contains a data member that contains a list" be enough to detect a low-level "data set" vs. low-level "group"?

I just want to avoid introducing too many non-standard attributes/sub-groups/naming conventions. It might be that we need some, since we don't have as in XML/HDF5/ADIOS/... attributes on a path that are distinct from attributes on data sets. But we should keep it minimal, e.g. for attributes it's reasonable to have this sub-group and also to have a "data" key.

franzpoeschel: >Why do we need extent? Can't we determine this from a list (of lists (of lists)) by querying their lengths, even in ND-cases?

In principle, yes. It would just require pre-filling the dataset with `null` values to be sure that it does actually have the specified dimensions. (Currently, the data array is just as big as necessary for holding the chunks that have actually been written).
Would you prefer this?

franzpoeschel: The latest commit eliminates the `extent` field, next up will be the elimination of `object_type`.

ax3l: > pre-filling the dataset with null values

makes probably sense, this only works in serial anyway, right?

ax3l: > The latest commit eliminates the extent field, next up will be the elimination of object_type.

Can you please update the checked-in example file? :)

franzpoeschel: >this only works in serial anyway, right?

right
`object_type` is no more and the example file and documentation are updated.

ax3l: cool, thanks! looks good :)

ax3l:
@franzpoeschel I think I can implement #389 now :)
That means: when it is merged you can remove the changes you made to the vendored `nlohmann-json` files, which is a good thing. If you like, rebase and squash some commits when you are on it? (But again, wait for both until #389 is merged) :)

ax3l:
@franzpoeschel ready for rebase, remove of the JSON third party patch and some squashing of commits :)

ax3l:
@franzpoeschel just do (in case your `git remote` for the main repo is called `mainline`, see `git remote -v`):
```bash
git fetch mainline
git rebase -i mainline/dev
git push -f
```

ax3l:
Ouch, looks like #389 did not trigger `-isystem` yet... (Sorry!)

ax3l:
@franzpoeschel I don't know why this did not work, but you can fix it by adding an extra commit changing the following:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 2c73e2d..418428f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,8 +127,10 @@ else()
 endif()
 if(openPMD_HAVE_JSON)
     add_library(openPMD::thirdparty::nlohmann_json INTERFACE IMPORTED)
-    target_link_libraries(openPMD::thirdparty::nlohmann_json
-        INTERFACE nlohmann_json::nlohmann_json)
+    #target_link_libraries(openPMD::thirdparty::nlohmann_json
+    #    INTERFACE nlohmann_json::nlohmann_json)
+    get_target_property(lib_include_dirs nlohmann_json::nlohmann_json INTERFACE_INCLUDE_DIRECTORIES)
+    target_include_directories(openPMD::thirdparty::nlohmann_json SYSTEM INTERFACE ${lib_include_dirs})
 endif()
 
 # external library: HDF5 (optional)
```

ax3l:
I pushed it for you.

ax3l (approved):
Hurray, thanks a lot this is great! :sparkles:

PR #339, opened 2018-09-10
[WIP] Adjust some descriptions of the methods in AbstractIOHandlerImpl

franzpoeschel:
The class `AbstractIOHandlerImpl` contains a number of methods where implementing them as mandated by their documentation resulted in erroneous behaviour.
* `listPaths`, `listDatasets`, `listAttributes` and `readAttribute` alike are mandated by the documentation to fail if the `Writable` passed to them was not marked written. Neither the ADIOS nor the HDF5 backends implement  such a check and these functions are indeed often called with `Writable`s that are *not* marked written by the frontend. 
* `openPath` mandates that the `path should be relative (i.e. it should not start with a slash "/")`. In reality, the function is also called with absolute paths (e.g. path == "/data"). Checking for this turns out to be necessary for a well-functioning implementation.

This PR adjusts the documentation accordingly.

### DIFF
--- a/include/openPMD/IO/AbstractIOHandlerImpl.hpp
+++ b/include/openPMD/IO/AbstractIOHandlerImpl.hpp
@@ -204,7 +204,6 @@ public:
   virtual void listDatasets(Writable*, Parameter< Operation::LIST_DATASETS > &) = 0;
   /** List all attributes associated with an object.
    *
-   * The operation should fail if the Writable was not marked written.
    * The attribute should be associated with the Writable.
    * The list of attribute names should be stored in the location indicated by the pointer parameters.attributes.
    */
--- a/include/openPMD/IO/AbstractIOHandlerImpl.hpp
+++ b/include/openPMD/IO/AbstractIOHandlerImpl.hpp
@@ -94,7 +94,7 @@ public:
   /** Open all contained groups in a path, possibly recursively.
    *
    * The operation should overwrite existing file positions, even when the Writable was already marked written.
-   * The path parameters.path may contain multiple levels (e.g. first/second/third/). This path should be relative (i.e. it should not start with a slash "/").
+   * The path parameters.path may contain multiple levels (e.g. first/second/third/). This path may be relative or absolute (indicated by a leading slash "/").
    * The Writables file position should correspond to the complete opened path (i.e. first/second/third/ should be assigned to the Writables file position).
    * The Writable should be marked written when the operation completes successfully.
    */
@@ -183,7 +183,6 @@ public:
   virtual void readDataset(Writable*, Parameter< Operation::READ_DATASET > &) = 0;
   /** Read the value of an existing attribute.
    *
-   * The operation should fail if the Writable was not marked written.
    * The operation should fail if the attribute does not exist.
    * The attribute should be associated with the Writable and have the name parameters.name. This name should not contain a slash ("/").
    * The attribute datatype should be stored in the location indicated by the pointer parameters.dtype.
@@ -193,14 +192,12 @@ public:
   virtual void readAttribute(Writable*, Parameter< Operation::READ_ATT > &) = 0;
   /** List all paths/sub-groups inside a group, non-recursively.
    *
-   * The operation should fail if the Writable was not marked written.
    * The operation should fail if the Writable is not a group.
    * The list of group names should be stored in the location indicated by the pointer parameters.paths.
    */
   virtual void listPaths(Writable*, Parameter< Operation::LIST_PATHS > &) = 0;
   /** List all datasets inside a group, non-recursively.
    *
-   * The operation should fail if the Writable was not marked written.
    * The operation should fail if the Writable is not a group.
    * The list of dataset names should be stored in the location indicated by the pointer parameters.datasets.
    */


anokfireball:
I am short on time for a debugging session for this problem at the moment, but I have a suspicion.

> these functions are indeed often called with `Writable`s that are not marked written by the frontend.

There's really two stages to IO operations:
 - the scheduling of an `IOTask` with an associated `Writable` and `AbstractParameter*` in the frontend
 - the execution of that `IOTask` (generally) at a deferred point in time in the backend

The constraints that you are changing here are all referring to the second stage. The operations might (and often will) be *scheduled* when the `Writable` is not written yet (as IO is an expensive operation, try to bunch as much of it together as possible). As the scheduling order must be honored in the backends, these inconsistencies should have been resolved once the *execution* of the operation starts.

> marked written by the frontend

*No* operation in the frontend is supposed to set the `written` status of a `Writable` (except for workarounds with `fileBased` that replicate the `written` status across all the files). That is the job of the individual backend, as mandated by the last constraint for [`AbstractIOHandlerImpl::openFile(...)`](
https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/IO/AbstractIOHandlerImpl.hpp#L91) for example.

So coming around to the inital concern: If such a check fails when implemented, this is actually a bigger problem in the scheduling order rather than a documentation problem (on the condition that all of the other constraints are honored).
If the latter is the case, I would ask you to provide a workflow where such a check (when implemented in HDF5 for example) fails.

franzpoeschel:
Thanks for your clarification! I was already suspecting that there are misunderstandings on my side, but I did not want to leave the JSON implementation in a state that does not conform with the documentation. I will look into it again and try to figure out where the problem is.

ax3l:
@franzpoeschel is this PR still active or did the question clarify? :)

franzpoeschel:
The second bullet point is still a thing. I just verified it again using the HDF5 backend. If gdb is not lying to me, in the first call to `openPath()` issued by the example code on [here](https://openpmd-api.readthedocs.io/en/latest/usage/serial.html#c), `parameters.path` is equal to `"/data"`, contradicting the documentation (`This path should be relative (i.e. it should not start with a slash "/").`).
I will look into the rest today and comment again.

anokfireball:
> contradicting the documentation

Very strictly speaking, this is not a contradiction. The word [`should`](https://en.oxforddictionaries.com/definition/should) was deliberately used to signal that the desired effect is not guaranteed in every single function call. And, hence, the [name sanitation](https://github.com/openPMD/openPMD-api/blob/dev/src/IO/HDF5/HDF5IOHandler.cpp#L389-L394) was put in place in the HDF5 backend (for example).

The problem you're referring to originates in [`Series::read()`](https://github.com/openPMD/openPMD-api/blob/4771137d831a0ebfa810dfe8ed96b87123617b48/src/Series.cpp#L827-L833), where the `Series::basePath` (which is always `/data/%T/`) is just stripped of `/%T/`, i.e. resulting in a `Parameter< Operation::OPEN_PATH >::path` equal to `/data`.

If you really think this sanitation problem is worth pursuing, it needs to be solved in the fronend logic. Relaxation of the backend invariants is probably redundant, as `should` already allows the condition to be false.

franzpoeschel:
After digging a bit deeper, I think I might have found (part of) the problem:
In `src/Series.cpp`, the methods `readGroupBased` and `readFileBased` both modify the `written` status:
```cpp
/* allow all attributes to be set */
written = false;
```
(lines 646 and 708). In both methods, the following will happen in this order:
1) Open the file, the backend will set the `written` status `true`.
2) The frontend sets the `written` status `false` again
3) In `readBase()`, an `IOTask` of type `READ_ATTR` is fed  to the `IOHandler`, the `Writable`'s `written` status still being `false`, but `AbstractIOHandlerImpl::readAttribute()` requires that the status be `true`. 

In the ADIOS and HDF5 backends, this does not lead to a problem since the status is not actually checked.
Leaving out those lines without modifying anything else will not work either, since both methods that contain said line, will later call `setIterationFormat()` which throws an exception if `written == true`.

So if I'm understanding it correctly, the frontend does actually meddle with the `written` status in a way that will interfere negatively with the backend. 
So maybe just changing the method descriptions in `AbstractIOHandlerImpl.hpp` as suggested in this PR will likely not be the way to go, I think the issue is rather with the frontend's implementation. In the meantime, skipping the check of the `written` status is probably a good workaround.

franzpoeschel:
@C0nsultant Ok, that clarifies things. I implemented `openPath()` in the JSON backend to distinguish between absolute and relative paths, but just stripping the leading slash from "/data" and treating it as a relative path works as well.

anokfireball:
Thank you for the report, @franzpoeschel!
Now I understand the situation, especially with regards to your first bullt-point.

And yes, your concern is absolutely valid. The lines ([here](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L500-L502), [here](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L646-L647), [here](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L708-709), [here](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L789-L790) and [here](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L807-L809)) you report as responsible are quick (and admittedly dirty) workarounds for the following situation:
 * a `Series` is opened as `READ_ONLY` and scheduled for readin from disk
 * a `Operation::OPEN_PATH` for the root `/` is scheduled
 * once that opeartion executes, `Series::written` will equal to `true` (as per [condition](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/IO/AbstractIOHandlerImpl.hpp#L99))
 * *only now* can we do certain reads and set in-memory `Attribute`s from those read values
 * certain `Attribute`s must not be set once their `Writable`s are marked `written`, as the corresponding action may not be available in backends (e.g. change from `fileBased` to `groupBased`)
 * to work around this issue (e.g. in [`Series::setIterationEncoding()`](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L292-L311)) when first reading an existing files (i.e. in the [`Series::Series()`](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/Series.cpp#L108-L116) constructor), the `written` status is emulated when potentially conflicting `Attribute`s may have to be set

Which leaves us... with a dirty hack.
This undoubtedly should be resolved, but IMHO removing the conditions from the backend is not a desirable way to go. The conditions should actually be veryfied once the `Series` is constructed. It is just during the construction when this problem may happen.

I'm very open to your input here, as you have dug deep into the code already.
If you don't want to, I'll take the problem and implement some cleaner solution. But maybe you have something elegant in mind?

I will probably pull the emulated status to the conflicting calls. Then, 
```cpp
void
Series::readFileBased(AccessType actualAccessType)
{
            /* ... */
            if( *aRead.dtype == DT::STRING )
                setIterationFormat(Attribute(*aRead.resource).get< std::string >());
            else
            /* ... */
}
```
becomes
```cpp
void
Series::readFileBased(AccessType actualAccessType)
{
            /* ... */
            if( *aRead.dtype == DT::STRING )
            {
                this->written = false,
                setIterationFormat(Attribute(*aRead.resource).get< std::string >());
                this->written = true;
            } else
            /* ... */
}
```
and operations in the backends can verify the condition.

anokfireball:
> just stripping the leading slash from "/data" and treating it as a relative path works as well

Do you think this would be a better explaination for the condition?
`paths should always be treated as relative (i.e. a leading slash "/" treated as synonymous  with "./")`

franzpoeschel:
> > just stripping the leading slash from "/data" and treating it as a relative path works as well
> 
> Do you think this would be a better explaination for the condition?
> `paths should always be treated as relative (i.e. a leading slash "/" treated as synonymous with "./")`

Yes, I think it would be. You can argue that the current description is technically correct, but it misled me into thinking that `openPath()` would not have to handle paths of this form at all.

> This undoubtedly should be resolved, but IMHO removing the conditions from the backend is not a desirable way to go. The conditions should actually be veryfied once the `Series` is constructed. It is just during the construction when this problem may happen.

Seconded. I don't think that the updated descriptions from this PR should find their way into dev. 
> I will probably pull the emulated status to the conflicting calls. Then,
> 
> ```c++
> void
> Series::readFileBased(AccessType actualAccessType)
> {
>             /* ... */
>             if( *aRead.dtype == DT::STRING )
>                 setIterationFormat(Attribute(*aRead.resource).get< std::string >());
>             else
>             /* ... */
> }
> ```
> 
> becomes
> 
> ```c++
> void
> Series::readFileBased(AccessType actualAccessType)
> {
>             /* ... */
>             if( *aRead.dtype == DT::STRING )
>             {
>                 this->written = false,
>                 setIterationFormat(Attribute(*aRead.resource).get< std::string >());
>                 this->written = true;
>             } else
>             /* ... */
> }
> ```
> 
> and operations in the backends can verify the condition.

Making the workaround as local as possible in this way is probably the best easy fix for the issue you described. 
I'm not quite confident that this fix will also fix the issue in asynchronous backends? Essentially, what we're trying to achieve is the possibility to force execution of `WRITE_ATTR` operations, regardless of `written` status. Changing the `written` status temporally will work in sequential backends, but can we be sure that it will not leak into other operations in an asynchronous backend?
Two cleaner options might be:
1) Flushing before and after every "forced" attribute write additionally to your suggested solution. Since it only affects construction of a `Series`, the overhead caused by this should be negligible. Still a hack though ;)
2) Adding a `bool force = false` field to `Parameter< Operation::WRITE_ATT >` (and probably a private member in `Series.cpp` akin to `setIterationEncoding` that skips the `written` check). Clean, but requires changing the backends.

franzpoeschel:
Also, I can take care of this, but it won't be until after next week.

anokfireball:
> what we're trying to achieve is the possibility to force execution of `WRITE_ATTR` operations, regardless of `written` status

That is not the ultimate goal, actually.
You probably interpreted the call to `Attributable::setAttribute()` as a write-through to persistent storage. That is not what is happenening in this concrete problem.
The value is already stored in a persistent file on disk and just needs to be injected into the in-memory representation of the `Series`. That is to say, there will be no `WRITE_ATT` for this value.
The problem is storing the value in frontend handles after they have been marked `written` by the backend.

The workaround is only for the logic in the frontend. Backends (neither sequential nor sequentially consistent) need not be considered as the values should not be triggered for write-through.

Anyway, I created a PR with my proposal in #352. @ax3l @franzpoeschel

franzpoeschel:
Closing as stale

PR #341, opened 2018-09-13
setAttribute: Explicit C-String Handling

ax3l:
Explicitly handle C-strings in order to avoid implicit casts. Makes bindings easier.

Split-off from #333

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,10 @@ Features
 Bug Fixes
 """""""""
 
-- Refactor integer type system  #337
+- Refactor type system and ``Attribute`` set/get
+  - integers #337
+  - support ``long double`` reads on MSVC #184
+- ``setAttribute``: explicit C-string handling #341
 - ``Dataset``: ``setCompression`` warning and error logic #326
 - avoid impact on unrelated classes in invasive tests #324
 - Python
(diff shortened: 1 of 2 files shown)


anokfireball (approved):

anokfireball:
Now here's one I don't understand:

The actual [unit test](https://travis-ci.org/openPMD/openPMD-api/jobs/428064863#L1027) with gcc@7.3.0+MPI succeeds, yet the same config (+Python) in the [coverage test](https://travis-ci.org/openPMD/openPMD-api/jobs/428064869#L2254) throws:
```
Test can only be run with at least two ranks
  what():  A Record can not be written without any contained RecordComponents: position
ParallelIOTests is a Catch v2.2.1 host application.
Run with -? for options
hdf5_write_test_zero_extent
/home/travis/build/openPMD/openPMD-api/test/ParallelIOTest.cpp:146
/home/travis/build/openPMD/openPMD-api/test/ParallelIOTest.cpp:146:FAILED:
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal
test cases:  3 | 2 passed |1 failed
assertions: 13 |12 passed |1 failed
```

**UPDATE**
This also happens on [```dev```](https://travis-ci.org/openPMD/openPMD-api/jobs/428060958#L2244), so it's not a blocker for this PR.

PR #344, opened 2018-09-13
SerialIO ADIOS1 Test: Bool

ax3l:
ADIOS1 bool testing. Probably needs a fix.

### DIFF
--- a/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
@@ -40,8 +40,13 @@ CommonADIOS1IOHandlerImpl::close(ADIOS_FILE* f)
 void
 CommonADIOS1IOHandlerImpl::flush_attribute(int64_t group, std::string const& name, Attribute const& att)
 {
+    auto dtype = att.dtype;
+    // https://github.com/ComputationalRadiationPhysics/picongpu/pull/1756
+    if( dtype == Datatype::BOOL )
+        dtype = Datatype::UCHAR;
+
     int nelems = 0;
-    switch( att.dtype )
+    switch( dtype )
     {
         using DT = Datatype;
         case DT::VEC_CHAR:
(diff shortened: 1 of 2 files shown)


src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -311,8 +311,8 @@ CommonADIOS1IOHandlerImpl::flush_attribute(int64_t group, std::string const& nam
         }
         case DT::BOOL:
         {
-            auto ptr = reinterpret_cast< unsigned char* >(values.get());
-            *ptr = static_cast< unsigned char >(att.get< unsigned char >());
+            auto ptr = reinterpret_cast< bool* >(values.get());
+            *ptr = static_cast< bool >(att.get< bool >());
```
anokfireball: `values` is a `void*` that is given to ADIOS, allocated [here](https://github.com/openPMD/openPMD-api/blob/dev/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L99).
To support bool writing (with the workaround), this will have to be allocated as `uint8_t`.
The `Attribute` is passed as `bool`, so that case needs to be handled manually.

```cpp
void
CommonADIOS1IOHandlerImpl::flush_attribute(int64_t group, std::string const& name, Attribute const& att)
{
    auto dtype = att.dtype;
    if( dtype == Datatype::BOOL )
        dtype = Datatype::UCHAR;

    switch( dtype ) /* NOTE - workaround dtype */
    /* ... */

    auto values = auxiliary::allocatePtr(dtype, nelems); /* NOTE - workaround dtype */
    switch( att.dtype ) /* NOTE - original dtype */
    {
        case DT::BOOL:
        {
            auto ptr = reinterpret_cast< unsigned char* >(values.get());
            *ptr = static_cast< unsigned char >(att.get< bool >());
            break;
        }
    }
    /* ... */
}
```

anokfireball (dismissed):

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
ax3l: @C0nsultant can you push the change you are suggesting on the PR branch? I am open to any changes here, just needed a demonstrator to show what I saw was causing the crash during write :)

(Was on the plane and road today, sorry for the missing annotations!)

ax3l:
@C0nsultant thx! :)

PR #345, opened 2018-09-13
Attribute: Implement get<>() Cast

### LINKED ISSUE #335: Unify Variant.hpp and Attribute.hpp
Left-over from #333: in order to work-around the issue of `Attribute(static_cast<short>(42u))` on OSX and MSVC, we should implement a [move constructor in openPMD::auxiliary::Variant](https://github.com/openPMD/openPMD-api/pull/333#issuecomment-419026604) and [make the allowed types](https://github.com/openPMD/openPMD-api/pull/333#issuecomment-419088770) known at this point, which are currently listed in `Attribute.hpp`.

https://github.com/openPMD/openPMD-api/pull/337#discussion_r216566489

ax3l:
Close #335: Implement `getCast<>()` in `Attribute::get<U>()`. Will now cast
during `::get<U>()` if convertible.

Implementations can now simply use `::get<>()` again :)
Final clean-up for refactoring started in #337


ax3l:
@C0nsultant Ping :)

anokfireball:
This make [`Attributable::readFloatingPoint()`](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/include/openPMD/backend/Attributable.hpp#L244-L275) redundant. The casting logic applied therein is now handled in `get` itself. I'll do a follow-up PR that removes this redundancy and just uses `get`.

include/openPMD/backend/Attribute.hpp
```diff
@@ -192,8 +209,12 @@ getCast( Attribute const a )
         return DoConvert<bool, U>{}(pvalue_b);
     else
         throw std::runtime_error("getCast: unknown Datatype.");
+}
 
-    return U{};
+template< typename U >
+U Attribute::get() const
+{
+    return getCast< U >( Variant::getResource() );
```
anokfireball, line 217: I would hate to cast silently here and not notify the user about possible loss of precision.
There already is a function that does this ([`warnWrongDtype()`](https://github.com/openPMD/openPMD-api/blob/0.4.0-alpha/src/backend/Attributable.cpp#L282-L290)). I'll push a patch shortly that will use that function if the resource dtype and `U` differ.

test/SerialIOTest.cpp
```diff
@@ -1526,45 +1526,45 @@ TEST_CASE( "adios1_dtype_test", "[serial][adios1]" )
...
     REQUIRE(s.getAttribute("vecUint32").get< std::vector< uint32_t > >() == std::vector< uint32_t >({4294967294u, 4294967295u}));
-    REQUIRE(getCast< std::vector< uint64_t > >(s.getAttribute("vecUint64")) == std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
+    REQUIRE(s.getAttribute("vecUint64").get< std::vector< uint64_t > >() == std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
     REQUIRE(s.getAttribute("vecFloat").get< std::vector< float > >() == std::vector< float >({0.f, 3.40282e+38f}));
     REQUIRE(s.getAttribute("vecDouble").get< std::vector< double > >() == std::vector< double >({0., 1.79769e+308}));
-    REQUIRE(getCast< std::vector< long double > >(s.getAttribute("vecLongdouble")) == std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
+    REQUIRE(s.getAttribute("vecLongdouble").get< std::vector< long double > >() == std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
     REQUIRE(s.getAttribute("vecString").get< std::vector< std::string > >() == std::vector< std::string >({"vector", "of", "strings"}));
     // REQUIRE(s.getAttribute("bool").get< bool >() == true);
```
anokfireball: This should work as expected now, as the stored `uint8_t` can be cast to `bool` in the new `get`.

anokfireball (changes requested):

include/openPMD/backend/Attribute.hpp
ax3l: That's actually not what I would like as a user.
If there is an attribute "attr" I can RT check it's dtype. If I want to `get<>` it as something, I want to get it as exactly such.

test/SerialIOTest.cpp
ax3l: Just need to rebase, done already in #344

include/openPMD/backend/Attribute.hpp
ax3l: Let's have a quick chat on the philosophy of that method if you like :) Gitter? VC?
In case you are busy today, we can also chat on Monday. I have to leave here around noon.

test/SerialIOTest.cpp
```diff
@@ -1527,25 +1527,25 @@ TEST_CASE( "adios1_dtype_test", "[serial][adios1]" )
...
-    REQUIRE(getCast< std::vector< uint64_t > >(s.getAttribute("vecUint64")) == std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
+    REQUIRE(s.getAttribute("vecUint64").get< std::vector< uint64_t > >() == std::vector< uint64_t >({18446744073709551614u, 18446744073709551615u}));
     REQUIRE(s.getAttribute("vecFloat").get< std::vector< float > >() == std::vector< float >({0.f, 3.40282e+38f}));
     REQUIRE(s.getAttribute("vecDouble").get< std::vector< double > >() == std::vector< double >({0., 1.79769e+308}));
-    REQUIRE(getCast< std::vector< long double > >(s.getAttribute("vecLongdouble")) == std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
+    REQUIRE(s.getAttribute("vecLongdouble").get< std::vector< long double > >() == std::vector< long double >({0.L, std::numeric_limits<long double>::max()}));
     REQUIRE(s.getAttribute("vecString").get< std::vector< std::string > >() == std::vector< std::string >({"vector", "of", "strings"}));
     REQUIRE(s.getAttribute("bool").get< unsigned char >() == true);
     REQUIRE(s.getAttribute("boolF").get< unsigned char >() == false);
```
anokfireball: This now can be retrieved as `bool`, thanks to the casting `get`.

Have this in my staged set, will push soon.

anokfireball (approved):

include/openPMD/backend/Attribute.hpp
anokfireball: Offline resolution:
C++ requires a certain effort from the user to put the *type* inside the brackets. Thus, he should be well aware of what he is requesting.
Information about the datatype can be obtained at run-time. If non-casting behavior is desired, we shift this obligation to the user.

ax3l: > If non-casting behavior is desired, we shift this obligation to the user.

e.g. by checking the `dtype` at runtime, taking the `resource` directly or casting back to the underlying `auxiliary::Variant` that will throw.

test/SerialIOTest.cpp
ax3l: That sounds great, thanks! :)

src/Datatype.cpp
```diff
@@ -22,6 +22,19 @@
 
 #include <iostream>
 
+namespace openPMD
+{
+void warnWrongDtype(std::string const& key,
+                    Datatype store,
+                    Datatype request)
+{
+    std::cerr << "Warning: Attribute '" << key
```
ax3l, line 33: loving MSVC:
```
error C2679: binary '<<': no operator found which takes a right-hand operand of type 'const std::string' (or there is no acceptable conversion
```

forgot to include `<string>`

ax3l: fixed it :)

Issue #37, opened 2018-02-09
Example: MPI Benchmark


ax3l:
We need to add a parallel read & write MPI benchmark at some point, that can be scaled with ranks e.g. for 3D meshes and large particle records.

It can then measure performance in terms of bandwidth, etc.

ax3l:
@franzpoeschel I found the following leftovers:

### Clang 6.0 Warning

```
warning: destructor called on non-final
      'openPMD::OneDimensionalBlockSlicer' that has virtual functions but non-virtual destructor [-Wdelete-non-virtual-dtor]

warning: destructor called on non-final
      'openPMD::RandomDatasetFiller<std::uniform_int_distribution<long>, long>' that has virtual functions but non-virtual destructor [-Wdelete-non-virtual-dtor]
        destroy(_Up* __p) { __p->~_Up(); }
```

Triggered when both are placed into a `std::make_shared` pointer, e.g. with `auto blockSlicer = std::make_shared<openPMD::OneDimensionalBlockSlicer>(0);`

### Sporadic ADIOS Issue in CI

[See here](https://github.com/openPMD/openPMD-api/pull/346#issuecomment-446927357)

```
what():  Internal error: Failed to create directories during ADIOS file creation
```

PR #346, opened 2018-09-14
MPI Benchmark

franzpoeschel:
Adds a generic MPI benchmark to openPMD (#37).

Todo:

- [x] The reading benchmark fails when using the ADIOS1 backend (`openPMD::unsupported_data_error`: Unsupported attribute datatype)
- [ ] Feature requests?
- [x] Resolve strange MPI behaviour (see MPIBenchmarkReport.hpp → addReport() method)
- [x] Adjust documentation to refactoring
- [x] Add user documentation (`*.rst`).
- [x] Add licensing information


ax3l:
Just as a general feature request: it would be cool if this tool allows easy, user-level performance benchmarks on clusters similar to what we did in Figure 1 in https://arxiv.org/abs/1706.00522 . So that we can easily switch backends, compression, written size per node and the like.

franzpoeschel:
Thanks for your suggestions, the figure gives a good guideline.
I am not yet fully done (currently having a little fight with MPI), but next to some refactoring I have added support for picking multiple backends and multiple numbers of threads to be used for the benchmark.
The total extent of the written dataset is still fixed during one benchmark to give comparable results. Do you think it would be useful being able to instruct the Benchmark to be executed for several dataset extents?
So in total, the user currently picks
1) compression schemes
2) backends (per filename extension)
3) number of threads to be used

and the benchmark is then executed for every possible combination of entered values.

↑

edit: this has changed now, look into the [documentation](https://github.com/franzpoeschel/openPMD-api/tree/topic-mpiBenchmark/docs/source/utilities)

ax3l:
> Do you think it would be useful being able to instruct the Benchmark to be executed for several dataset extents?

Yep, sounds useful since it controls the data per node.

We should write both particle and mesh data, but I am ok to start with the latter.

franzpoeschel:
> Yep, sounds useful since it controls the data per node.

I have expressed myself unclearly, the Benchmark class already allows choosing the part of the dataset to write per rank relatively freely. The user can choose this be passing to the constructor an implementation of the [BlockSlicer](https://github.com/openPMD/openPMD-api/blob/0022fbb1709c6088254e1894de6c36785f6d4d87/include/openPMD/benchmark/mpi/BlockSlicer.hpp) abstract class. Once the BlockSlicer is chosen, it is currently fixed to allow for comparable results – unlike compression strategies, backends and thread sizes which can be altered. 

>We should write both particle and mesh data, but I am ok to start with the latter.

To be fair, I just copied that part from [here](https://openpmd-api.readthedocs.io/en/latest/usage/parallel.html), I will look into it.

docs/source/index.rst
```diff
@@ -82,3 +82,13 @@ Development
...
+
+*********
+Utilities
+*********
+.. toctree::
+   :caption: UTILITIES
+   :maxdepth: 1
+   :hidden:
+
+   /utilities/benchmark.rst
```
ax3l: does not need an initial `/`

src/Datatype.cpp
```diff
@@ -144,3 +146,161 @@ std::operator<<(std::ostream& os, openPMD::Datatype d)
...
+
+
+    std::string datatypeToString( openPMD::Datatype dt )
+    {
+        std::stringbuf buf;
+        std::ostream os(&buf);
+        os << dt;
+        return buf.str();
+    }
+}
```
ax3l: missing single newline before EOF

src/benchmark/mpi/OneDimensionalBlockSlicer.cpp
```diff
@@ -0,0 +1,88 @@
...
+        return std::make_pair(
+            std::move( offs ),
+            std::move( localExtent )
+        );
+    }
+
+
+#endif
+}
+
```
ax3l: one newline too much ;)

include/openPMD/benchmark/mpi/RandomDatasetFiller.hpp
```diff
@@ -0,0 +1,129 @@
...
+            {
+                this->buffered
+                    .reset( );
+                this->buffered = this->produceData( );
+            }
+        }
+
+    };
+
+}
```
ax3l: missing newline before EOF

include/openPMD/benchmark/mpi/OneDimensionalBlockSlicer.hpp
```diff
@@ -0,0 +1,32 @@
...
+            Extent
+        > sliceBlock(
+            Extent & totalExtent,
+            int size,
+            MPI_Comm comm
+        ) override;
+    };
+}
+#endif
+
```
ax3l: one newline too much ;)

include/openPMD/benchmark/mpi/MPIBenchmarkReport.hpp
```diff
@@ -0,0 +1,354 @@
...
+        else
+        {
+            return it->second;
+        }
+    }
+
+}
+
+#endif
+
```
ax3l: one newline too much ;)

CMakeLists.txt
```diff
@@ -248,7 +248,8 @@ set(CORE_SOURCE
         src/backend/MeshRecordComponent.cpp
         src/backend/PatchRecord.cpp
         src/backend/PatchRecordComponent.cpp
-        src/backend/Writable.cpp)
+        src/backend/Writable.cpp
+        src/benchmark/mpi/OneDimensionalBlockSlicer.cpp)
```
ax3l, line 290: Is it possible that this file and `mpi_benchmark.cpp` should be just in an extra or example target list?

docs/source/utilities/mpi_benchmark.cpp
```diff
@@ -0,0 +1,88 @@
...
+                          time.first
+                      ).count() << "\tread time: "
+                      << std::chrono::duration_cast<std::chrono::milliseconds>(
+                          time.second
+                      ).count() << std::endl;
+        }
+    }
+
+    MPI_Finalize();
+}
```
ax3l: missing newline before EOF :)

CMakeLists.txt
franzpoeschel: I added `mpi_benchmark.cpp` (now `8_mpi_benchmark.cpp`) to the example target list.
For `OneDimensionalBlockSlicer.cpp` do you mean to create a new list for that?

ax3l: Yes, this `.cpp` file is not part of the library's `CORE_SOURCE` but of the examples/apps only.

include/openPMD/Datatype.hpp
```diff
@@ -518,6 +519,184 @@ isSame( openPMD::Datatype const d, openPMD::Datatype const e )
...
+ * @tparam Args The functors argument types.
+ * @param dt The openPMD datatype.
+ * @param action The functor.
+ * @param args The functor's arguments.
+ * @return The return value of the functor, when calling its <operator()>() with
+ * the passed arguments and the template parameter type corresponding to the
+ * openPMD type.
+ */
+
+#if _MSC_VER && !__INTEL_COMPILER
```
ax3l, line 540: Please place this macro block *above* the doxygen block (separated by an empty newline)

ax3l:
I can merge this for now, thanks a lot!

ax3l (approved):

ax3l:
@franzpoeschel in CI, we get sporadically the error
```
what():  Internal error: Failed to create directories during ADIOS file creation
```

in the MPI benchmark. Can you confirm and might there be a quirk we need to fix?

franzpoeschel:
I have seen this error already, yes. I thought it was an issue with the ADIOS backend, I will try finding the cause of it.

franzpoeschel:
Just to be sure, #418 is exactly this issue, right?

ax3l:
yes, #418 and #421 were sporadic CI issues with parallel I/O :)

PR #351, opened 2018-09-17
Numpy Dtype

ax3l:
Accept the Numpy Dtype as first class type in our python (numpy) bindings.

Handle translation from/to `openPMD.Datatype` behind bindings.

### DIFF
--- a/src/binding/python/BaseRecordComponent.cpp
+++ b/src/binding/python/BaseRecordComponent.cpp
@@ -24,6 +24,7 @@
 
 #include "openPMD/backend/BaseRecordComponent.hpp"
 #include "openPMD/Datatype.hpp"
+#include "openPMD/binding/python/Numpy.hpp"
 
 #include <sstream>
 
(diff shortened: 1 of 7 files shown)


src/binding/python/BaseRecordComponent.cpp
```diff
@@ -46,74 +47,7 @@ void init_BaseRecordComponent(py::module &m) {
 
         .def_property_readonly("unit_SI", &BaseRecordComponent::unitSI)
         .def_property_readonly("dtype", [](BaseRecordComponent & brc) {
```
ax3l, line 49: that's the only API breaking change 1/2

src/binding/python/Dataset.cpp
```diff
@@ -51,7 +58,9 @@ void init_Dataset(py::module &m) {
         .def_readonly("transform", &Dataset::transform)
         .def("set_custom_transform", &Dataset::setCustomTransform)
         .def_readonly("rank", &Dataset::rank)
-        .def_readonly("dtype", &Dataset::dtype)
+        .def_property_readonly("dtype", [](const Dataset &d) {
```
ax3l, line 61: that's the only API breaking change 2/2

PR #353, opened 2018-09-17
Refactor Numpy Stride Check

ax3l:
Refactor and clarify numpy stride checks.

### DIFF
--- a/src/binding/python/Attributable.cpp
+++ b/src/binding/python/Attributable.cpp
@@ -104,12 +104,20 @@ bool setAttributeFromBufferInfo(
         // stride handling
         for( auto d = 0; d < buf.ndim; ++d )
         {
-            // std::cout << "    stride '" << d << "': " << buf.strides[d] << " - " << buf.shape[d] << std::endl;
-            if( buf.strides[d] != buf.shape[d] ) // general criterial
-                if( d > 0 && buf.strides[d] != 1 ) // ok for byte strings
+            // std::cout << "    stride '" << d << "': "
+            //           << buf.strides[d] / buf.itemsize
+            //           << " - " << buf.shape[d] << std::endl;
+            if( buf.strides[d] / buf.itemsize != buf.shape[d] ) // general criteria
+            {
+                if( buf.ndim == 1u && buf.strides[0] / buf.itemsize > buf.shape[0] )
+                    ; // ok in 1D
+                else if( buf.ndim == 1u && buf.strides[0] / buf.itemsize == 1u )
+                    ; // ok and occurs for byte strings
+                else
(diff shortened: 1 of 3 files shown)


src/binding/python/Attributable.cpp
```diff
@@ -104,12 +104,19 @@ bool setAttributeFromBufferInfo(
         // stride handling
         for( auto d = 0; d < buf.ndim; ++d )
         {
-            // std::cout << "    stride '" << d << "': " << buf.strides[d] << " - " << buf.shape[d] << std::endl;
-            if( buf.strides[d] != buf.shape[d] ) // general criterial
-                if( d > 0 && buf.strides[d] != 1 ) // ok for byte strings
+            // std::cout << "    stride '" << d << "': " << buf.strides[d]
+            //           << " - " << buf.shape[d] << std::endl;
+            if( buf.strides[d] != buf.shape[d] ) // general criteria
```
ax3l: Uh, `strides` in in Bytes: https://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.strides.html

Can also simply check the `c_contiguous` flag (or divide the strides by `.itemsize()`)

src/binding/python/Attributable.cpp
```diff
@@ -104,12 +104,20 @@ bool setAttributeFromBufferInfo(
         // stride handling
         for( auto d = 0; d < buf.ndim; ++d )
         {
-            // std::cout << "    stride '" << d << "': " << buf.strides[d] << " - " << buf.shape[d] << std::endl;
-            if( buf.strides[d] != buf.shape[d] ) // general criterial
-                if( d > 0 && buf.strides[d] != 1 ) // ok for byte strings
+            // std::cout << "    stride '" << d << "': "
+            //           << buf.strides[d] / buf.itemsize
+            //           << " - " << buf.shape[d] << std::endl;
+            if( buf.strides[d] / buf.itemsize != buf.shape[d] ) // general criteria
```
anokfireball, line 110: Under what condition is this branch not `true`?

```python
import numpy as np

import openPMD

DATA = np.arange(100).reshape((4, 25))

series = openPMD.Series('path', openPMD.Access_Type.create)
rc = series.iterations[1].particles['e']['position']['x']
rc.store([0, 0], [2, 25], DATA[::2, :])
series.flush()
```
?

anokfireball (approved):

src/binding/python/Attributable.cpp
ax3l: Yes, views for example but  you can also allocate non-contiguous memory in numpy arrays itself.

PR #354, opened 2018-09-17
Python Record_Component: make_constant

ax3l:
Refactor numpy detection of Attributable to general method. Reuse in Record_Component bindings for make_constant.

Also fixes the lifetime of `Container` members (`iterations`, `meshes`, `particles`) in `Series` and `Iteration`. Returned (referenced) containers need to be freed before their owner classes.

- [x] depends on #358

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -51,20 +51,92 @@ void init_RecordComponent(py::module &m) {
         .def_property_readonly("ndim", &RecordComponent::getDimensionality)
         .def_property_readonly("shape", &RecordComponent::getExtent)
 
-        .def("make_constant", &RecordComponent::makeConstant<float>)
-        .def("make_constant", &RecordComponent::makeConstant<double>)
-        .def("make_constant", &RecordComponent::makeConstant<long double>)
-        .def("make_constant", &RecordComponent::makeConstant<short>)
-        .def("make_constant", &RecordComponent::makeConstant<int>)
-        .def("make_constant", &RecordComponent::makeConstant<long>)
-        .def("make_constant", &RecordComponent::makeConstant<long long>)
-        .def("make_constant", &RecordComponent::makeConstant<unsigned short>)
-        .def("make_constant", &RecordComponent::makeConstant<unsigned int>)
-        .def("make_constant", &RecordComponent::makeConstant<unsigned long>)
-        .def("make_constant", &RecordComponent::makeConstant<unsigned long long>)
-        .def("make_constant", &RecordComponent::makeConstant<char>)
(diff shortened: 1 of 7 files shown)


anokfireball (approved):
Unit tests would be nice down the line.

src/binding/python/RecordComponent.cpp
```diff
@@ -51,20 +51,86 @@ void init_RecordComponent(py::module &m) {
...
+            else
+            {
+                throw std::runtime_error("make_constant: "
+                    "Only scalar values supported!");
+            }
+
+        }, py::arg("value")
+        )
+        // allowed python intrinsics, after (!) buffer matching
+        .def("make_constant", &RecordComponent::makeConstant<char>,
```
ax3l, line 132: must be `std::string`: Python has no native `char`s and otherwise our strings are all casted to size-one char

ax3l: but a `std::string` makes no real sense as a value for a record. so let's keep `char`s here.

ax3l:
@C0nsultant unit test added as well :)

test/python/unittest/API/APITest.py
```diff
@@ -308,6 +308,121 @@ def testAttributes(self):
...
+
+            ms["single"][SCALAR].reset_dataset(DS(np.dtype("single"), extent))
+            ms["single"][SCALAR].make_constant(np.single(1.234))
+            ms["double"][SCALAR].reset_dataset(DS(np.dtype("double"), extent))
+            ms["double"][SCALAR].make_constant(np.double(1.234567))
+            ms["longdouble"][SCALAR].reset_dataset(DS(np.dtype("longdouble"),
+                                                      extent))
+            ms["longdouble"][SCALAR].make_constant(np.longdouble(1.23456789))
+
+        # flush and close file, TODO affected by #357
```
ax3l: seems to be affected by #357 in both HDF5 and ADIOS1

test/python/unittest/API/APITest.py
```diff
@@ -308,6 +308,122 @@ def testAttributes(self):
...
                 self.attributeRoundTrip(backend_filesupport[b])
 
+    def makeConstantRoundTrip(self, file_ending):
+        # write
+        series = openPMD.Series(
+            "unittest_py_constant_API." + file_ending,
+            openPMD.Access_Type.create
+        )
+
+        ms = series.iterations[0].meshes
```
ax3l, line 318: That object needs love via [return policies](https://pybind11.readthedocs.io/en/stable/advanced/functions.html). Using that shortcut variable corrupts things.

ax3l: Fixed: had to ensure returned `Container`s are freed before the object that exposed them in the garbage collector

PR #358, opened 2018-09-18
Fix and test constant scalar Records

### LINKED ISSUE #356: Read of Mesh with Constant Record
**Describe the bug**
Reading a files with mesh data with constant record components fails.

**To Reproduce**

I added the following sample file (constant, scalar mesh "rho"): [3_write_serial.h5.tar.gz](https://github.com/openPMD/openPMD-api/files/2392132/3_write_serial.h5.tar.gz)

```C++

int main()
{
    using openPMD;

    Series series = Series(
        "samples/3_write_serial.h5",
        AccessType::READ_ONLY
    );

    return 0;
}
```

fails with:
```
HDF5-DIAG: Error detected in HDF5 (1.10.0-patch1) thread 0:
  #000: ../../../src/H5A.c line 438 in H5Aopen(): unable to load attribute info from object header for attribute: 'position'
    major: Attribute
  #001: ../../../src/H5Oattribute.c line 530 in H5O_attr_open_by_name(): can't locate attribute: 'position'
    major: Attribute
    minor: Object not found
  what():  Internal error: Failed to open HDF5 attribute during attribute read
```

**Expected behavior**
Should be able to read the position attribute from a constant record component.

Also, if position *is* indeed missing, we should handle the backend (e..g HDF5) error gracefully and throw ourselves. The read should not crash in the HDF5 layer.

**Software Environment:**
 - version of openPMD-api: 0.5.0-alpha
 - installed openPMD-api via: from source
 - operating system: Debian GNU/Linux 9.5 (stretch)
 - machine: local laptop
 - name and version of Python implementation: CPython 3.6.6
 - version of HDF5: 1.10.0.1
 - version of ADIOS1: 1.13.1
 - name and version of MPI: OpenMPI 2.0.2

anokfireball:
For constant scalar `Record`s, both the `Record` and the contained sclar `RecordComponent` are the same group in the persistent datastructure. The `RecordComponent`has not been properly assigned for `Mesh`es, which made it impossible to read back files with constant scalar `MeshRecordComponent`s.

This addresses the issue by assigning the group to both the `Record` and the `RecordComponent` for both `Mesh` and `ParticleSpecies`.

Fix #356.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,12 +27,13 @@ Bug Fixes
 """""""""
 
 - dataOrder: mesh attribute is a string #355
+- constant scalar MeshRecords: reading corrected #358
 
 Other
 """""
 
+- test cases for mixed constant/non-constant Records #358
 - examples: close handles explicitly #359
-
 
 0.5.0-alpha
 -----------
--- a/src/ParticleSpecies.cpp
+++ b/src/ParticleSpecies.cpp
@@ -57,20 +57,17 @@ ParticleSpecies::read()
             IOHandler->enqueue(IOTask(&r, aList));
             IOHandler->flush();
 
-            auto attrBegin = aList.attributes->begin();
-            auto attrEnd = aList.attributes->end();
-            auto value = std::find(attrBegin, attrEnd, "value");
-            auto shape = std::find(attrBegin, attrEnd, "shape");
-            if( value != attrEnd && shape != attrEnd )
+            auto att_begin = aList.attributes->begin();
+            auto att_end = aList.attributes->end();
+            auto value = std::find(att_begin, att_end, "value");
+            auto shape = std::find(att_begin, att_end, "shape");
+            if( value != att_end && shape != att_end )
             {
                 RecordComponent& rc = r[RecordComponent::SCALAR];
-                *rc.m_isConstant = true;
-                rc.m_writable->parent = r.m_writable->parent;
                 rc.parent = r.parent;
-                rc.m_writable->abstractFilePosition = r.m_writable->abstractFilePosition;
-                rc.abstractFilePosition = r.abstractFilePosition;
-                rc.m_writable->written = true;
-                rc.written = true;
+                IOHandler->enqueue(IOTask(&rc, pOpen));
+                IOHandler->flush();
+                *rc.m_isConstant = true;
             }
             r.read();
         }
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -1445,6 +1445,40 @@ TEST_CASE( "hdf5_110_optional_paths", "[serial][hdf5]" )
         REQUIRE(s.iterations[1].particles.empty());
     }
 }
+
+TEST_CASE( "hdf5_constant_scalar", "[serial][hdf5]" )
+{
+    {
+        Series s = Series("../samples/constant_scalar.h5", AccessType::CREATE);
+        auto rho = s.iterations[1].meshes["rho"][MeshRecordComponent::SCALAR];
+        rho.resetDataset(Dataset(Datatype::CHAR, {1, 2, 3}));
+        rho.makeConstant(static_cast< char >('a'));
+
+        auto pos_x = s.iterations[1].particles["e"]["position"][RecordComponent::SCALAR];
+        pos_x.resetDataset(Dataset(Datatype::DOUBLE, {3, 2, 1}));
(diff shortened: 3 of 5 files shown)


test/SerialIOTest.cpp
```diff
@@ -1445,6 +1445,40 @@ TEST_CASE( "hdf5_110_optional_paths", "[serial][hdf5]" )
         REQUIRE(s.iterations[1].particles.empty());
     }
 }
+
+TEST_CASE( "hdf5_constant_scalar", "[serial][hdf5]" )
```
ax3l, line 1539: @C0nsultant should we also add a test for a non-scalar record that only has a constant record component (but not all record components are constant)? Or is this case already covered somewhere else?

src/ParticleSpecies.cpp
```diff
@@ -58,18 +58,17 @@ ParticleSpecies::read()
             IOHandler->enqueue(IOTask(&r, aList));
             IOHandler->flush();
 
-            auto attrBegin = aList.attributes->begin();
-            auto attrEnd = aList.attributes->end();
-            auto value = std::find(attrBegin, attrEnd, "value");
-            auto shape = std::find(attrBegin, attrEnd, "shape");
-            if( value != attrEnd && shape != attrEnd )
+            auto begin = aList.attributes->begin();
```
ax3l:
```
error: declaration of ‘begin’ shadows a member of 'this' [-Werror=shadow]
error: declaration of ‘end’ shadows a member of 'this' [-Werror=shadow]
```

ax3l (commented):
Annotated the shadowing Werror inline.
Feel free to also update the changelog with the fix for it.

test/SerialIOTest.cpp
anokfireball: Are you talking about a `Record`with mixed constant and non-constant `RecordComponent`s? I don't think we have such a case yet.

ax3l: Yep, that's what I refer to.

anokfireball: I included a little something for this.

ax3l (approved):

ax3l (approved):

ax3l:
@C0nsultant sorry, changelog rebase :)

ax3l (approved):

ax3l:
@C0nsultant I extended the test to also test the "bp" ADIOS1 backend. First I noted: attribute reading for string arrays (such as `axisLabels`) was missing, added it.

What is now missing: in ADIOS1, all attributes, records, etc. are just addressed by a simple map with the path as key. For constant, *scalar* record components, there seem to be the "directory" of the scalar record missing. Afair, ADIOS1 does not have any directories (see above) anyway, so we might need to add extra handling for this to emulate /  properly detect it anyway.

Concrete, there are attributes such as
```
  double              /data/0/meshes/char/gridGlobalOffset        attr
  unsigned long long  /data/0/meshes/char/shape                   attr
  string              /data/0/meshes/char/geometry                attr
  double              /data/0/meshes/char/gridSpacing             attr
  double              /data/0/meshes/char/unitSI                  attr
  byte                /data/0/meshes/char/value
// ...
```

in a scalar constant record, but the empty `/data/0/meshes/char/` is missing which we query for in `Iteration.cpp` and `ParticleSpecies.cpp`.

This can probably be fixed in `CommonADIOS1IOHandlerImpl::listPaths`.

CHANGELOG.rst
```diff
@@ -26,13 +26,14 @@ Bug Fixes
 """""""""
 
 - dataOrder: mesh attribute is a string #355
+- constant scalar MeshRecords: reading corrected #358
```
ax3l, line 30: note: also constant scalar ParticleRecords. so maybe `constant scalar records: reading corrected #358` ?

anokfireball: The logic for `ParticleRecords` was actually correct, it just did manually what the second `READ_PATH` now does automatically. It mainly is there to simplify and unify the code in both.

anokfireball:
> This can probably be fixed in `CommonADIOS1IOHandlerImpl::listPaths`

That's the only place where it can be fixed without major changes. I hacked together a solution and while I don't think it's too elegant... there's probably not many other ways to do this. Essentially, we have to verify that something is a group by checking that it is not a variable :fearful:.

ax3l:
Thank you, that's great! IMHO, for the ADIOS1 implementation fully lacking hierarchy awareness, your detection ist very gentle now. Love it!

Issue #149, opened 2018-04-26
Explicit user-exposed Series close/(re-)open functionality


anokfireball:
Currently, the lifetime of file handles used in the backends is tied to the lifetime of a ```Series``` they are contained in. It would be very convenient to offer explicit control over this lifetime by having a public interface to close all handles in a Series (e.g. through ```Series::close()``` and likewise to re-open a closed Series (e.g. through ```Series::reopen()```).
Sidenote: When reopening a ~file~ series, it would be wise to re-read ~the entire file~ the meta-data of all involved files instead of resuing old values stored in this logical API. The reasoning is that the file could be modified externally, resulting in an incosistent state between logical and physical representation.

ax3l:
> re-read the entire file

yes, meta-data wise - and even the whole series, not only a single file. Also, it would be useful to add an option to explicitly skip this for performance (taking the risk into account the files below changed - if one can control it by other means).

```C++
Series::reopen( bool verify = true );
```

PR #359, opened 2018-09-18
Examples: Close Handles

ax3l:
Add more explicit notes and handling of `Series` object in all examples. I saw users copying the (serial) examples into Jupyter notebook cells, wondering why files have not been written and/or are broken on still open HDF5 files.

Note: HDF5 and ADIOS1 do not implement proper journaling, corrupting them if not closed properly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,8 @@ Bug Fixes
 Other
 """""
 
+- examples: close handles explicitly #359
+
 
 0.5.0-alpha
 -----------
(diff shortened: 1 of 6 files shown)


anokfireball:
#149 is very much related to this issue.

anokfireball (approved):
As it seems this is not intuitive for users, we should put an effort into providing `Series::close()` and `Series::reopen()`.

For the meantime, this is the best thing we can do.

ax3l:
Yep, I have the same feeling. But if we implement `close()`/`reopen()` we will have to introduce an *invalid* state in out `Series` (and *all* other linked objects) that needs to be checked *every time* they are accessed... -.-

PR #362, opened 2018-09-19
Python: Particle Patches

ax3l:
Expose particle patches to python bindings.

- [x] storing
- [x] loading
- [x] new `load()` at once API (rebase after #364)

### DIFF
--- a/src/binding/python/PatchRecordComponent.cpp
+++ b/src/binding/python/PatchRecordComponent.cpp
@@ -1,0 +1,157 @@
+/* Copyright 2018 Axel Huebl
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * openPMD-api is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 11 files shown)


src/binding/python/PatchRecordComponent.cpp
```diff
@@ -0,0 +1,124 @@
...
+    py::class_<PatchRecordComponent, BaseRecordComponent >(m, "Patch_Record_Component")
+        .def("set_unit_SI", &PatchRecordComponent::setUnitSI)
+        .def("reset_dataset", &PatchRecordComponent::resetDataset)
+        .def_property_readonly("ndims", &PatchRecordComponent::getDimensionality)
+        .def_property_readonly("shape", &PatchRecordComponent::getExtent)
+
+        /*
+        template< typename T >
+        std::shared_ptr< T > load(uint64_t idx);
+        template< typename T >
```
ax3l: TODO implement

test/python/unittest/API/APITest.py
```diff
@@ -644,8 +644,92 @@ def testMesh_Container(self):
         self.assertRaises(TypeError, openPMD.Mesh_Container)
 
     def testParticlePatches(self):
-        """ Test openPMD.ParticlePatches. """
-        self.assertRaises(TypeError, openPMD.ParticlePatches)
+        """ Test openPMD.Particle_Patches. """
+        self.assertRaises(TypeError, openPMD.Particle_Patches)
+
+    def backend_particle_patches(self, file_ending):
```
ax3l: we need the same unit test in C++.
I think `load()` is likely broken.

test/python/unittest/API/APITest.py
```diff
@@ -643,9 +643,91 @@ def testMesh_Container(self):
         """ Test openPMD.Mesh_Container. """
         self.assertRaises(TypeError, openPMD.Mesh_Container)
 
+    def backend_particle_patches(self, file_ending):
```
ax3l, line 646: we need the same unit test in C++.
I think `load()` is likely broken.

ax3l: (was ok, forgot to `flush()`)

PR #367, opened 2018-09-20
Use Catch2 2.3.0+

ax3l:
Update internal include of Catch2 to 2.3.0.
Update all usage of Catch2 to the new, future-proof includes `<catch2/catch.hpp>` which started shipping in 2.3.0

https://github.com/catchorg/Catch2

First release that changed to the new `<catch2/catch.hpp>` header: https://github.com/catchorg/Catch2/releases/tag/v2.3.0


ax3l:
forgot  to bump version in `README.md`, will do now

PR #369, opened 2018-09-21
Python Stride Check: Too Strict

ax3l:
The stride check in Attribute and Record Component writing was too strict. One-element strides are allowed also for non-char types. Also, overlong 1D arrays are ok and should not use integer division for check.

### DIFF
--- a/src/binding/python/Attributable.cpp
+++ b/src/binding/python/Attributable.cpp
@@ -124,12 +124,12 @@ bool setAttributeFromBufferInfo(
             // std::cout << "    stride '" << d << "': "
             //           << buf.strides[d] / buf.itemsize
             //           << " - " << buf.shape[d] << std::endl;
-            if( buf.strides[d] / buf.itemsize != buf.shape[d] ) // general criteria
+            if( buf.strides[d] != buf.shape[d] * buf.itemsize ) // general criteria
             {
-                if( buf.ndim == 1u && buf.strides[0] / buf.itemsize > buf.shape[0] )
+                if( buf.ndim == 1u && buf.strides[0] > buf.shape[0] * buf.itemsize )
                     ; // ok in 1D
-                else if( buf.ndim == 1u && buf.strides[0] / buf.itemsize == 1u )
-                    ; // ok and occurs for byte strings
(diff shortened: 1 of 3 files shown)


ax3l:
thx to @HighIander for finding a data set that triggered this :)

src/binding/python/Attributable.cpp
```diff
@@ -124,12 +124,12 @@ bool setAttributeFromBufferInfo(
...
             //           << " - " << buf.shape[d] << std::endl;
-            if( buf.strides[d] / buf.itemsize != buf.shape[d] ) // general criteria
+            if( buf.strides[d] != buf.shape[d] * buf.itemsize ) // general criteria
             {
-                if( buf.ndim == 1u && buf.strides[0] / buf.itemsize > buf.shape[0] )
+                if( buf.ndim == 1u && buf.strides[0] > buf.shape[0] * buf.itemsize )
                     ; // ok in 1D
-                else if( buf.ndim == 1u && buf.strides[0] / buf.itemsize == 1u )
-                    ; // ok and occurs for byte strings
+                else if( buf.strides[0] == buf.itemsize )
```
ax3l, line 131: dang, this should have been `buf.strides[d]`

src/binding/python/RecordComponent.cpp
```diff
@@ -232,12 +232,12 @@ void init_RecordComponent(py::module &m) {
...
                 //           << " - " << a.shape()[d] << std::endl;
-                if( a.strides()[d] / a.itemsize() != a.shape()[d] ) // general criteria
+                if( a.strides()[d] != a.shape()[d] * a.itemsize() ) // general criteria
                 {
-                    if( a.ndim() == 1u && a.strides()[0] / a.itemsize() > a.shape()[0] )
+                    if( a.ndim() == 1u && a.strides()[0] > a.shape()[0] * a.itemsize() )
                         ; // ok in 1D
-                    else if( a.ndim() == 1u && a.strides()[0] / a.itemsize() == 1u )
-                        ; // ok and occurs for byte strings
+                    else if( a.strides()[0] == a.itemsize() )
```
ax3l, line 239: dang, this should have been `buf.strides[d]`

PR #379, opened 2018-10-10
requirements.txt: Upper bounds

ax3l:
New major releases of these dependencies would very likely cause problems and were not tested on release of this software. If they work, we can bump the bound (in packages recipes aka package builds).

For the bare software install, we don't define those bounds to allow bumping them in packaging alone without the need for a new mainline release.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-cmake>=3.10.0
-pybind11>=2.2.3
-numpy>=1.15.0
+cmake>=3.10.0,<4.0.0
+pybind11>=2.2.3,<3.0.0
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
-cmake>=3.10.0
-pybind11>=2.2.3
-numpy>=1.15.0
+cmake>=3.10.0,<4.0.0
+pybind11>=2.2.3,<3.0.0
+numpy>=1.15.0,<2.0.0
 setuptools>=38.6
```
ax3l, line 4: I do not get their versioning scheme, but today v40 works as well.

PR #381, opened 2018-10-19
[WIP] FindADIOS.cmake: Handle -lc++

ax3l:
In some cross-compile scenarios on conda-forge, a `-lc++` sneaks into the OSX build which we don't handle well and think it's a missing library: https://github.com/conda-forge/openpmd-api-feedstock/pull/11

Try to handle more gracefully.


ax3l:
Probably not needed and was only triggered by falsely building a lib for OSX and using it on Linux and vice versa on conda-forge.

PR #386, opened 2018-11-05
store_chunk order of arguments

### LINKED ISSUE #368: storeChunk argument order
The order of `storeChunk`/`store_chunk` should be changed to allow useful defaults to:
- `data`
- `offset`
- `extent`

or
- `data`
- `extent`
- `offset` = n-D {0}

For example `extent` can be derived from `data` if a container is given (needs more overloads in C++, e.g. if we add xtensor later on but also for `std::vector` useful) but also in the Python interface where we naturally work with the buffer protocol.

Not sure if `extent` or `offset` should be last. My thoughts are: in serial I/O where one most of the time only writes a single chunk, this does not matter much since both can be defaulted. In parallel I/O, when using the overloads described in the future, one could only default for extent, so the order is `data`, `offset`, `extent` = default is probably best.

ax3l:
The order of `storeChunk`/`store_chunk` should be changed to allow useful defaults to:
- `data`
- `offset`
- `extent`

Introduce defaults for the Python bindings to make serial I/O more simple.
Close #368 .

Also introduces an `openPMD::traits::IsContiguousContainer` trait that emulates the C++17 concept [ContiguousContainer](https://en.cppreference.com/w/cpp/named_req/ContiguousContainer). With that, the C++11 frontend `store_chunk` also gets default arguments for offset and extent for `std::array` and `std::vector` (and anything else the user implements a `openPMD::traits::IsContiguousContainer` trait for).

User-level upgrade guide documented in #385

### DIFF
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -21,6 +21,7 @@
 #pragma once
 
 #include "openPMD/backend/BaseRecordComponent.hpp"
+#include "openPMD/auxiliary/ShareRaw.hpp"
 #include "openPMD/Dataset.hpp"
 
 #include <cmath>
@@ -30,6 +31,9 @@
 #include <string>
 #include <sstream>
 #include <stdexcept>
+#include <type_traits>
+#include <vector>
+#include <array>
 
 // expose private and protected members for invasive testing
 #ifndef OPENPMD_protected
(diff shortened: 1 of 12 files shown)


include/openPMD/RecordComponent.hpp
```diff
@@ -82,8 +118,15 @@ class RecordComponent : public BaseRecordComponent
...
+
     template< typename T >
-    void storeChunk(Offset, Extent, std::shared_ptr< T >);
+    void storeChunk(std::shared_ptr< T >, Offset, Extent);
+
+    template< typename T_ContiguousContainer >
+    typename std::enable_if<
+        traits::IsContiguousContainer< T_ContiguousContainer >::value
+    >::type
+    storeChunk(T_ContiguousContainer, Offset = {0u}, Extent = {-1u});
```
ax3l, line 129: bug: container must be passed as reference to avoid copy to temporaries: #414 #416

PR #389, opened 2018-11-08
ThirdParty Libs: IMPORTED

ax3l:
CMake thirdParty libraries should be marked as imported. We do so by creating `openPMD::thirdparty::` targets that are imported only and depend on `find_package` or `add_subdirectory` targets to decouple them.

Imported targets are always `SYSTEM` libraries: https://cmake.org/cmake/help/v3.8/prop_tgt/NO_SYSTEM_FROM_IMPORTED.html

That allows us to use stricter warnings in our project than third party headers support.

Original issue:
The nlohmann-json library should be marked as `SYSTEM` linked lib. This will cause it to be linked with `-isystem`, allowing us to use warnings in openPMD that are stricter than the ones in the dependency (since their headers do not throw warnings anymore).

https://github.com/nlohmann/json/issues/1339


ax3l:
oh wait, that's not working with `target_link_libraries` and target objects...
Probably the dependency needs to be clearly marked as [`IMPORTED`](https://cmake.org/cmake/help/v3.8/prop_tgt/NO_SYSTEM_FROM_IMPORTED.html).

Dang, that's a [CMake 3.11.0+ feature](https://cmake.org/cmake/help/v3.11/release/3.11.html#commands)! Updated accordingly. Maybe even needs [3.11.2+](https://cmake.org/cmake/help/v3.11/release/3.11.html#id2)

franzpoeschel:
Hi @ax3l,
I am experiencing an issue including the openPMD API into other projects since this PR has been merged. Minimal failing CMake file:
```cmake
cmake_minimum_required(VERSION 3.0)
project(openPMD_test)
add_executable(openPMD_mpi_write mpi_write.cpp)
find_package(MPI REQUIRED)
add_definitions(-DOMPI_SKIP_MPICXX)
find_package(openPMD REQUIRED)
target_link_libraries(openPMD_mpi_write PRIVATE openPMD::openPMD MPI::MPI_C MPI::MPI_CXX)
```
Fails with the output:
```
-- Configuring done
CMake Error at CMakeLists.txt:3 (add_executable):
  Target "openPMD_mpi_write" links to target
  "openPMD::thirdparty::mpark_variant" but the target was not found.  Perhaps
  a find_package() call is missing for an IMPORTED target, or an ALIAS target
  is missing?


CMake Error at CMakeLists.txt:3 (add_executable):
  Target "openPMD_mpi_write" links to target
  "openPMD::thirdparty::mpark_variant" but the target was not found.  Perhaps
  a find_package() call is missing for an IMPORTED target, or an ALIAS target
  is missing?
```

The openPMD build configuration is as follows:
```
> cmake .
-- MPark.Variant: Using INTERNAL version 1.3.0
-- Note: run
   <src_dir>/.travis/download_samples.sh
to add example files to samples/git-sample/ directory!

openPMD build configuration:
  library Version: 0.6.2
  openPMD Standard: 1.1.0
  C++ Compiler: GNU 8.2.0 
    /opt/openmpi/bin/mpic++

  Installation prefix: /usr/local
        bin: bin
        lib: lib
    include: include
      cmake: lib/cmake/openPMD

  Additionally, install following third party libraries:
    MPark.Variant: ON

  Build Type: Release
  Testing: OFF
  Invasive Tests: OFF
  Internal VERIFY: ON
  Build Options:
    MPI: ON
    JSON: OFF
    HDF5: OFF
    ADIOS1: OFF
    ADIOS2: OFF
    PYTHON: OFF

-- Configuring done
-- Generating done
-- Build files have been written to: <build_dir>/openPMD-api-build-MPI
```
Do you have an idea whether the issue is on my side? When I checkout the last commit before this one has been merged, everything works fine.

ax3l:
I think it's on my side, indeed there might be a `find_package` missing after import or an export of the new targets. Will check!

PR #391, opened 2018-11-12
CMake 3.12.4+: _ROOT Vars

ax3l:
Honor `<PackageName>_ROOT` variables, see
  https://cmake.org/cmake/help/v3.12/policy/CMP0074.html

Should be set to avoid warnings in CMake 3.12.4+.

- [x] verify warning arises


ax3l:
Verified on Hypnos (HZDR). The warning kicks in as soon as a `<Package>_ROOT` variable is set and CMake 3.12.4+ is used without policy hint (which defaults to `OLD`). `NEW` is good, we want that :)

PR #395, opened 2018-11-14
[WIP] Fix Read-Only File Series

### LINKED ISSUE #394: Read-Only on Files Without Write Permissions
**Describe the bug**
Opening a `Series` as read-only does not work on files that do not have write permission for the user under Linux.

**To Reproduce**

Terminal:
```bash
$ chmod u-w simData_5000.h5

$ h5ls -v tmp_dir/simData_5000.h5 
Opened "tmp_dir/simData_5000.h5" with sec2 driver.
# ...

$ python -c 'import openPMD; openPMD.Series("simData_%T.h5", openPMD.Access_Type.read_only)'
HDF5-DIAG: Error detected in HDF5 (1.8.14) thread 0:
  #000: H5F.c line 604 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 990 in H5F_open(): unable to open file: time = Wed Nov 14 14:25:54 2018
, name = 'tmp_dir/simData_5000.h5', tent_flags = 1
    major: File accessibilty
    minor: Unable to open file
  #002: H5FD.c line 992 in H5FD_open(): open failed
    major: Virtual File Layer
  #003: H5FDsec2.c line 343 in H5FD_sec2_open(): unable to open file: name = 'tmp_dir/simData_5000.h5', errno = 13, error message = 'Permission denied', flags = 1, o_flags = 2
    major: File accessibilty
    minor: Unable to open file
Traceback (most recent call last):
  File "<string>", line 1, in <module>
RuntimeError: Failed to open HDF5 file tmp_dir/simData_5000.h5
```

**Expected behavior**
Read-only files should be readable with only read-permissions.
`h5dump` and `h5ls` seem to work with those as well.

**Software Environment:**
 - version of openPMD-api: 0.6.3-alpha
 - installed openPMD-api via: pip (sdist; compiled from source)
 - operating system: Ubuntu 14.04.5 LTS "trusty"
 - machine: Hypnos (HZDR)
 - name and version of Python implementation: CPython 3.6.2
 - version of HDF5: 1.8.14
 - version of ADIOS1: none
 - name and version of MPI: none

**Additional context**
First seen in #390

ax3l:
Fix #394

Opening a HDF5 `Series` as read-only fails if the files are not *writable*.

For some reason, on `open()` in HDF5 we thing that our `m_handler->accessType` is `READ_WRITE` when opened with `READ_ONLY`... [This hack](https://github.com/openPMD/openPMD-api/blob/0.6.3-alpha/src/Series.cpp#L487-L491) might be the origin. Hush, that will need a lot of refactoring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -195,6 +195,7 @@ jobs:
         - if [ ! -d samples/git-sample/ ]; then
             if [ "$USE_SAMPLES" = "ON" ]; then
               $TRAVIS_BUILD_DIR/.travis/download_samples.sh;
+              chmod u-w samples/git-sample/*.h5;
             fi
           fi
         - if [ "$USE_PYTHON" == "ON" ]; then
@@ -503,6 +504,7 @@ jobs:
         - cd $HOME/build
         - if [ ! -d samples/git-sample/ ]; then
             $TRAVIS_BUILD_DIR/.travis/download_samples.sh;
+            chmod u-w samples/git-sample/*.h5;
           fi
         - CXXFLAGS="-g -O0 -fprofile-arcs -ftest-coverage" CXX=$CXX CC=$CC
           cmake
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -101,6 +101,7 @@ Bug Fixes
 - fix ODR issue in ADIOS1 backend corrupting the ``AbstractIOHandler`` vtable #415
 - fix race condition in MPI-parallel directory creation #419
 - ADIOS1: fix use-after-free in parallel I/O method options #421
+- do not require write permissions to open ``Series`` read-only #395
 
 Other
 """""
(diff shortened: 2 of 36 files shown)


ax3l:
Hm, after testing around this could also be a HDF5 bug... (using 1.8.13 locally)

anokfireball:
Oh, boy... Sorry you have this mess at hand.
The problem indeed very likely results from the internal hack during `fileBased` reading (which was introduced to avoid raised exceptions when setting an `Attribute` value in read-only `Series`).

Seeing that (in the hack) I already introduced two `IOHandler::accessType`s
 - the one actually desired by the user (and to be used by the backend)
 - an apparent one for the frontend logic (that might not reflect the first one _only during the opening of a `Series`_ and can be safely ignored by the backend)

, it probably would be less of a hassle to modify `AbstractIOHandler`.

Let's see if my C++ is not too rusty yet. :wrench:

ax3l:
superseded by #460, hurray!

PR #398, opened 2018-11-30
Parallel Benchmark: Cleanup

ax3l:
Clean leftovers from new MPI benchmark in #346.

- properly add auxiliary benchmark sources
- properly name for build with MPI only
- fix BlockSlicer API: no MPI needed
- includes
- add changelog

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -580,7 +580,7 @@ set(openPMD_EXAMPLE_NAMES
     5_write_parallel
     6_dump_filebased_series
     7_extended_write_serial
-    8_mpi_benchmark
+    8_benchmark_parallel
 )
 set(openPMD_PYTHON_EXAMPLE_NAMES
     2_read_serial
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ Changelog
 
 [Summary]
 
-Changes to "0.6.2-alpha"
+Changes to "0.6.3-alpha"
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
 Features
@@ -27,11 +27,12 @@ Features
   - works with Python 3.7 #376
   - setup.py for sdist #240
 - Backends: JSON support added #384 #393 #338
+- Parallel benchmark added #346 #398
 
 Bug Fixes
 """""""""
 
-- support reading series with varying or no iteration padding in filename #388
+- spurious MPI C++11 API usage in ParallelIOTest removed #396
 
 Other
 """""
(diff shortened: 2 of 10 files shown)


CMakeLists.txt
```diff
@@ -286,8 +286,7 @@ set(CORE_SOURCE
         src/backend/MeshRecordComponent.cpp
         src/backend/PatchRecord.cpp
         src/backend/PatchRecordComponent.cpp
-        src/backend/Writable.cpp
-        src/benchmark/mpi/OneDimensionalBlockSlicer.cpp)
```
ax3l: hm, unless this shall be auxiliary as the documentation you wrote suggests (part of the library, not only part of the benchmark example...).

Maybe it's well-placed in `auxiliary` then. But then it should be included in our facade headers as well.

@franzpoeschel did you pass the whole communicator in case you want to determine e.g. a MPI grid position from it?

Maybe an `openPMD::benchmark` CMake target makes sense, outside of examples.

ax3l: side note on AppleClang: [travis](https://travis-ci.org/openPMD/openPMD-api/jobs/461721314) when MPI is off:
```
/Applications/Xcode-9.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib:
file: lib/libopenPMD.a(OneDimensionalBlockSlicer.cpp.o) has no symbols
```

Will redesign the API, this does not need to pull MPI (for the communicator) if the only thing it needs is size and rank.

franzpoeschel: Passing the rank directly is probably the more sensible solution, yes.

franzpoeschel: >part of the library, not only part of the benchmark example

I was intending for it to become part of the library to allow users to easily dissect a dataset into pieces. If we go with this, `OneDimensionalBlockSlicer.cpp` should probably be put back into the sources list again (possibly after moving it to `auxiliary`). If not, the complete class (i.e. `OneDimensionalBlockSlicer`) should probably go into the example file and be ridded from the API altogether otherwise. But this will result in users having to needlessly copy-paste the class if they want to use it..?

>Maybe it's well-placed in auxiliary then. 

So you suggest placing the whole parallel benchmark into `auxiliary` (headers as well as sources)? Sounds good for me.

ax3l: Yes, prospectively that would make a good place for that helper. I'll leave it in the dir for now. We also need to thing of the auxiliary headers for this workflow, they are currently not part of the public API but we still might want to add facade headers... dunno yet :)

PR #403, opened 2018-12-14
IMPORTED Targets: Fix Downstream Use

ax3l:
Follow-up to use CMake third party librarys as IMPORTED.
Make sure
- `PRIVATE` JSON target does not leak into install
- variant target is known

Follow-up to #389 in order to fix reported issues when used in third-party libs: https://github.com/openPMD/openPMD-api/pull/389#issuecomment-446970532


ax3l:
@franzpoeschel does the latest `dev` with this PR merged solve your problem?

PR #405, opened 2019-01-02
Clarify instructions for installing from source

sbastrakov:
Funnily, on Windows `mkdir -p openPMD-api-build` creates a directory named `-p`. Clarify that administrative rights are normally required - it is actually part of the error output when run without it, but can be easily missed in the wall of warnings. Also fix a typo with double `required`.

### DIFF
--- a/docs/source/install/install.rst
+++ b/docs/source/install/install.rst
@@ -70,7 +70,7 @@ Linux & OSX
 
    git clone https://github.com/openPMD/openPMD-api.git
 
-   mkdir -p openPMD-api-build
+   mkdir openPMD-api-build
    cd openPMD-api-build
 
    # optional: for full tests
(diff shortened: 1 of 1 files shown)


docs/source/install/install.rst
```diff
@@ -89,20 +89,21 @@ Linux & OSX
...
 
-   # sudo might be required required for system paths
+   # sudo might be required for system paths
    cmake --build . --target install
 
 Windows
 ^^^^^^^
 
-Replace the last commands with:
+The process is basically similar to Linux & OSX, with just a couple of minor tweaks. Use ``mkdir openPMD-api-build`` without the ``-p`` option and replace the last three commands with
```
ax3l: Thanks!
In order to simplify, can you please:
- start a new line with the second sentence
- remove `-p` from the Linux example, it's not needed and ok to fail if existing dir is in place
- replace "Use ``mkdir openPMD-api-build`` without the ``-p`` option" with "Use ``ps ..\openPMD-api\.travis\download_samples.ps1`` to download sample files for tests"

sbastrakov: Thanks, done and squashed.

ax3l (approved):

sbastrakov:
@ax3l I've tried the [current `dev`](https://github.com/openPMD/openPMD-api/commit/0af7408b7bf93c7b27b1d259f8835ca428f6f3a0) on Windows. Basically, it works.

The Windows helper command to download the sample files printed during the configuration looks reasonable, I could not check if it works since this is not allowed on my work laptop.

A tiny issue is that the change in this PR somehow got lost or reverted and copying from [the instructions](https://github.com/openPMD/openPMD-api/blob/dev/README.md#from-source) again creates a directory named `-p` in addition to the correct one.

ax3l:
Thanks for testing!

I was actually wondering if the `conda` (conda-forge channel) install works via the GUI. That caused some trouble for you last time and I was wondering of the latest updates to it helped.

sbastrakov:
Ah, I misunderstood you originally. Installing from the command line works. From GUI (Anadonda Navigator) installation fails on my machine with "Multiple errors occurred" and seemingly no further information.

ax3l:
Interesting, thanks!

PR #410, opened 2019-01-05
IOTask: AbstractParameter Slice Safety

ax3l:
The `IOTask` constructor caused [problems with Clang 4.0.1 on OSX](https://github.com/conda-forge/openpmd-api-feedstock/pull/15), where the `Parameter< Operation::CREATE_FILE >` does seem not to copy the `name` member. Thanks to @mingwandroid for first seeing this and debugging a lot with me!

~~It turns out that we copy the `AbstractParameter` base class, which leads to object slicing.~~
Probably implicit copy constructors are weirdly handled in this old Clang version, so we make them explicitly deleted as the core guidelines recommend.

Also, `IOTasks` might better have an explicit copy constructor since we copy it around (`std::queue::push()`, etc.)

https://stackoverflow.com/questions/274626/what-is-object-slicing
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-virtual
https://en.wikipedia.org/wiki/Object_slicing

### DIFF
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -28,6 +28,7 @@
 #include <map>
 #include <vector>
 #include <string>
+#include <utility>
 
 
 namespace openPMD
@@ -68,9 +69,16 @@ enum class Operation
 struct AbstractParameter
 {
     virtual ~AbstractParameter() = default;
+    AbstractParameter() = default;
+    //AbstractParameter(AbstractParameter&&) = default;
+
+    // avoid object slicing
+    AbstractParameter(const AbstractParameter&) = delete;
+    AbstractParameter& operator=(const AbstractParameter&) = delete;
+    virtual std::unique_ptr< AbstractParameter > clone() const = 0;
 };
 
-/** @brief Typesafe description of all required Arguments for a specified Operation.
+/** @brief Typesafe description of all required arguments for a specified Operation.
  *
  * @note    Input operations (i.e. ones that transfer data from persistent files
  *          to logical representations in openPMD-api) use shared pointers to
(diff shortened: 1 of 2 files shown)


include/openPMD/IO/IOTask.hpp
```diff
@@ -238,15 +238,15 @@ class IOTask
            Parameter< op > const& p)
             : writable{w},
               operation{op},
-              parameter{new Parameter< op >(p)}
```
ax3l, line 241: that abstract base class should not be copied to each other. That is object slicing:
https://stackoverflow.com/questions/274626/what-is-object-slicing

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-virtual
https://en.wikipedia.org/wiki/Object_slicing

ax3l: Hm, but since we explicitly new on the `Parameter< op >` this should not be slicing...
At least `clang-tidy-6.0` does not find it as such:
```bash
clang-tidy-6.0 test.cpp -checks='*,cppcoreguidelines-slicing' --
```

Could still be an AppleClang compiler bug that optimizes it to one.

include/openPMD/IO/IOTask.hpp
```diff
@@ -235,22 +440,22 @@ class IOTask
      */
     template< Operation op >
     IOTask(Writable* w,
```
ax3l: maybe also make `explicit`to avoid unintentional implicit conversions

include/openPMD/IO/IOTask.hpp
```diff
@@ -235,22 +440,22 @@ class IOTask
...
-           Parameter< op > const& p)
+           Parameter< op > const & p)
             : writable{w},
               operation{op},
-              parameter{new Parameter< op >(p)}
+              parameter{p.clone()}
     { }
 
     template< Operation op >
     IOTask(Attributable* a,
```
ax3l: maybe also make `explicit`to avoid unintentional implicit conversions

include/openPMD/IO/IOTask.hpp
ax3l: Nah, it is likely still slicing since we `push` `IOTasks` into `std::qeues` ... but they are only copies of `shared_ptr`s of `AbstractParameter`, so no copy.

Maybe it's just a compiler bug with the implicit/explicit copy constructors. Now it's safe.

ax3l (commented):
Maybe copy shared pointer as {other.parameters} (really share ownership by value)

PR #412, opened 2019-01-05
Travis-CI: Clang 7

ax3l:
Add clang 7 coverage in CI.

Also adds a `-fsanitze=address` entry which currently only enables a backend-free run.
After #414 is fixed we can enable JSON as well. Also uncovered an ODR issue with ADIOS: #413 

Fixing an unused var warning in the MPI benchmark and an unnecessary copy in the python bindings for `Container` in this PR as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,11 @@ cache:
 env:
   global:
     - BUILD: ~/buildTmp
-    - CXXFLAGS: "-std=c++11"
+    - CXXFLAGS: "-Werror"
     - SPACK_ROOT: $HOME/.cache/spack
     - PATH: $PATH:$HOME/.cache/spack/bin
     - PYBIND11_VERSION=@2.2.3
+    - USE_JSON: ON
 
 addons:
   apt:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -239,10 +241,44 @@ jobs:
...
+      addons:
+        apt:
+          <<: *apt_common_sources
+          packages:
+            - *clang50_deps
+            - libopenmpi-dev
+            - openmpi-bin
+      script: *script-cpp-unit
+      before_install: &clang_init
+        - CXX=clang++ && CC=clang && CXXFLAGS="-Werror -Wno-deprecated"
```
ax3l, line 262: `-Wno-deprecated` is needed since pybind11 is not yet included as `SYSTEM` and uses a deprecated Python API in 3.7+.
Both include and issue are fixed in upcoming releases of pybind11 (v2.3.0).

.travis.yml
```diff
@@ -239,10 +241,44 @@ jobs:
...
+      compiler: clang
+      script: *script-cpp-unit
+      before_install:
+        - CXX=clang++ && CC=clang && CXXFLAGS="-fsanitize=address"
+        - export ASAN_OPTIONS=detect_odr_violation=0  # this must be fixed: #413
     # Clang 9.1.0-apple + Python 2.7.15 @ OSX
     - <<: *test-cpp-unit
       name: AppleClang@9.1.0 -MPI +PY@2.7 +H5 +ADIOS1
       os: osx
+      osx_image: xcode9.4
```
ax3l, line 281: current default, just want to fixate it

PR #429, opened 2019-01-10
JSON Backend: Avoid Tmp Copies

ax3l:
Adds (`const`) references where applicable to avoid temporary copies in the JSON backend.

@franzpoeschel can you please review they are all safe and ok?

Found initially via clang-tidy `performance-*` checks.

### DIFF
--- a/src/IO/JSON/JSONIOHandlerImpl.cpp
+++ b/src/IO/JSON/JSONIOHandlerImpl.cpp
@@ -53,7 +53,6 @@ namespace openPMD
 
     std::future< void > JSONIOHandlerImpl::flush( )
     {
-
         AbstractIOHandlerImpl::flush( );
         for( auto const & file: m_dirty )
         {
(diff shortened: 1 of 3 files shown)


src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -1452,7 +1451,7 @@ namespace openPMD
     {
         nlohmann::json j;
         CppToJSON< T > ctj;
-        for( auto a: v )
+        for( auto const & a: v )
         {
             j.push_back( ctj( a ) );
```
ax3l: maybe we could also `emplace_back` here

src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -1475,7 +1474,7 @@ namespace openPMD
...
```
ax3l: maybe we could also `emplace_back` here

ax3l:
@franzpoeschel I'll merge this for now, but please feel free to comment still if something is off and we fix it in a follow-up PR.

PR #454, opened 2019-01-27
mpi4py

### LINKED ISSUE #99: Python MPI Support
Part of #32: Add MPI support to the Python API with [mpi4py](https://github.com/mpi4py/mpi4py).
As in the C++ API, MPI support must be provided optionally.

Example implementations of pybind11 + mpi4py:
- [ADIOS2](https://github.com/ornladios/ADIOS2/tree/master/bindings/Python)

ax3l:
Add support for MPI-parallel I/O in python3 bindings.

Close #99

### To Do

- [x] CMake Logic
- [x] min. versions in docs (should work with any mpi4py version, at least for > 1.0, from what the impls for the comm type look; but [mpi4py.run](https://mpi4py.readthedocs.io/en/3.0.0/mpi4py.run.html) options are new in 2.1+)
- [x] Examples
- [ ] Unit Tests
- [x] extras in `setup.py`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -574,10 +574,6 @@ if(openPMD_HAVE_PYTHON)
         PDB_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}
         COMPILE_PDB_OUTPUT_DIRECTORY ${CMAKE_PYTHON_OUTPUT_DIRECTORY}
     )
-
-    #if(openPMD_HAVE_MPI)
-    #    target_link_libraries(openPMD.py PRIVATE PythonModule::mpi4py)
-    #endif()
 endif()
 
 # tests
@@ -868,6 +864,16 @@ endif()
 # Python Examples
 if(openPMD_HAVE_PYTHON)
     if(EXISTS "${openPMD_BINARY_DIR}/samples/git-sample/")
+        execute_process(COMMAND ${PYTHON_EXECUTABLE}
+            -m mpi4py
+            -c "import mpi4py.MPI"
+            RESULT_VARIABLE MPI4PY_RETURN)
+
+        if(NOT MPI4PY_RETURN EQUAL 0)
+            message(STATUS "Note: mpi4py not found. "
+                           "Skipping MPI-parallel Python examples.")
+        endif()
+
         foreach(examplename ${openPMD_PYTHON_EXAMPLE_NAMES})
             add_custom_command(TARGET openPMD.py POST_BUILD
                 COMMAND ${CMAKE_COMMAND} -E copy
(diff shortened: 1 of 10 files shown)


src/binding/python/Series.cpp
```diff
@@ -30,7 +34,24 @@ using namespace openPMD;
...
+        .def(py::init<std::string const&, AccessType>(),
+            py::arg("filepath"), py::arg("access_type"))
+#if openPMD_HAVE_MPI
+        .def(py::init([](std::string const& filepath, AccessType at, py::object &comm){
+            //! @todo perform mpi4py import test?
+            //!       required C-API init?
+            //!       https://bitbucket.org/mpi4py/mpi4py/src/3.0.0/demo/wrap-c/helloworld.c
+            // if( import_mpi4py() < 0 ) { here be dragons }
+
+            MPI_Comm *mpiCommPtr = PyMPIComm_Get(comm.ptr());
```
ax3l: trying to avoid `mpi4py/mpi4py.h` include:
this is [just a cast](https://github.com/mpi4py/mpi4py/blob/3.0.0/src/mpi4py/MPI/CAPI.pxi#L118-L119), we can do this ourselves with a `dynamic_cast`.

Note: not all implementations implement `MPI_Comm` as `int`

src/binding/python/Series.cpp
```diff
@@ -30,7 +34,24 @@ using namespace openPMD;
...
 
-        .def(py::init<std::string const&, AccessType>())
+        .def(py::init<std::string const&, AccessType>(),
+            py::arg("filepath"), py::arg("access_type"))
+#if openPMD_HAVE_MPI
+        .def(py::init([](std::string const& filepath, AccessType at, py::object &comm){
+            //! @todo perform mpi4py import test?
+            //!       required C-API init?
+            //!       https://bitbucket.org/mpi4py/mpi4py/src/3.0.0/demo/wrap-c/helloworld.c
+            // if( import_mpi4py() < 0 ) { here be dragons }
```
ax3l, line 66: trying to avoid `mpi4py/mpi4py.h` include:
Function generated during build. This just seems to import and perform tests on functions and type signatures for binary compatibility (e.g. the usual: MPI during build time must be MPI at runtime).
Afterwards the imported test module is unloaded again. We can probably skip it and check e.g. versions with `pybind11`'s import method.

CMakeLists.txt
```diff
@@ -276,6 +276,11 @@ if(openPMD_HAVE_PYTHON)
         INTERFACE pybind11::pybind11)
 endif()
 
+# TODO: Check if mpi4py is available
+# ref: https://github.com/ornladios/ADIOS2/blob/v2.3.0/cmake/FindPythonModule.cmake
+# find_package(PythonModule REQUIRED COMPONENTS mpi4py mpi4py/mpi4py.h)
```
ax3l: maybe we can avoid his, see implementation of `src/binding/python/Series.cpp`

src/binding/python/Series.cpp
```diff
@@ -20,17 +20,43 @@
...
+
+            //MPI_Comm *mpiCommPtr = PyMPIComm_Get(comm.ptr());
+            MPI_Comm *mpiCommPtr = std::dynamic_pointer_cast<MPI_Comm>(comm.ptr());
+            if( mpiCommPtr == nullptr ) {
+                throw std::runtime_error("Series: MPI communicator cast failed. "
+                                         "(Mismatched MPI at compile vs. runtime?)");
+            }
+
+            return new Series(filepath, at, *mpiCommPtr); //! @todo do we already clone the communicator?
```
ax3l: `MPI_Comm_dup` missing in parallel HDF5 backend (done in ADIOS backend)

ax3l: nope, also done in HDF5

CMakeLists.txt
```diff
@@ -859,12 +865,23 @@ if(openPMD_HAVE_PYTHON)
...
-                )
+                if(${examplename} MATCHES "^.*_parallel$")
+                    if(openPMD_HAVE_MPI)
+                        # see https://mpi4py.readthedocs.io/en/stable/mpi4py.run.html
+                        add_test(NAME Example.py.${examplename}
+                            COMMAND ${MPI_TEST_EXE} ${PYTHON_EXECUTABLE} -m mpi4py
+                                ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${examplename}.py
+                            WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+                        )
+                    endif()
```
ax3l, line 896:
```cmake
else()
    continue()
endif()
```

src/binding/python/Series.cpp
```diff
@@ -20,17 +20,93 @@
...
+                throw std::runtime_error("Series: MPI communicator is a nullptr.");
+
+            // check type string to see if this is mpi4py
+            //   __str__ (pretty)
+            //   __repr__ (unambiguous)
+            //   mpi4py: <mpi4py.MPI.Intracomm object at 0x7f998e6e28d0>
+            //   pyMPI:  ... (todo)
+            py::str const comm_pystr = py::repr(comm);
+            std::string const comm_str = comm_pystr.cast<std::string>();
+            if( comm_str.substr(0, 12) != std::string("<mpi4py.MPI.") )
```
ax3l, line 80: Should probably only match `mpi4py` Communicators ([and derived](https://github.com/mpi4py/mpi4py/blob/3.0.0/src/mpi4py/MPI.pxd#L100-L141)/same-layouted objects)

Issue #138, opened 2018-04-20
load/storeChunk: slice view


ax3l:
The `A.storeChunk(data, offset, extent)` and `A.loadChunk(data, chunk_offset, chunk_extent)` methods should for syntactic sugar provide a slice-view access of the form:

```C++
// write (copy)
A[view] = data;

// returns: std::unique_ptr< A.dtype >
auto B = A[view];
// hm, maybe that's not possible, so let's say something like
auto B = A[view].astype< some_dtype >();
```

ax3l:
Hm, maybe we can also implements views when we add the `xtensor` frontend

https://xtensor.readthedocs.io/en/latest/basic_usage.html

related to #32

ax3l:
Hm, I think we will need this C++ `Slice` class even earlier if we want to have a seamless python API.
And two helpers for:
`Extent`, `Offset` -> `Slice` and `Slice` -> `Extent`, `Offset`

No need to expose it in the C++ frontend at this point, tough.

ax3l:
Implemented with #458

PR #458, opened 2019-02-04
Python: Slice Protocol for Record Component

ax3l:
Draft for supporting slices in Python `Record_Component` load and store operations.

Related to #32 and #138.

Also fixes the "padded stride" (contiguous) detection in user-provided arrays:
Strides are not the only way to create non-contiguous buffers in Python. Also, non-owned views might look like stride-free, but are actually padded strides in another buffer object. Use the underlying C API now to check if contiguous.

Refs:

- https://docs.scipy.org/doc/numpy-1.15.0/reference/arrays.indexing.html

### To Do

- [x] unit tests including exception checks
- [x] ellipsis support
- [ ] newaxis support
- [x] `[()]` and `[:, :]` support
- [x] one slice (or even one index) abbreviation support
- [x] index omission support
- [x] over-selected upper range support in slices
- [x] single index slice reduces `ndim`, but `N:N+1` and `...` does not
- [x] `load_chunk`
- [x] `store_chunk`
- [x] update more examples
- [x] changelog

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -258,38 +258,28 @@ store_chunk(RecordComponent & r, py::array & a, Offset const & offset, Extent co
             );
     }
 
-    // verify input array is not strided (aka contiguous)
-    // a pseudo stride in one dimension is ok as long as all other
-    // dimensions stride exactly by itemsize
-    bool strideFound = false;
-    // all strides but one are element strides?
-    bool allOthersOne = true;
-    for( auto d = 0; d < a.ndim(); ++d )
+    /* required are contiguous buffers
+     *
+     * - not strided with paddings
+     * - not a view in another buffer that results in striding
+     */
+    Py_buffer* view = new Py_buffer();
+    int flags = PyBUF_STRIDES | PyBUF_FORMAT;
+    if( PyObject_GetBuffer( a.ptr(), view, flags ) != 0 )
     {
-        // std::cout << "    stride '" << d << "': "
-        //           << a.strides()[d] / a.itemsize()
-        //           << " - " << a.shape()[d] << std::endl;
-
-        if( a.strides()[d] == a.itemsize() )
-            continue; // ok to stride on element level
-        else
-        {
-            if( a.strides()[d] == a.shape()[d] * a.itemsize() )
-                allOthersOne = false;
-            else                         // padded stride
-                if( strideFound )        // more than one stride found
-                    throw py::index_error(
-                        "strides in chunk are inefficient, not implemented!");
-                else
-                    strideFound = true;  // first found stride, ok with allOthersOne
-        }
+        delete view;
+        throw py::error_already_set();
     }
-    if( strideFound && !allOthersOne )   // one padded stride and others not element strides
+    bool isContiguous = ( PyBuffer_IsContiguous( view, 'A' ) != 0 );
+    PyBuffer_Release( view );
+    delete view;
+
+    if( !isContiguous )
         throw py::index_error(
             "strides in chunk are inefficient, not implemented!");
-
     // @todo in order to implement stride handling, one needs to
-    //       loop over the input data strides during write below
+    //       loop over the input data strides during write below,
+    //       also data might not be owned!
 
     // store
     auto const dtype = dtype_from_numpy( a.dtype() );
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -29,6 +29,7 @@
 
 #include <string>
 #include <algorithm>
+#include <tuple>
 #include <iostream>
 #include <cstdint>
 #include <exception>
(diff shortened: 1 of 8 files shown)


src/binding/python/RecordComponent.cpp
```diff
@@ -29,13 +29,155 @@
...
+parseTupleSlices(uint8_t const ndim, Extent const & full_extent, py::tuple const & slices) {
+    if( py::len(slices) > ndim )
+        throw std::runtime_error("__getitem__: "
+            "too many indices for dimension of record component!");
+
+    Offset offset(ndim, 0u);
+    Extent extent(ndim, 1u);
+    for( uint8_t i = 0u; i < py::len(slices); ++i )
+    {
+        py::slice slice = py::cast< py::slice >( slices[i] );
```
ax3l: This can be:
- a `py::slice`
- a signed integer for a single axis
- `np.newaxis` (=`None`)
- an ellipsis `py::ellipsis` (allowed once per tuple)

Already handled but possible as well: an empty tuple (select all: `[()]`)

test/python/unittest/API/APITest.py
```diff
@@ -522,6 +522,187 @@ def testLoadSeries(self):
...
+
+        # more exotic syntax
+        #   ref: https://docs.scipy.org/doc/numpy-1.15.0/reference/arrays.indexing.html
+
+        # - [ ]: [M, L:LARGE, K:LARGE] over-select upper range
+        #                              (is numpy-allowed: crop to max range)
+        #        (not implemented)
+        with self.assertRaises(IndexError):
+            np.testing.assert_allclose(
+                pos_y[0:pos_y.shape[0]+10],
```
ax3l: here and below: many `flush()`es missing for comparison of actually loaded values

src/binding/python/RecordComponent.cpp
```diff
@@ -29,13 +29,300 @@
...
+                throw std::runtime_error("UH OH!");
+            if( start != 0 )
+                throw std::runtime_error(std::string("start is: ") + std::to_string(start));
+            if( stop != 66032 )
+                throw std::runtime_error(std::string("stop is: ") + std::to_string(stop));
+            if( slicelength != 66032 )
+                throw std::runtime_error(std::string("slicelength is: ") + std::to_string(slicelength));
+
+            // verified for size later in C++ API
+            offset.at(curAxis) = 0;
```
ax3l: temporary: must be `start` again :)

src/binding/python/RecordComponent.cpp
```diff
@@ -29,13 +29,300 @@
...
+
+#if( ( PYBIND11_VERSION_MAJOR == 2 && PYBIND11_VERSION_MINOR >= 3 ) || ( PYBIND11_VERSION_MAJOR > 2 ) )
+        try
+        {
+            py::ellipsis newaxis = py::cast< py::ellipsis >( slices[i] );
+
+            // only allowed once
+            if( posEllipsis != -1 )
+                throw py::index_error("an index can only have a single ellipsis ('...')");
+            posEllipsis = setSlice - 1;  // zero-based index
```
ax3l: posEllipsis = curAxis;

src/binding/python/RecordComponent.cpp
```diff
@@ -29,13 +29,300 @@
...
+
+            uint8_t const indexEllipsis = uint8_t(posEllipsis);
```
ax3l: can be replaced with `curAxis`

src/binding/python/RecordComponent.cpp
```diff
@@ -29,13 +29,300 @@
...
+            // how many slices does the ellipsis represent
+            uint8_t const numSlicesEllipsis =
+                ndim
+                - indexEllipsis            // slices before
+                - numSlicesAfterEllipsis;  // slices after
+
+            // fill ellipsis indices
+            for( ; curAxis < curAxis + numSlicesEllipsis; ++curAxis )
+            {
+                offset.at(k) = 0;
```
ax3l: `k` -> `curAxis`

src/binding/python/RecordComponent.cpp
```diff
@@ -29,13 +29,300 @@
...
+                extent.at(k) = full_extent[i];
```
ax3l: `k`, `i` -> `curAxis`

test/python/unittest/API/APITest.py
```diff
@@ -522,6 +524,365 @@ def testLoadSeries(self):
...
+            d1 = E_x[0, 0, E_x.shape[2]]
+
+        with self.assertRaises(IndexError):
+            d1 = pos_y[pos_y.shape[0]]
+        with self.assertRaises(IndexError):
+            d1 = w[w.shape[0]]
+
+        # cropped to upper range
+        d1 = E_x[10:E_x.shape[0]+2, 0, 0]
+        d2 = pos_y[10:pos_y.shape[0]+3]
```
ax3l, line 765: doesn't seem to work: `heap-use-after-free`

ax3l: ah, `series` survives function, flush missing

PR #462, opened 2019-02-06
[WIP] spdlog for multi-level multi-sink logging

### LINKED ISSUE #461: Logging / Tracing / Debugging facilities
In order to better understand the internal flow when debugging or developing, it would be very helpful to have run-time switchable logging, including debugging and tracing levels.

For a simple and fast solution, https://github.com/gabime/spdlog could be a good option.
More ready-made C++ logging facilities are [available](https://github.com/fffaraz/awesome-cpp#logging).

anokfireball:
_TBD_

Close #461.

### To Do

- [x] push implementation
- [ ] use spdlog 1.4+ and include link compiled `.cpp` file instead of increasing compile-time via header-only lib?
- Docs:
  - [x] `README.md`
  - [x] `docs/source/dev/buildoptions.rst`
  - [x] `docs/source/dev/dependencies.rst`


.travis.yml
```diff
@@ -583,8 +593,8 @@ install:
       export TRAVIS_PYTHON_VERSION=3.7.2;
     fi
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then
-      if [ "$TRAVIS_PYTHON_VERSION" == "2.7" ]; then
-        export TRAVIS_PYTHON_VERSION=2.7.14;
+      if [ "$TRAVIS_PYTHON_VERSION" == "3.6" ]; then
+        export TRAVIS_PYTHON_VERSION=3.6.14;
```
anokfireball: @ax3l I'll need your help finding the spack-relevant version of Py3.

.travis.yml
```diff
@@ -43,11 +43,16 @@ jobs:
...
+    - <<: *style_python
+      name: PY@3.7 +PEP8
+      python: "3.7"
+      dist: xenial
+      sudo: true
     # - &style_cpp
     #   stage: 'Style'
     #   name: clang-format@5.0.0
     #   language: python
-    #   python: "2.7"
```
ax3l: Python 3 is only needed if the python bindings are build

.travis.yml
ax3l: I think there should be no changed to travis needed at all, as long as we test against the internal version of spdlog to start with

.travis.yml
anokfireball: Well in this case, we require Python due to `run-clang-format.py`. It's 2019 already and I'd say it's time to leave 2.7 behind altogether.

.travis.yml
anokfireball: I'll revert the Python-related changes and create a seperate PR for them.

.travis.yml
ax3l: Ok, then  you only need to change the entries in the matrix that use `run-clang-format.py` if it is not backwards compatible :)
But I think all of this can be done in another PR and is independent of this change, no?
update: just saw the comment above, yep.

CMakeLists.txt
```diff
@@ -292,6 +294,15 @@ if(openPMD_HAVE_PYTHON)
         INTERFACE pybind11::pybind11)
 endif()
 
+# external library: spdlog (optional)
+if(openPMD_USE_LOGGING STREQUAL AUTO)
+    set(openPMD_HAVE_LOGGING (${CMAKE_BUILD_TYPE} STREQUAL "Debug" OR ${CMAKE_BUILD_TYPE} STREQUAL "RelWithDebInfo"))
```
ax3l: We can do the control this way, but it's slightly different as the other dependencies then.

For example, if the `openPMD_USE_XY` is `AUTO` and set to `EXTERNAL` and *not* found, this will not fail in our current logic. Here it would.

docs/source/dev/buildoptions.rst
```diff
@@ -83,3 +85,10 @@ Tests
 
 By default, tests and examples are built.
 In order to skip building those, pass ``-DBUILD_TESTING=OFF`` or ``-DBUILD_EXAMPLES=OFF`` to your ``cmake`` command.
+
+
+Logging
+-------
+
+By default, logging is only enabled in ``Debug`` builds.
+In order to enable this feature unconditionally, pass ``-DopenPMD_USE_LOGGING=ON`` or ``-DopenPMD_USE_LOGGING=OFF`` to disable it unconditinoally.
```
ax3l, line 94: little typo:
```suggestion
In order to enable this feature unconditionally, pass ``-DopenPMD_USE_LOGGING=ON`` or ``-DopenPMD_USE_LOGGING=OFF`` to disable it unconditionally.
```

include/openPMD/auxiliary/Logging.hpp
```diff
@@ -0,0 +1,93 @@
...
+namespace openPMD
+{
+namespace auxiliary
+{
+namespace log
+{
+
+enum class Level
+{
+    TRACE,
```
ax3l, line 36: can we also add a "timing" mode in this or another level? "throughput" (bytes/s) would also be awesome.
cc @franzpoeschel

CMakeLists.txt
anokfireball: Shamelessly copied from Pybind11 as that probably is the most consistent with the other CMake `_USE_` logic.

include/openPMD/auxiliary/Logging.hpp
anokfireball: Interesting idea. That would mean we count bytes during all backend IO and take timestamps in the ~fronend~ triggering calls.

Intuitively, this would result in a different code structure to the other log facilities.

```cpp
auto series = openPMD::Series("data%T.h5", openPMD::AccessType::CREATE);
RecordComponent ideas = ...

std::shared_prt< double > patch = ...

// assume DEFER
ideas.store_chunk(patch, {0}, {patch_size});
LOG_TIME(series.flush());     /* example.cpp:8 series.flush() took 0.314s */

std::array< std::shared_prt< double >, 42 > lots_of_patches = ...
for( auto patch : lots_of_patches )
    ideas.store_chunk(patch, ...);
LOG_THROUGHPUT(series.flush());  /* example.cpp:13 series.flush() averaged 314MB/s */
```

I'll see what I can do. This will be pretty invasive, though.

CMakeLists.txt
ax3l: Yes, I just wonder if we should couple it to the Debug build logic since we will use log levels on various intensity also in Release builds and vice versa, might want to disable it for some Debug runs.

So maybe just handle it's default activation just like any other functional extension? Just a thought.

include/openPMD/auxiliary/Logging.hpp
ax3l: Doesn't need to go into the first log version, we can add those in a follow-up if you like.

PR #482, opened 2019-02-18
Add ADIOS2 IO Backend

### LINKED ISSUE #148: ADIOS2 Backend
We want to add a backend for [ADIOS2](https://github.com/ornladios/ADIOS2).

 Although it has not yet reached feature parity with ADIOS1 at the time of writing, version [2.1.1](https://github.com/ornladios/ADIOS2/releases) of ADIOS2 already has a stable API we can use.

Resources: [how to write a backend](https://openpmd-api.readthedocs.io/en/latest/dev/backend.html)

franzpoeschel:
ADIOS2 backend. Close #148

## To Do

- [x] Fix known bugs
- [x] Test the parallel implementation
- [x] The frontend currently distinguishes backends by filename extension – ADIOS1 and ADIOS2 have the same extension (`.bp`) – find and implement a better solution
     (ADIOS2 can currently not yet open all files written by ADIOS1)
- [x] Deletion is unimplemented – probably postpone to a later PR *– resolved for now by throwing upon deletion*
- [x] avoid auto-formatting unchanged code for smaller/focused `diff` in the PR
- [x] Licensing information
- [x] Update documentation, including (defaults & min-version): `README.md`, `docs/source/dev/[buildoptions|dependencies].rst`
- [x] ADIOS2 does currently not seem to understand 128bit integers – look into that
- [x] If ADIOS2 is built with support for MPI but openPMD is built without it, ADIOS2 will fail with a runtime error – likely an ADIOS2 issue, verify this.
- [x] Merge #512 before this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,7 +59,7 @@ openpmd_option(MPI            "Parallel, Multi-Node I/O for clusters"     AUTO)
 openpmd_option(JSON           "JSON backend (.json files)"                AUTO)
 openpmd_option(HDF5           "HDF5 backend (.h5 files)"                  AUTO)
 openpmd_option(ADIOS1         "ADIOS1 backend (.bp files)"                AUTO)
-openpmd_option(ADIOS2         "ADIOS2 backend (.bp files)"                 OFF)
+openpmd_option(ADIOS2         "ADIOS2 backend (.bp files)"                AUTO)
 openpmd_option(PYTHON         "Enable Python bindings"                    AUTO)
 
 option(openPMD_HAVE_PKGCONFIG        "Generate a .pc file for pkg-config"   ON)
@@ -341,7 +341,10 @@ set(IO_SOURCE
         src/IO/HDF5/ParallelHDF5IOHandler.cpp
         src/IO/JSON/JSONIOHandler.cpp
         src/IO/JSON/JSONIOHandlerImpl.cpp
-        src/IO/JSON/JSONFilePosition.cpp)
+        src/IO/JSON/JSONFilePosition.cpp
+        src/IO/ADIOS/ADIOS2IOHandler.cpp
+        src/IO/ADIOS/ADIOS2Auxiliary.cpp
+        src/IO/InvalidatableFile.cpp)
 set(IO_ADIOS1_SEQUENTIAL_SOURCE
         src/auxiliary/Filesystem.cpp
         src/IO/ADIOS/ADIOS1IOHandler.cpp)
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -1,0 +1,77 @@
+.. _backends-adios2:
+
+ADIOS2 Backend
+==============
+
+openPMD supports writing to and reading from ADIOS2 ``.bp`` files.
+For this, the installed copy of openPMD must have been built with support for the ADIOS2 backend.
+To build openPMD with support for ADIOS2, use the CMake option ``-DopenPMD_USE_ADIOS2=ON``.
+For further information, check out the :ref:`installation guide <install>`,
+:ref:`build dependencies <development-dependencies>` and the :ref:`build options <development-buildoptions>`.
+
+
+I/O Method
+----------
+
+ADIOS2 has several engines for alternative file formats and other kinds of backends, yet natively writes to ``.bp`` files. At the moment, the openPMD API exclusively uses the BPFile engine.
+We currently leverage the default ADIOS2 transport parameters, i.e. ``POSIX`` on Unix systems and ``FStream`` on Windows.
+
+
+Backend-Specific Controls
+-------------------------
+
+The following environment variables control ADIOS2 I/O behavior at runtime.
+Fine-tuning these is especially useful when running at large scale.
+
+===================================== ======= ===================================================================
+environment variable                  default description
+===================================== ======= ===================================================================
+``OPENPMD_ADIOS2_HAVE_PROFILING``      ``1``   Turns on/off profiling information right after a run.
+``OPENPMD_ADIOS2_HAVE_METADATA_FILE``  ``1``   Online creation of the adios journal file (``1``: yes, ``0``: no).
+``OPENPMD_ADIOS2_NUM_SUBSTREAMS``      ``0``   Number of files to be created, 0 indicates maximum number possible.
+===================================== ======= ===================================================================
+
+Please refer to the `ADIOS2 manual, section 5.1 <https://media.readthedocs.org/pdf/adios2/latest/adios2.pdf>`_ for details.
+
+
+Best Practice at Large Scale
+----------------------------
+
+A good practice at scale is to disable the online creation of the metadata file.
+After writing the data, run ``bpmeta`` on the (to-be-created) filename to generate the metadata file offline (repeat per iteration for file-based encoding). 
+This metadata file is needed for reading, while the actual heavy data resides in ``<metadata filename>.dir/`` directories.
+Note that such a tool is not yet available for ADIOS2, but the ``bpmeta`` utility provided by ADIOS1 is capable of processing files written by ADIOS2.
+
+Further options depend heavily on filesystem type, specific file striping, network infrastructure and available RAM on the aggregator nodes.
+A good number for substreams is usually the number of contributing nodes divided by four.
+
+For fine-tuning at extreme scale or for exotic systems, please refer to the ADIOS2 manual and talk to your filesystem admins and the ADIOS2 authors.
+Be aware that extreme-scale I/O is a research topic after all.
+
+Selected References
+-------------------
+
+* Hasan Abbasi, Matthew Wolf, Greg Eisenhauer, Scott Klasky, Karsten Schwan, and Fang Zheng.
+  *Datastager: scalable data staging services for petascale applications,*
+  Cluster Computing, 13(3):277–290, 2010.
+  `DOI:10.1007/s10586-010-0135-6 <https://doi.org/10.1007/s10586-010-0135-6>`_
+
+* Ciprian Docan, Manish Parashar, and Scott Klasky.
+  *DataSpaces: An interaction and coordination framework or coupled simulation workflows,*
+  In Proc. of 19th International Symposium on High Performance and Distributed Computing (HPDC’10), June 2010.
+  `DOI:10.1007/s10586-011-0162-y <https://doi.org/10.1007/s10586-011-0162-y>`_
+
+* Qing Liu, Jeremy Logan, Yuan Tian, Hasan Abbasi, Norbert Podhorszki, Jong Youl Choi, Scott Klasky, Roselyne Tchoua, Jay Lofstead, Ron Oldfield, Manish Parashar, Nagiza Samatova, Karsten Schwan, Arie Shoshani, Matthew Wolf, Kesheng Wu, and Weikuan Yu.
+  *Hello ADIOS: the challenges and lessons of developing leadership class I/O frameworks,*
+  Concurrency and Computation: Practice and Experience, 26(7):1453–1473, 2014.
+  `DOI:10.1002/cpe.3125 <https://doi.org/10.1002/cpe.3125>`_
+
+* Robert McLay, Doug James, Si Liu, John Cazes, and William Barth.
+  *A user-friendly approach for tuning parallel file operations,*
+  In Proceedings of the International Conference for High Performance Computing, Networking, Storage and Analysis, SC'14, pages 229–236, IEEE Press, 2014.
+  `DOI:10.1109/SC.2014.24 <https://doi.org/10.1109/SC.2014.24>`_
+
+* Axel Huebl, Rene Widera, Felix Schmitt, Alexander Matthes, Norbert Podhorszki, Jong Youl Choi, Scott Klasky, and Michael Bussmann.
+  *On the Scalability of Data Reduction Techniques in Current and Upcoming HPC Systems from an Application Perspective,*
+  ISC High Performance 2017: High Performance Computing, pp. 15-29, 2017.
+  `arXiv:1706.00522 <https://arxiv.org/abs/1706.00522>`_, `DOI:10.1007/978-3-319-67630-2_2 <https://doi.org/10.1007/978-3-319-67630-2_2>`_
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2017-2019 Fabian Koller
+/* Copyright 2017-2019 Fabian Koller and Franz Pöschel
  *
  * This file is part of openPMD-api.
  *
@@ -524,6 +524,7 @@ isSame( openPMD::Datatype const d, openPMD::Datatype const e )
 #else
 #define OPENPMD_TEMPLATE_OPERATOR template operator
 #endif
+
 /**
  * Generalizes switching over an openPMD datatype.
  *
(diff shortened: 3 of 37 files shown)


src/Datatype.cpp
```diff
@@ -303,4 +303,91 @@ namespace openPMD
...
+                if (d == Datatype::ARR_DBL_7)
+                    continue;
+                Datatype basic = basicDatatype(d);
+                if (basic == d)
+                    continue;
+                res[basic] = d;
+            }
+            return res;
+        };
+        static auto map {initializer()};
```
ax3l: probably just use `map(initalizer());` here as the default `&&` constructor: https://de.cppreference.com/w/cpp/container/map/map

Otherwise some compilers (clang 5.0, gcc 4.8-4.9) interpret the result as a `std::initializer_list` due to `auto`.

src/IO/InvalidatableFile.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    return fileState.operator bool( );
+}
+
+
+openPMD::InvalidatableFile::FileState::FileState( std::string s ) :
+    name { std::move( s ) }
+{}
+
+unsigned long
```
ax3l: `std::size_t` should be used here, as in it's declaration.
(not all C++ implementations, e.g. windows, implement `std::size_t` as `unsigned long` :-) )

include/openPMD/IO/InvalidatableFile.hpp
```diff
@@ -0,0 +1,76 @@
...
+        explicit operator bool( ) const;
+    };
+}
+
+namespace std
+{
+    template< >
+    struct hash< openPMD::InvalidatableFile >
+    {
+        typedef openPMD::InvalidatableFile argument_type;
```
ax3l: please use `using` in new code for `typedef`s :)

src/Datatype.cpp
franzpoeschel: fixed

src/IO/InvalidatableFile.cpp
franzpoeschel: Fixed. There were also two places in `ADIOS2IOHandler.hpp` with the same issue.

include/openPMD/IO/InvalidatableFile.hpp
franzpoeschel: fixed

franzpoeschel:
Does someone have suggestions on how to handle `The frontend currently distinguishes backends by filename extension – ADIOS1 and ADIOS2 have the same extension (.bp) – find and implement a better solution`? Possible solutions include:

- Leave the implementation as is, i.e. if openPMD has been built with support for ADIOS2, always use ADIOS2.
   **Pro**: Easy to implement, ADIOS2 is at some point probably going to fully replace ADIOS1. 
   **Con**: Building openPMD with support for ADIOS2 and ADIOS1 at the same time becomes useless, because ADIOS1 will never be chosen for a backend, ADIOS2 currently fails opening files that have been written with ADIOS1.
- Make openPMD prefer ADIOS2, but use ADIOS1 as a fallback if ADIOS2 fails opening a file. This solution implies that ADIOS1 would never be used for writing new files.
   **Pro**: Easy to use, "silent" transition to ADIOS2 that a frontend user would ideally not even notice
   **Con**: Could result in a hacky implementation?
- Let the user decide by explicitly specifying the backend version, e.g. by passing an argument map to the `Series` constructor
  **Pro**: Most control for the user; at some point we will need a mechanism to provide further arguments anyway
  **Con**: ?

ax3l:
Hm, besides current hiccups reading ADIOS1 files with ADIOS2 I think the first option is the cleanest and makes sense. If we have the choice to use the newer C++ ADIOS2 code base this is preferable over the ADIOS1 C code base.
We can talk to the ADIOS people next week and ask if the file ending will change at some point, e.g. to `.bp4` or what they think.

franzpoeschel:
[This](https://github.com/ornladios/ADIOS2/issues/1239#issuecomment-467951583) sounds related to the issues that I am encountering with opening BP files written by ADIOS1 with ADIOS2. For reference, I created a small sample file, once using the ADIOS1 and once the ADIOS2 backend of the openPMD API. Reading the ADIOS1 file with ADIOS1's `bpls` gives the following output.
```bash
bpls serial1.bp 
  unsigned long long  /tmp/data/100/meshes/E/x_chunkSize0    scalar
  unsigned long long  /tmp/data/100/meshes/E/x_chunkOffset0  scalar
  unsigned long long  /tmp/data/100/meshes/E/x_chunkSize1    scalar
  unsigned long long  /tmp/data/100/meshes/E/x_chunkOffset1  scalar
  double              /data/100/meshes/E/x                   {3, 3}
  byte                /__adios__/timer_labels_1              {8, 16}
  double              /__adios__/timers_1                    {1, 8}
```
The output contains statistical data written by ADIOS1, suggesting that this is the same corner case discussed in the referenced thread. Trying to use ADIOS2's `bpls` fails:
```
bpls serial1.bp 

Error: Could not open this file with any ADIOS2 file reading engines
```
The file written by the ADIOS2 backend can be read by either version of `bpls` with the same output:
```
bpls serial2.bp
  double   /data/100/meshes/E/x  {3, 3}
```
Statistical data is not present here.

Is it worth further pursuing the corner case (?) of reading ADIOS1 files by using the ADIOS2 backend?

ax3l:
> issues that I am encountering with opening BP files written by ADIOS1 with ADIOS2.

Looks to me like the counter reading of ADIOS1 statistics was fixed in ADIOS2 in your linked thread. Can you just try if it works with the latest `master` of ADIOS2? It contains this fix: https://github.com/ornladios/ADIOS2/commit/880649a7b051c418c24cc69f0b535f43b1c227cb
Otherwise, please open an issue in ADIOS2 with the little reproducer.

ax3l:
@franzpoeschel can you please rebase your PR against the latest `dev` and keep the CI unchanged? I will add a commit to modify `.travis.yaml` afterwards to your branch.

ax3l:
@franzpoeschel please do not change `.travis.yaml`, please revert those changes as I will help you with those

franzpoeschel:
Sorry, I accidentally picked the wrong version in rebasing

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1225 @@
...
-#endif
+#endif // openPMD_HAVE_MPI
 
-#if openPMD_HAVE_ADIOS2 && !openPMD_HAVE_MPI
-ADIOS2IOHandler::ADIOS2IOHandler(std::string path, AccessType at)
-        : AbstractIOHandler(std::move(path), at),
-          m_impl{new ADIOS2IOHandlerImpl(this)}
-{ }
+ADIOS2IOHandlerImpl::ADIOS2IOHandlerImpl( AbstractIOHandler * handler )
+: AbstractIOHandlerImplCommon( handler ), m_ADIOS{ADIOS2_DEBUG_MODE}
```
ax3l: Clang-tidy does complain about the `ADIOS2` constructor here.
  https://travis-ci.org/openPMD/openPMD-api/jobs/535234522

Maybe try:
```suggestion
: AbstractIOHandlerImplCommon( handler ),
#if openPMD_HAVE_MPI
    m_ADIOS{MPI_COMM_SELF, ADIOS2_DEBUG_MODE}
#else
    m_ADIOS{ADIOS2_DEBUG_MODE}
#endif
```
? I sincerely hope this does not require an `MPI_Init` or run with `mpirun/srun`...
Weird that the serial overload is missing if MPI is on: https://adios2.readthedocs.io/en/latest/components/components.html#adios

Or are you missing a second argument? https://github.com/ornladios/ADIOS2/blob/v2.3.1/source/adios2/core/ADIOS.h#L88
```suggestion
: AbstractIOHandlerImplCommon( handler ), m_ADIOS{ADIOS2_DEBUG_MODE, "C++"}
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1236 @@
...
+        filePositionToString( setAndGetFilePosition( writable->parent ) );
+    std::string suffix = auxiliary::removeSlashes( parameters.path );
+    std::string infix = auxiliary::ends_with( prefix, '/' ) ? "" : "/";
+
+    /* ADIOS has no concept for explicitly creating paths.
+     * They are implicitly created with the paths of variables/attributes. */
+
+    if ( writable->abstractFilePosition )
+    {
+        *writable->abstractFilePosition = ADIOS2FilePosition(
```
ax3l: How awesome is that?! [Clang-tidy](https://travis-ci.org/openPMD/openPMD-api/jobs/535698753) found an issue of [object slicing](https://en.wikipedia.org/wiki/Object_slicing), cool!

If you really need to copy to the abstract class here, maybe consider implementing a `clone()` method or rewrite/store it another way.

Refs:
- https://github.com/openPMD/openPMD-api/blob/dev/include/openPMD/IO/IOTask.hpp / #410
- https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-slicing.html

ax3l:
Are you sure the last commit avoids object slicing? Just looks to me like a trick to silence the check but actually still has the same issue.

franzpoeschel:
Do you think so? The last commit was definitely object slicing, since I dereferenced the old object and moved the new object into the old one. With the new implementation, I create a new object alltogether and the [shared pointer specialisation of `std::swap`](https://en.cppreference.com/w/cpp/memory/shared_ptr/swap2) only swaps pointers, so no copy/move assignment takes place actually.

franzpoeschel:
Alright, but thinking about it further, if I actually swap pointers, there's no use in reusing the old pointer so I will have to think of another fix.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1243 @@
...
+        filePositionToString( setAndGetFilePosition( writable->parent ) );
+    std::string suffix = auxiliary::removeSlashes( parameters.path );
+    std::string infix = auxiliary::ends_with( prefix, '/' ) ? "" : "/";
+
+    /* ADIOS has no concept for explicitly creating paths.
+     * They are implicitly created with the paths of variables/attributes. */
+
+    if ( writable->abstractFilePosition )
+    {
+        auto tempPtr = std::dynamic_pointer_cast<AbstractFilePosition>(
```
ax3l: Why not simply a
```C++
writable->abstractFilePosition = std::make_shared< ADIOS2FilePosition >(
    prefix + infix + suffix, ADIOS2FilePosition::GD::GROUP );
```
? (As below...)
Why do you differentiate at all of something is in `writable->abstractFilePosition` already?
http://www.cplusplus.com/reference/memory/shared_ptr/operator=/

franzpoeschel: I was not sure whether the frontend wants the old pointer replaced, so I added the distinction, but the tests run fine when dropping it, so I think this will be the easiest solution.

ax3l: You decrease the use-count (and drop the old value) either way, at least how you wrote it.

franzpoeschel: Yep, that's what I meant with 
>Alright, but thinking about it further, if I actually swap pointers, there's no use in reusing the old pointer so I will have to think of another fix.

So I guess it's good we have an easier fix now ;)

ax3l: Yep, looks good :) I am still amazed and happy that clang-tidy helps us with such tricky cases, that's super nice :)

ax3l:
The `SerialIOTest` does seem to fail in case of enabled MPI support.
https://travis-ci.org/openPMD/openPMD-api/jobs/535757755
```
*** The MPI_Comm_dup() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
```

franzpoeschel:
If you remember the discussion that we had during the ADIOS2 workshop, I think we have reached the point where it becomes relevant. Looking into the [ADIOS2 headers](https://github.com/ornladios/ADIOS2/blob/master/bindings/CXX11/cxx11/ADIOS.h), there exist no constructors without `MPI_Communicator` when having MPI enabled.
:/

ax3l:
It's not ideal but wrapped by us if MPI headers are needed for our implementation files. The question is, if they internally still call all operations with `MPI_COMM_SELF` or implement a second path...

Update: oh yes, the `#else` [here](https://github.com/ornladios/ADIOS2/blob/02f246eb328a524eefd00286ce88ed1c0f2172f9/bindings/CXX11/cxx11/ADIOS.h#L67-L80) is a problem, too.

ax3l:
Opened in https://github.com/ornladios/ADIOS2/issues/1443

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -606,11 +613,13 @@ class ADIOS2IOHandler : public AbstractIOHandler
 
 #if openPMD_HAVE_MPI
 
-    ADIOS2IOHandler( std::string path, AccessType, MPI_Comm );
+    ADIOS2IOHandler( std::string path, AccessType, MPI_Comm = MPI_COMM_WORLD );
```
ax3l: Why default to `MPI_COMM_WORLD`?
I think that's dangerous, maybe `MPI_COMM_SELF` or even no default. It's not a class a user interfaces with.

franzpoeschel:
> It's not ideal but wrapped by us if MPI headers are needed for our implementation files.

What do you mean by that?
Currently, if ADIOS2 has been built with support for MPI, you *must* pass an MPI communicator to the constructor of `adios2::ADIOS` (the entry point of the library). So what causes the above error is that ADIOS2 is given a communicator (the default one that you have spotted in the constructor of `openPMD::ADIOS2IOHandler` and that I added as a default for exactly this reason, hoping for an easy fix of the situation) – but MPI has never been initialized.
I guess switching `MPI_COMM_WORLD` out for `MPI_COMM_SELF` will not solve the issue either?

ax3l:
Please apply `MPI_COMM_SELF` for serial writes, as a library we do not own `MPI_COMM_WORLD` and might be called in an actual MPI environment although we want to write serial.

> > It's not ideal but wrapped by us if MPI headers are needed for our implementation files.
>
> What do you mean by that?

ADIOS2: a single, MPI-pulling header is not ideal yet we only burry it in a `.cpp` implementation file. The actual trouble is therefore if some MPI-commands that require `MPI_Init` are called even if `MPI_COMM_SELF` is passed. This prevents building serial tools from MPI-enabled adios2 libs if we are unlucky.

franzpoeschel:
Done that, but it does not fix the aforementioned issue (at least not on my machine).

ax3l:
Yep, this will probably limit the applicability of ADIOS2 to two incompatible serial and parallel variants :-S
The only thing I can think about that will sometimes work is to sneakily init MPI in the serial case like this:
```C++
// in serial constructor
#if openPMD_HAVE_MPI
    int isInitialized = (int)false;
    bool m_weInitialized = false;
    MPI_Initialized(&isInitialized);
    if(!isInitialized) {
        // TODO: verify if comm is truly only MPI_COMM_SELF
        MPI_Init(nullptr, nullptr);
        m_weInitialized = true;
    }
#endif

// ...
// Serial destructor
#if openPMD_HAVE_MPI
    if(m_weInitialized)
        MPI_Finalize();
#endif
```
Not very pretty nor expected and can still require an `mpirun` in some cases... :(

> Done that, but it does not fix the aforementioned issue (at least not on my machine).

No problem, but it already fixes the use cases of serial writing from an actual, initialized, mpi-parallel program. It just does not (yet) fix serial writing from a serial program with a parallel adios2 lib, which is an upstream design flaw.
Admittedly, I am surprised it works in `bpls`: https://github.com/ornladios/ADIOS2/blob/7b3939beec73029f0fe7e62e54f29e0832e697fa/source/utils/bpls/bpls.cpp#L1184-L1186 Probably because they do exactly that trick: https://github.com/ornladios/ADIOS2/blob/7b3939beec73029f0fe7e62e54f29e0832e697fa/source/utils/bpls/bpls.cpp#L2936-L2938

Anyway, until this is solved: with a serial ADIOS2 library and no MPI seems to be a little problem with attributes in `SerialIOTest` still: https://travis-ci.org/openPMD/openPMD-api/jobs/535817586

franzpoeschel:
Interestingly, the tests all pass when building the most recent ADIOS2 from their master branch, but not from their release branch. Will look further into this, but I assume this is a bug that has been fixed in ADIOS2. Running the tests on my machine with the release version, I get the same error.

franzpoeschel:
I pushed a version that completely deactivates non-MPI constructors if the library has been built with support for MPI. I don't think we should be sneaky and initialize MPI on the user's behalf, even if this means that the user needs to include MPI when wanting to use serial IO with an MPI-enabled openPMD API.

ax3l:
Admittedly, I think we should do the `MPI_Init`/`MPI_Finalize` [work-around](https://github.com/openPMD/openPMD-api/pull/482#issuecomment-494854971) for ADIOS2 until we have an upstream solution that does not require a MPI context.

franzpoeschel:
The failing tests are due to the fact that deletion of attributes is broken with the current release of ADIOS2 (fixed with [#1306](https://github.com/ornladios/ADIOS2/pull/1306) and [#1453](https://github.com/ornladios/ADIOS2/pull/1453)). The ADIOS2 backend internally deletes attributes to simulate overwriting existing ones, see [here](https://github.com/franzpoeschel/openPMD-api/blob/9d94d05b034b55babc95ab6056d2f08f52f4ee8d/src/IO/ADIOS/ADIOS2IOHandler.cpp#L733).

docs/source/backends/adios2.rst
```diff
@@ -0,0 +1,77 @@
...
+Backend-Specific Controls
+-------------------------
+
+The following environment variables control ADIOS1 I/O behavior at runtime.
+Fine-tuning these is especially useful when running at large scale.
+
+==================================== ======= ===================================================================
+environment variable                 default description
+==================================== ======= ===================================================================
+``OPENPMD_ADIOS_HAVE_PROFILING``     ``1``   Turns on/off profiling information right after a run.
```
ax3l: maybe make those vars `_ADIOS2_` suffixed (with the `2`)?

docs/source/backends/adios2.rst
```diff
@@ -0,0 +1,77 @@
+.. _backends-adios2:
+
+ADIOS2 Backend
+==============
+
+openPMD supports writing to and reading from ADIOS2 ``.bp`` files.
+For this, the installed copy of openPMD must have been built with support for the ADIOS2 backend.
+To build openPMD with support for ADIOS, use the CMake option ``-DopenPMD_USE_ADIOS2=ON``.
```
```suggestion
To build openPMD with support for ADIOS2, use the CMake option ``-DopenPMD_USE_ADIOS2=ON``.
```

docs/source/backends/adios2.rst
```diff
@@ -0,0 +1,77 @@
...
+----------
+
+ADIOS2 has several engines for alternative file formats and other kinds of backends, yet natively writes to ``.bp`` files. At the moment, the openPMD API exclusively uses the BPFile engine.
+We currently leverage the default ADIOS2 transport parameters, i.e. ``POSIX`` on Unix systems and ``FStream`` on Windows.
+
+
+Backend-Specific Controls
+-------------------------
+
+The following environment variables control ADIOS1 I/O behavior at runtime.
```
```suggestion
The following environment variables control ADIOS2 I/O behavior at runtime.
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1231 @@
...
+                 auxiliary::getEnvNum( "OPENPMD_ADIOS_HAVE_METADATA_FILE", 1 ) )
+            {
+                m_IO.SetParameter( "CollectiveMetadata", "On" );
+            }
+            else
+            {
+                m_IO.SetParameter( "CollectiveMetadata", "Off" );
+            }
+            if ( 1 ==
+                 auxiliary::getEnvNum( "OPENPMD_ADIOS_HAVE_PROFILING", 1 ) )
```
ax3l: see above: maybe name them `OPENPMD_ADIOS2_HAVE_PROFILING` with a `2`?

test/SerialIOTest.cpp
```diff
@@ -23,14 +23,17 @@ std::vector<std::tuple<std::string, bool>> getBackends() {
     // first component: backend file ending
     // second component: whether to test 128 bit values
     std::vector<std::tuple<std::string, bool>> res;
+#if openPMD_HAVE_JSON
+    res.emplace_back("json", false);
+#endif
 #if openPMD_HAVE_ADIOS1
     res.emplace_back("bp", true);
 #endif
+#if openPMD_HAVE_ADIOS2 && ! openPMD_HAVE_MPI
```
ax3l: two things:
- we should be able to do this with `openPMD_HAVE_MPI`, too (see `MPI_Init` work-around until upstream is fixed)
- if ADIOS1 *and* ADIOS2 are present, this seams to `emplace_back` `"bp"` twice

test/ParallelIOTest.cpp
```diff
@@ -17,77 +17,37 @@
...
-
-            for( int i = 0; i < 3; ++i )
-                REQUIRE(raw_ptr[i] == constant_value);
-        }
-    } catch (no_such_file_error& e)
-    {
-        std::cerr << "git sample not accessible. (" << e.what() << ")\n";
-        return;
-    }
+#if openPMD_HAVE_ADIOS2
```
ax3l: this might add `"bp"` twice if both ADIOS1 and ADIOS2 are present

CMakeLists.txt
```diff
@@ -248,7 +248,7 @@ endif()
 
 # external library: ADIOS2 (optional)
 if(openPMD_USE_ADIOS2 STREQUAL AUTO)
-    find_package(ADIOS2 2.1.0)
+    find_package(ADIOS2 2.4.0)
```
ax3l: must be updated in line 258 as well :)

ax3l:
I updated the caches on travis for ADIOS2 v2.4.0. In the morning after CI has [refreshed the last merge](https://travis-ci.org/openPMD/openPMD-api/builds/553055629), please rebase against the latest `dev` and force-push :)

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1261 @@
...
+
 #if openPMD_HAVE_MPI
-        : AbstractIOHandler(std::move(path), at, MPI_COMM_NULL)
+ADIOS2IOHandler::ADIOS2IOHandler( std::string path, AccessType at,
+                                  MPI_Comm comm )
+: AbstractIOHandler( std::move( path ), at, comm )
+{
+}
+
 #else
```
ax3l: is this `#else` still correct? Shouldn't this be an `#endif`?

ax3l (commented):
Little inline question.
Feel free to remove the `[WIP]` in the title and ping me when it's ready for merge :)

ax3l:
@franzpoeschel ping: any updates I should review? :)
See my last comment above :)

franzpoeschel:
I pushed two small commits, one to fix compression and another one to fix the `#endif`.

ax3l:
Cool, thx. Please feel free to remove the [WIP] in the title when you consider the PR ready :)

ax3l:
I introduced a `openPMD/config.hpp` header file for all `openPMD_HAVE_<feature>` defines in #532 which is causing a minor conflict. Feel free to rebase accordingly, it's just adding an include instead of adding the defines via the command line during compile.

ax3l:
@franzpoeschel Please rebase this PR against the latest `dev` to fix the merge conflicts :)
Just tell me when you consider it ready for review & merge.

ax3l:
@franzpoeschel ping :) What's the status?

franzpoeschel:
I am currently looking through my branches that build upon this in order to see whether some commits may also be relevant for this. After that, I think you can review it.

franzpoeschel:
@ax3l Ready for review from my side :)

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
```diff
@@ -0,0 +1,98 @@
...
+        template < int n, typename... Params >
+        size_t operator( )( Params &&... );
+    };
+
+    Datatype attributeInfo( adios2::IO &, std::string const & attributeName );
+} // namespace detail
+
+} // namespace openPMD
+
+#endif
```
ax3l: missing end-of-file newline and please add a comment `// openPMD_HAVE_ADIOS2`

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include "openPMD/Datatype.hpp"
+#include <adios2.h>
+#include <utility>
+#include <vector>
+
+namespace openPMD
+{
+namespace detail
+{
+    // ADIOS2 does not natively support boolean values
```
ax3l, line 35: urgh, [right](https://github.com/ornladios/ADIOS2/blob/v2.4.0/source/adios2/common/ADIOSMacros.h) -.-

but for which attributes do we use `bool` in openPMD? At least in the standard and extensions, I tried to avoid them due to the lack of a C99 implementation, lack of standard way in HDF5/ADIOS1 and tricky way in `std::vector<bool>`.

include/openPMD/Datatype.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2017-2019 Fabian Koller
+/* Copyright 2017-2019 Fabian Koller and Franz Pöschel
```
ax3l: Do you mind replacing ö with `oe` in source code that is seen by the compiler? Had some issues in the past with that... (reference missing...).

Also, since `ADIOS/ADIOS2*` files were stubs so far, feel free to list yourself as first author in those :+1:

include/openPMD/IO/ADIOS/ADIOS2FilePosition.hpp
```diff
@@ -20,14 +20,49 @@
  */
 #pragma once
 
+
 #include "openPMD/IO/AbstractFilePosition.hpp"
+#include <string>
```
ax3l, line 27: please include `<utility>` for `std::move`

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -21,64 +21,595 @@
...
-#include <memory>
+#include <memory> // shared_ptr
 #include <string>
+#include <unordered_map>
+#include <utility> // pair
+#include <vector>
+
+
+#if openPMD_HAVE_ADIOS2
+#include <adios2.h>
```
ax3l: pls align with three spaces to fourth position, looks better.
```C++
#if openPMD_HAVE_ADIOS2
#   include <adios2.h>
#   include "openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp"
#endif
```
https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#7-preprocessor

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -21,64 +21,595 @@
...
+#include "openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp"
+#endif
+
+#if openPMD_HAVE_MPI
+#include <mpi.h>
```
```suggestion
#   include <mpi.h>
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -21,64 +21,595 @@
...
-
-    AbstractIOHandler* m_handler;
-};  //ADIOS2IOHandlerImpl
-#else
-class ADIOS2IOHandlerImpl
-{ };
-#endif
+    static_assert(
+        sizeof( bool ) == 1,
+        "ADIOS2 backend needs a platform with boolean size equals one byte." );
```
ax3l: this makes me shiver. even if a `bool` is one size, a [`std::vector< bool >`](https://en.cppreference.com/w/cpp/container/vector_bool) can be anything. Are we safe? :)

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -21,64 +21,595 @@
...
+
+    /*
+     * IO-heavy actions are deferred to a later point. This map stores for
+     * each open file (identified by an InvalidatableFile object) an object
+     * that manages IO-heavy actions, as well as its ADIOS2 objects, i.e.
+     * IO and Engine object.
+     * Not to be accessed directly, use getFileData().
+     */
+    std::unordered_map< InvalidatableFile,
+                        std::unique_ptr< detail::BufferedActions > >
```
ax3l: I would put the closing `>` on a new line and write
```C++
std::unordered_map< InvalidatableFile,
                        std::unique_ptr< detail::BufferedActions >
> m_fileData;
```
for better readability.

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -21,64 +21,595 @@
...
+     * (2) the file's deferred IO-heavy actions
+     */
+    struct BufferedActions
+    {
+        BufferedActions( BufferedActions const & ) = delete;
+
+        std::string m_file;
+        adios2::IO m_IO;
+        std::vector< std::unique_ptr< BufferedAction > > m_buffer;
+        // std::optional would be more idiomatic, but it's not in
```
ax3l: feel free to make this a doxygen string and add a `@todo ...` here, so we can refactor it when we switch at some point to C++17...

include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
```diff
@@ -0,0 +1,259 @@
...
+    enum PossiblyExisting
+    {
+        PE_InvalidatableFile = 0,
+        PE_Iterator,
+        PE_NewlyCreated,
+    };
+
+    std::tuple< InvalidatableFile,
+                std::unordered_map< Writable *, InvalidatableFile >::iterator,
+                bool >
```
ax3l: see above, potentially on new line as `> getPossiblyExisting( std::string file );`?

include/openPMD/auxiliary/Environment.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
ax3l, line 23: please add the copyright header

include/openPMD/auxiliary/StringManip.hpp
```diff
@@ -183,5 +203,34 @@ join(std::vector< std::string > const& vs, std::string const& delimiter)
             return ss.str();
     }
 }
+
+inline std::string
+removeSlashes( std::string s )
```
ax3l, line 214: please add a doxygen string: e.g. explain where in the string slashes are removed

src/Datatype.cpp
```diff
@@ -303,4 +303,90 @@ namespace openPMD
         os << dt;
         return buf.str();
     }
+
+    std::vector<Datatype> openPMD_Datatypes{
+           Datatype::CHAR ,
```
ax3l, line 308: do the arguments here have to correspond to the exact order and number of arguments as elsewhere? Can we bring them closer together or at least cross-link them via a doxygen comment above each of them?

src/IO/ADIOS/ADIOS2Auxiliary.cpp
```diff
@@ -0,0 +1,172 @@
...
+            { "int", Datatype::INT },
+            { "unsigned int", Datatype::UINT },
+            { "long int", Datatype::LONG },
+            { "unsigned long int", Datatype::ULONG },
+            { "long long int", Datatype::LONGLONG },
+            { "unsigned long long int", Datatype::ULONGLONG },
+            { "float", Datatype::FLOAT },
+            { "double", Datatype::DOUBLE },
+            { "long double", Datatype::LONG_DOUBLE },
+            { "uint8_t", Datatype::UCHAR },
```
ax3l, line 96: no (u)char type in adios2?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1244 @@
...
 
-#include <utility>
-
+#include "openPMD/Datatype.hpp"
+#include "openPMD/IO/ADIOS/ADIOS2FilePosition.hpp"
+#include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
+#include "openPMD/auxiliary/Environment.hpp"
+#include "openPMD/auxiliary/Filesystem.hpp"
+#include "openPMD/auxiliary/StringManip.hpp"
+#include <iostream>
```
ax3l, line 29: pls add an empty newline. also add to header as author pls :)

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1244 @@
...
+        if ( !auxiliary::directory_exists( dir ) )
+        {
+            auto success = auxiliary::create_directories( dir );
+            VERIFY( success, "Could not create directory." );
+        }
+
+        associateWithFile( writable, shared_name );
+        this->m_dirty.emplace( shared_name );
+        getFileData( shared_name ).m_mode = adios2::Mode::Write; // WORKAROUND
+        // ↑ ADIOS2 does not yet implement ReadWrite Mode
```
ax3l: - pls remove non-ascii in code seen by compiler
- work-around: pls open and link corresponding adios2 issue :)

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1244 @@
...
+
+    /* ADIOS has no concept for explicitly creating paths.
+     * They are implicitly created with the paths of variables/attributes. */
+
+    writable->written = true;
+    writable->abstractFilePosition = std::make_shared< ADIOS2FilePosition >(
+        path, ADIOS2FilePosition::GD::GROUP );
+
+    auto varName = nameOfVariable( writable );
+    // m_IO.DefineVariable< void >( varName );
```
ax3l: dead code from debugging?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,58 +20,1244 @@
...
+    if ( !writable->written )
+    {
+        /* Sanitize name */
+        std::string name = auxiliary::removeSlashes( parameters.name );
+
+        auto file = refreshFileFromParent( writable );
+        auto filePos = setAndGetFilePosition( writable, name );
+        filePos->gd = ADIOS2FilePosition::GD::DATASET;
+        auto varName = filePositionToString( filePos );
+        // we use a unique_ptr to circumvent the fact that std::optional
```
ax3l: add same `@todo as above?`

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
```diff
@@ -0,0 +1,98 @@
+#pragma once
```
ax3l, line 22: please add copyright header :)

src/IO/ADIOS/ADIOS2Auxiliary.cpp
```diff
@@ -0,0 +1,172 @@
+#include "openPMD/config.hpp"
```
ax3l, line 22: please add copyright header :)

src/IO/InvalidatableFile.cpp
```diff
@@ -0,0 +1,70 @@
+
```
ax3l, line 21: please add copyright header :)

ax3l (commented):
Really nice work, thank you so much.
I added minor comments inline and then we should be able to merge this :)

Also, [ADIOS2 v2.5.0](https://github.com/ornladios/ADIOS2/releases/tag/v2.5.0) was just released. Any new features we want and love, to start with 2.5.0+ only?

Just for curiosity, when do you think a PR with the SST work is likely? Can you open the further features as `[WIP]` as well?

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l: Btw, I you really want it for internal use, I can add optional-lite:
https://gist.githubusercontent.com/ax3l/b4ae0d439c8503cfc08b9c04d65f1327/raw/f750ada9a6460f4df05657827f70dea574165c9c/awesome_standalone.md

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: There is no `VEC_BOOL` [datatype](https://github.com/openPMD/openPMD-api/blob/b4a904b6f1d0ce38294ab499066acfdd53fdc00a/include/openPMD/Datatype.hpp#L38) in openPMD, so this issue does not arise.
I added this check just to be sure that a single bool can be stored in an unsigned char, but I think the check is a bit pointless and can be removed.

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
franzpoeschel: I think we don't use it yet. I have used it in my `topic-streaming` branch to indicate whether an iteration has been finalized, but I noticed that it is not so straightforward to do so. Since `bool`s are coerced to `unsigned char`, you need to [pay attention](https://github.com/franzpoeschel/openPMD-api/blob/909fa5efcd6a692535aa1cee4f905b6d439b7b24/src/IO/ADIOS/ADIOS2IOHandler.cpp#L686) when reading whether an attribute was `bool` or whether it was genuinely an `unsigned char`. If I remember correctly, the other backends don't handle this case yet. My assumption would be that this hasn't come up yet since bools have not been used?
We might think about removing `BOOLEAN` alltogether from `openPMD::Datatypes`?

src/Datatype.cpp
franzpoeschel: The declaration is in [Datatype.hpp](https://github.com/franzpoeschel/openPMD-api/blob/909fa5efcd6a692535aa1cee4f905b6d439b7b24/include/openPMD/Datatype.hpp#L700). I think, it's better to move it below the `Datatype` enum, will do that.

src/IO/ADIOS/ADIOS2Auxiliary.cpp
franzpoeschel: Not any longer, see [here](https://github.com/ornladios/ADIOS2/pull/1452).

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: I think this would be helpful to have. At least, I missed having something similar to this several times now, and it's arguably better than a type that also forces us to do heap allocations.

include/openPMD/Datatype.hpp
franzpoeschel: I replaced my name with Poeschel also in other files of the project.

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
ax3l: It's probably not a bad idea, let's document this in an issue. At least for HDF5, there are some conventions to express bools. But many users are unaware of the implications for their data and it opens a lot of corner-cases for little gain. Needs probably more input.

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
ax3l: Sounds good, thanks for double-checking!

src/IO/ADIOS/ADIOS2Auxiliary.cpp
ax3l: hm...

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l: I can provide you that, documented in #563.

We must just avoid to use it in user-facing header files.

ax3l (approved):
Thank you so much, this is really great work! :sparkles: :rocket: 
We have ADIOS2 support, yay!!

docs/source/backends/adios2.rst
```diff
@@ -0,0 +1,77 @@
...
+-------------------------
+
+The following environment variables control ADIOS2 I/O behavior at runtime.
+Fine-tuning these is especially useful when running at large scale.
+
+===================================== ======= ===================================================================
+environment variable                  default description
+===================================== ======= ===================================================================
+``OPENPMD_ADIOS2_HAVE_PROFILING``      ``1``   Turns on/off profiling information right after a run.
+``OPENPMD_ADIOS2_HAVE_METADATA_FILE``  ``1``   Online creation of the adios journal file (``1``: yes, ``0``: no).
```
ax3l, line 30: @franzpoeschel are you sure that is still correct?

By default, this now creates a `<name>.bp/` directory in ADIOS2 and `bpls` (among others) seams to be able to operate directly on that - without an extra metadata file.

docs/source/backends/adios2.rst
```diff
@@ -0,0 +1,77 @@
...
+===================================== ======= ===================================================================
+
+Please refer to the `ADIOS2 manual, section 5.1 <https://media.readthedocs.org/pdf/adios2/latest/adios2.pdf>`_ for details.
+
+
+Best Practice at Large Scale
+----------------------------
+
+A good practice at scale is to disable the online creation of the metadata file.
+After writing the data, run ``bpmeta`` on the (to-be-created) filename to generate the metadata file offline (repeat per iteration for file-based encoding).
```
ax3l, line 41: see above, maybe we double-check the new workflow.

PR #483, opened 2019-02-19
Doc: AccessType

### LINKED ISSUE #152: AccessType Docstring
The options in the enum `AccessType` need a doxygen docstring for each allowed value :)

Is there a mode similar to `h5py`'s `File` object default (`a` - "append", not to a data set but a series)? [doc](http://docs.h5py.org/en/latest/high/file.html): *Read/write if exists, create otherwise*

ax3l:
Document the behavior of `AccessType` in a Series.
Close #152

@C0nsultant Did I document those correctly? The following questions in behavior arise to me:

### `CREATE`

If we open a new base `filePath` do we actively seek for existing files (for file-based encoding) with this pattern and remove them? I think we should, otherwise funnily mixed series might be created (imagine someone re-running a simulation and overwriting existing data half-way).

We might also want to add a mode that does create and truncates only iterations that are newly created (I think that's what we do right now in file-based access, right?). We should made both `create` modes behave consistent for file- and group-based series (from a user point of view).

Idea:
- `CREATE` truncates all existing iterations (independent if in file- or grou-based mode)
- `CREATE_ITERATION` truncates only newly written iterations

Alternatively and maybe better: allow/document to express the second mode in `READ_WRITE` (by just deleting an iteration explicitly).

### `READ_WRITE`

Anything I missed here?

### DIFF
--- a/include/openPMD/IO/AccessType.hpp
+++ b/include/openPMD/IO/AccessType.hpp
@@ -23,12 +23,12 @@
 
 namespace openPMD
 {
-/** File access permissions to use during IO.
- */
-enum class AccessType
-{
-    READ_ONLY,
-    READ_WRITE,
-    CREATE
-};  //AccessType
-} // openPMD
+    /** File access mode to use during IO.
+     */
+    enum class AccessType
+    {
+        READ_ONLY,  //!< open series as read-only, fails if series is not found
+        READ_WRITE, //!< open existing series as writable
+        CREATE      //!< create new series and truncate existing (files)
+    }; // AccessType
+} // namespace openPMD
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,6 +46,7 @@ Other
   - more info on MPI #449
   - new "first steps" section #473 #478
   - update invasive test info #474
+  - more info on ``AccessType`` #483
 
 
 0.7.1-alpha


anokfireball:
> If we open a new base `filePath` do we actively seek for existing files (for file-based encoding) with this pattern ...

No verification of exisitng files is done (if not implicitly done by backend implementation). 

> ... and remove them?

Accordingly, this does not happen either.
I don't  think deletion is the right way to go here. If the concern is interference with existing data on `CREATE`, writing should fail instead of silently deleting files.

Note that in plain UNIX, this is not the default behaviour and only happens when both [`O_CREAT`](http://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION) and [`O_EXCL`](http://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION) are specified. [HDF5](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5F.html#File-Create) reflects this UNIX behaviour.

> a mode that does create and truncates only iterations that are newly created

I am unsure about what you refer to as `truncation` here. Does truncation of newly created files mean the existing, unmodified files/`Iterarions` from a `fileBased` `Series` are not touched while existing, modified files/`Iterations` are overwritten at the file-level (i.e. all in-file data is deleted)?

_If so_, yes that is the current behaviour with `CREATE`. I can only vouch for HDF5 and ADIOS1, but would assume JSON does the same as that is the intended behaviour of [`O_CREAT`](http://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION).

> Alternatively and maybe better: allow/document to express the second mode in `READ_WRITE` (by just deleting an iteration explicitly).

I would very much prefer this to implicitly and silently deleting whole `Series`. If the user is aware that he is partially overwriting data, he should do it explicitly.
One option I am considering right now is to introduce a new mode `UPDATE`/`APPEND`/`OVERWRITE` and restrict `CREATE` to completely new Series without any colliding files. (Though I don't think `UPDATE` and `APPEND` satifactorily describe the behaviour. `OVERWRITE` might not, either.)

include/openPMD/IO/AccessType.hpp
```diff
@@ -23,12 +23,12 @@
...
-    READ_ONLY,
-    READ_WRITE,
-    CREATE
-};  //AccessType
-} // openPMD
+    /** File access mode to use during IO.
+     */
+    enum class AccessType
+    {
+        READ_ONLY,  //!< open series as read-only
```
anokfireball: One more piece of information that might be useful to the user:
`READ_ONLY` raises an Error when no matching files exist, while the other two don't.

anokfireball (approved):

ax3l:
@C0nsultant thank you for the detailed answer!

Regarding the behavior of create and read-write, we *should treat* access *as if* "Series" *were* a single file. We should then apply this not only to "groupBased" but also "fileBased". 

The though behind this is as follows: for a user, they only know the "Series" object and not how we store them. Therefore, just as with adios `.bp.dir/` files we should treat all files of a series as a single "database" that is "opened, created or extended". (We should avoid user-facing "file-speech", it is not consistent and might soon be streaming/[staging](https://github.com/ornladios/ADIOS/tree/master/tutorial/201402_OLCF/staging).)

You are right, for exact unix defaults we would:

- `CREATE`: [If pathname does not exist, create it as a regular file.](http://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION)

but I like the idea to make EXCL in there as well.

> I am unsure about what you refer to as `truncation` here.  [...] (i.e. all in-file data is deleted)?

Yes, I meant: `O_TRUNC` [If the file already exists and is a regular file and the access mode allows writing (i.e., is O_RDWR or O_WRONLY) it will be truncated to length 0.](http://man7.org/linux/man-pages/man2/open.2.html#DESCRIPTION)

As above, `groupBased` and `fileBased` should be made consistent regarding truncation of all or the newly created iterations (e.g. if we truncate in one of the modes, truncate all).

> I would very much prefer this [explicit mode] ...

Yes, makes sense to me.

I think what you propose and applied *to the whole Series* (and therefore all iterations, independent of how they are stored) makes sense, I try to summarize and correct me where I am off:
- `CREATE`: equal to UNIX `O_CREAT` + `O_EXCL`
- `CREATE_TRUNC`: equal to UNIX `O_CREAT` + `O_TRUNC`
- `READ_WRITE`: similar to UNIX `O_APPEND`, might overwrite iterations partially (or just records? or just record components? why not extend with more records to an existing iteration as well?) if written again and is *not* `O_EXCL`; create if missing
- `READ_ONLY`: well, read only :) fail if Series is missing

Issue #470, opened 2019-02-16
Parameter Read/Write: Data Policy


ax3l:
It would probably be useful if we added a trait / policy or runtime converter to the [`Parameter< Operation::READ_DATASET >`](https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/include/openPMD/IO/IOTask.hpp#L321-L340) (and write) data members of shared pointers in order to share more complex data objects.

For example returning to python a `py::array`, that the user might not handle, might result in it getting garbage collected before our `flush()` can be called, which in turn causes in invalid memory access in the backend. We could keep that `py::array` alive until `flush()` as in C++ by returning a `shared_ptr< py::array >`.

Overloads for [`RecordComponent::loadChunk<T>`](https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/include/openPMD/RecordComponent.hpp#L276) are already okay and well designed, but we cannot share yet an object that needs a more indirect data access pattern than a shared_ptr of raw memory, since we access it right away:
https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/src/IO/HDF5/HDF5IOHandler.cpp#L1017

ax3l:
cc @C0nsultant @franzpoeschel do you have thoughts on this? We could improve data safety for user-provided containers and python bindings a lot if we could store more complex `T` types in our `std::shared_ptr<T> data` members.

ax3l:
Some first support for 1D contiguous containers: #204
ADIOS2 has an interface providing C++20 `std::span` and we could also think about [mdspan](https://github.com/kokkos/mdspan).

Issue #361, opened 2018-09-19
High-Level Particle Patches


ax3l:
Particle Patches should get a high-level interface for convenient usage.

The data type is fixed for number of particles and offset in the particle array and can be as attributes taken for the position "offset" and "extent".

I am thinking of something a `Patch` object and an API like:
```C++
// for 123 particles, decomposed in x
// x and y range: [0.:123.)x[0.:123.), equally distributed
Patch patch0(
  {10, 0}, // uint64_t  numParticles & numParticlesOffset
  {{0., 0.}, {10., 123.}}  // T offset & T extent in position
);
Patch patch1(
  {113, 10}, // uint64_t numParticles & numParticlesOffset
  {{10., 0.}, {113., 123.}}  // T offset & T extent in position
);
// of course as well with default constructor and setter/getters

electrons.particlePatches.reset( 2 );
//  patch ID needs to be consistent if multiple MPI processes write those
// in MPI reality, two ways are possible
//   after an MPI_Scan, only the master process writes those
//   or each process writes at its MPI_Rank index
//   or every Nth process writes N patches ...
// so counting the add()s locally would not be sufficient

electrons.particlePatches.add(0, patch0); // or insert or ...
electrons.particlePatches.add(1, patch1);
```

Also, we need a `load()` method that returns a list/vector of all patches. Particle patches are small records, users will need them only *at once* during reads, so we can remove the `load( idx )` signature. -> done in #364

Note: Also in general, `idx` must not be shown to users during load. They shall handle patches as list of unsorted patches without any implicit order which would cause a lot of confusion. E.g. users must not assume that their MPI_Rank after a restart/between different jobs stays the same in the chosen domain decomposition, which is a big rabbit hole of errors one can make.

During refactoring, I think a `ParticleRecordComponent` should not inherit from a `BaseRecordComponent` but from a regular `RecordComponent` as well: we want for example support for constant records components in patches.

ax3l:
Alternatively, one could also try to provide helpers filling patch infos alongside `storeChunk` operations on `ParticleRecordComponents`.

PR #484, opened 2019-02-19
Series Flush Type

ax3l:
Add a new potential data flush mode to `Series`.

Although we implement a more general model of data passing and accumulating, allowing a "directly synchronous" mode will allow us to target the typical h5py Python user who just quickly wants to explore small data sets without frustration ("why is my data all zeros?").

We definitely have a continued need for our generalized, deferred, asynchronous chunk load/store mechanisms for our applications in parallel I/O and upcoming staging (think: streaming) modes. Just let us enable them explicitly as an "advanced" feature (therefore switch the default to the simple `DIRECT` mode). This is purely a user-interface change in order to target a wider audience.

The new `DIRECT` mode should also help somewhat in debugging.

Stylistic note: I think `AccessType` and `FlushType` might better be called `...Mode`?

@franzpoeschel @C0nsultant what do you think? I hope that description is not too brief, otherwise let's chat :-)
PRs to this branch that implement that mode, also in parts, very welcome.

## To Do

- [x] python bindings
- [ ] bugfix of #490
- Docs
  - [x] update first read / write
  - [x] update store/loadChunk doc strings
  - [x] add changelog
  - [x] breaking: add in upgrade guide
- [x] update examples
- [x] update and add more tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -244,6 +244,7 @@ Features
 
   - ``manylinux2010`` wheels for PyPI #523
   - add ``pyproject.toml`` for build dependencies (PEP-518) #527
+- Series: flush type "direct" (new default) and "defer" (previous default) #484
 
 Bug Fixes
 """""""""
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -28,6 +28,52 @@ The CMake option ``-DopenPMD_USE_JSON`` has been removed (as it is always ``ON``
 Previously, omitting a file ending in the ``Series`` constructor chose a "dummy" no-operation file backend.
 This was confusing and instead a runtime error is now thrown.
 
+Python
+^^^^^^
+
+``.flush()`` Method
+"""""""""""""""""""
+
+The ``Series.flush()`` method for ``load_chunk`` and ``store_chunk`` as well as particle patch ``load``/``store`` is now only needed in ``defer``-red flush mode.
+By default, a series will be opened with the new ``direct`` mode, which can be less performant for large-scale or many-var I/O but is more convenient.
+
+In order to restore the old default, set the new ``Series.set_flush`` method to ``Flush_Type.defer``.
+Both modes will be supported in the future and ``Flush_Type.defer`` is recommended for all high-throughput workloads.
+
+.. code-block:: python3
+
+   import openPMD_api as api
+
+   s = api.Series(
+       "data_%T.h5",
+       api.Access_Type.create)
+
+   # new default: api.Flush_Type.direct
+   s.set_flush(api.Flush_Type.defer)
+
+C++
+^^^
+
+``.flush()`` Method
+"""""""""""""""""""
+
+The ``Series::flush()`` method for ``loadChunk`` and ``storeChunk`` as well as particle patch ``load``/``store`` is now only needed in ``DEFER``-red flush mode.
+By default, a series will be opened with the new ``DIRECT`` mode, which can be less performant for large-scale or many-var I/O but is more convenient.
+
+In order to restore the old default, set the new ``Series::setFlush`` method to ``FlushType::DEFER``.
+Both modes will be supported in the future and ``FlushType::DEFER`` is recommended for all high-throughput workloads.
+
+.. code-block:: python3
+
+   #include <openPMD/openPMD.hpp>
+
+   s = openPMD::Series(
+       "data_%T.h5",
+       openPMD::AccessType::CREATE);
+
+   // new default: openPMD::FlushType::DIRECT
+   s.setFlush(openPMD::FlushType::DEFER);
+
 
 0.9.0-alpha
 -----------
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -48,6 +48,9 @@ int main(int argc, char *argv[])
     );
     cout << "Created an empty " << series.iterationEncoding() << " Series\n";
 
+    // accumulate load/store I/O operations until .flush()
+    // series.setFlush(FlushType::DEFER);
+
     MeshRecordComponent rho =
       series
           .iterations[1]
@@ -63,7 +66,8 @@ int main(int argc, char *argv[])
     rho.resetDataset(dataset);
     cout << "Set the dataset properties for the scalar field rho in iteration 1\n";
 
-    series.flush();
+    // in FlushType::DEFER, call now:
+    // series.flush();
     cout << "File structure and required attributes have been written\n";
 
     Offset offset = {0, 0};
@@ -71,7 +75,8 @@ int main(int argc, char *argv[])
     cout << "Stored the whole Dataset contents as a single chunk, "
             "ready to write content\n";
 
-    series.flush();
+    // in FlushType::DEFER, call now:
+    // series.flush();
     cout << "Dataset content has been fully written\n";
 
     /* The files in 'series' are still open until the object is destroyed, on
--- a/examples/3_write_serial.py
+++ b/examples/3_write_serial.py
@@ -26,6 +26,9 @@ if __name__ == "__main__":
         openpmd_api.Access_Type.create
     )
 
+    # accumulate load/store I/O operations until .flush()
+    # series.set_flush(openpmd_api.Flush_Type.defer)
+
     print("Created an empty {0} Series".format(series.iteration_encoding))
 
     print(len(series.iterations))
@@ -40,16 +43,12 @@ if __name__ == "__main__":
     rho.reset_dataset(dataset)
     print("Set the dataset properties for the scalar field rho in iteration 1")
 
-    series.flush()
-    print("File structure has been written")
-
     rho[()] = data
 
-    print("Stored the whole Dataset contents as a single chunk, " +
-          "ready to write content")
+    # in Flush_Type.defer, call now:
+    # series.flush()
 
-    series.flush()
-    print("Dataset content has been fully written")
+    print("Stored the whole Dataset contents as a single chunk")
 
     # The files in 'series' are still open until the object is destroyed, on
     # which it cleanly flushes and closes all open file handles.
--- a/include/openPMD/IO/FlushType.hpp
+++ b/include/openPMD/IO/FlushType.hpp
@@ -1,0 +1,42 @@
+/* Copyright 2019 Axel Huebl
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * openPMD-api is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+
+namespace openPMD
+{
+    /** Data flush mode.
+     *
+     * Control if RecordComponent::load_chunk and RecordComponent::store_chunk
+     * operations are flushed immediately or delayed until Series::flush
+     *
+     * DIRECT-ly populating passed data buffers is considered a safe and
+     * convienient default for all small-scale I/O operations. For large-scale
+     * IO or load/store of many small chunks, consider switching to DEFER
+     * mode and ensure passed data buffers are valid and unmodified until
+     * Series::flush is called.
+     */
+    enum class FlushType //! @todo we should call this AccessMode and FlushMode ...
+    {
+        DIRECT,  //!< immediately load/store data to passed data buffers (default)
+        DEFER,   //!< load/store data to passed data buffers on Series::flush
+    }; // FlushType
+} // namespace openPMD
--- a/include/openPMD/IO/HDF5/HDF5Auxiliary.hpp
+++ b/include/openPMD/IO/HDF5/HDF5Auxiliary.hpp
@@ -28,6 +28,7 @@
 #include <hdf5.h>
 
 #include <stack>
+#include <iostream>
 
 
 namespace openPMD
@@ -253,7 +254,13 @@ concrete_h5_file_position(Writable* w)
     std::string pos;
     while( !hierarchy.empty() )
     {
-        pos += std::dynamic_pointer_cast< HDF5FilePosition >(hierarchy.top()->abstractFilePosition)->location;
+        auto afp = hierarchy.top()->abstractFilePosition;
+
+        auto afp_h5 = std::dynamic_pointer_cast< HDF5FilePosition >(afp);
+        if( afp_h5 == nullptr )
+            std::cerr << "WARNING: invalid file position in writable stack!" << std::endl;
+        else
+            pos.append(afp_h5->location);
         hierarchy.pop();
     }
 
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -21,6 +21,8 @@
 #pragma once
 
 #include "openPMD/backend/BaseRecordComponent.hpp"
+#include "openPMD/IO/AbstractIOHandler.hpp"
+#include "openPMD/IO/FlushType.hpp"
 #include "openPMD/auxiliary/ShareRaw.hpp"
 #include "openPMD/Dataset.hpp"
 
@@ -125,6 +127,14 @@ public:
      *
      * If offset is non-zero and extent is {-1u} the leftover extent in the
      * record component will be selected.
+     *
+     * In FlushType::DEFER, the data managed by the returned
+     * std::shared_ptr< T > MUST NOT be read, modified or deleted until
+     * Series::flush is called.
+     *
+     * @todo implement: if the shared pointers' `use_count` drops to one
+     *       before Series::flush is called, the load will be skipped during
+     *       Series::flush.
      */
     template< typename T >
     std::shared_ptr< T > loadChunk(
@@ -140,6 +150,15 @@ public:
      *
      * If offset is non-zero and extent is {-1u} the leftover extent in the
      * record component will be selected.
+     *
+     * In FlushType::DEFER, the data managed by the passed
+     * std::shared_ptr< T > MUST NOT be read, modified or deleted until
+     * Series::flush is called.
+     *
+     * @todo implement: if the shared pointers' `use_count` drops to one
+     *       before Series::flush is called, the load will be skipped during
+     *       Series::flush.
+     *       Depends on https://github.com/openPMD/openPMD-api/issues/470
      */
     template< typename T >
     void loadChunk(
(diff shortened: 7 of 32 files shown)


include/openPMD/RecordComponent.hpp
```diff
@@ -129,6 +136,14 @@ class RecordComponent : public BaseRecordComponent
      *
      * If offset is non-zero and extent is {-1u} the leftover extent in the
      * record component will be selected.
+     *
+     * In FlushType::DEFER, the data managed by the passed
+     * std::shared_ptr< T > MUST NOT be read, modified or deleted until
+     * Series::flush is called.
+     * @todo implement: if the shared pointers' `use_count` drops to one
+     *       before Series::flush is called, the load will be skipped during
+     *       Series::flush. (check if this works with our `shareRaw()`
```
ax3l: this will become possible when #470 is implemented

include/openPMD/RecordComponent.hpp
```diff
@@ -137,9 +152,37 @@ class RecordComponent : public BaseRecordComponent
...
+    /** Store a chunk of data
+     *
+     * shared_ptr< T > for data must be contiguous and large enough for extent
+     *
+     * In FlushType::DEFER, the data managed by the passed
+     * std::shared_ptr< T > MUST NOT be modified or deleted until
+     * Series::flush is called.
+     * @todo implement: if the shared pointers' `use_count` drops to one
+     *       before Series::flush is called, the store will be skipped during
+     *       Series::flush. (check if this works with our `shareRaw()`
```
ax3l: this will become possible when #470 is implemented

ax3l:
@C0nsultant do you think the right location to store the `flush mode` would be with `AbstractIOHandler`?

test/ParallelIOTest.cpp
```diff
@@ -86,6 +86,7 @@ TEST_CASE( "hdf5_write_test", "[parallel][hdf5]" )
     uint64_t mpi_size = static_cast<uint64_t>(mpi_s);
     uint64_t mpi_rank = static_cast<uint64_t>(mpi_r);
     Series o = Series("../samples/parallel_write.h5", AccessType::CREATE, MPI_COMM_WORLD);
+    o.setFlush(FlushType::DEFER);  // @todo crashes in DIRECT flush mode
```
ax3l: see bug #490

test/ParallelIOTest.cpp
```diff
@@ -120,6 +123,7 @@ TEST_CASE( "hdf5_write_test_zero_extent", "[parallel][hdf5]" )
     uint64_t size = static_cast<uint64_t>(mpi_s);
     uint64_t rank = static_cast<uint64_t>(mpi_r);
     Series o = Series("../samples/parallel_write_zero_extent.h5", AccessType::CREATE, MPI_COMM_WORLD);
+    o.setFlush(FlushType::DEFER);  // @todo crashes in DIRECT flush mode
```
ax3l: see bug #490

anokfireball:
> @C0nsultant do you think the right location to store the `flush mode` would be with `AbstractIOHandler`?

There (currently) is a 1:1 mapping of `Series` to `IOHandlers`, but in theory this could become n:1 in the future (i.e. multiple versions of a `Series` in the same directory with the same access properties). One could think of scenarios where the different versions have different `flush` strategies, but I would not worry much about that case as there's one too many _if_ s involved for serious dedication.

Strictly speaking separation of concern, the IO backend (and in this case, the handler) is a perfectly suitable place for it. The `flush()` calls in the front- and backend have slightly different intentions. The one you care about (actually swapping bytes between RAM and NVstorage, without traversing and syncing the whole frontend tree) is the one the backend is concerned with.

ax3l:
Thanks, sounds good.

I actually pushed a first working draft in this PR, for in-code comments/review :)

ax3l:
@C0nsultant ready for review :)
Any idea why `direct` writing in `3_write_serial` might be segfaulting? Do we need to flush some path creations? Seems the `Record::flush_impl` path creation is a bit late as well as the parent/writable registration for new record components if they are not scalar.

ax3l:
@C0nsultant if you have the time, please feel free to add additional commits

NEWS.rst
```diff
@@ -3,6 +3,56 @@
...
+0.8.0-beta
+----------
+
+Python
+^^^^^^
+
+``.flush()`` Method
+"""""""""""""""""""
+
+The ``Series.flush()`` method for ``load_chunk`` and ``store_chunk`` as well as particle patch ``load``/``store`` is now only needed in ``defer``-red flush mode.
```
anokfireball, line 37: MINOR
I find it a bit misleading to limit the use of `Series.flush()` to dataset IO here.
This sentence is undoubtedly geared toward the user (rather than the dev), so what do you think about 

> The call to ``Series.flush()`` required for [...]

?

NEWS.rst
```diff
@@ -3,6 +3,56 @@
...
+   # new default: api.Flush_Type.direct
+   s.set_flush(api.Flush_Type.defer)
+
+C++
+^^^
+
+``.flush()`` Method
+"""""""""""""""""""
+
+The ``Series::flush()`` method for ``loadChunk`` and ``storeChunk`` as well as particle patch ``load``/``store`` is now only needed in ``DEFER``-red flush mode.
```
anokfireball, line 60: Same here.

examples/3_write_serial.py
```diff
@@ -43,16 +46,12 @@
...
     print("Set the dataset properties for the scalar field rho in iteration 1")
 
-    series.flush()
-    print("File structure has been written")
-
     rho[()] = global_data
 
-    print("Stored the whole Dataset contents as a single chunk, " +
-          "ready to write content")
+    # in Flush_Type.defer, call now:
```
anokfireball, line 48: Maybe keep the old code and branch it depending on `FlushType`

```Python
if series.flush == openpmd_api.Flush_Type.defer:
    print("Stored the whole Dataset contents as a single chunk, ready to write content")
    # need to manually initiate transfer of accumulated operations
    series.flush()
    print("Dataset content has been fully written")
else:
    # transfer was implicitly done
    print("Stored the whole Dataset contents as a single chunk")
```

(similar branching code could be included in all comments concerning `.flush()` for `DEFER`)

include/openPMD/IO/FlushType.hpp
```diff
@@ -0,0 +1,42 @@
...
+ */
+#pragma once
+
+
+namespace openPMD
+{
+    /** Data flush mode.
+     *
+     * Control if RecordComponent::load_chunk and RecordComponent::store_chunk
+     * operations are flushed immediately or delayed until Series::flush
```
anokfireball, line 29: Some of the `.flush()`es you removed in the documentation/examples do IO that is not concerned with chunk transfers:

```cpp
/* 3_write_serial.cpp:69 */

// in FlushType::DEFER, call now:
// series.flush();
cout << "File structure and required attributes have been written\n";
```

As I think this docstring is correct, some of the `print()`s after the commented-out code are now invalid.

include/openPMD/IO/FlushType.hpp
```diff
@@ -0,0 +1,42 @@
...
+     *
+     * DIRECT-ly populating passed data buffers is considered a safe and
+     * convienient default for all small-scale I/O operations. For large-scale
+     * IO or load/store of many small chunks, consider switching to DEFER
+     * mode and ensure passed data buffers are valid and unmodified until
+     * Series::flush is called.
+     */
+    enum class FlushType //! @todo we should call this AccessMode and FlushMode ...
```
anokfireball, line 37: Another suggestion w.r.t. naming: `FlushStrategy`.

Generally, I support dropping `-Type`.

include/openPMD/IO/HDF5/HDF5Auxiliary.hpp
```diff
@@ -253,7 +254,13 @@ concrete_h5_file_position(Writable* w)
     std::string pos;
     while( !hierarchy.empty() )
     {
-        pos += std::dynamic_pointer_cast< HDF5FilePosition >(hierarchy.top()->abstractFilePosition)->location;
+        auto afp = hierarchy.top()->abstractFilePosition;
+
+        auto afp_h5 = std::dynamic_pointer_cast< HDF5FilePosition >(afp);
+        if( afp_h5 == nullptr )
+            std::cerr << "WARNING: invalid file position in writable stack!" << std::endl;
```
anokfireball, line 261: Once we hit the `nullptr`, we're in an undetermined and likely unrecoverable state.
Rather than merely `cerr`ing and failing later down the line (i.e. when HDF5 fails to open a path), this makes a great candidate for `VERIFY`.

include/openPMD/RecordComponent.hpp
```diff
@@ -114,6 +116,14 @@ class RecordComponent : public BaseRecordComponent
      *
      * If offset is non-zero and extent is {-1u} the leftover extent in the
      * record component will be selected.
+     *
+     * In FlushType::DEFER, the data managed by the returned
+     * std::shared_ptr< T > MUST NOT be read, modified or deleted until
```
anokfireball, line 132: Modification would probably not even be detrimental, but rather useless.

include/openPMD/RecordComponent.hpp
```diff
@@ -114,6 +116,14 @@ class RecordComponent : public BaseRecordComponent
...
+     * Series::flush is called.
+     *
+     * @todo implement: if the shared pointers' `use_count` drops to one
+     *       before Series::flush is called, the load will be skipped during
+     *       Series::flush.
```
anokfireball, line 137: Whoop! :raised_hands:  This is a great optimization.

include/openPMD/RecordComponent.hpp
```diff
@@ -137,9 +156,38 @@ class RecordComponent : public BaseRecordComponent
...
+     *
+     * shared_ptr< T > for data must be contiguous and large enough for extent
+     *
+     * In FlushType::DEFER, the data managed by the passed
+     * std::shared_ptr< T > MUST NOT be modified or deleted until
+     * Series::flush is called.
+     *
+     * @todo implement: if the shared pointers' `use_count` drops to one
+     *       before Series::flush is called, the store will be skipped during
+     *       Series::flush.
```
anokfireball, line 180: I don't think this works, though.
The user schedules a chunk store and then drops *his* ref to the data. So the write would still be valid.

include/openPMD/RecordComponent.hpp
```diff
@@ -137,9 +156,38 @@ class RecordComponent : public BaseRecordComponent
         Extent,
         double targetUnitSI = std::numeric_limits< double >::quiet_NaN() );
 
+    /** Store a chunk of data
+     *
+     * shared_ptr< T > for data must be contiguous and large enough for extent
```
anokfireball, line 172: > and large enough for extent

This does not apply here or is misworded. It must have the same number of entries as the cumulative product of extent. *large enough* does not portray this fact.

**EDIT** Thinking more about it, you could only schedule part of a chunk in RAM. Still, I think *large enough* is not sufficient here.

include/openPMD/RecordComponent.hpp
```diff
@@ -137,9 +156,38 @@ class RecordComponent : public BaseRecordComponent
...
+     *       before Series::flush is called, the store will be skipped during
+     *       Series::flush.
+     *       Depends on https://github.com/openPMD/openPMD-api/issues/470
+     */
     template< typename T >
     void storeChunk(std::shared_ptr< T >, Offset, Extent);
 
+    /** Store a chunk of data
+     *
+     * Passed contiguous container must be large enough for extent.
```
anokfireball, line 188: > and large enough for extent

See above.

include/openPMD/IO/FlushType.hpp
```diff
@@ -0,0 +1,42 @@
...
+    {
+        DIRECT,  //!< immediately load/store data to passed data buffers (default)
```
anokfireball, line 39: Another suggestion w.r.t. naming: `IMMEDIATE`.

include/openPMD/RecordComponent.hpp
```diff
@@ -275,6 +324,16 @@ RecordComponent::loadChunk(std::shared_ptr< T > data, Offset o, Extent e, double
         dRead.dtype = getDatatype();
         dRead.data = std::static_pointer_cast< void >(data);
         m_chunks->push(IOTask(this, dRead));
+
+        if( this->IOHandler->flushType == FlushType::DIRECT )
+        {
+            while( !m_chunks->empty() )
```
anokfireball, line 361: The loop is probably redundant. The only scenario in which there would be more than one chunks in the `m_chunks` queue is when one chunk (or more) is scheduled in `DEFER`, the Series is switched to `DIRECT` and another chunk is scheduled.

include/openPMD/Series.hpp
```diff
@@ -230,6 +231,25 @@ class Series : public Attributable
...
+     *
+     * DIRECT-ly populating passed data buffers is considered a safe and
+     * convienient default for all small-scale I/O operations. For large-scale
+     * IO or load/store of many small chunks, consider switching to DEFER
+     * mode and ensure passed data buffers are valid and unmodified until
+     * Series::flush is called.
+     *
+     * Default: FlushType::DIRECT
+     */
+    void setFlush(FlushType flushType);
```
anokfireball, line 257: Nomenclature:
```cpp
void setFlushType(FlushType flushType);
```
?

include/openPMD/backend/PatchRecordComponent.hpp
```diff
@@ -56,14 +57,37 @@ class PatchRecordComponent : public BaseRecordComponent
...
+     * In FlushType::DEFER, the passed data MUST NOT be modified or deleted
+     * until Series::flush is called.
+     */
     template< typename T >
     void store(uint64_t idx, T);
+
     //! @todo add support for constant patch record components
 
+    //! @todo allow to read all patch records at once in a convienient object
+    //!       to iterate over
```
anokfireball, line 89: There's an issue for this already (https://github.com/openPMD/openPMD-api/issues/361), maybe mention it here.

src/Series.cpp
```diff
@@ -350,6 +350,16 @@ Series::setName(std::string const& n)
     return *this;
 }
 
+void Series::setFlush(FlushType flushType)
+{
+    IOHandler->flushType = flushType;
```
anokfireball, line 364: To solve the redundant loop, we could `Series::flush()` when switching from `DEFER` TO `DIRECT`. This will schedule and perform all chunk transfers that would be done in the loops.

src/Series.cpp
```diff
@@ -350,6 +350,16 @@ Series::setName(std::string const& n)
     return *this;
 }
 
+void Series::setFlush(FlushType flushType)
```
anokfireball, line 362: Maybe we want to include `FlushType` as an argument in the `Series::Series()` constructor with `DIRECT` as default to make the behaviour a little more obvious?

anokfireball (approved):
Thank you for improving user-approachability!

examples/3_write_serial.cpp
```diff
@@ -63,15 +66,17 @@ int main(int argc, char *argv[])
...
     cout << "File structure and required attributes have been written\n";
 
     Offset offset = {0, 0};
     rho.storeChunk(shareRaw(global_data), offset, extent);
     cout << "Stored the whole Dataset contents as a single chunk, "
             "ready to write content\n";
 
-    series.flush();
+    // in FlushType::DEFER, call now:
+    // series.flush();
```
anokfireball, line 79: This commented out `flush()` is very likely what's causing the SEGFAULT.
In `RecordComponent::store_chunk()`, you only schedule a single `DATASET_WRITE` (well, multiple ones potentially).
Literally no other IO operations have been scheduled or completed at that point. No HDF5 handle exists, no structure inside the HDF5 file. So naturally the call in the backend fails ([here](https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/src/IO/HDF5/HDF5IOHandler.cpp#L670-L673)).

anokfireball (changes requested):
Your current changes do not entirely spare the user of calling `.flush()`.
As a very crude solution, you would call `Series.flush()` inside `store_chunk()` and `load_chunk()` (avoiding infinite recursion).

examples/3_write_serial.py
ax3l: The idea is to keep the code valid and compact. Branching here is not be best impression for copy-&-paste examples for my taste, since 100+ lines of code examples are hard to read. Just subjective feeling.
But one can do that, yep. Will think about it.

include/openPMD/IO/FlushType.hpp
ax3l: Uh, strategy is a looong word :D
If we rename, I would do so for both in a follow-up PR. Just dropping the suffix is also a great idea!

include/openPMD/RecordComponent.hpp
ax3l: Yes, I basically just describe the contract here. Both is not allowed from a user-side for various reasons (either ignored or even crashes).

include/openPMD/RecordComponent.hpp
ax3l: Yes, but really needs #470 to work :)

include/openPMD/RecordComponent.hpp
ax3l: Ah true for writes! So we can leave it here as is.

include/openPMD/RecordComponent.hpp
ax3l: ok, makes sense to be a more explicit.

include/openPMD/IO/FlushType.hpp
ax3l: yes, I was looking for short, concise words so users have fun typing them.

include/openPMD/RecordComponent.hpp
ax3l: Yes, that could happen imho and is what I had in mind.

include/openPMD/Series.hpp
ax3l: and in the future
```C++
constructor(..., Access access, ...);
void setFlush(Flush flush);
Flush getFlush();
Access getAccess();
void flush();
```
?

src/Series.cpp
ax3l: I think of the loop as a safety net at negligible costs.
Nevertheless, we can start flushing around here as well, that makes sense.

src/Series.cpp
ax3l: I thought about this and felt it has enough arguments.
If one wants to activate the "advanced" usage, I kinda like that it is an explicit step.

examples/3_write_serial.cpp
ax3l: > This commented out `flush()` is very likely what's causing the SEGFAULT.

Yes, I know, that's why I was wondering if we want to [re-structure some of the logic](https://github.com/openPMD/openPMD-api/pull/484#issuecomment-466646641) in side `Record::flush_impl` so `RecordComponent` knows about it?

NEWS.rst
```diff
@@ -3,6 +3,56 @@
 Upgrade Guide
 =============
 
+0.8.0-beta
```
ax3l: memo to myself: update me

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ Changes to "0.7.1-alpha"
 Features
 """"""""
 
+- Series: flush type "direct" (new default) and "defer" (previous default) #484
```
ax3l: move to newer release

ax3l:
@C0nsultant if you have the time, the logic in flushing here and in #490 needs to be refactored by us to be more flexible with flushes triggered from a record component (instead of top-down from a series-iteration-record-...) :-)

Issue #452, opened 2019-01-27
CI: UndefinedBehaviorSanitizer


ax3l:
We already run the CI with AddressSanitizers but should also run an entry with [UndefinedBehaviorSanitizer](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html).

Also, if we [suppress]() some of the memory leaks in dependencies we might be able to check more backends in AddressSanitizer.

Check all but memory leaks: `-fsantize=address` with export:
```
export ASAN_OPTIONS=detect_stack_use_after_return=1:detect_leaks=0:check_initialization_order=true:strict_init_order=true:detect_stack_use_after_scope=1
```

We can also activate the python backend for address sanitizer checks, as long as we export the `LD_PRELOAD` path to the falsely hidden `asan` symbols: #479

For MPI-parallel checks, we might be able to write a [suppression file](https://releases.llvm.org/7.0.0/tools/clang/docs/AddressSanitizer.html#suppressing-reports-in-external-libraries), excluding the whole external lib.

Refs:
(clang ubsan) https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
(gcc ubsan) https://developers.redhat.com/blog/2014/10/16/gcc-undefined-behavior-sanitizer-ubsan/
https://github.com/google/sanitizers/wiki/AddressSanitizer
https://github.com/google/sanitizers/wiki/AddressSanitizerExampleUseAfterReturn

PR #488, opened 2019-02-21
CI: +UBSan +Python +MPI +HF5

ax3l:
Perform als UndefinedBehavior Sanitizer tests and run for Python bindings, MPI and HDF5 as well.
Related to #452 

- Python: due to shared nature of modules, `-shared-libsan` and `LD_PRELOAD` is needed
- ADIOS1: [our shared vptr setup](https://travis-ci.org/openPMD/openPMD-api/builds/496789844) confuses UBsan (maybe rightfully... :-S)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
     - CXXFLAGS: "-Werror"
     - SPACK_ROOT: $HOME/.cache/spack
     - PATH: $PATH:$HOME/.cache/spack/bin
+    - BUILD_TYPE: "Debug"
     - PYBIND11_VERSION=@2.2.4
     - USE_JSON: ON
     - USE_SHARED: OFF
@@ -112,7 +113,7 @@ jobs:
     #     - cd $HOME/build
     #     - cmake
     #         -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
-    #         -DCMAKE_BUILD_TYPE=Debug
+    #         -DCMAKE_BUILD_TYPE=${BUILD_TYPE}
     #         -DopenPMD_USE_MPI=$USE_MPI
     #         -DopenPMD_USE_JSON=$USE_JSON
     #         -DopenPMD_USE_HDF5=$USE_HDF5
@@ -201,7 +202,7 @@ jobs:
       addons:
         apt:
           <<: *apt_common_sources
-          packages: &clang50_deps
+          packages: &clang_deps
             - environment-modules
             - gfortran-4.9
             - g++-4.9  # CMake build
@@ -209,7 +210,7 @@ jobs:
         - mkdir -p $HOME/build
         - cd $HOME/build
         - if [ ! -d samples/git-sample/ ]; then
-            if [ "$USE_SAMPLES" = "ON" ]; then
+            if [ "$USE_SAMPLES" == "ON" ]; then
               $TRAVIS_BUILD_DIR/.travis/download_samples.sh;
               chmod u-w samples/git-sample/*.h5;
             fi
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -289,18 +289,20 @@ jobs:
...
       env:
-        - CXXSPEC="%clang@7.0.0" USE_MPI=OFF USE_PYTHON=OFF USE_HDF5=OFF USE_ADIOS1=OFF USE_ADIOS2=OFF USE_JSON=ON USE_SAMPLES=ON
+        - CXXSPEC="%clang@7.0.0" USE_MPI=OFF USE_PYTHON=ON USE_HDF5=OFF USE_ADIOS1=OFF USE_ADIOS2=OFF USE_JSON=ON USE_SAMPLES=ON ASAN_OPTIONS=detect_stack_use_after_return=1:detect_leaks=1:check_initialization_order=true:strict_init_order=true:detect_stack_use_after_scope=1
       compiler: clang
       script: *script-cpp-unit
       before_install:
-        - CXX=clang++ && CC=clang && CXXFLAGS="-Werror -fsanitize=address"
+        - CXX=clang++ && CC=clang && CXXFLAGS="-Werror -fsanitize=address,undefined" && LDFLAGS="-fsanitize=address,undefined"
+        # pybind11: work-around for missing symbols with sanitizers
+        - export LD_PRELOAD=/usr/local/clang-7.0.0/lib/clang/7.0.0/lib/linux/libclang_rt.asan-x86_64.so:/usr/local/clang-7.0.0/lib/clang/7.0.0/lib/linux/libclang_rt.ubsan_minimal-x86_64.so
```
ax3l: that's a bit early... should be directly before tests run

- maybe set rpath/runpath instead, in case shared libs on linux can load deps like this as well (not on only executables)
- debug with `export LD_DEBUG=libs`
- GCC supports `-static-libasan` (clang doesn't? no [wait](https://stackoverflow.com/a/47022141/2719194))
- maybe `chrpath` and `patchelf` on created cpython module

```
patchelf --add-needed /usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.asan-x86_64.so lib/python3.6/site-packages/openpmd_api.cpython-36m-x86_64-linux-gnu.so
patchelf --add-needed /usr/lib/llvm-6.0/lib/clang/6.0.0/lib/linux/libclang_rt.ubsan_minimal-x86_64.so lib/python3.6/site-packages/openpmd_api.cpython-36m-x86_64-linux-gnu.so
```

> ==8216==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.

Following [FAQs](https://github.com/google/sanitizers/wiki/AddressSanitizer#faq), `LD_PRELOAD` seems to be the recommended way...

> Q: I've built my shared library with ASan. Can I run it with unsanitized executable?
>
> A: Yes! You'll need to build your library with dynamic version of ASan and then run executable with LD_PRELOAD=path/to/asan/runtime/lib.

with [experimental shared ASAN](https://github.com/google/sanitizers/wiki/AddressSanitizerAsDso) (experimental in clang, default in gcc)

.travis.yml
```diff
@@ -229,7 +230,17 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=$HOME/openPMD-test-install
             $TRAVIS_BUILD_DIR
         - make -j 2
+        # run tests
+        #   pybind11: work-around for missing symbols with sanitizers
+        #   /usr/local/clang-7.0.0/lib/clang/7.0.0/lib/linux/libclang_rt.asan-x86_64.so
+        - if [ "$USE_PYTHON" == "ON" ] && [ ! -z ${ASAN_OPTIONS+x} ]; then
+            export LD_PRELOAD=/usr/local/clang-7.0.0/lib/clang/7.0.0/lib/linux/libclang_rt.ubsan_standalone-x86_64.so;
```
ax3l: hm, actually `minimal`, but was still lacking `vptr` symbols...

PR #496, opened 2019-03-09
Examples: 2D MPI Write

### LINKED ISSUE #495: C++ example with parallel write of 2D/3D mesh
Right now, the documentation contains an example for parallel write in C++, but only with a 1d array and one element per MPI rank.

Would it be possible to add an example (or transform the current example) to include writing of 2D/3D mesh in parallel?

Thanks!

ax3l:
Enhance the parallel write example to demonstrate a 1D-decomposed, MPI-parallel write of a 2D mesh.

Fix #495

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -47,6 +47,7 @@ Other
   - new "first steps" section #473 #478
   - update invasive test info #474
   - more info on ``AccessType`` #483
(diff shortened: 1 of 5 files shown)


RemiLehe (approved):
Looks great! Thanks!

ax3l:
Cool, glad I could help!

PR #500, opened 2019-03-09
Doxyfile: add Feature Macros

ax3l:
Enable all frontends and backends for Doxygen parsing.

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -13,6 +13,16 @@ RECURSIVE         = YES
 JAVADOC_AUTOBRIEF = YES
 USE_MDFILE_AS_MAINPAGE = README.md
 
+# "enable" all frontends and backends
+PREDEFINED        = openPMD_HAVE_MPI=1 \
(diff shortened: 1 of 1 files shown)


docs/Doxyfile
```diff
@@ -13,6 +13,16 @@ RECURSIVE         = YES
...
 USE_MDFILE_AS_MAINPAGE = README.md
 
+# "enable" all frontends and backends
+PREDEFINED        = openPMD_HAVE_MPI=1 \
+                    openPMD_HAVE_JSON=1 \
+                    openPMD_HAVE_HDF5=1 \
+                    openPMD_HAVE_ADIOS1=1 \
+                    openPMD_HAVE_ADIOS2=1 \
+                    openPMD_HAVE_PYTHON=1 \
+                    OPENPMD_private:=private \
```
ax3l, line 23: last hack on visibility for invasive test doesn't seem to work with doxygen :(
https://www.star.bnl.gov/public/comp/sofi/doxygen/preprocessing.html

PR #511, opened 2019-05-13
Fix linker error concerning Mesh::setTimeOffset method template

franzpoeschel:
`Mesh::setTimeOffset< T >( )`, declared in [Mesh.hpp](https://github.com/openPMD/openPMD-api/blob/dev/include/openPMD/Mesh.hpp) is defined in [Mesh.cpp](https://github.com/openPMD/openPMD-api/blob/dev/src/Mesh.cpp) without explicit instantiations. This triggers a linker error when using this method.
First commit fails the CI, to highlight the issue.

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -181,6 +181,18 @@ Mesh::setTimeOffset(T to)
     return *this;
 }
 
+template
+Mesh&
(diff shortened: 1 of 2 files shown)


ax3l (approved):
Good catch! Thanks!

PR #518, opened 2019-06-05
Allow flushing before defining position and positionOffset components of particle species

franzpoeschel:
**Issue** The fact that the class [`ParticleSpecies`](https://github.com/openPMD/openPMD-api/blob/dev/include/openPMD/ParticleSpecies.hpp) will enforce openPMD requirements currently leads to the awkward situation that those records have to be defined using [`RecordComponent::resetDataset( … )`](https://github.com/openPMD/openPMD-api/blob/dev/include/openPMD/RecordComponent.hpp#L103) before issuing the first flush on the `Series` object. A later `RecordComponent::storeChunk( … )` is fine, but the datasets need to be defined up front.
This is (1) an awkward pitfall, (2) leads to rather confusing errors if done wrongly (`what():  A Record can not be written without any contained RecordComponents: position`) and (3) leads to ugly workarounds in workflows where the single records of a particle species are defined and written one after another (e.g. the [openPMD plugin for picongpu](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2966) that I am currently working on).

**Suggested Change** Move [these two lines](https://github.com/openPMD/openPMD-api/blob/37b525dff1fb1bdd50274d229353d8226f973065/include/openPMD/ParticleSpecies.hpp#L59) into the `flush` method override of `ParticleSpecies`, but do not create the records if not present.

**Open questions** The above solution will ensure openPMD requirements for position[Offset] records if the user remembers to create them. I have not (yet) implemented an alternative check if the user forgets to create those records. The logical location for such a check would be the destructor, but throwing exceptions in destructors is discouraged.

First commit adds new tests (and removes some existing ones), second commit will implement the suggested change.

### DIFF
--- a/include/openPMD/ParticleSpecies.hpp
+++ b/include/openPMD/ParticleSpecies.hpp
@@ -55,9 +55,6 @@ namespace traits
         template< typename T >
         void operator()(T & ret)
         {
-            /* enforce these two RecordComponents as required by the standard */
-            ret["position"].setUnitDimension({{UnitDimension::L, 1}});
-            ret["positionOffset"].setUnitDimension({{UnitDimension::L, 1}});
             ret.particlePatches.linkHierarchy(ret.m_writable);
 
             auto& np = ret.particlePatches["numParticles"];
--- a/src/ParticleSpecies.cpp
+++ b/src/ParticleSpecies.cpp
@@ -122,6 +122,17 @@ ParticleSpecies::flush(std::string const& path)
             for( auto& patch : particlePatches )
                 patch.second.flush(patch.first);
         }
+
+        iterator it = find("position");
+        if ( it != end() )
+        {
+            it->second.setUnitDimension({{UnitDimension::L, 1}});
+        }
+        it = find("positionOffset");
+        if ( it != end() )
+        {
+            it->second.setUnitDimension({{UnitDimension::L, 1}});
+        }
     }
 }
 } // openPMD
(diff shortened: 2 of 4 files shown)


ax3l:
Thank you, great this gets addressed.

I think in general that we should remove the creation logic from the objects and divide "openPMD objects" from "required and default attribute creation" in our implementation. This is too interwinded still and causes issues.

Furthermore, I think we should relax what happens on `.flush()` to purely flush buffers and check and create default structures only on `Series` close (destruction) or any additional member function that we introduce for that, e.g. `.complete()`/`.close()` (e.g. also on iterations!) or so. cc @C0nsultant 

> The logical location for such a check would be the destructor, but throwing exceptions in destructors is discouraged.

Good point. Maybe we just need an explicit `.close()` on `Series` and even `Iteration` objects. #149

ax3l (approved):

ax3l:
Before we merge this, did you verify the created files still pass the validator? https://github.com/openPMD/openPMD-validator

franzpoeschel:
Did that just now, the files yield exactly the same errors as those on the dev branch do, so I guess there's further things to investigate. I checked the JSON files with diff now and there are actually differences for the `serial_patch` and `constant_scalar` tests where the unit dimensions are erroneously all zero, so I'll need to check that.

ax3l:
Ok, let's do that in a follow-up PR.

note: `unitDimensions` can be 7x zero, e.g. for unitless counts.

PR #519, opened 2019-06-06
Follow-up – Allow flushing before defining position and positionOffset components of particle species

franzpoeschel:
Pull request #518 contains a little bug, the unit dimensions need to be set *before* flushing.

### DIFF
--- a/src/ParticleSpecies.cpp
+++ b/src/ParticleSpecies.cpp
@@ -109,6 +109,13 @@ ParticleSpecies::flush(std::string const& path)
             patch.second.flush(patch.first);
     } else
     {
+        iterator it = find("position");
+        if ( it != end() )
+            it->second.setUnitDimension({{UnitDimension::L, 1}});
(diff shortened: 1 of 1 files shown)


src/ParticleSpecies.cpp
```diff
@@ -109,6 +109,17 @@ ParticleSpecies::flush(std::string const& path)
             patch.second.flush(patch.first);
     } else
     {
+        iterator it = find("position");
+        if ( it != end() )
+        {
```
ax3l: you can actually spare the `{}` for a single following line :)

franzpoeschel: I need to admit that I don't like that style that much, but since it's part of the [20 commandments](https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md), I will comply ;)

ax3l (approved):

Issue #489, opened 2019-02-21
Flush Before storeChunk: Triggers Write


ax3l:
@franzpoeschel just realized that a `series.flush()` for new record components, before a `storeChunk` is even called, queues a one-element dataset-write to the backends.

This is visible in the parallel write example:

  https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/examples/5_write_parallel.cpp#L86

and doesn't seem right from our internal handling logic:
- it will write unnecessary, random data to location (0,[0,...0])
- this data will be overwritten later on
- all backends have to handle this unusual corner case

Likely origin:
https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/src/backend/PatchRecordComponent.cpp#L97

franzpoeschel:
After following things around in the debugger, I can confirm that the code you are pointing to is the origin of this behavior.
First problem: `auxiliary::allocatePtr` will only allocate memory, but not *initialize* it, so we read uninitialized memory in the backends while performing the bogus writes. An easy fix would be to make `auxiliary::allocatePtr` actually *initialize* the memory. Apart from this file, the function is only ever used twice in [`CommonADIOS1Handler.cpp`](https://github.com/openPMD/openPMD-api/blob/dev/src/IO/ADIOS/CommonADIOS1IOHandler.cpp) and never for large values of `numPoints`. If we do this, the function should probably be renamed to avoid confusion.

The behavior of ADIOS not to create variables if no actual values have been written persists in ADIOS2. We could discuss whether it should be ok for our API to simply leverage this behavior. After all, when creating a dataset but not writing actual data to it, not seeing the dataset in the bpfile is arguably less surprising than seeing uninitialized memory at position 0 – especially since the behavior leaks into the other backends, resulting in such things:
```json
{
  "attributes": {
    ...
  },
  "data": [
    [
      43,
      null,
      null,
      null,
      null
    ],
    [
      null,
      null,
      null,
      null,
      null
    ],
    [
      null,
      null,
      null,
      null,
      null
    ]
  ],
  "datatype": "UINT"
}
```
It is likely impossible in either version of ADIOS to force the creation of a variable without doing tricks like this, but before we investigate it: should we? I would suggest just removing the lines that @ax3l pointed to and calling it a day.

Issue #516, opened 2019-05-28
makeEmpty: Overload / Support Zero-Extends


ax3l:
In some cases, as described in #514, it makes sense to explicitly write a zero-extend record component. We can already express this with constant record components but must allow this in the API properly.

Additionally, we can already add a new API `make_empty` instead of the "value-expecting" `make_constant` and automatically zero-initialize the `value` attribute of a constant record component.

In the future (https://github.com/openPMD/openPMD-standard/issues/211), we might want to allow more complicated types which are not zero-initializable. But for the API, we can keep the `value` with zero for now and expose the functionality already as described.

Python:
```python
import openpmd_api as api
import numpy as np

series = api.Series(
    "myOutput/data_%05T.h5",
    api.Access_Type.create)
i = series.iterations[42]

B = i.meshes["B"]
B_x = B["x"]

dataset = api.Dataset(
    np.double,
    [0, 0, 0, ])
B_x.reset_dataset(dataset)

# new (could also be implied if a `reset_dataset()` with zero-size shape is created
B_y.make_empty()

del series
```
or C++
```C++
#include <openPMD/openPMD.hpp>

namespace api = openPMD;

auto series = api::Series(
    "myOutput/data_%05T.h5",
    api::AccessType::CREATE);
auto i = series.iterations[42];

auto B = i.meshes["B"];
auto B_x = B["x"];

auto dataset = api::Dataset(
    api::determineDatatype<float>(),
    {0, 0});
B_x.resetDataset(dataset);

// new (could also be implied if a `resetDataset()` with zero-size shape is created
B_y.makeEmpty()
```

PR #528, opened 2019-06-17
Remove dummy dataset writing from RecordComponent::flush()

franzpoeschel:
See #489 and #512 for a description. The exactly same thing happens also in [`RecordComponent.cpp`](https://github.com/openPMD/openPMD-api/blob/dev/src/RecordComponent.cpp#L116). PR #512 did not catch this.

### DIFF
--- a/src/RecordComponent.cpp
+++ b/src/RecordComponent.cpp
@@ -137,21 +137,6 @@ RecordComponent::flush(std::string const& name)
                 dCreate.compression = m_dataset->compression;
                 dCreate.transform = m_dataset->transform;
                 IOHandler->enqueue(IOTask(this, dCreate));
-                if( m_chunks->empty() )
-                {
-                    /* Ensure at least one WRITE_DATASET per dataset occurs
-                     * ADIOS1 backend only creates a dataset after at least once cell has been written */
-                    Parameter< Operation::WRITE_DATASET > dWrite;
-                    auto uptr = auxiliary::allocatePtr(dCreate.dtype, 1);
-                    std::shared_ptr< void > data{std::move(uptr)};
-                    dWrite.data = data;
-                    dWrite.dtype = dCreate.dtype;
-                    if( dWrite.dtype == Datatype::UNDEFINED )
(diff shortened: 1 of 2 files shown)


ax3l:
CI seams unhappy.
You might want to implement the default for such records and attributes on `Series::close()` or similar.

franzpoeschel:
The problem is with the `fileBased_write_empty_test` which triggers exactly the situation that I have described [here](https://github.com/openPMD/openPMD-api/pull/512#issue-278721748):
> It is a design decision in ADIOS (also ADIOS2) not to physically create variables (used for openPMD datasets) unless data is actually written and openPMD can leverage this behavior. The workaround removed in this PR tries to circumvent this behavior. _Needs to be documented in this case_.

Relevant part of the test code:
```cpp
    Dataset dset = Dataset(Datatype::DOUBLE, {2});
    {
        Series o = Series("../samples/subdir/serial_fileBased_write%T." + backend, AccessType::CREATE);

        ParticleSpecies& e_1 = o.iterations[1].particles["e"];
        e_1["position"][RecordComponent::SCALAR].resetDataset(dset);
        e_1["positionOffset"][RecordComponent::SCALAR].resetDataset(dset);
        o.iterations[1].setTime(1.f);
        ParticleSpecies& e_2 = o.iterations[2].particles["e"];
        e_2["position"][RecordComponent::SCALAR].resetDataset(dset);
        e_2["positionOffset"][RecordComponent::SCALAR].resetDataset(dset);
        o.iterations[2].setTime(2.f);
        ParticleSpecies& e_3 = o.iterations[3].particles["e"];
        e_3["position"][RecordComponent::SCALAR].resetDataset(dset);
        e_3["positionOffset"][RecordComponent::SCALAR].resetDataset(dset);
        o.iterations[3].setTime(3.f);
    }
```
Since no data is ever written, the ADIOS variable is not created in storage and not read back in.

Implementing a default as you suggest would be equivalent to writing dummy data, just at a later point.
If we go for this, we should keep #516 in mind. Should we?

ax3l (approved):

PR #529, opened 2019-06-20
Allow to write empty datasets

### LINKED ISSUE #514: Writing datasets with zero extent in at least one dimension
**Problem**: For implementing the openPMD plugin in PIConGPU, it would be very helpful to be able to write datasets that have an extent of zero in at least one dimension, since this is currently also done by the raw ADIOS plugin. This can e.g. be seen in the [WarmCopper](https://github.com/ComputationalRadiationPhysics/picongpu/tree/master/share/picongpu/examples/WarmCopper) example whose output bp files contain lines such as:
```
  real                /data/0/fields/ph_particleMomentumComponent    {32, 32, 32}
  real                /data/0/fields/eth_particleMomentumComponent   {32, 32, 32}
  real                /data/0/fields/ehot_particleMomentumComponent  {32, 32, 32}
  real                /data/0/fields/Cu_particleMomentumComponent    {32, 32, 32}
  real                /data/0/particles/ph/position/x                {0}
  real                /data/0/particles/ph/position/y                {0}
  real                /data/0/particles/ph/position/z                {0}
  real                /data/0/particles/ph/momentum/x                {0}
  real                /data/0/particles/ph/momentum/y                {0}
  real                /data/0/particles/ph/momentum/z                {0}
  real                /data/0/particles/ph/weighting                 {0}
  unsigned long long  /data/0/particles/ph/particleId                {0}
  integer             /data/0/particles/ph/positionOffset/x          {0}
  integer             /data/0/particles/ph/positionOffset/y          {0}
  integer             /data/0/particles/ph/positionOffset/z          {0}
```
(output of `bpls`, the last column indicates the shape). Such data is currently impossible to write using the openPMD API since it [explicitly throws errors](https://github.com/openPMD/openPMD-api/blob/b560652881bdab1cd573e54f91e285d2ecf93295/src/RecordComponent.cpp#L56) upon attempting to write an empty dataset.

It should be made possible to write empty datasets in the natural way in the openPMD API.
With PR #512 having been merged, one potential problem to that end is already eliminated. Two possible solutions:
- Quick and dirty: (ab)use constant record components. Advantage: Should work with every backend. Disadvantage: A dummy value needs to be given, maybe not standard compliant.
- Cleaner: Allow to write as regular dataset. Advantage: Write generic code more easily, corresponds with the current state of affairs in PIConGPU's ADIOS plugin. Disadvantage: Backends do currently not expect empty datasets. 

As a further problem, unlike ADIOS1, ADIOS2 does not remember the dimensionality of a dataset that has zero extent in *all* dimensions:
```cpp
#include <adios2.h>

int
main( int argc, char * argv[] )
{
    using dt = long long;
    adios2::ADIOS adios( adios2::DebugON );
    adios2::IO bpIO = adios.DeclareIO( "IO" );
    adios2::Engine bpWriter = bpIO.Open( "empty.bp", adios2::Mode::Write );

    auto var = bpIO.DefineVariable< dt >( "empty",
        adios2::Dims{ 0, 0 },
        adios2::Dims{ 0, 0 },
        adios2::Dims{ 0, 0 } );
    auto var2 = bpIO.DefineVariable< dt >( "also_empty",
        adios2::Dims{ 0, 5 },
        adios2::Dims{ 0, 0 },
        adios2::Dims{ 0, 0 } );

    bpWriter.Put< dt >( var, nullptr );
    bpWriter.Put< dt >( var2, nullptr );
    bpWriter.Close();
}
```
`bpls empty.bp`:
```
  long long int  also_empty  {__, 5}
  long long int  empty       scalar
```
Hence, when writing a dataset with dimensionality `{ 0 }` (as is the case in the above PIConGPU example), the ADIOS2 backend would currently read this back in as a dataset with dimensionality *zero* and errors will be thrown if not treating it as such.

Is there another suggested way of writing data of this form? (@ax3l)

franzpoeschel:
Close #514 and #516.
This PR adds the following member to the class `RecordComponent`:
```cpp
    template< typename T >
    RecordComponent& makeEmpty( uint8_t dimensions );
```
This lets the user explicitly create an empty dataset.
Upon detecting an empty dataset, the member `RecordComponent& resetDataset(Dataset);` will not fail any more, but also create an empty dataset.

Since backends do currently not expect empty datasets, I implemented this feature by using constant record components. Zero-dimensional datasets are still not allowed.

### DIFF
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -107,6 +107,9 @@ public:
 
     template< typename T >
     RecordComponent& makeConstant(T);
+    
+    template< typename T >
+    RecordComponent& makeEmpty( uint8_t dimensions );
 
     /** Load and allocate a chunk of data
      *
@@ -159,6 +162,7 @@ OPENPMD_protected:
 private:
     void flush(std::string const&);
     virtual void read();
+    RecordComponent& makeEmpty( Dataset );
 }; // RecordComponent
 
 
@@ -172,6 +176,15 @@ RecordComponent::makeConstant(T value)
     *m_constantValue = Attribute(value);
     *m_isConstant = true;
     return *this;
+}
+
+template< typename T >
+inline RecordComponent&
+RecordComponent::makeEmpty( uint8_t dimensions )
+{
+    return makeEmpty( Dataset(
+        determineDatatype< T >(),
+        Extent( dimensions, 0 ) ) );
 }
 
 template< typename T >
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -182,6 +182,8 @@ template< typename T >
 inline RecordComponent&
 RecordComponent::makeEmpty( uint8_t dimensions )
 {
+    if ( dimensions == 0 )
+        throw std::runtime_error("Dataset extent must be at least 1D.");
     return makeEmpty( Dataset(
         determineDatatype< T >(),
         Extent( dimensions, 0 ) ) );
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -313,8 +313,10 @@ RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
 {
     if( *m_isConstant && !*m_isEmpty )
         throw std::runtime_error("Chunks can not be written for a constant RecordComponent.");
-    if( !data )
+    if( !data && !*m_isEmpty )
         throw std::runtime_error("Unallocated pointer passed during chunk store.");
+    if( *m_isEmpty && data )
+        throw std::runtime_error("Non-null pointer passed during chunk store of an empty RecordComponent.");
     Datatype dtype = determineDatatype(data);
     if( dtype != getDatatype() )
     {
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -107,7 +107,7 @@ public:
 
     template< typename T >
     RecordComponent& makeConstant(T);
-    
+
     template< typename T >
     RecordComponent& makeEmpty( uint8_t dimensions );
 
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -311,12 +311,12 @@ template< typename T >
 inline void
 RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
 {
-    if( *m_isConstant && !*m_isEmpty )
-        throw std::runtime_error("Chunks can not be written for a constant RecordComponent.");
-    if( !data && !*m_isEmpty )
-        throw std::runtime_error("Unallocated pointer passed during chunk store.");
-    if( *m_isEmpty && data )
-        throw std::runtime_error("Non-null pointer passed during chunk store of an empty RecordComponent.");
+    if( *m_isConstant )
+        throw std::runtime_error("Chunks cannot be written for a constant RecordComponent.");
+    if( *m_isEmpty )
+        throw std::runtime_error("Chunks cannot be written for an empty RecordComponent.");
+    if( !data )
+throw std::runtime_error("Unallocated pointer passed during chunk store.");
     Datatype dtype = determineDatatype(data);
     if( dtype != getDatatype() )
     {
@@ -348,8 +348,6 @@ RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
                                      + " - Chunk: " + std::to_string(o[i] + e[i])
                                      + ")");
 
-    if (*m_isEmpty)
-        return;
     Parameter< Operation::WRITE_DATASET > dWrite;
     dWrite.offset = o;
     dWrite.extent = e;
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -158,6 +158,9 @@ OPENPMD_protected:
 
     std::shared_ptr< std::queue< IOTask > > m_chunks;
     std::shared_ptr< Attribute > m_constantValue;
+    std::shared_ptr< bool > m_isEmpty{
+        std::make_shared< bool >( false )
+    };
 
 private:
     void flush(std::string const&);
@@ -297,7 +300,7 @@ template< typename T >
 inline void
 RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
 {
-    if( *m_isConstant )
+    if( *m_isConstant && !*m_isEmpty )
         throw std::runtime_error("Chunks can not be written for a constant RecordComponent.");
     if( !data )
         throw std::runtime_error("Unallocated pointer passed during chunk store.");
@@ -332,6 +335,8 @@ RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
                                      + " - Chunk: " + std::to_string(o[i] + e[i])
                                      + ")");
 
+    if (*m_isEmpty)
+        return;
     Parameter< Operation::WRITE_DATASET > dWrite;
     dWrite.offset = o;
     dWrite.extent = e;
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -108,6 +108,14 @@ public:
     template< typename T >
     RecordComponent& makeConstant(T);
 
+    /**
+     * Create a dataset with zero extent in each dimension.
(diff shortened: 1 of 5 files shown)


include/openPMD/RecordComponent.hpp
```diff
@@ -108,6 +108,9 @@ class RecordComponent : public BaseRecordComponent
     template< typename T >
     RecordComponent& makeConstant(T);
 
+    template< typename T >
+    RecordComponent& makeEmpty( uint8_t dimensions );
```
ax3l, line 120: please add a Doxygen string for this new functionality :)

include/openPMD/RecordComponent.hpp
```diff
@@ -155,10 +158,14 @@ class RecordComponent : public BaseRecordComponent
 
     std::shared_ptr< std::queue< IOTask > > m_chunks;
     std::shared_ptr< Attribute > m_constantValue;
+    std::shared_ptr< bool > m_isEmpty{
```
ax3l: shouldn't a simple `= ...` work as well here?

include/openPMD/RecordComponent.hpp
```diff
@@ -155,10 +158,14 @@ class RecordComponent : public BaseRecordComponent
...
+        std::make_shared< bool >( false )
+    };
 
 private:
     void flush(std::string const&);
     virtual void read();
+    RecordComponent& makeEmpty( Dataset );
```
ax3l, line 181: please also add a doxygen string here :)

include/openPMD/RecordComponent.hpp
```diff
@@ -282,7 +300,7 @@ template< typename T >
 inline void
 RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
 {
-    if( *m_isConstant )
+    if( *m_isConstant && !*m_isEmpty )
```
ax3l: shouldn't we throw a runtime error on `m_isEmpty` if `storeChunk` is called on it?

include/openPMD/backend/BaseRecordComponent.hpp
```diff
@@ -61,4 +61,25 @@ class BaseRecordComponent : public Attributable
     std::shared_ptr< Dataset > m_dataset;
     std::shared_ptr< bool > m_isConstant;
 }; // BaseRecordComponent
+namespace detail
+{
+template< typename RecordComponent_T >
+struct DefaultValue
```
ax3l, line 76: please give this a doxygen string on where and how this is used.

include/openPMD/backend/BaseRecordComponent.hpp
```diff
@@ -61,4 +61,25 @@ class BaseRecordComponent : public Attributable
     std::shared_ptr< Dataset > m_dataset;
     std::shared_ptr< bool > m_isConstant;
 }; // BaseRecordComponent
+namespace detail
```
ax3l, line 65: maybe add an empty newline here for readability?

src/RecordComponent.cpp
```diff
@@ -46,16 +46,19 @@ RecordComponent::setUnitSI(double usi)
...
 }
 
-RecordComponent&
-RecordComponent::resetDataset(Dataset d)
+RecordComponent &
+RecordComponent::resetDataset( Dataset d )
 {
     if( written )
-        throw std::runtime_error("A Records Dataset can not (yet) be changed after it has been written.");
+        throw std::runtime_error( "A Records Dataset can not (yet) be changed "
```
ax3l: typos:
```suggestion
        throw std::runtime_error( "A record's Dataset cannot (yet) be changed "
```

src/RecordComponent.cpp
```diff
@@ -74,6 +77,25 @@ RecordComponent::getExtent() const
     return m_dataset->extent;
 }
 
+RecordComponent&
+RecordComponent::makeEmpty( Dataset d )
+{
+    if( written )
+        throw std::runtime_error(
+            "A recordComponent can not (yet) be made"
```
ax3l: `RecordComponent` is a type, so let's keep the spelling to the typename as above:
```suggestion
            "A RecordComponent cannot (yet) be made"
```

test/SerialIOTest.cpp
```diff
@@ -37,6 +37,71 @@ std::vector<std::tuple<std::string, bool>> getBackends() {
...
+        REQUIRE(makeEmpty_resetDataset_dim3_notallzero.getExtent() == Extent{1,2,0});
+        REQUIRE(isSame(makeEmpty_resetDataset_dim3_notallzero.getDatatype(), Datatype::LONG_DOUBLE));
+    }
+}
+
+TEST_CASE( "empty_dataset_test", "[serial]" )
+{
+    for (auto const & t: backends)
+    {
+        //if (std::get<0>(t) == "json")
```
ax3l: dead code?

ax3l (commented):
Just a few little inline comments.

Can you please expose the new `makeEmpty` functionality in the Python bindings as well? :)

include/openPMD/RecordComponent.hpp
franzpoeschel: My intention was to allow writing generic code that does not need to distinguish empty datasets from non-empty ones. The offset and extent are still sanity-checked, meaning that the call to `storeChunk` will throw anyway if a non-empty chunk is passed. If all checks pass, this operation is a no-op.

ax3l: So just to understand this better: what do you consider a valid value for `data` in this case?

franzpoeschel: I guess you are hinting at the following lines? 
```cpp
if( !data )
    throw std::runtime_error("Unallocated pointer passed during chunk store.");
```
You are raising a point that I didn't think about thoroughly enough apparently. C++ does not allow zero-sized arrays (I was under the impression it did), so the user will need to make some distinction. (Meaning I'll have to check my PIConGPU plugin)
I guess, we should let nullpointers slide in this case?

ax3l: I mean I follow your wish to have simple code paths, but just contextually it seams wrong to me to call `storeChunk` on a constant or empty record component.

If we decided to allow this, which I feel makes the API actually more corner-casy than simple, then the only logically allowed value for `data` should be a `nullptr` because everything else is quite surprising.

franzpoeschel: Using `storeChunk` on a constant record component is still forbidden because its intention in the openPMD standard is to be an *alternative* to dealing with chunks.
Here, on the other hand, constant record components are merely an (arguably quick and dirty) *implementation* to allow storing chunks that are incidentally empty. 

But given this implementation, I see your point that `storeChunk` may give the wrong impression that the backend will eventually receive a dataset-related task which is definitely not the case. And since the C++ standard already makes a unified code path impossible, I would be ok with removing this functionality again if you prefer.

Related question: `storeChunk` does not yet support workflows where one rank contributes no data to a dataset that has data, right? We should consider this, if we decide to keep the change.

ax3l:
Agreed.

> Related question: `storeChunk` does not yet support workflows where one rank contributes no data to a dataset that has data, right? We should consider this, if we decide to keep the change.

Regarding the last point: no, storeChunk supports partial zero contributions (if not even all-zero for most backends). That's an important use case, e.g. for particles.

franzpoeschel:
Agreed = agreed to removing it again?
>storeChunk supports partial zero contributions (if not even all-zero for most backends)

Isn't the nullpointer check problematic in this case?

include/openPMD/RecordComponent.hpp
ax3l: (answered below since mail answer was wrapped around)

include/openPMD/RecordComponent.hpp
```diff
@@ -108,6 +108,17 @@ class RecordComponent : public BaseRecordComponent
     template< typename T >
     RecordComponent& makeConstant(T);
 
+    /**
+     * Create a dataset with zero extent in each dimension.
+     * Implemented by creating a constant record component with
```
ax3l, line 113: potentially add an empty comment line to separate title and brief description in doxygen :)

include/openPMD/backend/BaseRecordComponent.hpp
```diff
@@ -61,4 +61,34 @@ class BaseRecordComponent : public Attributable
     std::shared_ptr< Dataset > m_dataset;
     std::shared_ptr< bool > m_isConstant;
 }; // BaseRecordComponent
+
+namespace detail
+{
+/**
+ * Functor template to be used in combination with <switchType>"()"
```
ax3l, line 68: Maybe also keep a one-line title and a separate brief description here.

ax3l:
> Isn't the nullpointer check problematic in this case?

No, no contribution means: do not call `storeChunk`, as it already is.

ax3l (approved):

Issue #505, opened 2019-03-10
Default: Shared Lib


ax3l:
Let's change the CMake default to build a shared lib. That's more unix-ish and works by default with `cling`.

Follow-up to #503 

Way to do:
https://github.com/ornladios/ADIOS2/blob/v2.3.1/CMakeLists.txt#L96-L106

PR #532, opened 2019-07-18
CMake: Generate pkg-config .pc file

ax3l:
Generate `.pc` files for consumption via [`pkg-config`](https://people.freedesktop.org/~dbn/pkg-config-guide.html). This is useful for projects with legacy build environments that do not use a build system generator such as CMake and instead write `Makefile`s by hand or even compile manually by calling the compiler directly.

Example to update downstream: WarpX https://github.com/ECP-WarpX/WarpX/pull/80

Also removes the public defines for "variants" of the installed library from the build system and instead pulls them automatically in the facade headers via a `openPMD/config.hpp` file. Since the library "variant" cannot be changed after the fact anyway, this avoids having to set magic defines downstream.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,5 +290,5 @@ pkg-config --libs openPMD
 # -L${HOME}/somepath/lib -lopenPMD
 
 pkg-config --cflags openPMD
-# -I${HOME}/somepath/include -DopenPMD_HAVE_MPI=1 -DopenPMD_HAVE_JSON=1 -D_FORTIFY_SOURCE=2 -DopenPMD_HAVE_HDF5=1 -DopenPMD_HAVE_ADIOS1=0 -DopenPMD_HAVE_ADIOS2=0
+# -I${HOME}/somepath/include
 ```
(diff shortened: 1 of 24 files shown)


README.md
```diff
@@ -272,3 +272,23 @@ add_subdirectory("path/to/source/of/openPMD-api")
...
+
+```bash
+# optional: only needed if installed outside of system paths
+export PKG_CONFIG_PATH=$HOME/somepath/lib/pkgconfig:$PKG_CONFIG_PATH
+```
+
+Additional linker and compiler flags for your project are available via:
+```bash
+pkg-config --libs openPMD
+# -L${HOME}/somepath/lib -lopenPMD
```
ax3l, line 290: note: the `--static` output requires proper listing of `Libs.private` which is non-trivial. Luckily, we changed the default build variant to `shared` #505 for the next release, so `pkg-config` installs are only supported with shared builds for now, which is a sane limitation.

PR #538, opened 2019-07-25
Record_Component.make_empty()

franzpoeschel:
See [this PR](https://github.com/openPMD/openPMD-api/pull/529), this adds bindings for Python accordingly.

### DIFF
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -128,12 +128,12 @@ public:
     RecordComponent& makeEmpty( uint8_t dimensions );
 
     /**
-     * @brief Non-template overload of RecordComponent::makeEmpty(). 
+     * @brief Non-template overload of RecordComponent::makeEmpty().
      * Uses the passed openPMD datatype to determine the template parameter.
-     * 
+     *
      * @param dt The datatype of which to create an empty dataset.
      * @param dimensions The dimensionality of the dataset.
-     * @return RecordComponent& 
+     * @return RecordComponent&
      */
     RecordComponent& makeEmpty( Datatype dt, uint8_t dimensions );
 
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -127,6 +127,16 @@ public:
     template< typename T >
     RecordComponent& makeEmpty( uint8_t dimensions );
 
+    /**
+     * @brief Non-template overload of RecordComponent::makeEmpty(). 
+     * Uses the passed openPMD datatype to determine the template parameter.
+     * 
+     * @param dt The datatype of which to create an empty dataset.
+     * @param dimensions The dimensionality of the dataset.
+     * @return RecordComponent& 
+     */
+    RecordComponent& makeEmpty( Datatype dt, uint8_t dimensions );
+
     /** Returns true if this is an empty record component
      *
      * An empty record component has a defined dimensionality but zero extent
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -550,6 +550,162 @@ class APITest(unittest.TestCase):
         for ext in io.file_extensions:
             self.makeConstantRoundTrip(ext)
 
+    def makeEmptyRoundTrip(self, file_ending):
+        # write
+        series = io.Series(
+            "unittest_py_empty_API." + file_ending,
+            io.Access_Type.create
+        )
+
+        ms = series.iterations[0].meshes
+        SCALAR = io.Mesh_Record_Component.SCALAR
+        DT = io.Datatype
+
+        ms["CHAR"][SCALAR].make_empty(DT.CHAR, 1)
+        ms["UCHAR"][SCALAR].make_empty(DT.UCHAR, 2)
+        ms["SHORT"][SCALAR].make_empty(DT.SHORT, 3)
+        ms["INT"][SCALAR].make_empty(DT.INT, 4)
+        ms["LONG"][SCALAR].make_empty(DT.LONG, 5)
+        ms["LONGLONG"][SCALAR].make_empty(DT.LONGLONG, 6)
+        ms["USHORT"][SCALAR].make_empty(DT.USHORT, 7)
+        ms["UINT"][SCALAR].make_empty(DT.UINT, 8)
+        ms["ULONG"][SCALAR].make_empty(DT.ULONG, 9)
+        ms["ULONGLONG"][SCALAR].make_empty(DT.ULONGLONG, 10)
+        ms["FLOAT"][SCALAR].make_empty(DT.FLOAT, 11)
+        ms["DOUBLE"][SCALAR].make_empty(DT.DOUBLE, 12)
+        ms["LONG_DOUBLE"][SCALAR].make_empty(DT.LONG_DOUBLE, 13)
+
+        if found_numpy:
+            ms["int16"][SCALAR].make_empty(np.dtype("int16"), 14)
+            ms["int32"][SCALAR].make_empty(np.dtype("int32"), 15)
+            ms["int64"][SCALAR].make_empty(np.dtype("int64"), 16)
+            ms["uint16"][SCALAR].make_empty(np.dtype("uint16"), 17)
+            ms["uint32"][SCALAR].make_empty(np.dtype("uint32"), 18)
+            ms["uint64"][SCALAR].make_empty(np.dtype("uint64"), 19)
+            ms["single"][SCALAR].make_empty(np.dtype("single"), 20)
+            ms["np_double"][SCALAR].make_empty(np.dtype("double"), 21)
+
+        # flush and close file
+        del series
+
+        # read back
+        series = io.Series(
+            "unittest_py_empty_API." + file_ending,
+            io.Access_Type.read_only
+        )
+
+        ms = series.iterations[0].meshes
+
+        self.assertEqual(
+            ms["CHAR"][SCALAR].shape,
+            [0 for _ in range(1)]
+        )
+        self.assertEqual(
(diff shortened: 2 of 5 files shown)


ax3l:
Please also add tests :)

test/python/unittest/API/APITest.py
```diff
@@ -413,6 +413,158 @@ def testConstantRecords(self):
...
+        series = api.Series(
+            "unittest_py_empty_API." + file_ending,
+            api.Access_Type.create
+        )
+
+        ms = series.iterations[0].meshes
+        SCALAR = api.Mesh_Record_Component.SCALAR
+        DT = api.Datatype
+
+        ms["CHAR"][SCALAR].make_empty(DT.CHAR, 1)
```
ax3l, line 564: if you do this (as well) with `np.dtype`'s as above this would be ideal

test/python/unittest/API/APITest.py
```diff
@@ -413,6 +413,158 @@ def testConstantRecords(self):
...
+        self.assertEqual(
+            ms["DOUBLE"][SCALAR].shape,
+            [0 for _ in range(12)]
+        )
+        self.assertEqual(
+            ms["LONG_DOUBLE"][SCALAR].shape,
+            [0 for _ in range(13)]
+        )
+
+        self.assertEqual(
```
ax3l: `self.assertIs`?

test/python/unittest/API/APITest.py
franzpoeschel: The method `RecordComponent.make_empty` currently expects a `openpmd_api.Datatype`. Should I add an overload for numpy datatypes so the following is legal?
```python
series.iterations[0].meshes["E"]["x"].make_empty(numpy.dtype("int16"), 5)
```

test/python/unittest/API/APITest.py
franzpoeschel: I used explicit casting now. Wasn't aware that the objects would not be identical.

test/python/unittest/API/APITest.py
ax3l: yep, such an additional overload would be good for usability.

franzpoeschel:
The [datatype tests](https://github.com/franzpoeschel/openPMD-api/blob/7c84fa64dc73fc0b99c5acabdb1f29bcfc834cb2/test/python/unittest/API/APITest.py#L552) apparently fail for the HDF5 backend. I don't think this is a bug with the current PR.

ax3l:
CMake backport issue in Spack is going to be fixed in https://github.com/spack/spack/pull/12241
Just push again after the merge.

Maybe check already to Windows builds, they seem actually broken.

ax3l:
@franzpoeschel ping: please rebase

test/python/unittest/API/APITest.py
```diff
@@ -413,6 +413,215 @@ def testConstantRecords(self):
...
+        self.assertTrue(
+            api.determine_datatype(ms["INT"][SCALAR].dtype) ==
+            DT.INT
+        )
+        self.assertTrue(
+            api.determine_datatype(ms["LONG"][SCALAR].dtype) ==
+            DT.LONG
+        )
+        self.assertTrue(
+            api.determine_datatype(ms["LONGLONG"][SCALAR].dtype) ==
```
ax3l: careful, this is not how `int`/`long`/`long long` works in C/C++: https://github.com/openPMD/openPMD-api/pull/333#issuecomment-419257519 / #342 

backends likely serialize them to the closest fixed size type, leading to a round-trip that depends on the [data model of a platform](https://en.cppreference.com/w/cpp/language/types#Integer_types).

The easiest way to fix this is to either:
- only check fixed-type round-trips or
- check the corresponding fixed-size type on read of `short`/`int`/`long`/`long long` and compare to that.

You can check the other tests in the same file how it is done there.

franzpoeschel: Thanks for the hint, I have changed things accordingly!

franzpoeschel:
It seems there is an issue in the ADIOS1 backend, the [recent commit](https://github.com/openPMD/openPMD-api/pull/538/commits/0a6377f1990938313484caa8c32a2ef00c25394c) has made the checks pass. Leaving this line inside
```cpp
ms["double"][SCALAR].make_empty(np.dtype("double"), 21)
```
will create datasets that cannot even be processed by `bpls` (in either version of ADIOS). 

EDIT: It seems that the reason for this behavior is the fact that ADIOS is apparently case insensitive, choosing another name for the field will apparently work. I will push a commit in a minute.

ax3l:
Huh, I wasn't aware of that :-o

include/openPMD/RecordComponent.hpp
```diff
@@ -119,6 +119,15 @@ class RecordComponent : public BaseRecordComponent
...
     RecordComponent& makeEmpty( uint8_t dimensions );
 
+    /**
+     * @brief Non-template overload of RecordComponent::makeEmpty(). 
+     * Uses the passed openPMD datatype to determine the template parameter.
+     * 
+     * @param dimensions 
+     * @return RecordComponent& 
+     */
+    RecordComponent& makeEmpty( Datatype, uint8_t dimensions );
```
ax3l: Just an ascetic question: would it be better to have the `Datatype` last?
Also: doxygen does not describe both parameters (second name is missing, `dt`?)

ax3l:
@franzpoeschel I fixed the one test that ran into timeouts with #558.
Please rebase this PR against the latest `dev` and and remove the `[WIP]` in the title when it's ready for review (looks like it).

ax3l:
@franzpoeschel pls rebase me one more time :)

include/openPMD/RecordComponent.hpp
franzpoeschel: I am used to putting that parameter first since that's how `switchType` does it, so I did that without a lot of thinking. You mean because it would align better with the templated `makeEmpty`? I actually like the order that it has now, since you first specify the datatype (as template or as normal parameter), thus also specifying whether you use the templated `makeEmpty` or the non-template, and next you specify the dimensions.
```cpp
rc.makeEmpty< float >( 3u );
rc.makeEmpty( Datatype::FLOAT, 3u);
```
Think that's pretty straightforward this way around?

I've completed the doxygen description.

ax3l: Okay, that looks legit as well.

ax3l: Orthogonal question for a follow-up PR about the return type:
with handles implemented, could we make the coder safer by returning in all those places just a `RecordComponent` instead of a `RecordComponent&`? What do you think?

include/openPMD/RecordComponent.hpp
```diff
@@ -127,6 +127,16 @@ class RecordComponent : public BaseRecordComponent
     template< typename T >
     RecordComponent& makeEmpty( uint8_t dimensions );
 
+    /**
+     * @brief Non-template overload of RecordComponent::makeEmpty(). 
+     * Uses the passed openPMD datatype to determine the template parameter.
+     * 
+     * @param dt The datatype of which to create an empty dataset.
+     * @param dimensions The dimensionality of the dataset.
+     * @return RecordComponent&
```
ax3l:
```suggestion
     * @return A reference to this RecordComponent.
```

ax3l: Orthogonal question for a follow-up PR about the return type:
with handles implemented, could we make the code usage safer by returning in all those places just a `RecordComponent` instead of a `RecordComponent&`? What do you think?

test/python/unittest/API/APITest.py
```diff
@@ -447,6 +447,161 @@ def testConstantRecords(self):
...
+            self.assertTrue(ms["uint64"][SCALAR].dtype == np.dtype("uint64"))
+            self.assertTrue(ms["single"][SCALAR].dtype == np.dtype("single"))
+            self.assertTrue(
+                ms["np_double"][SCALAR].dtype == np.dtype("double"))
+
+    def testEmptyRecords(self):
+        backend_filesupport = {
+            'json': 'json',
+            'hdf5': 'h5',
+            'adios1': 'bp'
```
ax3l: Missing?
```suggestion
            'adios1': 'bp',
            'adios2': 'bp'
```
or
```suggestion
            'adios': 'bp'
```
?

include/openPMD/RecordComponent.hpp
franzpoeschel: The intent of returning this at all is chaining methods, right? So, things such as `recordComponent.makeEmpty(3).setUnitSI(3.0)`.
By forcing a copy there, we would also needlessly force copying the contained `shared_ptr`s, which we already do somewhat excessively at times, so I'm not really convinced.

I'm in general not a huge fan of how C++ does references and the problems in returning references is one of the major reasons (especially implicit copying upon `auto x = <reference>`), but do you see a safety issue in our case?

test/python/unittest/API/APITest.py
franzpoeschel:
```
In [2]: openpmd_api.variants
Out[2]: {'adios1': False, 'adios2': True, 'hdf5': False, 'json': True, 'mpi': True}
```
The first suggestion is the right one, yep I forgot ADIOS2.

include/openPMD/RecordComponent.hpp
ax3l: Yes, the idea is to chain methods.

I think copying a shared_ptr is okay if it prevents that a user can create a dangling reference (to `RecordComponent`). It makes the API safer and keeps the syntactic sugar of chaining.

franzpoeschel:
I think this one is finished @ax3l ;)

ax3l (approved):
Thank you, this looks great! :rocket: :sparkles:

Issue #150, opened 2018-04-26
Passing of advanced backend parameters


anokfireball:
An option to provide advanced parameters specific to a backend is desirable. This can include MPI_Info, file access strategies,communication strategies, buffer sizes among other things. This should be *optional* with sane defaults for all values.

Specifically, this matters in two places:
 - Series, where one would provide general backend parameters shared across all files, paths, datasets and attributes in the Series (e.g. MPI_Info, ADIOS_AGGREGATE, nometa)
```cpp
namespace openPMD
{
Series::Series(std::string const& filepath,
              AccessType at,
              MPI_Comm comm,
              AbstractSeriesParam* = nullptr /* default */);
} // openPMD
```
 - Datasets, where one would provide parameters different for specific datasets (e.g. compression/transform)
```cpp
namespace openPMD
{
Dataset&
Dataset::setParameters(AbstractDatasetParam*);
} // openPMD
```

Implementation can be done via polymorphic parameter structs (similar to ```IO::Parameter```) where paramters are passed named and typesafe (preferred) or as a map (similar to ```backend::Attributable```) (not preferred).

ax3l:
I would suggest to pass an object by value for `AbstractSeriesParam`. One can still add a default dummy as default arg:
```C++
Series::Series(std::string const& filepath,
              AccessType at,
              MPI_Comm comm,
              AbstractSeriesParam = SeriesDefaultParam{});
```

We can also add a `Series::setOptions()` method for later adjustment, but some flags might already be needed at construction time.

franzpoeschel:
To get back on this, a further option could be using JSON. We already ship with [Niels Lohmann's JSON library](https://github.com/nlohmann/json). This would allow to flexibly express backend-specific configuration options while at the same time not requiring to implement backend-specific classes like the solution with `AbstractSeriesParam`.

As an example, choosing the ADIOS2 engine with a JSON configuration:
```json
{
  "Engine": {
    "Type": "SST",
    "Params": {
      "QueueLimit": 3,
      "QueueFullPolicy": "Block"
    }
  }
}
```
This can be easily exposed to the Python API by passing the JSON configuration [as a simple string](https://github.com/nlohmann/json#tofrom-strings).

ax3l:
[...]

PR #554, opened 2019-08-22
storeChunk: independent calls

ax3l:
Parallel test: allow to call `storeChunk()` an independent number of times from various ranks. This currently deadlocks on `Series::flush()` for ~~both ADIOS1 and~~ HDF5 (ok!).

@C0nsultant @franzpoeschel can you please take a look at this? At least for ADIOS1 this should be a totally fine workflow. I think that is a bug in our library to require a matched number of calls to `storeChunk()` from all participating ranks.

### To Do

- add more test cases. Currently test the corner-case
  - [x] one calls `storeChunk`, another does not (fails); but we also need
  - [ ] some ranks call `storeChunk` <N> times, others `<M>!=<N>` times
  - [ ] and the same tests for `loadChunk`

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -177,7 +177,9 @@ TEST_CASE( "hdf5_write_test_zero_extent", "[parallel][hdf5]" )
 
 TEST_CASE( "hdf5_write_test_skip_chunk", "[parallel][hdf5]" )
 {
-    write_test_zero_extent( "h5", false );
+    //! @todo add with H5FD_MPIO_INDEPENDENT
+    // write_test_zero_extent( "h5", false );
+    REQUIRE(true);
 }
 
 #else
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -175,6 +175,10 @@ TEST_CASE( "hdf5_write_test_zero_extent", "[parallel][hdf5]" )
     write_test_zero_extent( "h5", true );
 }
 
+TEST_CASE( "hdf5_write_test_skip_chunk", "[parallel][hdf5]" )
+{
+    write_test_zero_extent( "h5", false );
+}
 
 #else
 
@@ -224,6 +228,11 @@ TEST_CASE( "adios_write_test", "[parallel][adios]" )
 TEST_CASE( "adios_write_test_zero_extent", "[parallel][adios]" )
 {
     write_test_zero_extent( "bp", true );
+}
+
+TEST_CASE( "adios_write_test_skip_chunk", "[parallel][adios]" )
+{
+    write_test_zero_extent( "bp", false );
 }
 
 TEST_CASE( "hzdr_adios_sample_content_test", "[parallel][adios1]" )
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,7 @@ Bug Fixes
 
 - Clang: fix pybind11 compile on older releases, such as AppleClang 7.3-9.0, Clang 3.9 #543
 - Python: skip examples using HDF5 if backend is missing #544
+- ADIOS1: fix deadlock in MPI-parallel, non-collective calls to ``storeChunk()`` #554
 
 Other
 """""
--- a/src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
@@ -58,10 +58,6 @@ ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
 
     if( this->m_handler->accessTypeBackend != AccessType::READ_ONLY )
     {
-        for( auto& f : m_openWriteFileHandles )
-            close(f.second);
-        m_openWriteFileHandles.clear();
-
         for( auto& group : m_attributeWrites )
             for( auto& att : group.second )
                 flush_attribute(group.first, att.first, att.second);
@@ -75,6 +71,7 @@ ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
 
         for( auto& f : m_openWriteFileHandles )
             close(f.second);
+        m_openWriteFileHandles.clear();
     }
 
     int status;


ax3l:
@guj just quickly mentioned the list of collective HDF5 calls again, so if we do not want to re-implement the buffer functionality of ADIOS for HDF5, this backend might just not be as flexible: https://support.hdfgroup.org/HDF5/doc/RM/CollectiveCalls.html

Nevertheless, I seam to remember for ADIOS1 that calls such as `adios_write_byid`/`adios_schedule_read` are [not collective per-se](https://github.com/ComputationalRadiationPhysics/picongpu/commit/7fadc570ebdb0ae0e056a399bc3386b09b44c59b) (only in corner cases, such as transports to PHDF5) and should be callable independently. What we might be doing right now is that we accidentally group them together with collective calls in the backend?

@RemiLehe also [mentioned](https://github.com/openPMD/openPMD-api/issues/150#issuecomment-538503104) we could try `H5FD_MPIO_INDEPENDENT` again for HDF5, but from our past experiences (see mention) this does not work well (deadlocks or performance penalties at scale)... (maybe better these days).

guj:
It should work if one calls adios_open/close methods in all ranks, and write var in some ranks.

RemiLehe:
@guj Thanks! Is that true for both adios1 and adios2?

guj:
Yes - Junmin

test/ParallelIOTest.cpp
```diff
@@ -175,6 +175,10 @@ TEST_CASE( "hdf5_write_test_zero_extent", "[parallel][hdf5]" )
     write_test_zero_extent( "h5", true );
 }
 
+TEST_CASE( "hdf5_write_test_skip_chunk", "[parallel][hdf5]" )
+{
+    write_test_zero_extent( "h5", false );
```
ax3l: according to @RemiLehe's tests these days, only HDF5 seems to be problematic.
So let's expose `H5FD_MPIO_INDEPENDENT` via runtime options #569

ax3l:
I am not sure why this is not deadlocking for BP3/ADIOS1 anymore, but that is good news. Potentially I mismatched it while running and designing the tests.

~~Update: funny, ADIOS1 BP works for me locally (mpich@3.3.1) but deadlocks in CI (ancient openmpi@1.6.5). Could be still a bug in there or it's just the really old MPI. For old OpenMPI versions... currently testing <= 2.1.6, I can reproduce the deadlock locally as well.~~

test/ParallelIOTest.cpp
ax3l: ~~funny, ADIOS1 BP works for me locally but fails in CI. Could be still a bug in there.~~

ax3l:
Honestly, I bet it's another OpenMPI bug... Urgh.

ax3l:
So, these are my own checks on my local machine, Ubuntu 18.04.3 LTS, GCC 7.4.0, CMake 3.15.4. ADIOS1 and ADIOS2 build via Spack, PHDF5 taken from system.

| MPI         | backend        | result             |
|-------------|----------------|--------------------|
| MPICH 3.3.1 | ADIOS2 2.5.0   | :heavy_check_mark: |
| MPICH 3.3.1 | ADIOS 1.13.1   | :x:                |
| MPICH 3.3.1 | ADIOS 1.13.1 (**) | :x:                |
| MPICH 3.3a2 | PHDF5 1.10.0.1 | :x:                |
| MPICH 3.3.1 | PHDF5 1.8.13  | :x:                |
| MPICH 3.3.1 | PHDF5 1.8.13 (*) | :heavy_check_mark: |

(*) with `export OPENPMD_HDF5_INDEPENDENT=ON`
(**) with [ROMIO hints](https://www.mcs.anl.gov/projects/romio/2008/09/26/system-hints-hints-via-config-file/), although ADIOS should not be using MPI-I/O but rather POSIX I/O and MPI communication, such as
```
romio_cb_write disable
romio_no_indep_rw false
```

Now ADIOS1, ADIOS2, HDF5 build via Spack for OpenMPI 3.1.4:

| MPI           | backend        | result             |
|---------------|----------------|--------------------|
| OpenMPI 3.1.4 | ADIOS2 2.5.0   | :heavy_check_mark: |
| OpenMPI 3.1.4 | ADIOS 1.13.1   | :x:                |
| OpenMPI 3.1.4 | PHDF5 1.10.5   |  :x:               |
| OpenMPI 4.0.2 | PHDF5 1.10.5   |  :x:               |
| OpenMPI 4.0.2 | PHDF5 1.10.5 (*) | :heavy_check_mark: |

(*) with `export OPENPMD_HDF5_INDEPENDENT=ON`

But, when executed on Cori-KNL (NERSC), with the following commands

```bash
module swap craype-haswell craype-mic-knl
module swap PrgEnv-intel PrgEnv-gnu  # GCC 8.2.0
module load adios/1.13.1

export CC="$(which cc)"
export CXX="$(which CC)"
export CRAYPE_LINK_TYPE=dynamic

cd $SCRATCH
git clone https://github.com/openPMD/openPMD-api.git
cd openPMD-api
git remote add ax3l https://github.com/ax3l/openPMD-api.git
git checkout ax3l/fix-storeChunkCalls
mkdir build
cd build
cmake .. -DopenPMD_USE_HDF5=OFF -DopenPMD_USE_ADIOS1=ON -DopenPMD_USE_ADIOS2=OFF -DopenPMD_USE_PYTHON=OFF -DMPIEXEC_EXECUTABLE=$(which srun)
make -j 8

salloc --time=1:00:00 -N 1 -C knl -q interactive
CTEST_OUTPUT_ON_FAILURE=1 ctest -V -R ParallelIO
# or manually
srun -n 2 bin/ParallelIOTests
```

ADIOS1 does not deadlock:

| MPI           | backend        | result             |
|---------------|----------------|--------------------|
| cray-mpich/7.7.6 | ADIOS 1.13.1    | :heavy_check_mark: |

### Legend
:x: : deadlock
:heavy_check_mark: : works

ax3l:
For PHDF5, I added a new environment variable, `OPENPMD_HDF5_INDEPENDENT=[OFF]`, that allows to set the transfer to non-collective I/O #576. With that, PHDF5 I/O with non-collective calls to `storeChunk` worked for me. (We will also add a `Series` constructor option for this, but #569 still needs some love.)

I am still puzzled why the ADIOS1 implementation deadlocks. Maybe @franzpoeschel can spot it quicker than I do.

.travis.yml
```diff
@@ -238,6 +238,9 @@ jobs:
             $TRAVIS_BUILD_DIR
         - make -j 2
         # run tests
+        #   Independent I/O in PHDF5
+        #   https://github.com/openPMD/openPMD-api/pull/554
+        - OPENPMD_HDF5_INDEPENDENT=ON
```
ax3l: it would be better if we add this for the specific test as JSON option #569, because otherwise we wildcard a good amount of potential issues in CI, but we do not have this in place yet.

test/ParallelIOTest.cpp
ax3l: potentially exclude this test for now, see CI note.
Otherwise with the current defaults, collective, PHDF5 tests will fail for users.

ax3l:
`printf` debugging on my MPICH 3.3.1 setup locally:
```
$ mpiexec -prepend-rank -n 2 bin/ParallelIOTests adios_write_test_skip_chunk
[0] adios_init_noxml
[1] adios_init_noxml
[0] adios_read_init_method
[1] adios_read_init_method
[0] adios_declare_group
[0] adios_select_method
[1] adios_declare_group
[1] adios_select_method
[0] adios_define_var
[1] adios_define_var
[1] adios_define_var
[1] adios_define_var[0] adios_define_var
[0] adios_define_var
[1] 
[0] adios_define_var
[0] adios_define_var
[0] adios_define_var
[1] adios_define_var
[1] adios_define_var
[1] adios_define_var[1] 
[1] adios_open
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue[0] 
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue
[0] adios_define_attribute_byvalue[0] 
[0] adios_define_attribute_byvalue
[0] adios_open
[0] WARN : MPI_AMR method: No OST to use. Set num_ost=NNN in the adios config xml file.
[0] adios_close
[1] adios_write
[1] adios_write
[1] adios_write
[1] adios_write
[1] adios_write
[1] adios_write[1] 
[1] adios_close
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue[1] 
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue[1] 
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue[1] 
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_define_attribute_byvalue
[1] adios_open
```
For some reason, I see 3x `adios_open` in there.

Looks like a rouge work-around inside `ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()` is causing the problems.

[This block](https://github.com/openPMD/openPMD-api/blob/0.9.0-alpha/src/IO/ADIOS/ParallelADIOS1IOHandler.cpp#L69-L74) which is also present in the serial backend seams to be the culpit:
```C++
for( auto& f : m_filePaths )
    if( m_openWriteFileHandles.find(f.second) == m_openWriteFileHandles.end() )
        m_openWriteFileHandles[f.second] = open_write(f.first);
```

Note: The `m_openWriteFileHandles.find(f.second) == m_openWriteFileHandles.end()` comparison has some code smell, it compares a shared pointer of string instead of the strings itself... This does not seam right.
Why is this even a [`std::unordered_map< std::shared_ptr< std::string >, int64_t >`](https://github.com/openPMD/openPMD-api/blob/0.9.0-alpha/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp#L91) instead of a simple `std::unordered_map< std::string, int64_t >`?
This `.find()` construct is used at several places in the ADIOS1 implementation.

guj:
if a Series is constructed,  there are two places  adios_open() is called.  
(1) Series::~Series(). this  destructor calls a flush(), and this flush() leads to adios_open()  when  "while( !handler->m_work.empty() )". is true (ParallelADIOS1IOHandler.cpp: 135).   
 (2). Attributable::~Attributable(), because it is super class of Series.  

In your test,  rank 1 called adios_open() in both places,  rank 0 has no work, so called at (2) only.
when (1) is true, CommonADIOS1IOHandler.cpp::766 calls open_write()
it is valid because no file has been created

I am wondering why (2) is necessary to call file creation in a destructor. Might make more sense to construct the file in the Series. constructor.

ax3l:
Thank you for finding these details!

I think although issuing an early `flush()` after series open will already work-around this quickly, we should investigate not to delay the `adios_open`. I think we will generally not benefit from delaying that command in the ADIOS1 backend after `Series()` creation since that's a point where we are collective anyway.

`.flush()` and `storeChunk()` on the other hand should ideally be non-collective/independent operations.

ax3l:
I think I might have now spotted the issue in `ParallelADIOS1IOHandler.cpp` and pushed a fix in the last commit :)

PR #568, opened 2019-10-07
ADIOS2: Infrastructure & Docs Update

ax3l:
Further updates now that we have ADIOS2 #482 taking off :rocket: :sparkles: 

- [x] update documentation further, including changes and upgrades
- [x] update CI to dominantly check ADIOS2 runtime
- [x] update `Dockerfile` for `manylinux2010` build https://github.com/ornladios/ADIOS2/issues/1810
- [x] update parallel benchmark example
- [x] make selected backend for ``.bp`` files a runtime option (currently via environment variable)
- [x] fix OSX `Extent` cast causing a build error

## Found bug during the process

ADIOS2 VariableDefiner: Fix OSX Cast Error
    
Add an explicit cast from `openPMD::Extent` (aka `std::vector< std::uint64_t >`) to `adios2::Dims` (aka `std::vector< std::size_t >`) to avoid a compile error on OSX:
    
```
    include/openPMD/Datatype.hpp:558:23: error: no matching member function for call to 'operator()'
            return action.OPENPMD_TEMPLATE_OPERATOR( )< char >(
                   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    
    include/openPMD/Datatype.hpp:532:44: note: expanded from macro 'OPENPMD_TEMPLATE_OPERATOR'
                                               ^
    
    src/IO/ADIOS/ADIOS2IOHandler.cpp:201:9: note: in instantiation of function template specialization 'openPMD::switchType<void, openPMD::detail::VariableDefiner, adios2::IO &, std::__1::basic_string<char> &, std::__1::unique_ptr<adios2::Operator, std::__1::default_delete<adios2::Operator> >, const std::__1::vector<unsigned long long, std::__1::allocator<unsigned long long> > &>' requested here
            switchType( parameters.dtype, detail::VariableDefiner( ),
            ^
    
    src/IO/ADIOS/ADIOS2IOHandler.cpp:794:5: note: candidate function not viable: no known conversion from 'const vector<unsigned long long, allocator<unsigned long long>>' to 'const vector<size_t, allocator<unsigned long>>' for 4th argument
        operator( )( adios2::IO & IO, const std::string & name,
        ^
    
    src/IO/ADIOS/ADIOS2IOHandler.cpp:804:27: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'n'
        void VariableDefiner::operator( )( adios2::IO &, Params &&... )
```

PR #569, opened 2019-10-10
Advanced backend configuration via JSON

### LINKED ISSUE #150: Passing of advanced backend parameters

franzpoeschel:
This PR adds backend-specific configuration options. Close #150. The configuration is passed as an optional string parameter in JSON format to the constructor of `openPMD::Series`.
The top-level JSON maps backends to the corresponding backend's configuration. This allows to configure several backends at once.

At the moment, only the ADIOS2 backend has been made aware of this. A possible configuration looks like this:
```cpp
std::string options = R"(
{
  "adios2": {
    "engine": {
      "type": "sst",
      "parameters": ["BufferGrowthFactor": "2.0", "QueueLimit": "2"]
    }
  }
}
)";
```
TODO:

- [x] Safety measure: Warn upon reading an unexpected key.
- [x] Add Python bindings
- [x] Code cleanup
- [x] Add user documentation, e.g. in `docs/source/backends/adios2.rst`

Probably follow-up for a later PR:

- [ ] Implement the suggestions in issue #688 
- [ ] Further configuration options in ADIOS2, e.g. switching on and off debug mode at runtime.
- [ ] Implement option reading in other backends
- [ ] What about case insensitivity? Should the JSON keys that are read by the openPMD API and not directly passed through to the backend (such as the ADIOS2 "parameters" list) be made case insensitive?
- [ ] Override choice of backend from JSON. Choosing the backend by filename extension is simple, yet inflexible if one backend has several possible extensions (ADIOS2) or two backens share an extension (ADIOS in versions 1 and 2). Just overwrite setting introduced in #568

### DIFF
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -1,6 +1,6 @@
 .. _backendconfig
 
-Backend-Specific configuration
+Backend-Specific Configuration
 ==============================
 
 While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is sometimes useful to pass configuration parameters to the specific backend in use.
@@ -10,6 +10,7 @@ A JSON option always takes precedence over an environment variable.
 The fundamental structure of this JSON configuration string is given as follows:
 
 .. literalinclude:: config_layout.json
+   :language: json
 
 This structure allows keeping one configuration string for several backends at once, with the concrete backend configuration being chosen upon choosing the backend itself.
 
@@ -39,12 +40,17 @@ ADIOS2
 A full configuration of the ADIOS2 backend:
 
 .. literalinclude:: adios2.json
+   :language: json
 
-All keys found under ``adios2.dataset`` are applicable globally as well as per dataset, keys found under ``adios2.engine`` only globally. Explanation of the single keys:
+All keys found under ``adios2.dataset`` are applicable globally as well as per dataset, keys found under ``adios2.engine`` only globally.
+Explanation of the single keys:
 
-* ``adios2.engine.type``: A string that is passed directly to ``adios2::IO:::SetEngine`` for choosing the ADIOS2 engine to be used. Please refer to the `official ADIOS2 documentation <https://adios2.readthedocs.io/en/latest/engines/engines.html>`_ for a list of available engines.
-* ``adios2.engine.type``: An associative array of string-formatted engine parameters, passed directly through to ``adios2::IO::SetParameters``. Please refer to the official ADIOS2 documentation for the allowable engine parameters.
-* ``adios2.dataset.operators``: (*currently unimplemented* – please use the ``openPMD::Dataset::compression`` for the meantime) This key contains a list of ADIOS2 `operators <https://adios2.readthedocs.io/en/latest/components/components.html#operator>`_, used to enable compression or dataset transformations. Each object in the list has three keys:
+* ``adios2.engine.type``: A string that is passed directly to ``adios2::IO:::SetEngine`` for choosing the ADIOS2 engine to be used.
+  Please refer to the `official ADIOS2 documentation <https://adios2.readthedocs.io/en/latest/engines/engines.html>`_ for a list of available engines.
+* ``adios2.engine.type``: An associative array of string-formatted engine parameters, passed directly through to ``adios2::IO::SetParameters``.
+  Please refer to the official ADIOS2 documentation for the allowable engine parameters.
+* ``adios2.dataset.operators``: (*currently unimplemented* – please use the ``openPMD::Dataset::compression`` for the meantime) This key contains a list of ADIOS2 `operators <https://adios2.readthedocs.io/en/latest/components/components.html#operator>`_, used to enable compression or dataset transformations.
+  Each object in the list has three keys:
 
   * ``type`` supported ADIOS operator type, e.g. zfp, sz
   * ``parameters`` is an associative map of string parameters for the operator (e.g. compression levels)
@@ -52,4 +58,5 @@ All keys found under ``adios2.dataset`` are applicable globally as well as per d
 Other backends
 ^^^^^^^^^^^^^^
 
-Do currently not read the configuration string. Please refer to the respective backends' documentations for further information on their configuration.
+Do currently not read the configuration string.
+Please refer to the respective backends' documentations for further information on their configuration.
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -1,9 +1,11 @@
 .. _backendconfig
 
-Backend-specific configuration
+Backend-Specific configuration
 ==============================
 
-While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is often necessary to pass configuration parameters to the concrete backend in use. Some backends recognize a number of configuration options via environment variables (see the backends' documentations), the preferred way to do so, however, *– as far as it is implemented –* is via a JSON-formatted string.
+While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is sometimes useful to pass configuration parameters to the specific backend in use.
+:ref:`For each backend <backends-overview>`, configuration options can be passed via a JSON-formatted string or via environment variables.
+A JSON option always takes precedence over an environment variable.
 
 The fundamental structure of this JSON configuration string is given as follows:
 
@@ -11,18 +13,24 @@ The fundamental structure of this JSON configuration string is given as follows:
 
 This structure allows keeping one configuration string for several backends at once, with the concrete backend configuration being chosen upon choosing the backend itself.
 
-The configuration is currently read in a case-sensitive manner. Since this may change in the future, all parts of the configuration are given in *lower case*. Parameters that are directly passed through to an external library and not interpreted within the openPMD API (e.g. ``adios2.engine.parameters``) are unaffected by this and follow the respective library's conventions.
+The configuration is read in a case-sensitive manner.
+Generally, keys of the configuration are *lower case*.
+Parameters that are directly passed through to an external library and not interpreted within openPMD API (e.g. ``adios2.engine.parameters``) are unaffected by this and follow the respective library's conventions.
 
-The configuration string may refer to the complete ``openPMD::Series`` and additionally be specified per ``openPMD::Dataset``, passed in the respective constructors. *Configuration per dataset is currently not yet implemented.* This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
+The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
+*A configuration per dataset is currently not yet implemented.*
+This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
 
 For a consistent user interface, backends shall follow the following rules:
 
 * The configuration structures for the Series and for each dataset should be defined equivalently.
 * Any setting referring to single datasets should also be applicable globally, affecting all datasets.
 * If a setting is defined globally, but also for a concrete dataset, the dataset-specific setting should override the global one.
-* If a setting is passed to a dataset that only makes sense globally (such as the storage engine), the setting should be ignored except for printing a warning. Backends should define clearly which keys are applicable to datasets and which are not.
+* If a setting is passed to a dataset that only makes sense globally (such as the storage engine), the setting should be ignored except for printing a warning.
+  Backends should define clearly which keys are applicable to datasets and which are not.
 
-Configuration structure per backend
+
+Configuration Structure per Backend
 -----------------------------------
 
 ADIOS2
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -1,0 +1,47 @@
+.. _backendconfig
+
+Backend-specific configuration
+==============================
+
+While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is often necessary to pass configuration parameters to the concrete backend in use. Some backends recognize a number of configuration options via environment variables (see the backends' documentations), the preferred way to do so, however, *– as far as it is implemented –* is via a JSON-formatted string.
+
+The fundamental structure of this JSON configuration string is given as follows:
+
+.. literalinclude:: config_layout.json
+
+This structure allows keeping one configuration string for several backends at once, with the concrete backend configuration being chosen upon choosing the backend itself.
+
+The configuration is currently read in a case-sensitive manner. Since this may change in the future, all parts of the configuration are given in *lower case*. Parameters that are directly passed through to an external library and not interpreted within the openPMD API (e.g. ``adios2.engine.parameters``) are unaffected by this and follow the respective library's conventions.
+
+The configuration string may refer to the complete ``openPMD::Series`` and additionally be specified per ``openPMD::Dataset``, passed in the respective constructors. *Configuration per dataset is currently not yet implemented.* This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
+
+For a consistent user interface, backends shall follow the following rules:
+
+* The configuration structures for the Series and for each dataset should be defined equivalently.
+* Any setting referring to single datasets should also be applicable globally, affecting all datasets.
+* If a setting is defined globally, but also for a concrete dataset, the dataset-specific setting should override the global one.
+* If a setting is passed to a dataset that only makes sense globally (such as the storage engine), the setting should be ignored except for printing a warning. Backends should define clearly which keys are applicable to datasets and which are not.
+
+Configuration structure per backend
+-----------------------------------
+
+ADIOS2
+^^^^^^
+
+A full configuration of the ADIOS2 backend:
+
+.. literalinclude:: adios2.json
+
+All keys found under ``adios2.dataset`` are applicable globally as well as per dataset, keys found under ``adios2.engine`` only globally. Explanation of the single keys:
+
+* ``adios2.engine.type``: A string that is passed directly to ``adios2::IO:::SetEngine`` for choosing the ADIOS2 engine to be used. Please refer to the `official ADIOS2 documentation <https://adios2.readthedocs.io/en/latest/engines/engines.html>`_ for a list of available engines.
+* ``adios2.engine.type``: An associative array of string-formatted engine parameters, passed directly through to ``adios2::IO::SetParameters``. Please refer to the official ADIOS2 documentation for the allowable engine parameters.
+* ``adios2.dataset.operators``: (*currently unimplemented* – please use the ``openPMD::Dataset::compression`` for the meantime) This key contains a list of ADIOS2 `operators <https://adios2.readthedocs.io/en/latest/components/components.html#operator>`_, used to enable compression or dataset transformations. Each object in the list has three keys:
+
+  * ``type`` supported ADIOS operator type, e.g. zfp, sz
+  * ``parameters`` is an associative map of string parameters for the operator (e.g. compression levels)
+
+Other backends
+^^^^^^^^^^^^^^
+
+Do currently not read the configuration string. Please refer to the respective backends' documentations for further information on their configuration.
--- a/docs/source/details/config_layout.json
+++ b/docs/source/details/config_layout.json
@@ -1,5 +1,6 @@
 {
     "adios": "put ADIOS config here",
     "adios2": "put ADIOS2 config here",
-    "hdf5": "put HDF5 config here"
+    "hdf5": "put HDF5 config here",
+    "json": "put JSON config here"
 }
--- a/docs/source/details/config_layout.json
+++ b/docs/source/details/config_layout.json
@@ -1,0 +1,5 @@
+{
+    "adios": "put ADIOS config here",
+    "adios2": "put ADIOS2 config here",
+    "hdf5": "put HDF5 config here"
+}
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -187,16 +187,16 @@ public:
 private:
     adios2::ADIOS m_ADIOS;
 
-#   if openPMD_HAVE_JSON
-    nlohmann::json m_config;
-    static nlohmann::json nullvalue;
+#    if openPMD_HAVE_JSON
+    auxiliary::TracingJSON m_config;
+    static auxiliary::TracingJSON nullvalue;
 
     void
     init( nlohmann::json config );
 
     template< typename Key >
-    nlohmann::json &
-    config( Key && key, nlohmann::json & cfg )
+    auxiliary::TracingJSON
+    config( Key && key, auxiliary::TracingJSON & cfg )
     {
         if( cfg.is_object() && cfg.contains( key ) )
         {
@@ -209,7 +209,7 @@ private:
     }
 
     template< typename Key >
-    nlohmann::json &
+    auxiliary::TracingJSON
     config( Key && key )
     {
         return config< Key >( std::forward< Key >( key ), m_config );
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -96,13 +96,24 @@ public:
 
 #if openPMD_HAVE_MPI
 
-    ADIOS2IOHandlerImpl( AbstractIOHandler *, MPI_Comm, nlohmann::json config );
+    ADIOS2IOHandlerImpl(
+        AbstractIOHandler *,
+        MPI_Comm
+#   if openPMD_HAVE_JSON
+        , nlohmann::json config
+#   endif
+    );
 
     MPI_Comm m_comm;
 
 #endif // openPMD_HAVE_MPI
 
-    explicit ADIOS2IOHandlerImpl( AbstractIOHandler *, nlohmann::json config );
+    explicit ADIOS2IOHandlerImpl(
+        AbstractIOHandler *
+#   if openPMD_HAVE_JSON
+        , nlohmann::json config
+#   endif
+    );
 
 
     ~ADIOS2IOHandlerImpl( ) override = default;
@@ -175,6 +186,8 @@ public:
 
 private:
     adios2::ADIOS m_ADIOS;
+
+#   if openPMD_HAVE_JSON
     nlohmann::json m_config;
     static nlohmann::json nullvalue;
 
@@ -201,7 +214,7 @@ private:
     {
         return config< Key >( std::forward< Key >( key ), m_config );
     }
-
+#   endif // openPMD_HAVE_JSON
     /*
      * We need to give names to IO objects. These names are irrelevant
      * within this application, since:
@@ -705,13 +718,21 @@ public:
     ADIOS2IOHandler(
         std::string path,
         AccessType,
-        MPI_Comm,
-        nlohmann::json options );
+        MPI_Comm
+#   if openPMD_HAVE_JSON
+        , nlohmann::json options
+#   endif
+    );
 
 #endif
 
-    ADIOS2IOHandler( std::string path, AccessType,
-        nlohmann::json options);
+    ADIOS2IOHandler(
+        std::string path,
+        AccessType
+#if openPMD_HAVE_JSON
+        , nlohmann::json options
+#endif
+    );
 
     std::string backendName() const override { return "ADIOS2"; }
 
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -287,16 +287,22 @@ private:
                                          std::string const & var );
 }; // ADIOS2IOHandlerImpl
 
+/*
+ * The following strings are used during parsing of the JSON configuration
+ * string for the ADIOS2 backend.
+ * Note that this struct's members additionally need to be defined at namespace
+ * scope up until C++17, see ADIOS2IOHandlerImpl.cpp
+ */
+struct ADIOS2Defaults
+{
+    using const_str = char const * const;
+    constexpr static const_str str_engine = "engine";
+    constexpr static const_str str_type = "type";
+    constexpr static const_str str_params = "parameters";
+};
+
 namespace detail
 {
-    /*
-     * The following strings are used during parsing of the JSON configuration
-     * string for the ADIOS2 backend.
-     */
-    constexpr char const * const str_engine = "engine";
-    constexpr char const * str_type = "type";
-    constexpr char const * str_params = "parameters";
-
     // Helper structs for calls to the switchType function
 
     struct DatasetReader
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -96,13 +96,7 @@ public:
 
 #if openPMD_HAVE_MPI
 
-    ADIOS2IOHandlerImpl(
-        AbstractIOHandler *,
-        MPI_Comm
-#   if openPMD_HAVE_JSON
-        , nlohmann::json config
-#   endif
-    );
+    ADIOS2IOHandlerImpl( AbstractIOHandler *, MPI_Comm, nlohmann::json config );
 
     MPI_Comm m_comm;
 
@@ -110,15 +104,14 @@ public:
 
     explicit ADIOS2IOHandlerImpl(
         AbstractIOHandler *
-#   if openPMD_HAVE_JSON
         , nlohmann::json config
-#   endif
     );
 
 
-    ~ADIOS2IOHandlerImpl( ) override = default;
+    ~ADIOS2IOHandlerImpl() override = default;
 
-    std::future< void > flush( ) override;
+    std::future< void >
+    flush() override;
 
     void createFile( Writable *,
                      Parameter< Operation::CREATE_FILE > const & ) override;
@@ -187,7 +180,6 @@ public:
 private:
     adios2::ADIOS m_ADIOS;
 
-#    if openPMD_HAVE_JSON
     auxiliary::TracingJSON m_config;
     static auxiliary::TracingJSON nullvalue;
 
@@ -214,7 +206,6 @@ private:
     {
         return config< Key >( std::forward< Key >( key ), m_config );
     }
-#   endif // openPMD_HAVE_JSON
     /*
      * We need to give names to IO objects. These names are irrelevant
      * within this application, since:
@@ -719,20 +710,12 @@ public:
         std::string path,
         AccessType,
         MPI_Comm
-#   if openPMD_HAVE_JSON
         , nlohmann::json options
-#   endif
     );
 
 #endif
 
-    ADIOS2IOHandler(
-        std::string path,
-        AccessType
-#if openPMD_HAVE_JSON
-        , nlohmann::json options
-#endif
-    );
+    ADIOS2IOHandler( std::string path, AccessType, nlohmann::json options );
 
     std::string backendName() const override { return "ADIOS2"; }
 
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -180,6 +180,14 @@ public:
 private:
     adios2::ADIOS m_ADIOS;
 
+    struct ParameterizedOperator
+    {
+        adios2::Operator const op;
+        adios2::Params const params;
+    };
+
+    std::vector< ParameterizedOperator > defaultOperators;
+
     auxiliary::TracingJSON m_config;
     static auxiliary::TracingJSON nullvalue;
 
@@ -206,6 +214,21 @@ private:
     {
         return config< Key >( std::forward< Key >( key ), m_config );
     }
+
+    /**
+     *
+     * @param config The top-level of the ADIOS2 configuration JSON object
+     * with operators to be found under dataset.operators
+     * @return first parameter: the operators, second parameters: whether
+     * operators have been configured
+     */
+    std::pair< std::vector< ParameterizedOperator >, bool >
+    getOperators( auxiliary::TracingJSON config );
+
+    // use m_config
+    std::pair< std::vector< ParameterizedOperator >, bool >
+    getOperators();
+
     /*
      * We need to give names to IO objects. These names are irrelevant
      * within this application, since:
@@ -516,7 +539,8 @@ namespace detail
         defineVariable(
             adios2::IO & IO,
             std::string const & name,
-            std::unique_ptr< adios2::Operator > compression,
+            std::vector< ADIOS2IOHandlerImpl::ParameterizedOperator >
+                compressions,
             adios2::Dims const & shape = adios2::Dims(),
             adios2::Dims const & start = adios2::Dims(),
             adios2::Dims const & count = adios2::Dims(),
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -21,6 +21,7 @@
 #pragma once
 
 #include "ADIOS2FilePosition.hpp"
+#include "openPMD/config.hpp"
 #include "openPMD/IO/AbstractIOHandler.hpp"
 #include "openPMD/IO/AbstractIOHandlerImpl.hpp"
 #include "openPMD/IO/AbstractIOHandlerImplCommon.hpp"
@@ -28,22 +29,6 @@
 #include "openPMD/IO/InvalidatableFile.hpp"
 #include "openPMD/auxiliary/JSON.hpp"
 #include "openPMD/backend/Writable.hpp"
-
-#include <array>
-#include <exception>
-#include <future>
-#include <iostream>
-#include <memory> // shared_ptr
-#include <string>
-#include <unordered_map>
-#include <utility> // pair
-#include <vector>
-
-#include <nlohmann/json.hpp>
-
-#include "openPMD/config.hpp"
-
-#include <nlohmann/json.hpp>
 
 #if openPMD_HAVE_ADIOS2
 #   include <adios2.h>
@@ -54,6 +39,17 @@
 #   include <mpi.h>
 #endif
 
+#include <nlohmann/json.hpp>
+
+#include <array>
+#include <exception>
+#include <future>
+#include <iostream>
+#include <memory> // shared_ptr
+#include <string>
+#include <unordered_map>
+#include <utility> // pair
+#include <vector>
 
 namespace openPMD
 {
@@ -111,10 +107,9 @@ public:
     );
 
 
-    ~ADIOS2IOHandlerImpl() override = default;
+    ~ADIOS2IOHandlerImpl( ) override = default;
 
-    std::future< void >
-    flush() override;
+    std::future< void > flush( ) override;
 
     void createFile( Writable *,
                      Parameter< Operation::CREATE_FILE > const & ) override;
@@ -316,8 +311,6 @@ private:
 /*
  * The following strings are used during parsing of the JSON configuration
  * string for the ADIOS2 backend.
- * Note that this struct's members additionally need to be defined at namespace
- * scope up until C++17, see ADIOS2IOHandlerImpl.cpp
  */
 namespace ADIOS2Defaults
 {
@@ -757,8 +750,8 @@ public:
     ADIOS2IOHandler(
         std::string path,
         AccessType,
-        MPI_Comm
-        , nlohmann::json options
+        MPI_Comm,
+        nlohmann::json options
     );
 
 #endif
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -316,13 +316,13 @@ private:
  * Note that this struct's members additionally need to be defined at namespace
  * scope up until C++17, see ADIOS2IOHandlerImpl.cpp
  */
-struct ADIOS2Defaults
+namespace ADIOS2Defaults
 {
     using const_str = char const * const;
-    constexpr static const_str str_engine = "engine";
-    constexpr static const_str str_type = "type";
-    constexpr static const_str str_params = "parameters";
-};
+    constexpr const_str str_engine = "engine";
+    constexpr const_str str_type = "type";
+    constexpr const_str str_params = "parameters";
+}
 
 namespace detail
 {
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -375,17 +375,13 @@ namespace detail
 
     struct VariableDefiner
     {
+        // Parameters such as DatasetHelper< T >::defineVariable
+        template < typename T, typename... Params >
+        void operator( )( Params &&... params );
 
-        template < typename T >
-        void operator( )( adios2::IO & IO, const std::string & name,
-                          std::unique_ptr< adios2::Operator > compression,
-                          const adios2::Dims & shape = adios2::Dims( ),
-                          const adios2::Dims & start = adios2::Dims( ),
-                          const adios2::Dims & count = adios2::Dims( ),
-                          const bool constantDims = false );
-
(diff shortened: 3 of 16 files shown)


ax3l:
I updated the CI with further ADIOS2 tests in #568.
Please rebase against the current `dev` :)

include/openPMD/IO/AbstractIOHandlerHelper.hpp
```diff
@@ -36,26 +36,28 @@ namespace openPMD
...
-    );
-} // openPMD
+/** Construct an appropriate specific IOHandler for the desired IO mode.
+ *
+ * @param   path        Path to root folder for all operations associated with
+ * the desired handler.
+ * @param   accessType  AccessType describing desired operations and permissions
+ * of the desired handler.
+ * @param   format      Format describing the IO backend of the desired handler.
+ * @return  Smart pointer to created IOHandler.
```
ax3l, line 59: `options` param is still missing :)

include/openPMD/IO/AbstractIOHandlerHelper.hpp
```diff
@@ -36,26 +36,28 @@ namespace openPMD
      * @param   comm        MPI communicator used for IO.
      * @return  Smart pointer to created IOHandler.
```
ax3l, line 39: `options` param is still missing :)

include/openPMD/Series.hpp
```diff
@@ -226,19 +234,21 @@ class Series : public Attributable
     std::string name() const;
     /** Set the pattern for file names.
      *
-     * @param   name    String of the pattern for file names. Must include iteration regex <CODE>\%T</CODE> for fileBased data.
+     * @param   name    String of the pattern for file names. Must include
+     * iteration regex <CODE>\%T</CODE> for fileBased data.
```
ax3l: Maybe align this to beginning of description?

include/openPMD/Series.hpp
```diff
@@ -226,19 +234,21 @@ class Series : public Attributable
...
      */
-    void flush();
+    void
+    flush();
 
     Container< Iteration, uint64_t > iterations;
 
-OPENPMD_private:
-    struct ParsedInput;
+    OPENPMD_private : struct ParsedInput;
```
ax3l: this syntax confuses me, probably an auto-formatting issue?

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -18,66 +18,73 @@
  * and the GNU Lesser General Public License along with openPMD-api.
  * If not, see <http://www.gnu.org/licenses/>.
  */
-#include <openPMD/IO/ADIOS/ADIOS2IOHandler.hpp>
 #include "openPMD/IO/AbstractIOHandlerHelper.hpp"
-#include "openPMD/IO/DummyIOHandler.hpp"
+
+#include <openPMD/IO/ADIOS/ADIOS2IOHandler.hpp>
```
ax3l:
```suggestion
#include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
```
and no extra empty line

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -18,66 +18,73 @@
...
+
 #include "openPMD/IO/ADIOS/ADIOS1IOHandler.hpp"
 #include "openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp"
+#include "openPMD/IO/DummyIOHandler.hpp"
 #include "openPMD/IO/HDF5/HDF5IOHandler.hpp"
 #include "openPMD/IO/HDF5/ParallelHDF5IOHandler.hpp"
 #include "openPMD/IO/JSON/JSONIOHandler.hpp"
+#include <nlohmann/json.hpp>
```
ax3l, line 30: extra empty line

ax3l:
@franzpoeschel I rebased this for you.

ax3l:
We should decide if we want to make `<nlohmann/json.hpp>` a non-optional dependency, which looks potentially solid enough to me to make this call, or if we want to cleanly error out if a user tries to pass runtime options without having JSON available.

I will commit a patch for the latter for now, but it's not pretty.

ax3l:
I implemented the selection of the preferred `BP` backend via an environment variable here:
https://github.com/openPMD/openPMD-api/blob/731a4073bf5661da853001bb54ad13709a38c18d/src/Series.cpp#L868-L887

We can also overwrite that by a runtime option, like the other options you added.

Adding control to HDF5 for collective/independent IO, via `H5FD_MPIO_INDEPENDENT`, to address #554 would be a good thing to implement right away as well.

ax3l:
ping @franzpoeschel any thoughts on this? :)
Would like to make this available for the next release, so I can address the HDF5 control for independent I/O via a runtime API option as well.

franzpoeschel:
I'm fine with controlling the `.bp` backend with an environment variable. I think that we should add a parameter to the JSON string to explicitly choose backends that takes precedence over other options.
Making the other backends aware of this configuration style would be good (as you suggested for HDF5), I just don't have the time at the moment.
I think there is no issue with making `<nlohmann/json.jpp>` a non-optional dependency? It is currently not included in user-facing code, and this would also mean that the API will compile with at least one usable backend, simplifying accessibility.

ax3l:
> I'm fine with controlling the .bp backend with an environment variable. I think that we should add a parameter to the JSON string to explicitly choose backends that takes precedence over other options.

Agreed, the env var was a quick hack of mine but I want users to be able to control it via an option, which is more friendly and which is why I try to merge this PR soon :)

> Making the other backends aware of this configuration style would be good (as you suggested for HDF5), I just don't have the time at the moment.

Argh, too bad. Can you potentially finish the ADIOS impl. with it and I take a look at the other backends?

> I think there is no issue with making <nlohmann/json.jpp> a non-optional dependency.

Excellent, agreed. Will do that in an independent PR then (#587).

Edit: branch updated and rebased so you can finish the ADIOS2 controls.

ax3l:
ping @franzpoeschel finishing this one would also be really nice, please :)

franzpoeschel:
My plan is to have topic-streaming go on top of this one, so this is part of the endeavours ;)

ax3l:
Sounds great, I already made the JSON backend required so you just need to rebase to continue.

ax3l:
@franzpoeschel just since I need this more urgently for HDF5 option control and compression support: when do you think you can continue to work on this?

franzpoeschel:
@ax3l My defense talk is next wednesday, I can fully focus on openPMD again after that :)

ax3l:
Wish you all the best for your defense, please feel free to send me your slides on my lbl e-mail if you want me to take a look. Fingers crossed! :crossed_fingers: :muscle:

franzpoeschel:
For testing purposes, it might be sensible to link the `nlohmann_json::nlohmann_json` target also with the testing targets, so tests that directly use that library can be written as well.

ax3l:
I admittedly do not want to introduce this as a public library dependency and would prefer to keep this a (cmake-lingo-)private dep.
The reason behind this is that this would otherwise complicate downstream builds that want to use openPMD-api (e.g. number of to-be-set include paths) for users without CMake toolchains. Also, I want to limit the number of (public) dependencies that have to compile when included in a `nvcc` TU, which is basically everything that goes through the public includes to our users.

But we can certainly provide an auxiliary target in CMake that can be optionally be used.

Anyway, in the tests I would prefer just to write the json options as strings (or nested init lists) for now, to guarantee that the above mentioned aspect works.

franzpoeschel:
The reason why I am suggesting this is because with the latest commit, I introduced a [JSON class](https://github.com/openPMD/openPMD-api/blob/5c0086a000c189fa96b66dc4b621e8b15200b8e6/include/openPMD/auxiliary/JSON.hpp) depending on `nlohmann::json` that will trace accesses (in order to give warnings for unused keys). With the current state of the build system, I have only been able to test this externally by including that class and also N. Lohmann's JSON library in another project. Since it's only for warnings and not "mission-critical", I could leave it at that – or do you have another suggestion?

ax3l:
Is it possible you did not rebase against the latest `dev` yet? I adjusted the build system already a few months back to make `nlohmann_json` a mandatory, private dependency of openPMD-api #587 

Notes:
- please rebase against the latest `dev` branch
- you don't need `#if openPMD_HAVE_JSON` guards anymore, the `<nlohmann/json.hpp>` include is now always guaranteed to be there by the build system, as long as you build the primary source files of the CMake `openPMD` target, which your auxiliary `JSON.cpp` file is part of.
- as you do already: include `openPMD/auxiliary/JSON.hpp` only in `.cpp` files of the openPMD library target and do not add this include to the facade header `include/openPMD/openPMD.hpp` to avoid a public dependency on `<nlohmann/json.hpp>` for users

ax3l:
@franzpoeschel when this is ready for review: please remove the `[WIP]` tag from the PR description.

Also: can you add a small test, e.g. using compression, setting an engine, etc.?

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
+.. _backendconfig
+
+Backend-specific configuration
```
ax3l: English titles:
```suggestion
Backend-Specific Configuration
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
...
+==============================
+
+While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is often necessary to pass configuration parameters to the concrete backend in use. Some backends recognize a number of configuration options via environment variables (see the backends' documentations), the preferred way to do so, however, *– as far as it is implemented –* is via a JSON-formatted string.
```
```suggestion
While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is sometimes useful to to pass configuration parameters to the specific backend in use.
:ref:`For each backend <backends-overview>`, configuration options can be passed via a JSON-formatted string or via environment variables.
A JSON option always takes precedence over an environment variable.
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
...
+
+The fundamental structure of this JSON configuration string is given as follows:
+
+.. literalinclude:: config_layout.json
+
+This structure allows keeping one configuration string for several backends at once, with the concrete backend configuration being chosen upon choosing the backend itself.
+
+The configuration is currently read in a case-sensitive manner. Since this may change in the future, all parts of the configuration are given in *lower case*. Parameters that are directly passed through to an external library and not interpreted within the openPMD API (e.g. ``adios2.engine.parameters``) are unaffected by this and follow the respective library's conventions.
```
ax3l: One sentence per line please:
```suggestion
The configuration is read in a case-sensitive manner.
Generally, keys of the configuration are *lower case*.
Parameters that are directly passed through to an external library and not interpreted within openPMD API (e.g. ``adios2.engine.parameters``) are unaffected by this and follow the respective library's conventions.
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
...
+
+The configuration string may refer to the complete ``openPMD::Series`` and additionally be specified per ``openPMD::Dataset``, passed in the respective constructors. *Configuration per dataset is currently not yet implemented.* This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
```
```suggestion
The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
*A configuration per dataset is currently not yet implemented.*
This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
...
+
+For a consistent user interface, backends shall follow the following rules:
+
+* The configuration structures for the Series and for each dataset should be defined equivalently.
+* Any setting referring to single datasets should also be applicable globally, affecting all datasets.
+* If a setting is defined globally, but also for a concrete dataset, the dataset-specific setting should override the global one.
+* If a setting is passed to a dataset that only makes sense globally (such as the storage engine), the setting should be ignored except for printing a warning. Backends should define clearly which keys are applicable to datasets and which are not.
```
```suggestion
* If a setting is passed to a dataset that only makes sense globally (such as the storage engine), the setting should be ignored except for printing a warning.
  Backends should define clearly which keys are applicable to datasets and which are not.
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Configuration structure per backend
```
```suggestion
Configuration Structure per Backend
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,47 @@
+.. _backendconfig
+
+Backend-specific configuration
+==============================
+
+While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is often necessary to pass configuration parameters to the concrete backend in use. Some backends recognize a number of configuration options via environment variables (see the backends' documentations), the preferred way to do so, however, *– as far as it is implemented –* is via a JSON-formatted string.
+
+The fundamental structure of this JSON configuration string is given as follows:
+
+.. literalinclude:: config_layout.json
```
ax3l, line 12: Follow-up PR: can we add a [schema](https://json-schema.org/) for this?

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -258,6 +298,14 @@ class ADIOS2IOHandlerImpl
 
 namespace detail
 {
+    /*
+     * The following strings are used during parsing of the JSON configuration
+     * string for the ADIOS2 backend.
+     */
+    constexpr char const * const str_engine = "engine";
```
ax3l: global vars: :scream: 

can we place them in a struct or so? e.g. `struct ADIOS2defaults { ... };`?

include/openPMD/auxiliary/JSON.hpp
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
+#include "openPMD/config.hpp"
+
+#if openPMD_HAVE_JSON
```
ax3l: see previous comments: `openPMD_HAVE_JSON` guards are not used in the code anymore.

include/openPMD/auxiliary/JSON.hpp
```diff
@@ -0,0 +1,116 @@
...
+            m_originalJSON,
+            m_shadow,
+            newPositionInOriginal,
+            newPositionInShadow,
+            traceFurther );
+    }
+} // namespace auxiliary
+} // namespace openPMD
+
+#endif // openPMD_HAVE_JSON
```
ax3l: missing EOF newline

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,18 +20,21 @@
...
+#include <iterator>
+#include <set>
+#include <string>
+
 #include "openPMD/Datatype.hpp"
 #include "openPMD/IO/ADIOS/ADIOS2FilePosition.hpp"
 #include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
 #include "openPMD/auxiliary/Environment.hpp"
 #include "openPMD/auxiliary/Filesystem.hpp"
+// #include "openPMD/auxiliary/JSON.hpp"
```
ax3l: dead code?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -58,25 +61,59 @@ namespace openPMD
...
+ADIOS2IOHandlerImpl::ADIOS2IOHandlerImpl(
+    AbstractIOHandler * handler
+#    if openPMD_HAVE_JSON
+    ,
+    nlohmann::json cfg
+#    endif // openPMD_HAVE_JSON
+    )
     : AbstractIOHandlerImplCommon( handler ), m_ADIOS{ ADIOS2_DEBUG_MODE }
 {
+#    if openPMD_HAVE_JSON
```
ax3l: see above

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -515,6 +551,10 @@ ADIOS2IOHandlerImpl::adios2Accesstype()
     }
 }
 
+#    if openPMD_HAVE_JSON
```
ax3l: see above

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1116,58 +1157,114 @@ namespace detail
...
+            m_engine->Close();
+        }
+    }
+
+    void
+    BufferedActions::configure_IO( ADIOS2IOHandlerImpl & impl )
+    {
+        (void)impl;
+        std::set< std::string > alreadyConfigured;
+#    if openPMD_HAVE_JSON
```
ax3l: see above

docs/source/details/backendconfig.rst
franzpoeschel: Probably a good idea, yeah

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: They are `constexpr`, so not really variables ;)
The alternative would be using `#define` or using literals directly in the code, I think this is the cleanest solution. Putting them in a struct is a good idea, will do that.

franzpoeschel:
In the [test for this PR](https://github.com/openPMD/openPMD-api/blob/28ebb36f64ec0d7f449db69dd0c9d5e7fbf11ab2/test/SerialIOTest.cpp#L2192), I have not been able to catch an expected exception thrown by ADIOS2, neither with `REQUIRE_THROWS`, nor with `try..catch`. Might this be related to [this issue](https://stackoverflow.com/questions/38878999/try-catch-doesnt-work-in-shared-library)? 
EDIT: Recompiling with `-shared-libgcc` did not change this, the exception is not caught. Output from `ctest`:
```
3: terminate called after throwing an instance of 'std::ios_base::failure[abi:cxx11]'
3:   what():  ERROR: couldn't read from file ../samples/jsonConfigured.bp, in call to POSIX IO read                                                                                           
3: : iostream error         
3:                                             
3: ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
3: SerialIOTests is a Catch v2.6.1 host application.                                           
3: Run with -? for options
3:   
3: -------------------------------------------------------------------------------            
3: serial_adios2_json_config
3: -------------------------------------------------------------------------------             
3: /home/franz/workspace/openPMD-api/test/SerialIOTest.cpp:2149                               
3: ...............................................................................             
3:   
3: /home/franz/workspace/openPMD-api/test/SerialIOTest.cpp:2149: FAILED:                      
3: due to a fatal error condition:                                                             
3:   SIGABRT - Abort (abnormal termination) signal                                            
3:                                                                                             
3: ===============================================================================            
3: test cases:  14 |  13 passed | 1 failed                                                     
3: assertions: 487 | 486 passed | 1 failed                                                     
3:                                                                                             
3/4 Test #3: Serial.SerialIO ..................Child aborted***Exception:   0.24 sec
```

I have deactivated the corresponding check for now.

guj:
Hi Franz,
It is due to ADIOS2IOHandler::flush( )
returns a future object.

Would future be needed here?

franzpoeschel:
Hi @guj,
`ADIOS2IOHandler::flush` returns a `std::future` since it overrides `AbstractIOHandlerImpl::flush` which also does, in order to support future backends that may work asynchronously. The ADIOS2 backend does not make use of that and returns the default-constructed `std::future` created in [`AbstractIOHandlerImpl.hpp`](https://github.com/openPMD/openPMD-api/blob/c16ffd22993027eef46281724c7fa920f7851ca0/include/openPMD/IO/AbstractIOHandlerImpl.hpp#L115).

If I understand you correctly, you are suggesting that the exception is thrown in asynchronous code (i.e. inside the `std::future`) which I verified is not the case. I also tried catching the exception directly upon opening the engine in [`detail::BufferedActions::getEngine`](https://github.com/openPMD/openPMD-api/blob/c16ffd22993027eef46281724c7fa920f7851ca0/src/IO/ADIOS/ADIOS2IOHandler.cpp#L1174) which did not work either.

I think I will simply rewrite the test to not rely on throwing and catching exceptions for now.

franzpoeschel:
I fixed the test now.

Does the CI currently have some issues?
```
HDF5-DIAG: Error detected in HDF5 (1.10.1) thread 0:

  #000: H5F.c line 586 in H5Fopen(): unable to open file

    major: File accessibilty

    minor: Unable to open file

  #001: H5Fint.c line 1236 in H5F_open(): unable to open file: time = Wed Mar  4 14:34:08 2020

, name = '../samples/git-sample/data00000100.h5', tent_flags = 1

    major: File accessibilty

    minor: Unable to open file

  #002: H5FD.c line 809 in H5FD_open(): open failed

    major: Virtual File Layer

  #003: H5FDsec2.c line 346 in H5FD_sec2_open(): unable to open file: name = '../samples/git-sample/data00000100.h5', errno = 13, error message = 'Permission denied', flags = 1, o_flags = 2

    major: File accessibilty

    minor: Unable to open file

git sample not accessible. ([HDF5] Failed to open HDF5 file ../samples/git-sample/data00000100.h5)
```
(https://travis-ci.org/openPMD/openPMD-api/jobs/658223474)

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -36,9 +39,15 @@ namespace openPMD
         std::string path,
         AccessType accessTypeBackend,
         Format format,
-        MPI_Comm comm
-    )
+        MPI_Comm comm,
+        std::string const & options )
     {
+#   if openPMD_HAVE_JSON
```
ax3l: we always have `openPMD_HAVE_JSON`

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,55 @@
...
+==============================
+
+While the openPMD API intends to be a backend-*independent* implementation of the openPMD standard, it is sometimes useful to pass configuration parameters to the specific backend in use.
+:ref:`For each backend <backends-overview>`, configuration options can be passed via a JSON-formatted string or via environment variables.
+A JSON option always takes precedence over an environment variable.
+
+The fundamental structure of this JSON configuration string is given as follows:
+
+.. literalinclude:: config_layout.json
+
```
ax3l, line 14: Not sure if this will be already nicely highlighted, otherwise add:
```suggestion
   :language: json
   
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,55 @@
+.. _backendconfig
+
+Backend-Specific configuration
```
ax3l: Titles:
```suggestion
Backend-Specific Configuration
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,55 @@
...
+Configuration Structure per Backend
+-----------------------------------
+
+ADIOS2
+^^^^^^
+
+A full configuration of the ADIOS2 backend:
+
+.. literalinclude:: adios2.json
+
```
ax3l, line 44: See above, if needed:
```suggestion
   :language: json
   
```

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,55 @@
...
+All keys found under ``adios2.dataset`` are applicable globally as well as per dataset, keys found under ``adios2.engine`` only globally. Explanation of the single keys:
```
ax3l: One sentence per line:
```suggestion
All keys found under ``adios2.dataset`` are applicable globally as well as per dataset, keys found under ``adios2.engine`` only globally.
Explanation of the single keys:
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -48,6 +38,17 @@
...
 #endif
 
+#include <array>
+#include <future>
+#include <memory> // shared_ptr
+#include <string>
+#include <utility> // pair
+#include <vector>
+
+#include "openPMD/auxiliary/JSON.hpp"
```
ax3l: this should be in the first block of includes

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -48,6 +38,17 @@
...
+#include <nlohmann/json.hpp>
```
ax3l: this should be before the std lib includes

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -256,6 +308,20 @@ class ADIOS2IOHandlerImpl
                                          std::string const & var );
 }; // ADIOS2IOHandlerImpl
 
+/*
+ * The following strings are used during parsing of the JSON configuration
+ * string for the ADIOS2 backend.
+ * Note that this struct's members additionally need to be defined at namespace
+ * scope up until C++17, see ADIOS2IOHandlerImpl.cpp
+ */
+struct ADIOS2Defaults
```
ax3l: Is there an advantage by doing this in a struct with `static` members over just taking a `namespace` with the same name and avoiding all the `static` handling?

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -472,10 +534,15 @@ namespace detail
                           std::string const & fileName );
 
         static void
-        defineVariable( adios2::IO & IO, const std::string & name,
-                        std::unique_ptr< adios2::Operator > compression,
-                        const adios2::Dims & shape, const adios2::Dims & start,
-                        const adios2::Dims & count, bool constantDims );
+        defineVariable(
```
ax3l, line 550: Maybe add a doxygen parameter description here?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -990,11 +1098,12 @@ namespace detail
...
         }
-        // check whether the unique_ptr has an element
-        // and whether the held operator is valid
-        if ( compression && *compression )
+        for( auto const & compression : compressions )
         {
-            var.AddOperation( *compression );
+            if( compression.op )
+            {
+                var.AddOperation( compression.op, compression.params );
```
ax3l, line 1093: Just to make sure this works:
How would I set parameters of blosc, e.g.
```
    // add operations
    adios2::Operator BloscOp =
        adios.DefineOperator("BloscCompressor", adios2::ops::LosslessBlosc);

    varBig.AddOperation(BloscOp,
                         {{adios2::ops::blosc::key::compressor, "zstd"}});
    varBig.AddOperation(BloscOp,
                         {{adios2::ops::blosc::key::clevel, "1"}});
    varBig.AddOperation(BloscOp,
                         {{adios2::ops::blosc::key::doshuffle, "BLOSC_BITSHUFFLE"}});
```

via json?

guj:
> Hi @guj,
> `ADIOS2IOHandler::flush` returns a `std::future` since it overrides `AbstractIOHandlerImpl::flush` which also does, in order to support future backends that may work asynchronously. The ADIOS2 backend does not make use of that and returns the default-constructed `std::future` created in [`AbstractIOHandlerImpl.hpp`](https://github.com/openPMD/openPMD-api/blob/c16ffd22993027eef46281724c7fa920f7851ca0/include/openPMD/IO/AbstractIOHandlerImpl.hpp#L115).
> 
> If I understand you correctly, you are suggesting that the exception is thrown in asynchronous code (i.e. inside the `std::future`) which I verified is not the case. I also tried catching the exception directly upon opening the engine in [`detail::BufferedActions::getEngine`](https://github.com/openPMD/openPMD-api/blob/c16ffd22993027eef46281724c7fa920f7851ca0/src/IO/ADIOS/ADIOS2IOHandler.cpp#L1174) which did not work either.
> 
> I think I will simply rewrite the test to not rely on throwing and catching exceptions for now.

Hi Franz, yes you are right std::future is not the cause of this.  Especially ADIOS2IOHandler is not really doing async operation. Thanks for double checking.  I took a closer look today. The issue is still related to the flush() function.   It turns out when exception is thrown in flush(), it delegates and eventually the smart pointers are cleaning up the memory. When the ADIOS2IOHandler destructor is called,  the flush() is invoked again.  This time the same exception is thrown again and no where to go. One quick workaround is to do a try and catch in the destructor of ADIOS2IOHandler.  I did a quick test(reading a series from non existing bp file) with this workaround  and the exception from adios2 is then caught by my test.  Hope this makes sense.

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: I updated [the test](https://github.com/openPMD/openPMD-api/blob/f21074558c308398e6dda2cca3507a423db95ee0/test/SerialIOTest.cpp#L2166) accordingly. I skipped setting the compressor to `zstd` since ADIOS2 will fail if that is not available.

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: Yeah, using a namespace is better. Fixed it.

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: done

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: Is it [better now](https://github.com/openPMD/openPMD-api/blob/61be5f724229a056da50c1959519016eb3f7fbc2/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp#L23)?

docs/source/details/backendconfig.rst
franzpoeschel: I already applied the same fix to the documentation of the JSON backend.

src/IO/AbstractIOHandlerHelper.cpp
franzpoeschel: All traces of this should be removed from the source tree now, according to `grep`.

ax3l:
CI is up again, ready for rebase against `dev`.

ax3l:
@franzpoeschel looks like this PR reproducibly breaks the `clang@5.0.0 +OMPI -PY +H5 +ADIOS1 (BP3) +ADIOS2` HDF5 tests.

franzpoeschel:
Yep. I still need to figure out what's up with that.

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
ax3l: Not 100% yet. I would
- prefix `ADIOS2FilePosition.hpp` with the usual full path
- include `openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp` before third party libs

```C++
#include "openPMD/config.hpp"
#include "openPMD/IO/AbstractIOHandler.hpp"
#include "openPMD/IO/AbstractIOHandlerImpl.hpp"
#include "openPMD/IO/AbstractIOHandlerImplCommon.hpp"
#include "openPMD/IO/IOTask.hpp"
#include "openPMD/IO/InvalidatableFile.hpp"
#include "openPMD/IO/ADIOS/ADIOS2FilePosition.hpp"
#include "openPMD/auxiliary/JSON.hpp"
#include "openPMD/backend/Writable.hpp"

#if openPMD_HAVE_ADIOS2
#   include "openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp"
#   include <adios2.h>
#endif

#include <nlohmann/json.hpp>

#if openPMD_HAVE_MPI
#   include <mpi.h>
#endif

#include <array>
// ...
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l: Okay... but it's our default compressor we use all the time in production. Maybe add it in some way to tests nonetheless...

@chuckatkins @guj is it possible to query the ADIOS2 API at runtime to be sure that e.g. blosc was compiled in?

ax3l:
Re: ADIOS2 exception problem.
This is confusing, since I am able to catch shared library exceptions of openPMD-api (shared) in the python bindings (also shared)...
How did you build ADIOS2 and which version? Manually or is this what you observed both locally and in CI? If you can isolate the few lines for a minimal reproducer, we can ask upstream with ADIOS2 as well because its likely build system related.

guj:
@ax3l 
I am using shared library,  the master branch (I don't think branch matters). 
Try this simple test in cpp. See if it crashes (It should.) 
Next you put try/catch block, see if it still crashes (will)  

{
    MPI_Init(&argc, &argv);
    Series series = Series("no_such_file.bp",  AccessType::READ_ONLY,  MPI_COMM_WORLD);        
}

ax3l:
I think the full issue we talk about is more
```C++

int main()
{
    try
    {
        auto series = openPMD::Series(
            "test.bp",
            openPMD::AccessType::READ_ONLY
        );
    } catch( ... ) {}
    return 0;
}
```

which works with H5 (intentional verbose output but no crash) and fails with ADIOS2 (core dump due to missing catch). But I get the point, good demonstrator :)

src/IO/ADIOS/ADIOS2IOHandler.cpp
guj: Hi Axel,  Chuck will add it in.

guj:
> I think the full issue we talk about is more
> 
> ```c++
> #include <openPMD/openPMD.hpp>
> 
> int main()
> {
>     try
>     {
>         auto series = openPMD::Series(
>             "test.bp",
>             openPMD::AccessType::READ_ONLY
>         );
>     } catch( ... ) {}
>     return 0;
> }
> ```
> 
> which works with H5 (intentional verbose output but no crash) and fails with ADIOS2 (core dump due to missing catch). But I get the point, good demonstrator :)

Actually, my full argument is (assuming the bp file does not exist) 
1)  Series::Init(),  eventually calls ADIOS2IOHandler::flush()  from readBase(), around line  Series.cpp 782
2)  it calls m_impl.flush().  Exception is thrown. Msg looks like "ERROR: couldn't open file ...". And  currently this exception is not caught.  
3) plus, ADIOS2IOHandler.hpp:647, flush() is called again in the destructor when shared pointer releases resources when test ends. The same exception in 2) will be thrown again. This will not be caught in the test, so catch block is needed in either ADIOS2IOHandler/Impl. I would just add it in the destructor for a quick workaround.

ax3l:
Ups, clicked the wrong button.

Ah, the problem is in flush! of course! The destructor must not throw, we have to indeed catch in `Series::~Series`. Will do a PR.

franzpoeschel:
[This line](https://github.com/openPMD/openPMD-api/pull/569/commits/e7022771b7fe0c4300b4291251918a392a99b6cd) seems to be the cause for the broken build. Apparently in that build configuration, the BP4 engine will not write a directory? It can still be opened without any issue (the file will be read from in a later line).
Is there some way to inspect the files created by the CI builds?

ax3l:
> This line seems to be the cause for the broken build. Apparently in that build configuration, the BP4 engine will not write a directory?

Careful, there is an ADIOS2 bug that does not write an ADIOS2 directory if there is already an old ADIOS1 `.bp` file (BP3) with the same name in place. When testing locally, better disable ADIOS1 to be definitely sure. in CI, we only execute one or the other. I think we have not yet reported this upstream...

>  Is there some way to inspect the files created by the CI builds?

not really, besides adding some temporary `ls`/`cat`/`find` lines to `.travis.yml`

franzpoeschel:
From that bug description, this shouldn't really happen during CI builds then?

franzpoeschel:
I think I've found the issue. The test that broke explicitly set `OPENPMD_BP_BACKEND=ADIOS1`, so the JSON string was completely ignored.
EDIT: After fixing, that test passes now. `gcc@6.5.0 -MPI +PY@3.5 +H5 +ADIOS1@1.13.1` fails during `apt-get install`. Should I trigger a new build?

ax3l:
Awesome!

> gcc@6.5.0 -MPI +PY@3.5 +H5 +ADIOS1@1.13.1 fails during apt-get install. Should I trigger a new build?

I just re-scheduled on travis the one build that failed (and the coverage test in the stage behind it). Sometimes fetch/install fails on travis due to network issues.

docs/source/details/backendconfig.rst
```diff
@@ -0,0 +1,62 @@
...
+* Any setting referring to single datasets should also be applicable globally, affecting all datasets.
+* If a setting is defined globally, but also for a concrete dataset, the dataset-specific setting should override the global one.
+* If a setting is passed to a dataset that only makes sense globally (such as the storage engine), the setting should be ignored except for printing a warning.
+  Backends should define clearly which keys are applicable to datasets and which are not.
+
+
+Configuration Structure per Backend
+-----------------------------------
+
+ADIOS2
```
ax3l, line 37: Generally, and I think this can be a follow-up PR as well, I think these backend-specific options should go into the respective backend docs page (where we also document the current environment vars).

docs/source/details/config_layout.json
```diff
@@ -0,0 +1,5 @@
+{
+    "adios": "put ADIOS config here",
+    "adios2": "put ADIOS2 config here",
+    "hdf5": "put HDF5 config here"
```
ax3l: Could be relevant, because nlohmann-json also supports a couple of binary JSON representations that we might want to use:
```suggestion
    "hdf5": "put HDF5 config here",
    "json": "put JSON config here"
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -40,6 +39,11 @@
 #include <utility> // pair
 #include <vector>
 
+#include <nlohmann/json.hpp>
```
ax3l: - `"openPMD/config.hpp"` should be in the first block of includes :)
- json: somehow included 2x :)
- stdlibs: should be included last

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -95,18 +99,22 @@ class ADIOS2IOHandlerImpl
...
+        , nlohmann::json config
+    );
 
 
-    ~ADIOS2IOHandlerImpl( ) override = default;
+    ~ADIOS2IOHandlerImpl() override = default;
 
-    std::future< void > flush( ) override;
+    std::future< void >
+    flush() override;
```
ax3l: this block has unnecessary whitespace changes and somehow an odd linebreak in the constructor parameters

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -258,6 +313,20 @@ class ADIOS2IOHandlerImpl
                                          std::string const & var );
 }; // ADIOS2IOHandlerImpl
 
+/*
+ * The following strings are used during parsing of the JSON configuration
+ * string for the ADIOS2 backend.
+ * Note that this struct's members additionally need to be defined at namespace
+ * scope up until C++17, see ADIOS2IOHandlerImpl.cpp
+ */
```
ax3l, line 321: Is this `Note` outdated?

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -258,6 +313,20 @@ class ADIOS2IOHandlerImpl
                                          std::string const & var );
 }; // ADIOS2IOHandlerImpl
 
+/*
+ * The following strings are used during parsing of the JSON configuration
+ * string for the ADIOS2 backend.
+ * Note that this struct's members additionally need to be defined at namespace
+ * scope up until C++17, see ADIOS2IOHandlerImpl.cpp
```
ax3l: Is this `Note` outdated?

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -666,11 +754,16 @@ friend class ADIOS2IOHandlerImpl;
 
 #if openPMD_HAVE_MPI
 
-    ADIOS2IOHandler( std::string path, AccessType, MPI_Comm );
+    ADIOS2IOHandler(
+        std::string path,
+        AccessType,
+        MPI_Comm
+        , nlohmann::json options
```
```suggestion
        MPI_Comm,
        nlohmann::json options
```

include/openPMD/auxiliary/JSON.hpp
```diff
@@ -0,0 +1,112 @@
+#pragma once
+
+#include <memory>  // std::shared_ptr
+#include <utility> // std::forward
```
ax3l: Include order:
- own header (in `.cpp`)
- openPMD-api headers
- third party libs
- std libs

include/openPMD/auxiliary/JSON.hpp
```diff
@@ -0,0 +1,112 @@
...
+         */
+        void
+        declareFullyRead();
+
+    private:
+        std::shared_ptr< nlohmann::json > m_originalJSON;
+        std::shared_ptr< nlohmann::json > m_shadow;
+        nlohmann::json * m_positionInOriginal;
+        nlohmann::json * m_positionInShadow;
+        bool m_trace = true;
```
ax3l: Please add `//! member variable docs` for these private members

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -20,18 +20,20 @@
  */
 
 #include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
+
+#include <algorithm>
+#include <iostream>
+#include <iterator>
+#include <set>
+#include <string>
```
ax3l, line 28: Stdlib includes last please

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l: @guj do we already have an issue or PR to link here, for reference?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1315,41 +1464,55 @@ namespace detail
...
+#    if openPMD_HAVE_MPI
 
-                                                    }
+ADIOS2IOHandler::ADIOS2IOHandler(
+    std::string path,
+    openPMD::AccessType at,
+    MPI_Comm comm,
+    nlohmann::json options )
+    : AbstractIOHandler( std::move( path ), at, comm )
+    , m_impl{ this, comm, std::move( options ) }
```
```suggestion
    : AbstractIOHandler( std::move( path ), at, comm ),
      m_impl{ this, comm, std::move( options ) }
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1315,41 +1464,55 @@ namespace detail
...
 #endif
 
-ADIOS2IOHandler::ADIOS2IOHandler( std::string path, AccessType at )
-: AbstractIOHandler( std::move( path ), at ), m_impl{this}
+ADIOS2IOHandler::ADIOS2IOHandler(
+    std::string path,
+    AccessType at,
+    nlohmann::json options )
+    : AbstractIOHandler( std::move( path ), at )
+    , m_impl{ this, std::move( options ) }
```
```suggestion
    : AbstractIOHandler( std::move( path ), at ),
      m_impl{ this, std::move( options ) }
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1315,41 +1464,55 @@ namespace detail
...
-#if openPMD_HAVE_MPI
-ADIOS2IOHandler::ADIOS2IOHandler( std::string path, AccessType at,
-                                  MPI_Comm comm )
-: AbstractIOHandler( std::move( path ), at, comm )
+#    if openPMD_HAVE_MPI
+ADIOS2IOHandler::ADIOS2IOHandler(
+    std::string path,
+    AccessType at,
+    MPI_Comm comm
+    , nlohmann::json
```
```suggestion
    MPI_Comm comm,
    nlohmann::json
```

src/auxiliary/JSON.cpp
```diff
@@ -0,0 +1,94 @@
+#include "openPMD/auxiliary/JSON.hpp"
+
+#include <vector>
+
+#include "openPMD/config.hpp"
```
ax3l:
```suggestion
#include "openPMD/auxiliary/JSON.hpp"
#include "openPMD/config.hpp"

#include <vector>
```

ax3l: and missing copyright header

src/auxiliary/JSON.cpp
```diff
@@ -0,0 +1,94 @@
...
+    TracingJSON::TracingJSON() : TracingJSON( nlohmann::json() )
+    {
+    }
+
+    TracingJSON::TracingJSON( nlohmann::json originalJSON )
+        : m_originalJSON(
+              std::make_shared< nlohmann::json >( std::move( originalJSON ) ) )
+        , m_shadow( std::make_shared< nlohmann::json >() )
+        , m_positionInOriginal( &*m_originalJSON )
+        , m_positionInShadow( &*m_shadow )
```
ax3l:
```suggestion
        : m_originalJSON(
              std::make_shared< nlohmann::json >( std::move( originalJSON ) ) ),
          m_shadow( std::make_shared< nlohmann::json >() ),
          m_positionInOriginal( &*m_originalJSON ),
          m_positionInShadow( &*m_shadow )
```

src/auxiliary/JSON.cpp
```diff
@@ -0,0 +1,94 @@
...
+    {
+        nlohmann::json inverted = *m_positionInOriginal;
+        invertShadow( inverted, *m_positionInShadow );
+        return inverted;
+    }
+
+    void
+    TracingJSON::invertShadow(
+        nlohmann::json & result,
+        nlohmann::json & shadow )
```
ax3l: is `shadow` a `const &`?

src/auxiliary/JSON.cpp
```diff
@@ -0,0 +1,94 @@
...
+        std::shared_ptr< nlohmann::json > shadow,
+        nlohmann::json * positionInOriginal,
+        nlohmann::json * positionInShadow,
+        bool trace )
+        : m_originalJSON( std::move( originalJSON ) )
+        , m_shadow( std::move( shadow ) )
+        , m_positionInOriginal( positionInOriginal )
+        , m_positionInShadow( positionInShadow )
+        , m_trace( trace )
+    {
```
ax3l:
```suggestion
        : m_originalJSON( std::move( originalJSON ) ),
          m_shadow( std::move( shadow ) ),
          m_positionInOriginal( positionInOriginal ),
          m_positionInShadow( positionInShadow ),
          m_trace( trace )
    {
```

test/SerialIOTest.cpp
```diff
@@ -3,23 +3,23 @@
...
 #include <numeric>
+#include <sstream>
+#include <string>
 #include <tuple>
+#include <vector>
+
+#include "openPMD/auxiliary/Filesystem.hpp"
+#include "openPMD/auxiliary/Environment.hpp"
+#include "openPMD/openPMD.hpp"
+#include <catch2/catch.hpp>
```
ax3l: include order is off

test/SerialIOTest.cpp
```diff
@@ -2320,3 +2320,119 @@ TEST_CASE( "no_serial_adios1", "[serial][adios]")
...
+{
+  "adios2": {
+    "engine": {
+      "type": "bp4",
+      "unused": "parameter"
+    }
+  }
+}
+)END";
+    auto read = []( std::string const & filename, std::string const & config ) {
```
```suggestion
    auto const read = []( std::string const & filename, std::string const & config ) {
```

src/auxiliary/JSON.cpp
franzpoeschel: Yes, this method mustn't modify `shadow`. I added the `const`.

docs/source/details/backendconfig.rst
ax3l: @franzpoeschel please try to split this up in a follow-up PR, we already have backend-specific docs where this should go :)

ax3l (approved):

PR #570, opened 2019-10-10
Streaming Support

franzpoeschel:
Add support for streaming IO to the openPMD API, including frontend semantics and a backend implementation for the ADIOS2 IO backend.

> The basic idea is to split the former concept of `flush`ing a Series into two separate concepts:
> * `flush`: frontend guarantees, application-side buffers may be used after this call returns
> * `advance`: backend guarantees, data has moved to/from the backend storage after this call returns. In a streaming context, this means that data packets are sent/received.

This concept is now purely internal, with a much simpler API exposed to users. The API is documented in [streaming.rst](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/docs/source/usage/streaming.rst). It is more restrictive than the traditional random-access based API and hence more generic, being usable for file-based as well as for streaming-based workflows.

Review entry points:
* Streaming user-api: `Series.hpp` classes `ReadIterations` and `WriteIterations`
* new IO tasks
* `Iteration::beginStep()/endStep()/close()`

Major additions include:
* ~~Additional functionality to deal with partial datasets in [Dataset.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/Dataset.hpp) and [RecordComponent.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/RecordComponent.hpp)~~ Since the high-level API now correlates streaming steps with iterations, single datasets are expected to be fully available or not available at all. Finer-grained accessing methods have hence been removed again.
* Implementation of streaming mode for the [ADIOS2 backend](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp). This does currently not include a Read-Write mode as provided by ADIOS2.
* Four new [IO tasks](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/IO/IOTask.hpp): Advance, ~~inquiring available chunks, closing files (see below for a reason why this is not a separate PR)~~ and closing paths. The latter command indicates that this group will not be written to any longer, allowing for optimizations in the backend.
  **Update:** Closing files is already merged separately, inquiring available chunks extracted to PR #802.~~
> * The `Series::advance` method as described above. Just like `flush`, this returns a `std::future`, but for the first time in the openPMD API, those futures are actually used. If discarding them, the destructor of `std::future` will immediately run, blocking until the future returns. If storing them, the result can be awaited at a later time, continuing with computation asynchronously. This currently makes sense mostly on the reader end, where the future will block until a next data packet has been received.
> CAUTION: As described in  #534, the `Series` object should currently not be copied due to unclear semantics, the the returned future wraps a **reference to** the `Series`. Deleting the Series before the returned future will result in use after free.
> Since the class `std::future` in the standard library is rather limited at the moment, this includes an implementation of an [extended future class](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/auxiliary/Future.hpp) that allows to chain futures.
> TODO: "Advance" tasks are currently only passed through to the backend if choosing group-based layouts. In file-based layouts, this currently falls back to flushing the Series.

**Update:** Now internal, not part of the public API.
> * Additionally to the random-access based chunk loading procedures and their aforementioned extensions, functionality to automatically distribute incoming chunks in [Streaming.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/Streaming.hpp). This obviously requires that reading applications can deal with having chunks arranged by the IO library. The strategies are flexibly interchangeable. They receive as input meta information about the MPI ranks of writers and readers, e.g. the hostnames. [MPI.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/auxiliary/MPI.hpp) contains functionality to collect this meta information, the writer can then set its meta information as a top-level attribute in [Series.hpp](https://github.com/franzpoeschel/openPMD-api/blob/3fb8cb216e94d7817a441ad3d87eab992014aea7/include/openPMD/Series.hpp#L222).

**Update:** Not part of this PR, will either come as a later PR or as part of external tooling.
> * A hopefully easy-to-use [class](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/benchmark/utils/DumpTimes.hpp) to dump times to a file. Can be deactivated at compile time via a template parameter.

**Update:** Removed.

* Adaptations for re-parsing a Series after receiving a new packet.
* **Update:** A high-level iterator-based Streaming API for reading (`ReadIterations` class) and a high-level container-based Streaming-API for writing (`WriteIterations` class)
* **Update:** Python bindings

TODO:
- [x] Document all of this
- [x] Add new examples for streaming
- [x] I mostly did ad-hoc testing so far, integrate those tests with the real ones.
- [x] Formatting and cleanup
- [x] Figure out why Windows tests hang – MSVC used the wrong method overload in one place, leading to an infinite recursion. Explicit typecasting solved the issue.

(I will be on holidays in the second half of october and likely without Internet during that time, just leave comments, but I might take a while to react :) )


ax3l:
I updated the CI with further ADIOS2 tests in #568.
Please rebase against the current `dev` :)

ax3l:
Hi @franzpoeschel, happy new year! :star2: 
Are you continuing on this PR this week? :)

franzpoeschel:
Hi, @ax3l, I will continue on thursday/friday, yes.
Happy new year to you, too!

franzpoeschel:
I have temporarily moved this to [topic-streaming-bak](https://github.com/franzpoeschel/openPMD-api/tree/topic-streaming-bak) to perform cleanup work on this branch.

ax3l:
Can we implement that a `Series::advance()` for regular file operations closes the opened file(s), e.g. via `H5Fclose()`? Currently files are kept open "too long" during simulations, which especially for HDF5 files increases the time frame on which corruption of the written data can occur. (And it's problematic for semi-live post-processing, too.)
Related to: https://github.com/ECP-WarpX/WarpX/issues/645

franzpoeschel:
I am currently planning to move `Series::advance()` to `Iteration::advance()` since that makes more sense for file-based layout (and maybe keep `Series::advance()` as a short-cut operation that advances all iterations). This being said, using `Series::advance()` to close a file would give different semantics to this call depending on the backend, making it potentially impossible to quickly switch backends. I have used `Iteration::finalize()` to pass a `CLOSE_FILE` command through to the backends and implemented that in the ADIOS2 backend already. 
Hence, in order to make use of `HF5Close()`, you should be fine with implementing [`AbstractIOHandlerImpl::closeFile()`](https://github.com/openPMD/openPMD-api/blob/aba435e8ad7e7a8a11dc4ce877ba9bea9f5463c0/include/openPMD/IO/AbstractIOHandlerImpl.hpp#L187) which is triggered by `Iteration::finalize()` independently of the backend.

ax3l:
`Iteration::finalize()` sounds legit, too.

franzpoeschel:
**Status Update**

> The basic idea is to split the former concept of `flush`ing a Series into two separate concepts:
> 
>     * `flush`: frontend guarantees, application-side buffers may be used after this call returns
> 
>     * `advance`: backend guarantees, data has moved to/from the backend storage after this call returns. In a streaming context, this means that data packets are sent/received.

This concept is now purely internal, with a much simpler API exposed to users. The API is documented in [streaming.rst](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/docs/source/usage/streaming.rst). It is more restrictive than the traditional random-access based API and hence more generic, being usable for file-based as well as for streaming-based workflows.
> 
> 
> Major additions include:
> 
>     * Additional functionality to deal with partial datasets in [Dataset.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/Dataset.hpp) and [RecordComponent.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/RecordComponent.hpp)

Since the high-level API now correlates streaming steps with iterations, single datasets are expected to be fully available or not available at all. Finer-grained accessing methods have hence been removed again.
> 
>     * Implementation of streaming mode for the [ADIOS2 backend](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp). This does currently not include a Read-Write mode as provided by ADIOS2.
> 
>     * Four new [IO tasks](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/IO/IOTask.hpp): Advance, inquiring available chunks, closing files (see below for a reason why this is not a separate PR) and putting an openPMD group to stale mode. The latter command indicates that this group will not be written to any longer, allowing for optimizations in the backend.

Inquiring available chunks is postponed to a later PR. "putting an openPMD group to stale mode" has been renamed "closing a group".
> 
>     * Declaring an iteration finalized from the frontend. […]

This has been merged already as `Iteration::close()`.
>     * The `Series::advance` method […]

Now internal, not part of the public API.
> 
>     * Additionally to the random-access based chunk loading procedures and their aforementioned extensions, functionality to automatically distribute incoming chunks in [Streaming.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/Streaming.hpp). This obviously requires that reading applications can deal with having chunks arranged by the IO library. The strategies are flexibly interchangeable. They receive as input meta information about the MPI ranks of writers and readers, e.g. the hostnames. [MPI.hpp](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/auxiliary/MPI.hpp) contains functionality to collect this meta information, the writer can then set its meta information as a top-level attribute in [Series.hpp](https://github.com/franzpoeschel/openPMD-api/blob/3fb8cb216e94d7817a441ad3d87eab992014aea7/include/openPMD/Series.hpp#L222).

Not part of this PR, will either come as a later PR or as part of external tooling.
> 
>     * A hopefully easy-to-use [class](https://github.com/franzpoeschel/openPMD-api/blob/topic-streaming/include/openPMD/benchmark/utils/DumpTimes.hpp) to dump times to a file. Can be deactivated at compile time via a template parameter.

Removed.

franzpoeschel:
The [hanging test](https://github.com/openPMD/openPMD-api/pull/570/checks?check_run_id=1183013046) might not be this PR's fault? Seeing the same issue on [other PRs](https://github.com/openPMD/openPMD-api/pull/570/checks?check_run_id=1183013046). This is the only remaining test that fails for this one.

ax3l:
Yes, the ASAN/UBSAN test is currently running into timeouts, since travis cuts it at 50min.

Trying to solve it, but it's tricky.

franzpoeschel:
I think we can start wrapping things up :) @ax3l

franzpoeschel:
@ax3l I added everything we talked about yesterday, so things should be ready.

.github/ci/sanitizer/clang/Leak.supp
```diff
@@ -11,3 +11,10 @@ leak:/usr/bin/make*
 # ADIOS 1.13.1
 leak:adios_read_bp_open_file
 leak:adios_inq_var
+# ADIOS2
```
ax3l, line 14: Are those still up-to-date?
@chuckatkins et al. added ASan/UBSan checks to ADIOS2 end of last year.

docs/source/backends/adios2.rst
```diff
@@ -13,13 +13,25 @@ For further information, check out the :ref:`installation guide <install>`,
...
+ADIOS2 has several engines for alternative file formats and other kinds of backends, yet natively writes to ``.bp`` files.
+The openPMD API uses the BP4 engine as the default file engine and the SST engine for streaming support.
 We currently leverage the default ADIOS2 transport parameters, i.e. ``POSIX`` on Unix systems and ``FStream`` on Windows.
 
+Steps
+-----
+
+ADIOS2 is optimized towards organizing the process of reading/writing data into IO steps.
+In order to activate steps, it is imperative to use the :ref:`Streaming API <usage-streaming>` (which can be used for either file-based or streaming-based workflows).
+With ADIOS2 releases 2.5.0 and 2.6.0 containing a bug (fixed in development versions) that disallows random-accessing steps in file-based engines, step-based processing must currently be opted in to via use of the :ref:`JSON parameter<backendconfig>` ``adios2.engine.usesteps = true`` when using a file-based engine such as BP3 or BP4.
```
ax3l: A few minor details to update here:
- don't mention 2.5.0 anymore since we only support 2.6.0+
- please link the pull request for `contains a bug` for context
- when does one have to set `adios2.engine.usesteps` as a work-around? During read of files if the write process was using SST?
- Should we turn on `adios2.engine.usesteps` by default (for streaming)?

examples/10_streaming_read.cpp
```diff
@@ -0,0 +1,63 @@
+#define BUILD_STREAMING_EXAMPLE false
+#if BUILD_STREAMING_EXAMPLE
```
ax3l, line 2: Is this intentionally untested, dead code?

examples/10_streaming_read.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
+import openpmd_api as io
+
+run_streaming_example = False
+
+if not run_streaming_example:
+    exit(0)
```
ax3l, line 7: Is this intentionally untested, dead code?

examples/10_streaming_write.cpp
```diff
@@ -0,0 +1,55 @@
+#define BUILD_STREAMING_EXAMPLE false
+#if BUILD_STREAMING_EXAMPLE
```
ax3l, line 2: Is this intentionally untested, dead code?

docs/source/usage/streaming.rst
```diff
@@ -0,0 +1,88 @@
+.. _usage-streaming:
+
+Streaming
+=========
+
+The openPMD API includes a streaming-aware API as well as streaming-enabled backends (currently: ADIOS2).
```
ax3l, line 10: Do we make guarantees that scientific production data written today will be readible with the next openPMD-api release when `sst`/streaming is used?

Otherwise, let's add a big warning, please
```suggestion
.. note::

   Data streaming is a novel backend and under active development.
   At the moment, the internal data format is still changing rapidly and is likely not compatible between releases of openPMD-api.

The openPMD API includes a streaming-aware API as well as streaming-enabled backends (currently: ADIOS2).
```

src/Series.cpp
```diff
@@ -748,7 +766,7 @@ Series::flushParticlesPath()
 }
 
 void
-Series::readFileBased()
+Series::readFileBased( bool __init )
```
ax3l: Please avoid naming with prefix `_`, especially `__` is reserved for the stdlib in C++.
https://stackoverflow.com/a/3651336/2719194

If the name collides with a method, consider calling the parameter `do_init`. If a member variable exists, consider renaming the member variable to `m_<name>`.

src/Series.cpp
```diff
@@ -834,54 +863,51 @@ Series::readFileBased()
 }
 
 void
-Series::readGroupBased()
+Series::readGroupBased( bool __init )
```
ax3l: See above, please rename the parameter.

src/Series.cpp
```diff
@@ -1067,4 +1189,193 @@ namespace
         }
     }
 } // namespace [anonymous]
+
+SeriesIterator::SeriesIterator() : m_series()
+{
+}
+
+SeriesIterator::SeriesIterator( Series * _series ) : m_series( _series )
```
ax3l: please avoid `_` here, too.

src/RecordComponent.cpp
```diff
@@ -186,7 +186,12 @@ RecordComponent::flush(std::string const& name)
 void
 RecordComponent::read()
 {
+    if ( *hasBeenRead )
```
ax3l, line 190: Do you want to use `Option`(al) for this?

src/Iteration.cpp
```diff
@@ -447,6 +437,119 @@ Iteration::read()
...
+        "[Iteration::close] Iteration not found in Series." );
+}
+
+
+StepStatus *
+Iteration::stepStatus()
+{
+    Series * s =
+        dynamic_cast< Series * >( parent->attributable->parent->attributable );
+    StepStatus * flag = nullptr;;
```
ax3l:
```suggestion
    StepStatus * flag = nullptr;
```

src/Iteration.cpp
```diff
@@ -447,6 +437,119 @@ Iteration::read()
...
+    throw std::runtime_error(
+        "[Iteration::close] Iteration not found in Series." );
+}
+
+
+StepStatus *
+Iteration::stepStatus()
+{
+    Series * s =
+        dynamic_cast< Series * >( parent->attributable->parent->attributable );
```
ax3l: This construct always makes me nervous: #789

See the method above for comparison, too.

src/Iteration.cpp
```diff
@@ -447,6 +437,119 @@ Iteration::read()
...
+    series.advance(
+        AdvanceMode::ENDSTEP, *file, myIteration< iterator_t >(), *this );
+}
+
+template< typename Iterator >
+Iterator
+Iteration::myIteration()
+{
+    /* Find the root point [Series] of this file,
+        * meshesPath and particlesPath are stored there */
```
ax3l:
```suggestion
    /* Find the root point [Series] of this file,
     * meshesPath and particlesPath are stored there */
```

src/Iteration.cpp
```diff
@@ -447,6 +437,119 @@ Iteration::read()
...
+Iteration::stepStatus()
+{
+    Series * s =
+        dynamic_cast< Series * >( parent->attributable->parent->attributable );
+    StepStatus * flag = nullptr;;
+    switch( *s->m_iterationEncoding )
+    {
+        using IE = IterationEncoding;
+        case IE::fileBased:
+            flag = &*this->m_stepStatus;
```
ax3l: Is this creating an unprotected pointer to a shared pointer's value and returning it? :-o
Can we return `StepStatus` by value? You could define an additional invalid value in it, if you need something for the `nullptr` indicator.

src/Iteration.cpp
```diff
@@ -447,6 +437,119 @@ Iteration::read()
...
+
+    return status;
+}
+
+void
+Iteration::endStep()
+{
+    using IE = IterationEncoding;
+    Series & series =
+        *dynamic_cast< Series * >( parent->attributable->parent->attributable );
```
ax3l: This construct always makes me nervous: #789

See the method below for comparison, too.

src/Iteration.cpp
```diff
@@ -112,47 +115,34 @@ Iteration::close( bool _flush )
...
+            *flag = StepStatus::NoStep;
         }
-        if( !found )
+        else
         {
-            throw std::runtime_error(
-                "[Iteration::close] Iteration not found in Series." );
+            // flush things manually
+            Series * s = &auxiliary::deref_dynamic_cast< Series >(
+                parent->attributable->parent->attributable );
```
ax3l, line 157: This construct always makes me nervous: #789

See the methods further below for comparison, too.

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -82,7 +85,10 @@ namespace openPMD
 #if openPMD_HAVE_ADIOS2
             case Format::ADIOS2:
                 return std::make_shared< ADIOS2IOHandler >(
-                    path, access, std::move( optionsJson ) );
+                    path, access, std::move( optionsJson ), "bp4" );
```
ax3l, line 88: I wonder if we should generally de-associate ADIOS2 with ".bp" and move to "bp4" == ADIOS2... Another issue/PR/discussion.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1244,20 +1364,61 @@ namespace detail
...
+        static std::set< std::string > streamingEngines = {
+            "sst", "insitumpi", "inline", "staging"
+        };
+        static std::set< std::string > fileEngines = {
+            "bp4", "bp3", "hdf5", "file"
+        };
+
+        // set engine type
+        {
+            // allow overriding through environment variable
```
ax3l, line 1536: Is the logic here: json option overwrites env var?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -592,6 +635,49 @@ void ADIOS2IOHandlerImpl::listAttributes(
...
+}
+
+void
+ADIOS2IOHandlerImpl::closePath(
+    Writable * writable,
+    Parameter< Operation::CLOSE_PATH > const & )
+{
+    VERIFY_ALWAYS(
+        writable->written,
+        "Cannot close a path that has not been written yet." );
```
```suggestion
        "[ADIOS2] Cannot close a path that has not been written yet." );
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -592,6 +635,49 @@ void ADIOS2IOHandlerImpl::listAttributes(
...
+    auto & fileData = getFileData( file );
+    if( !fileData.optimizeAttributesStreaming )
+    {
+        return;
+    }
+    auto position = setAndGetFilePosition( writable );
+    auto const positionString = filePositionToString( position );
+    VERIFY(
+        !auxiliary::ends_with( positionString, '/' ),
+        "ADIOS2 backend: Position string has unexpected format. This is a bug "
```
```suggestion
        "[ADIOS2] Position string has unexpected format. This is a bug "
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -77,8 +79,11 @@ ADIOS2IOHandlerImpl::ADIOS2IOHandlerImpl(
 
 ADIOS2IOHandlerImpl::ADIOS2IOHandlerImpl(
     AbstractIOHandler * handler,
-    nlohmann::json cfg )
-    : AbstractIOHandlerImplCommon( handler ), m_ADIOS{ ADIOS2_DEBUG_MODE }
+    nlohmann::json cfg,
+    std::string _engineType )
```
ax3l: Please dont `_` prefix (see other comments).

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -77,8 +79,11 @@ ADIOS2IOHandlerImpl::ADIOS2IOHandlerImpl(
...
+    : AbstractIOHandlerImplCommon( handler )
+    , m_ADIOS{ ADIOS2_DEBUG_MODE }
+    , engineType( std::move( _engineType ) )
```
ax3l: `engineType` -> `m_engineType` ?

include/openPMD/Streaming.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include <utility>
+#include <vector>
+
+#include "openPMD/Dataset.hpp"
+#include "openPMD/benchmark/mpi/BlockSlicer.hpp"
+#include <unordered_map>
+
+namespace openPMD
+{
+enum class AdvanceStatus : unsigned char
```
ax3l, line 20: Please document the purpose of each enum and each member in those enums with Doxygen. This is a good place to mention high-level concepts where they are used :)

include/openPMD/Series.hpp
```diff
@@ -282,5 +348,139 @@ class Series : public Attributable
...
+        SharedResources( iterations_t );
+        ~SharedResources();
+    };
+
+    using key_type = typename iterations_t::key_type;
+    using value_type = typename iterations_t::key_type;
+    WriteIterations( iterations_t );
+    explicit WriteIterations() = default;
+    // Index of the last opened iteration
+    std::shared_ptr< SharedResources > shared;
```
ax3l, line 504: Would `m_openIterationIndices` be a better name?

include/openPMD/Series.hpp
```diff
@@ -282,5 +348,139 @@ class Series : public Attributable
...
+
+        SharedResources( iterations_t );
+        ~SharedResources();
+    };
+
+    using key_type = typename iterations_t::key_type;
+    using value_type = typename iterations_t::key_type;
+    WriteIterations( iterations_t );
+    explicit WriteIterations() = default;
+    // Index of the last opened iteration
```
ax3l:
```suggestion
    //! Index of the last opened iteration
```

include/openPMD/Series.hpp
```diff
@@ -282,5 +348,139 @@ class Series : public Attributable
...
+ *        Designed to allow reading any kind of Series, streaming and non-
+ *        streaming alike. Calling Iteration::close() manually before opening
+ *        the next iteration is encouraged and will implicitly flush all
+ *        deferred IO actions. Otherwise, Iteration::close() will be implicitly
+ *        called upon SeriesIterator::operator++(), i.e. upon going to the next
+ *        iteration in the foreach loop.
+ *
+ *        Since this is designed for streaming mode, reopening an iteration is
+ *        not possible once it has been closed.
+ *
```
ax3l: Curious formatting, you can do title + description like this:
```suggestion
/** Writing side of the streaming API.
 *
 * Create instance via Series::writeIterations().
 * For use via WriteIterations::operator[]().
 * Designed to allow reading any kind of Series, streaming and non-
 * streaming alike. Calling Iteration::close() manually before opening
 * the next iteration is encouraged and will implicitly flush all
 * deferred IO actions. Otherwise, Iteration::close() will be implicitly
 * called upon SeriesIterator::operator++(), i.e. upon going to the next
 * iteration in the foreach loop.
 *
 * Since this is designed for streaming mode, reopening an iteration is
 * not possible once it has been closed.
```

include/openPMD/Series.hpp
```diff
@@ -282,5 +348,139 @@ class Series : public Attributable
...
+
+    bool
+    operator!=( SeriesIterator const & other ) const;
+
+    static SeriesIterator
+    end();
+};
+
+/**
+ * @brief Reading side of the streaming API. Create instance via
```
ax3l: See details below, please.

```suggestion
/** Reading side of the streaming API.
 *
 * Create instance via
```

include/openPMD/Series.hpp
```diff
@@ -282,5 +348,139 @@ class Series : public Attributable
...
+ *        Series::readIterations(). For use in a C++11-style foreach loop.
+ *        Designed to allow reading any kind of Series, streaming and non-
```
ax3l:
```suggestion
 *        Series::readIterations(). For use in a C++11-style foreach loop
 *        over iterations.
 *        Designed to allow reading any kind of Series, streaming and non-
```

include/openPMD/Series.hpp
```diff
@@ -282,5 +348,139 @@ class Series : public Attributable
...
+class SeriesIterator
+{
+    using iteration_index_t = IndexedIteration::index_t;
+
+    using maybe_series_t = auxiliary::Option< Series * >;
+
+    maybe_series_t m_series;
+    iteration_index_t m_currentIteration = 0;
+
+    // construct the end() iterator
```
ax3l:
```suggestion
    //! construct the end() iterator
```

include/openPMD/Series.hpp
```diff
@@ -255,23 +262,82 @@ class Series : public Attributable
...
+     */
+    void readFileBased( bool init = true );
+    void readGroupBased( bool init = true );
     void readBase();
     void read();
+    std::string iterationFilename( uint64_t i );
+
+    /**
+     * @brief In step-based IO mode, begin or end an IO step for the given
+     *        iteration.
```
ax3l, line 336: Can you please add who is calling this - reader, writer or both?

include/openPMD/Series.hpp
```diff
@@ -255,23 +262,82 @@ class Series : public Attributable
...
+    std::future< void > flush_impl(
+        iterations_iterator begin, iterations_iterator end );
+    void flushFileBased( iterations_iterator begin, iterations_iterator end );
+    void flushGroupBased( iterations_iterator begin, iterations_iterator end );
     void flushMeshesPath();
     void flushParticlesPath();
-    void readFileBased();
-    void readGroupBased();
+    /*
+     * Note:
```
ax3l: Can you please note with doxygen groups for which functions this note applies?

include/openPMD/Series.hpp
```diff
@@ -255,23 +262,82 @@ class Series : public Attributable
      */
     void flush();
 
+    /**
```
ax3l, line 268: For both functions, please add around a sentence on what it does before re-directing.
This is the direct, user-facing API call after all.

include/openPMD/Series.hpp
```diff
@@ -20,22 +20,25 @@
...
+#include "openPMD/backend/Container.hpp"
+#include "openPMD/config.hpp"
 #include "openPMD/version.hpp"
 
 #if openPMD_HAVE_MPI
 #   include <mpi.h>
 #endif
 
-#include <string>
+#include <map>
```
ax3l, line 43: `std::string` is still used in this file:
```suggestion
#include <map>
#include <string>
```

include/openPMD/RecordComponent.hpp
```diff
@@ -219,6 +219,12 @@ class RecordComponent : public BaseRecordComponent
      */
     bool
     dirtyRecursive() const;
+
+protected:
+    /*
```
ax3l:
```suggestion
    /**
```

include/openPMD/Iteration.hpp
```diff
@@ -158,10 +163,59 @@ class Iteration : public Attributable
...
+     *        layout, this will be the complete Series.
+     *
+     * @return AdvanceStatus
+     */
+    void
+    endStep();
+
+    /*
+     * We cannot give the return type yet, since Iteration is still an
+     * incomplete type.
```
ax3l: I don't understand what `myIteration` is doing from the Doxygen string. Can you please clarify? What does this iterate over if `Iteration` is a single openPMD iteration?

It this maybe better implemented as an external function like `iter`/`next` or `std::begin()` specialization, etc.? Bit curious, looks like something that would belong to `Series`.

include/openPMD/IO/IOTask.hpp
```diff
@@ -459,6 +464,51 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::LIST_ATTS > : public AbstractPara
...
+    std::unique_ptr< AbstractParameter >
+    clone() const override
+    {
+        return std::unique_ptr< AbstractParameter >(
+            new Parameter< Operation::ADVANCE >( *this ) );
+    }
+
+    // input parameter
+    AdvanceMode mode;
+    // output parameter
```
```suggestion
    //! input parameter
    AdvanceMode mode;
    //! output parameter
```

include/openPMD/IO/IOTask.hpp
```diff
@@ -65,8 +68,10 @@ OPENPMDAPI_EXPORT_ENUM_CLASS(Operation)
     DELETE_ATT,
     WRITE_ATT,
     READ_ATT,
-    LIST_ATTS
-};  //Operation
+    LIST_ATTS,
+
+    ADVANCE
```
ax3l: We missed the opportunity to document the operations, but let's start with the new ones:
```suggestion
    LIST_ATTS,

    ADVANCE //!< ...
```

If it's too brief here, then let's add it to the specializations below.

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -127,6 +133,37 @@ class AbstractIOHandlerImpl
    */
   virtual void
   closeFile( Writable *, Parameter< Operation::CLOSE_FILE > const & ) = 0;
+  /** Advance the file/stream that this writable belongs to.
```
ax3l, line 140:
```suggestion

  /** Advance the file/stream that this writable belongs to.
```

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -127,6 +133,37 @@ class AbstractIOHandlerImpl
...
+   * by splitting data into packets that are written to and read from transport.
+   *
+   * IO actions up to the point of closing a step must be performed now.
+   * 
+   * The advance mode is determined by parameters.mode.
+   * The return status code shall be stored as parameters.status.
+   */
+  virtual void advance(Writable*, Parameter< Operation::ADVANCE > &)
+  {}
+  /** Close an openPMD group.
```
ax3l, line 159:
```suggestion

  /** Close an openPMD group.
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -740,6 +890,40 @@ namespace detail
         invalidateVariablesMap();
 
     private:
+        // ADIOS engine
+        auxiliary::Option< adios2::Engine > m_engine;
+        std::string m_engineType;
+        /*
```
```suggestion
        //! ADIOS engine
        auxiliary::Option< adios2::Engine > m_engine;
        //! ... doc string ...
        std::string m_engineType;

        /*
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -680,36 +798,68 @@ namespace detail
...
+         * on chosen ADIOS2 engine and can not be explicitly overridden by user.
+         */
+        bool optimizeAttributesStreaming = false;
+        enum class Steps
+        {
+            UseSteps,
+            DontUseSteps,
+            Undecided
+        };
+        /*
```
```suggestion

        /*
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -680,36 +798,68 @@ namespace detail
...
+         * Will be automatically set by BufferedActions::configure_IO depending
+         * on chosen ADIOS2 engine and can not be explicitly overridden by user.
+         */
+        bool optimizeAttributesStreaming = false;
+        enum class Steps
+        {
+            UseSteps,
+            DontUseSteps,
+            Undecided
+        };
```
```suggestion

        /** ... Doxygen string ... */
        enum class Steps
        {
            UseSteps, //!< ... meaning ...
            DontUseSteps, //!< ... meaning ...
            Undecided //!< ... meaning ...
        };
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -680,36 +798,68 @@ namespace detail
...
+        detail::DatasetReader const m_readDataset;
+        detail::AttributeReader const m_attributeReader;
+
+        /*
+         * We call an attribute committed if the step during which it was
+         * written has been closed.
+         * A committed attribute cannot be modified.
+         */
+        std::set< std::string > uncommittedAttributes;
+        /*
```
ax3l, line 864:
```suggestion

        /*
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -680,36 +798,68 @@ namespace detail
         BufferedActions( BufferedActions const & ) = delete;
 
         std::string m_file;
+        std::string const m_IOName;
+        adios2::ADIOS & m_ADIOS;
         adios2::IO m_IO;
         std::vector< std::unique_ptr< BufferedAction > > m_buffer;
```
ax3l: Please document
```suggestion
        std::string m_file; //!< ... meaning ...
        std::string const m_IOName; //!< ... meaning ...
        adios2::ADIOS & m_ADIOS;
        adios2::IO m_IO;
        
        /** ... meaning ... */
        std::vector< std::unique_ptr< BufferedAction > > m_buffer;
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -323,7 +336,9 @@ namespace ADIOS2Defaults
     constexpr const_str str_engine = "engine";
     constexpr const_str str_type = "type";
     constexpr const_str str_params = "parameters";
-}
+    constexpr const_str str_usesteps = "usesteps";
+    constexpr const_str str_usesstepsAttribute = "__openPMD_internal/usesSteps";
```
ax3l: use-steps and uses-steps: once with two s once with one - good to unify?

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -182,6 +191,7 @@ class ADIOS2IOHandlerImpl
 
 private:
     adios2::ADIOS m_ADIOS;
+    std::string engineType;
```
```suggestion
    std::string engineType; //!< ... meaning & values ...
```

examples/10_streaming_write.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
+import openpmd_api as io
+import numpy as np
+
+run_streaming_example = False
+
+if not run_streaming_example:
+    exit(0)
```
ax3l, line 8: Is this intentionally untested, dead code?

ax3l (commented):
Very nice work :sparkles: 

I added a inline suggestions that can improve the long-term maintainability and clarify some things I did not understand. Thanks a lot! :+1:

examples/10_streaming_read.cpp
franzpoeschel: Our CI will automatically run all the examples, one after another. But the streaming examples need to be run concurrently, so the easiest solution for now was to just deactivate them by default. Otherwise, the CI will just run into a timeout, waiting for the reader/writer indefinitely.

examples/10_streaming_read.py
franzpoeschel: See above.

examples/10_streaming_write.cpp
franzpoeschel: See above.

src/Iteration.cpp
franzpoeschel: I knew there'd be a comment on this one :D
> Can we return StepStatus by value?

No, the intent of this is that some methods (`Iteration::close` as well as methods in `(Read|Write)Iterations`) need to modify the current `stepStatus`. Depending on the chosen iteration layout, `stepStatus` is a property of the whole `Series` or of each `Iteration`, so I'll need to decide which one to modify dynamically. So, similar to our `Attributable::written` method, this one must hand out a modifiable reference/pointer. Just that I have admittedly developed a strong dislike towards returning references, due to numerous accidental copying bugs I have had because of that...

Thinking about it, maybe a regular setter would do?

PrometheusPi:
@ax3l and @franzpoeschel Great to see that this amazing pull request is getting closer to go mainline. Are you planning to release it with 0.13.0? (And if yes, is there already a approximate date when this this will happen?) 

Currently, streaming works excellent in PIConGPU. I am just a bit afraid that the longer this is not getting into mainline both here and then in PIConGPU, the more adjustment have to be made to to the pull request to keep it working/mergeable.

franzpoeschel:
The plan is to merge #802 first (that one was originally part of the Streaming PR but I made it a separate one) and then this one. Given that both are relatively finished, I would suppose it should not take too long. On the plans for the next release, Axel will know more on that.

docs/source/backends/adios2.rst
franzpoeschel: > Should we turn on adios2.engine.usesteps by default (for streaming)?

For streaming, we must and do. For non-streaming, we should only do that upon requiring ADIOS 2.7.0 due to the inability to read such files back without the streaming API with ADIOS 2.6.0. (Or realistically, throw that option out and make it always true.)

> when does one have to set adios2.engine.usesteps as a work-around? During read of files if the write process was using SST?

Added documentation for that: `Upon reading a file, the ADIOS2 backend will automatically recognize whether it has been written with or without steps, ignoring the JSON option mentioned above.`
This is also what [this test](https://github.com/openPMD/openPMD-api/blob/5ca416835f99abae91eae77e4f42ce4f722fd625/test/SerialIOTest.cpp#L2820) tests.

Applied the first two bullet points, too.

docs/source/usage/streaming.rst
franzpoeschel: Adding that note is probably a wise thing to do regardless of that.

That being said, what you said is exactly the thing that [this test](https://github.com/openPMD/openPMD-api/blob/5ca416835f99abae91eae77e4f42ce4f722fd625/test/SerialIOTest.cpp#L2898) covers. Use conventional API to write a file and use streaming API to read it back in. The opposite test is currently impossible due to the bug described in [here](https://github.com/openPMD/openPMD-api/blob/5ca416835f99abae91eae77e4f42ce4f722fd625/docs/source/backends/adios2.rst).

.github/ci/sanitizer/clang/Leak.supp
franzpoeschel: Just tried it out, seems I can throw them out.

src/RecordComponent.cpp
franzpoeschel: No, this is just a regular flag. Just that the flag lives behind a shared pointer, since `RecordComponent` is a handle and the flag should be shared between instances.

src/Iteration.cpp
franzpoeschel: This one's already using the good cast ;)

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: Other way around actually. By default, pick the one given in the JSON string (that one's written to `m_engineType` earlier in `ADIOS2IOHandlerImpl::init`. Environment variable allows overwriting.

include/openPMD/Series.hpp
franzpoeschel: Hm, my idea was that `WriteIterations` is a handle and that `SharedResources` is the internal class containing all members shared between instances. Which in this case, happen to be `iterations` and `currentlyOpen`, but that's coincidence ;)

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: Hm, iirc my idea at that was that the first is used as a setter (tell the backend to use steps) and the second is used as a getter (whether the backend uses steps). But I suppose that gets lost in translation somewhere hm

src/Iteration.cpp
franzpoeschel: > Thinking about it, maybe a regular setter would do?

A regular setter now does.

include/openPMD/Iteration.hpp
franzpoeschel: Yeah, moved this one to `Series` now.

franzpoeschel:
Ooo found a doxygen bug
```
error: documented symbol 'openPMD::decltype' was not declared or defined. (warning treated as error, aborting now)
```

.github/ci/sanitizer/clang/Leak.supp
franzpoeschel: Well, it seems [that](https://travis-ci.com/github/openPMD/openPMD-api/jobs/431253555) answers that.
```
==10270==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 176 byte(s) in 1 object(s) allocated from:

    #0 0x7fbdfe615043 in __interceptor_malloc /tmp/final/llvm.src/projects/compiler-rt/lib/asan/asan_malloc_linux.cc:146:3

    #1 0x18445d6 in load_transport (/home/travis/build/bin/ParallelIOTests+0x18445d6)

    #2 0x18379e2 in INT_CMlisten_specific (/home/travis/build/bin/ParallelIOTests+0x18379e2)

    #3 0x180ed32 in CMlisten (/home/travis/build/bin/ParallelIOTests+0x180ed32)

    #4 0x17d31c6 in CP_getCPInfo (/home/travis/build/bin/ParallelIOTests+0x17d31c6)

    #5 0x17cbed7 in SstReaderOpen (/home/travis/build/bin/ParallelIOTests+0x17cbed7)

    #6 0x17731ae in adios2::core::engine::SstReader::SstReader(adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm) (/home/travis/build/bin/ParallelIOTests+0x17731ae)

    #7 0x147fdc5 in std::shared_ptr<adios2::core::Engine> adios2::core::IO::MakeEngine<adios2::core::engine::SstReader>(adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm) (/home/travis/build/bin/ParallelIOTests+0x147fdc5)

    #8 0x13eb5d8 in std::_Function_handler<std::shared_ptr<adios2::core::Engine> (adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm), std::shared_ptr<adios2::core::Engine> (*)(adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm)>::_M_invoke(std::_Any_data const&, adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode&&, adios2::helper::Comm&&) (/home/travis/build/bin/ParallelIOTests+0x13eb5d8)
```

src/IO/AbstractIOHandlerHelper.cpp
franzpoeschel: What about a JSON parameter `.backend`? We can leave the filename extensions in as shortcuts, including a `bp4` one. We would then of course need to actually store those things as `*.bp4` on disk, otherwise this trick won't work when reading things back in.

include/openPMD/Streaming.hpp
franzpoeschel: I could not think of a non-redundant documentation for `AdvanceMode::BEGINSTEP` and `AdvanceMode::ENDSTEP`, but I added the others.

franzpoeschel:
I think I have most of your remarks covered by now.

.github/ci/sanitizer/clang/Leak.supp
franzpoeschel:
```
leak:adios2::core::engine::SstReader::*
leak:adios2::core::engine::BP4Reader::*
leak:adios2::core::engine::SstWriter::*
leak:adios2::core::engine::BP4Writer::*
leak:CMpoll_forever
leak:SstWriterClose
```
Of those, the `SstWriter` and `SstReader` lines are still needed, the other ones not any longer.

franzpoeschel:
I've implemented the streaming API in `openPMD-ls`. Given that draining a stream is definitely a a non-`const` thing to do, this implied removing some `const`s.

examples/10_streaming_read.cpp
ax3l: Discussed today: automatic coverage will be enabled in a follow-up PR, so things don't break over time with different contributors to the code base.

Idea: Single MPI context and split for a test. Or: ctest running a script that starts two executables.

test/ParallelIOTest.cpp
```diff
@@ -524,3 +524,102 @@ TEST_CASE( "close_iteration_test", "[parallel]" )
...
+        }
+        REQUIRE( last_iteration_index == 9 );
+    }
+}
+
+TEST_CASE( "adios2_streaming", "[pseudoserial][adios2]" )
+{
+    adios2_streaming();
+}
+#endif
```
ax3l, line 625:
```suggestion
#endif

```

ax3l, line 625:
```suggestion
#endif

```

ax3l (approved):
This is fantastic work, thank you for your dedication and the high quality. Congratulations and happy holidays! :christmas_tree:

Issue #490, opened 2019-02-22
Segfault in Flush of Particle Record Component


ax3l:
**Describe the bug**
An early flush during parallel `CREATE` causes a segfault.

**To Reproduce**

Taken from `ParallelIOTest` section `hdf5_write_test` `[parallel][hdf5]` with an added `o.flush()` [here](https://github.com/openPMD/openPMD-api/blob/e3d816a615caba711c8ef85d82791a09387d72bd/test/ParallelIOTest.cpp#L101) after the first `storeChunk()`.

```C++
TEST_CASE( "hdf5_write_test", "[parallel][hdf5]" )
{
    int mpi_s{-1};
    int mpi_r{-1};
    MPI_Comm_size(MPI_COMM_WORLD, &mpi_s);
    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_r);
    uint64_t mpi_size = static_cast<uint64_t>(mpi_s);
    uint64_t mpi_rank = static_cast<uint64_t>(mpi_r);
    Series o = Series("../samples/parallel_write.h5", AccessType::CREATE, MPI_COMM_WORLD);

    o.setAuthor("Parallel HDF5");
    ParticleSpecies& e = o.iterations[1].particles["e"];

    std::vector< double > position_global(mpi_size);
    double pos{0.};
    std::generate(position_global.begin(), position_global.end(), [&pos]{ return pos++; });
    std::shared_ptr< double > position_local(new double);
    *position_local = position_global[mpi_rank];

    e["position"]["x"].resetDataset(Dataset(determineDatatype(position_local), {mpi_size}));
    e["position"]["x"].storeChunk(position_local, {mpi_rank}, {1});

    o.flush();  //! @todo why does this flush cause a segfault?
}
```
**Expected behavior**
Should not segfault :-)

**Software Environment:**
 - version of openPMD-api: 0.8.0-dev (as of e3d816a615caba711c8ef85d82791a09387d72bd)
 - installed openPMD-api via: from source
 - operating system: Debian 9.7
 - machine: local laptop
 - name and version of Python implementation: N/A
 - version of HDF5: 1.10.0-patch1
 - version of ADIOS1: N/A
 - name and version of MPI: OpenMPI 2.0.2

anokfireball:
Does serial execution of the same code also result in a segfault? If so, I might have a hunch (https://github.com/openPMD/openPMD-api/issues/321#issuecomment-417298702) about the origin.

[...]

Issue #321, opened 2018-08-30
Particles: Default for `positionOffset`


ax3l:
**Is your feature request related to a problem? If so, please describe.**
When writing particle data, we should define the default of `positionOffset` to a *constant record (component)* with default `0` (zero) offset.

This makes it easier for users to write position data if they do not care about fine-vs-coarse precision that `position` + `positionOffset` offer.

Python:
```python
import openPMD


num_cereals = 3

# cereal positions in my breakfast bowl
x = np.arange(num_cereals, dtype=np.double)
y = np.arange(num_cereals, dtype=np.double)

# open file for writing
series = openPMD.Series(
    "cereals.h5",
    openPMD.Access_Type.create
)

cereals = series.iterations[1].particles["cereals"]

datatype = openPMD.Datatype.DOUBLE
extent = [num_cereals]
dataset = openPMD.Dataset(datatype, extent)

cereals["position"]["x"].reset_dataset(dataset)
cereals["position"]["y"].reset_dataset(dataset)

offset = [0]
cereals["position"]["x"].store_chunk(offset, extent, x)
cereals["position"]["y"].store_chunk(offset, extent, y)

# terminate called after throwing an instance of 'std::runtime_error'
# what():  A Record can not be written without any contained RecordComponents: positionOffset
```

**Describe the solution you'd like**

Unless the user sets it explicitly, we should implicitly default in the above example to:
```python
cereals["positionOffset"]["x"].make_constant(0.0)
cereals["positionOffset"]["y"].make_constant(0.0)
```

**Additional context**

Also, this would solve a small bug: if I execute the above example and add the lines
```python
# cereals["positionOffset"]["x"].make_constant(0.0)
cereals["positionOffset"]["y"].make_constant(0.0)
```
the check `A Record can not be written without any contained RecordComponents: positionOffset` does *not throw* anymore, although not every `position` component has a `positionOffset` component counterpart:

```bash
$ h5ls -r cereals.h5
/                        Group
/data                    Group
/data/1                  Group
/data/1/particles        Group
/data/1/particles/cereals Group
/data/1/particles/cereals/position Group
/data/1/particles/cereals/position/x Dataset {3}
/data/1/particles/cereals/position/y Dataset {3}
/data/1/particles/cereals/positionOffset Group
/data/1/particles/cereals/positionOffset/y Group
```

anokfireball:
[...]

PR #583, opened 2019-11-05
Docs: MPI Contracts

ax3l:
Start writing API contract for users.
Which functions are collective and which are non-collective?

### DIFF
--- a/docs/source/details/mpi.rst
+++ b/docs/source/details/mpi.rst
@@ -1,0 +1,55 @@
+.. _details-mpi:
+
+MPI Behavior
+============
+
+openPMD-api can be used within a MPI-parallel environment.
+The following API contracts need to be followed.
+
+Collective
+----------
+
+Series
+^^^^^^
+
+Generally, opening and closing the ``Series`` object with a MPI communicator is a collective operation.
+The communicator will be duplicated via ``MPI_Comm_dup`` in its constructor and freed in the destructor.
+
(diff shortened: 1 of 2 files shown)


docs/source/details/mpi.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Generally, opening and closing the ``Series`` object with a MPI communicator is a collective operation.
+The communicator will be duplicated via ``MPI_Comm_dup`` in its constructor and freed in the destructor.
+
+The first access to ``Series::iterations::operator[]`` as well as ``Iteration::meshes::operator[]`` and ``Iteration::particles::operator[]`` are collective.
+
+Flush
+^^^^^
+
+Flush should be treated as collective operation.
```
ax3l, line 27: Besides bugs like #490, #484, and #321, I think this is a design bug.
`flush()` should only ensure that user-provided buffers can be manipulated or freed again, if still held by the user. We should make no assumptions about when something is written to disk.

Adding defaults such as [these](https://github.com/openPMD/openPMD-api/issues/490#issuecomment-466420004) should be done on `Series::~Series` (or `Series::close()` #149). (We can also add an `Iteration::close()` or individual `Mesh::/Particle::close()` routines if we want to signal explicitly that we are done manipulating those and do not intend to add more things.)

ax3l:
@franzpoeschel @C0nsultant Do you have thoughts on these docs? I try to express our MPI API contract to users here.

PR #588, opened 2019-11-08
Temporary fix for "ADIOS2 backend broken on post-2.5.0 master"

### LINKED ISSUE #586: ADIOS2 backend broken on post-2.5.0 master
PR [#1821](https://github.com/ornladios/ADIOS2/pull/1821) breaks the ADIOS2 backend, see [my comment there](https://github.com/ornladios/ADIOS2/issues/1816#issuecomment-551105186) for an explanation. 
I will shortly open PR with a temporary fix that reimplements the old prefix search, until I find the time for a more sensible solution (see the described solution in my comment) that makes use of the ADIOS2 PR.

franzpoeschel:
Temporarily close #586 until a [better solution](https://github.com/ornladios/ADIOS2/issues/1816#issuecomment-551105186) is implemented.

### DIFF
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -578,6 +578,9 @@ namespace detail
          * Delete all buffered actions without running them.
          */
         void drop( );
+
+        std::map< std::string, adios2::Params >
+        availableAttributesTemporary( std::string const & variable );
     };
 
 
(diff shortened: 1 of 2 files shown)


ax3l:
Thanks a lot!
Does this work with ADIOS 2.4.0 - ADIOS post-2.5.0 `master`?

Can you please quickly summarize the performance implications of this work-around in the PR description?

franzpoeschel:
Yes, it works also on the most recent version of ADIOS2 on the master branch. The performance implications coming from the previously working versions are that performance stays the same, since this PR quickly reimplements the linear search that ADIOS2 did previously. Putting it negatively, we temporarily opt out of the increased performance coming with the ADIOS2 PR.

ax3l (approved):

ax3l:
@franzpoeschel can you please open a new issue documenting the steps towards a *better solution*? :)

franzpoeschel:
@ax3l Following today's discussion in [#1816](https://github.com/ornladios/ADIOS2/issues/1816), it might turn out that we will not use this new feature at all or at most partially. We can either switch lookup strategy depending on whether we are parsing for a group's or a dataset's attributes – or we can simply go for the option that always works and use the fact that the returned map is sorted by lexicographic order.

Issue #593, opened 2019-11-11
Installing error messages in MacOS with conda/pip


LDAmorim:
I am trying to follow up on the answer to issue #591 and install OpenPMD-api with MPI support through ` conda install -c conda-forge openpmd-api=*=mpi_mpich*`.
But found an error that I don't know how to fix. I then tried the pip installation commands, but still encountered errors.
I will try spack next, but decided to leave this issue here in case I am doing something wrong that you can spot and correct me on.
(Sorry if this is related to issues already here or on `conda`'s GitHub repo, but I couldn't find something that seemed similar to me.)

Initially I tried the command and when I found the error reported below, I decided to try its dependencies so I did:
```bash
$ conda clean --all
$ conda update --all
$ conda install -c conda-forge hdf5
$ conda install -c conda-forge adios
$ conda install -c conda-forge adios2
conda install -c conda-forge mpich 
$ conda install -c conda-forge openpmd-api=*=mpi_mpich*
``` 
From the latter command, I got the following (and the same as the one I saw initially) error messages:
```bash
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

PackagesNotFoundError: The following packages are not available from current channels:

  - openpmd-api[build=mpi_mpich*]

Current channels:

  - https://conda.anaconda.org/conda-forge/osx-64
  - https://conda.anaconda.org/conda-forge/noarch
  - https://repo.anaconda.com/pkgs/main/osx-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/osx-64
  - https://repo.anaconda.com/pkgs/r/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

So then I tried to use `openPMD_USE_MPI=ON python3 -m pip install openpmd-api  --no-binary :all:` instead. This led to the error messages below:
```bash
$ openPMD_USE_MPI=ON pip install openpmd-api --no-binary :all:
Collecting openpmd-api
  Using cached https://files.pythonhosted.org/packages/10/e7/575daf3fe06d430713e5bf32ea68077c693a1aa82138209a85f8cbb715aa/openPMD-api-0.9.0a1.tar.gz
  Installing build dependencies ... error
[...]

PR #594, opened 2019-11-11
Doc: Improve Pip Install

ax3l:
- general: update pip first, so we definitely have pip version 19+
- MPI build: only build itself from source
- syntax: use pip as python(3) module

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,14 +161,21 @@ The current status for this install method is *experimental*.
 Please feel free to [report](https://github.com/openPMD/openPMD-api/issues/new/choose) how this works for you.
 
 ```bash
-# optional:             --user
-pip install openpmd-api
+# we need pip 19 or newer
+# optional:                   --user
+python3 -m pip install -U pip
+
+# optional:                        --user
+python3 -m pip install openpmd-api
 ```
 
 or with MPI support:
 ```bash
-# optional:                                                  --user
-openPMD_USE_MPI=ON pip install openpmd-api --no-binary :all:
+# optional:                   --user
+python3 -m pip install -U pip
+
+# optional:                                                                   --user
+openPMD_USE_MPI=ON python3 -m pip install openpmd-api --no-binary openpmd-api
 ```
 
 ### From Source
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -161,14 +161,21 @@ The current status for this install method is *experimental*.
 Please feel free to [report](https://github.com/openPMD/openPMD-api/issues/new/choose) how this works for you.
 
 ```bash
-# optional:             --user
-pip install openpmd-api
+# we need pip 19 or newer
+# optional:                   --user
+python3 -m pip install -U pip
```
LDAmorim, line 166: I usually add the `python3 -m` part of the command to make sure I am installing for the correct python version and I am thinking you did it too. Your change here helps to confirm that we use `Python 3.5 - 3.7` as explained in the build dependencies entry of your [documentation](https://openpmd-api.readthedocs.io/en/0.9.0-alpha/dev/dependencies.html#optional-language-bindings), which is very nice!

LDAmorim (approved):
Thanks for this PR!
It will help to make sure we always update pip.

docs/source/install/install.rst
```diff
@@ -69,15 +69,22 @@ Please feel free to `report <https://github.com/openPMD/openPMD-api/issues/new/c
...
 
 .. code-block:: bash
 
-   # optional:                                                  --user
-   openPMD_USE_MPI=ON pip install openpmd-api --no-binary :all:
+   # optional:                   --user
+   python3 -m pip install -U pip
+
+   # optional:                                                                   --user
+   openPMD_USE_MPI=ON python3 -m pip install openpmd-api --no-binary openpmd-api
```
LDAmorim, line 87: I am afraid that, so far this one is not exactly 100% operational with the MPI support in my (frustrating) computer ...

README.md
ax3l: Yes, exactly. As far as I know it's also the recommend way to use `pip` these days.

docs/source/install/install.rst
ax3l: Let's take a look! :)

ax3l: Ok, we found a work-around for your problem which is unrelated to the instructions herein for now: https://github.com/openPMD/openPMD-api/issues/593#issuecomment-552690470

PR #595, opened 2019-11-12
Fix OSX setup.py: @loader_path

ax3l:
On OSX, `$ORIGIN` is not a known concept for relative RPATHs. Instead, `@loader_path/` should be the right path (contrary to `@rpath/`) to find relative dependencies of a shared library.

Refs.:

- https://github.com/openPMD/openPMD-api/issues/593#issuecomment-552690470
- https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling
- https://cmake.org/pipermail/cmake/2014-May/057654.html
- https://docs.python.org/3.8/library/sys.html#platform

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Bug Fixes
 - Clang: fix pybind11 compile on older releases, such as AppleClang 7.3-9.0, Clang 3.9 #543
 - Python:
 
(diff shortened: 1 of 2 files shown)


LDAmorim (approved):
This works perfectly in my MacOS! Thanks for this PR

ax3l:
Hooray! Thank you so much for testing this with me! :rocket: :sparkles: :medal_sports:

PR #596, opened 2019-11-12
Fix mpi4py Examples: import first

ax3l:
At least on OSX and at least with MPICH, `mpi4py` must be included first in order to avoid an error of the kind:
```
Attempting to use an MPI routine before initializing MPICH
```
and in Jupyter
```
The kernel appears to have died. It will restart automatically.
```

First seen by @LDAmorim :sparkles: :muscle:


ax3l:
Likely related to https://github.com/conda-forge/openpmd-api-feedstock/pull/28

PR #597, opened 2019-11-12
[WIP] CMake: Enable IPO/LTO

ax3l:
[Interprocedural optimization (IPO) also known as link-time optimization (LTO)](https://en.wikipedia.org/wiki/Interprocedural_optimization) is a great feature to reduce binary size and to improve code performance, e.g. through extended inlining.

This feature is specifically useful for modern C++ code, which is also the reason pybind11 enables it automatically for our python bindings as soon as it is detected.

We now expose this feature for the rest of the library build as well and auto-enable it when supported by the compiler.

Binary size before:
```
$ du -hs lib/*
1.0M    lib/libCatchMain.so
1.0M    lib/libCatchRunner.so
1.3M    lib/libopenPMD.ADIOS1.Parallel.so
1.1M    lib/libopenPMD.ADIOS1.Serial.so
2.2M    lib/libopenPMD.so
968K    lib/python3.6
```
and with IPO/LTO enabled (new default):
```
$ du -hs lib/*
876K    lib/libCatchMain.so
876K    lib/libCatchRunner.so
1.2M    lib/libopenPMD.ADIOS1.Parallel.so
1.1M    lib/libopenPMD.ADIOS1.Serial.so
1.5M    lib/libopenPMD.so
968K    lib/python3.6
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -266,6 +266,7 @@ Features
   - Python 3.8 support #581
   - support empty datasets via ``Record_Component.make_empty`` #538
 - pkg-config: add ``static`` variable (``true``/``false``) to ``openPMD.pc`` package #580
+- CMake: interprocedural/link-time optimization #597
 
 Bug Fixes
 """""""""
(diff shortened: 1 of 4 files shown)


ax3l:
Problems occur with:
- MSVC on shared library builds (VS 2015)
- tooling such as clang-tidy

PR #599, opened 2019-11-12
Python3 in docs cmake instructions section

LDAmorim:
The installation in MacOS Mojave is successful with the command that is in [https://openpmd-api.readthedocs.io/en/latest/install/install.html](https://openpmd-api.readthedocs.io/en/latest/install/install.html):
```cmake ../openPMD-api -DCMALE_INSTALL_PREFIX=(...)/openPMD-api-install -DopenPMD_USE_PYTHON=ON -DPYTHON_EXECUTABLE=$(which python3)```

But can cause the error message below if both python2 and python3 versions are installed in that machine and `$(which python)` is used instead.

```bash
CMake Error at CMakeLists.txt:275 (message):
 python: Found version 2.7 is NOT supported.  Set for example
 -DPYTHON_EXECUTABLE=$(which python3) to use 3.5+.
-- Configuring incomplete, errors occurred!
```

- Changed `$(which python)` to `$(which python3)` in docs to prevent error of competing python < 3.5 versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ cd openPMD-api-build
 # for options append:
 #   -DopenPMD_USE_...=...
 # e.g. for python support add:
-#   -DopenPMD_USE_PYTHON=ON -DPYTHON_EXECUTABLE=$(which python)
+#   -DopenPMD_USE_PYTHON=ON -DPYTHON_EXECUTABLE=$(which python3)
 cmake ../openPMD-api
 
 cmake --build .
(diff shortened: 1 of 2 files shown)


ax3l (commented):
Looks great to me, good catch!

Let's also update the same line in `README.md` please :)

ax3l (approved):
Thanks a lot! :sparkles:

LDAmorim:
Thanks for noticing it 😃 We looked for `(which python)` in all files via
```git
git grep -i "(which python3)" .
````
 just to be sure and there are no other instances of it.

PR #605, opened 2019-11-14
ADIOS1: Keep Default for Now

ax3l:
Since we are planning to add more refactoring in the internal (variables vs. attributes) representation of ADIOS2, let's not make this default over ADIOS1 yet to avoid writing files that soon cannot be read again.

Delay this just for a little for this upcoming release, so we can draft a new version without changing defaults yet. (I want to release a new version very soon to fix some installation issues and other updates that we accumulated.)

Delays the new defaults to ADIOS2 in #568 for a later release.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -6,9 +6,10 @@ Upgrade Guide
 0.10.0-alpha
 ------------
 
-We added support for ADIOS2 in this release.
-As soon as the ADIOS2 backend is enabled it will take precedence for ``.bp`` files over a potentially also enabled ADIOS1 backend.
-In order to prefer the legacy ADIOS1 backend in such a situation, set an environment variable: ``export OPENPMD_BP_BACKEND="ADIOS1"``.
+We added preliminary support for ADIOS2 in this release.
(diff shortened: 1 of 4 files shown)


ax3l:
ping @franzpoeschel is this fine with you, too?

Just want to push a bug-fix release and the ADIOS2 files we write will probably change with your further updates for read speed, I guess?

franzpoeschel:
@ax3l This is ok as long as the ADIOS2 backend hasn't stabilized yet. I have already implemented the necessary changes on [topic-streaming](https://github.com/franzpoeschel/openPMD-api/tree/topic-streaming) and will backport them to dev soon. Those changes do not affect the layout of the BP4 file, since the implementation does not add ADIOS variables for groups, unlike initially planned. (see [here](https://github.com/ornladios/ADIOS2/issues/1816))

franzpoeschel:
[see this PR](https://github.com/openPMD/openPMD-api/pull/613)

ax3l:
> Those changes do not affect the layout of the BP4 file, since the implementation does not add ADIOS variables for groups, unlike initially planned.

Okay, thanks for the heads-up on this. I wasn't sure anymore.

PR #611, opened 2019-11-15
Python: Get Unit Dimension

ax3l:
Expose access to a `.unit_dimension` member in records, just as in `.unit_SI`. Avoids having to query this openPMD required member by attribute.

Thanks to @LDAmorim for spotting!

### DIFF
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -489,6 +489,11 @@ class APITest(unittest.TestCase):
         E_x = E["x"]
         shape = E_x.shape
 
+        if found_numpy:
+            np.testing.assert_allclose(E.unit_dimension,
+                                       [1., 1., -3., -1., 0., 0., 0.])
+        self.assertAlmostEqual(E_x.unit_SI, 1.0)
+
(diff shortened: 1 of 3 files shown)


LDAmorim (approved):
Thank you for these changes!!
It is really nice to have the unit_dimension accessible in python without calling the `attribs`.

test/python/unittest/API/APITest.py
```diff
@@ -540,6 +545,11 @@ def testSliceRead(self):
         pos_y = electrons["position"]["y"]
         w = electrons["weighting"][api.Record_Component.SCALAR]
 
+        if found_numpy:
+            np.testing.assert_allclose(electrons["position"].unit_dimension,
```
LDAmorim, line 549: Does this work for momentum as well?

LDAmorim (approved):

test/python/unittest/API/APITest.py
ax3l: yep, all openPMD records, both for meshes and particles :)

The line you are seeing here is not the implementation but a test case. I just tested it on a sample mesh record and a sample particle species record to see if it works.

PR #617, opened 2019-11-26
NVCC: No Visibility Attribute

ax3l:
This project is provides host-side APIs [1]. Some downstream projects are unable to split their translation units properly in host+device and host-only, and just pass all code through `nvcc`. Doing so is not appreciated (by me) but tolerated (by me).

Unfortunately, `nvcc` is a bit behind in supporting visibility attributes (Nvidia bug report: 2767443 as of v10.2). Assuming that openPMD-api itself is built with a proper host-compiler instead of `nvcc`, we can live with defining our `EXPORT` macro to nothing during downstream consumption.

The previous implementation caused a compile error when build with GCC <6 and a warning with newer GCC in `include/openPMD/IO/IOTask.hpp` for `enum class Operation`.

[1] Note: We might support passing device-side pointers in the future, though. If HDF5 or ADIOS2 start to support consuming those...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,9 @@ Bug Fixes
 """""""""
 
 - ADIOS2: fix C++17 build #614
+- nvcc:
+
+  - ignore export of ``enum class Operation`` #617
 
 Other
 """""
--- a/include/openPMD/IO/ADIOS/ADIOS1IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS1IOHandler.hpp
@@ -30,8 +30,10 @@
 #   include <queue>
 #endif
 
-#if _MSC_VER
+#ifdef _MSC_VER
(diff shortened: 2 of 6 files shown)


ax3l:
Just for the record, a typical compile error of such kind reads:
```
include/openPMD/IO/IOTask.hpp:52:1: warning: elaborated-type-specifier for a scoped enum must not use the 'class' keyword
 enum class Operation : EXPORT unsigned int
 ^
include/openPMD/IO/IOTask.hpp:52:12: error: use of enum 'Operation' without previous declaration
 enum class Operation : EXPORT unsigned int
            ^
include/openPMD/IO/IOTask.hpp:52:21: error: expected unqualified-id before ':' token
 enum class Operation : EXPORT unsigned int
                     ^
include/openPMD/IO/IOTask.hpp:100:11: error: 'Operation' has not been declared
 template< Operation >
```

PR #622, opened 2019-12-05
Add Container::contains

ax3l:
Add a [`contains` method](https://en.cppreference.com/w/cpp/container/map/contains) as in C++20 to our `openPMD::Container`. This allows to check the existence of keys in iterations, records, etc. without having to use the exception system.

Suggested by @LDAmorim, thank you! :)


LDAmorim (commented):
Thanks for this PR Axel!

PR #643, opened 2019-12-18
Allows multiple flush on the same file

### LINKED ISSUE #561: reading the multi-mode RZ output
Hi, I wanted to give a try to multi-mode RZ functionality (on CPU), but having some problems with output. 

I'm not sure at which stage #353 now, but for me the openPMD did not work in RZ -- everything compiles fine, the calculation gets stuck at the output stage.

Alternatively I've checked the plotfile option, but its not straightforward neither -- the `ds.covering_grid` seems to only contain 0th mode. 

Could you point my in the direction to look for a solution?

guj:
This is to resolve [WarpX issue 561](https://github.com/ECP-WarpX/WarpX/issues/561), which flush is called more than once.  And the later times not all processes participated while H5Fopen is a collective call.

### DIFF
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -124,6 +124,7 @@ HDF5IOHandlerImpl::createFile(Writable* writable,
         writable->abstractFilePosition = std::make_shared< HDF5FilePosition >("/");
 
         m_fileIDs[writable] = id;
+        m_fileNamesWithID[std::move(name)]=id;
         m_openFileIDs.insert(id);
     }
 }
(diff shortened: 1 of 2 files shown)


src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -352,6 +352,11 @@ HDF5IOHandlerImpl::openFile(Writable* writable,
     if( !auxiliary::ends_with(name, ".h5") )
         name += ".h5";
 
+    auto search = m_fileIDs.find(writable);
```
ax3l: This is effectively implementing the `//TODO` above, isn't it?
Maybe we can add this as a comment before this block?

guj: Yes,  looks like so.  Well  CI failed.  let me address that first

guj:
@ax3l  Hi Axel,  please review and merge  it when you get a chance.  Thanks.

ax3l (approved):
Thank you, this looks great to me!
Did this already address the issue in WarpX or do you need to add similar support to the iteration, as mentioned in the TODO comment?

PR #648, opened 2020-01-06
BaseRecord: Overwrite Erase with Iterator

ax3l:
Fix warning of the form:
```
openPMD/backend/BaseRecord.hpp(39):
  warning: overloaded virtual function
  "openPMD::Container<T, T_key, T_container>::erase
    [with T=openPMD::MeshRecordComponent, T_key=std::__cxx11::string, T_container=std::map<std::__cxx11::string, openPMD::MeshRecordComponent, std::less<std::__cxx11::string>, std::allocator<std::pair<const std::__cxx11::string, openPMD::MeshRecordComponent>>>]"
    is only partially overridden in class
      "openPMD::BaseRecord<openPMD::MeshRecordComponent>"
  detected during instantiation of class
    "openPMD::BaseRecord<T_elem>
      [with T_elem=openPMD::MeshRecordComponent]"
    openPMD/Mesh.hpp(38): here
```

Introduced in #89

Seen on Summit (PPC64le) with WarpX (C++14) with NVCC 10.1.168, GCC 6.4.0.

### DIFF
--- a/include/openPMD/backend/BaseRecord.hpp
+++ b/include/openPMD/backend/BaseRecord.hpp
@@ -62,6 +62,9 @@ public:
     mapped_type& operator[](key_type const& key) override;
     mapped_type& operator[](key_type&& key) override;
     size_type erase(key_type const& key) override;
+    iterator erase(iterator res) override;
+    //! @todo add also, as soon as added in Container:
+    // iterator erase(const_iterator first, const_iterator last) override;
 
     virtual std::array< double, 7 > unitDimension() const;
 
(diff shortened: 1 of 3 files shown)


include/openPMD/backend/BaseRecord.hpp
```diff
@@ -181,6 +184,36 @@ BaseRecord< T_elem >::erase(key_type const& key)
     return res;
 }
 
+template< typename T_elem >
+inline typename BaseRecord< T_elem >::iterator
+BaseRecord< T_elem >::erase(iterator res)
+{
+    bool scalar = (res->first == RecordComponent::SCALAR);
```
ax3l, line 191: maybe just calling `erase(res->first)` (after checking `res` is not `end()`) is easier and avoids code duplication.

PR #649, opened 2020-01-08
Docs first read python typo

LDAmorim:
Hi,
This PR addresses a really minor detail.
But, just in case anyone (who is not 100% familiar with python) tries to follow `docs/source/usage/firstread.rst` by copy pasting it's sections, I think there is an issue with the attributes part:
```
if( series.contains_attribute("author") )
    print("Author: ",
          series.author)
```
That I think is just a missing ":" at the end of the `if(...):` line.
This PR fixes that.
Cheers

### DIFF
--- a/docs/source/usage/firstread.rst
+++ b/docs/source/usage/firstread.rst
@@ -145,7 +145,7 @@ Python
    print("openPMD version: ",
          series.openPMD)
 
-   if( series.contains_attribute("author") )
+   if( series.contains_attribute("author") ):
        print("Author: ",
              series.author)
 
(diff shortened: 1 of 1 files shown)


docs/source/usage/firstread.rst
```diff
@@ -145,7 +145,7 @@ Python
    print("openPMD version: ",
          series.openPMD)
 
-   if( series.contains_attribute("author") )
+   if( series.contains_attribute("author") ):
```
ax3l: Can we even write this? :)
```suggestion
   if series.contains_attribute("author"):
```

ax3l (approved):
Thank you, I slipped in some C++ into the Python section :)

docs/source/usage/firstread.rst
LDAmorim: Yes, of course :) Thanks for checking it!

PR #652, opened 2020-01-09
Clang & GCC: Woverloaded-virtual, Wextra-semi

ax3l:
GCC & Clang: `-Woverloaded-virtual`
Add warnings for partly overwritten virtual methods (GCC only).
Thrown anyway as a warning by nvcc (even without any flags): #648 

Clang: `-Wextra-semi`
Catch things like superfluous semicolons.

Fix some doxygen and clang `-Wdocumentation` warnings.


ax3l:
Hm, some warnings are truly only occurring when including our headers in a `.cu` file for nvcc.

PR #653, opened 2020-01-09
MPI Benchmark: Doxygen Params

ax3l:
Start adding missing parameter descriptions.

@franzpoeschel can you please push to this branch and fill the `...` descriptions accordingly? :)

### DIFF
--- a/include/openPMD/benchmark/mpi/MPIBenchmark.hpp
+++ b/include/openPMD/benchmark/mpi/MPIBenchmark.hpp
@@ -46,7 +46,7 @@ namespace openPMD
     /**
      * Class representing a benchmark.
      * Allows to configure a benchmark and execute it.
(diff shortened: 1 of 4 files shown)


include/openPMD/benchmark/mpi/MPIBenchmark.hpp
```diff
@@ -180,9 +182,13 @@ namespace openPMD
 
             /**
          * Execute a single read benchmark.
-         * @tparam Clock Clock type to use.
+         * @tparam T Clock type to use.
```
ax3l: that's wrong... `T` is the data type? The `Clock` `tparam` must move up as part of the `BenchmarkExecution` doxygen block.

ax3l:
@franzpoeschel thanks for the update! :)

PR #655, opened 2020-01-13
LGTM: Ubuntu Update adds MPI

ax3l:
Has the image of Ubuntu been updated on LGTM?
If so, we can also test MPI features on.


ax3l:
Nope, not yet.

PR #656, opened 2020-01-14
ADIOS2: Environment Variable for Engine

ax3l:
Allows us to switch to e.g. [`Null` engine](https://adios2.readthedocs.io/en/latest/engines/engines.html) to measure our own (or user-preparation-) overhead.

Example:
```bash
export OPENPMD_ADIOS2_ENGINE="Null"
```

Require 2.5.0+ for ADIOS2 since there the engine default was set to "File" (and BP4 instead of BP3) first.


ax3l:
@williamfgc The speed of the `Null` engine does not disappoint :p

williamfgc:
@ax3l it's the best for write-only data

guj:
Looks good!  Thanks!

guj (approved):

PR #657, opened 2020-01-14
Defaults for Date & Software Attributes

ax3l:
Adds defaults for the date and software attributes, which are recommended openPMD attributes (missing is a warning in our validator).

Deprecate old API that allows to set the software version independent of the software name, which can lead to inconsistencies, e.g. if the default version is kept and only the software name is changed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ Features
 """"""""
 
 - ADIOS2: expose engine #656
+- defaults for ``date`` and software base attributes #657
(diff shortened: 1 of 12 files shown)


ax3l:
C/C++ localtime...
- [so](https://en.cppreference.com/w/cpp/chrono/c/localtime)
- [much](https://en.cppreference.com/w/c/chrono/localtime)
- [wow](https://sourceware.org/bugzilla/show_bug.cgi?id=16145)

PR #663, opened 2020-01-18
Attribute Constructor: Move Argument

ax3l:
The value-passed argument for the constructor of `Attribute` can be moved into place to avoid double copies.

### DIFF
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -23,14 +23,15 @@
 #include "openPMD/auxiliary/Variant.hpp"
 #include "openPMD/Datatype.hpp"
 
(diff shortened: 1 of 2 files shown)


include/openPMD/backend/Attribute.hpp
```diff
@@ -70,7 +71,7 @@ class Attribute :
                             bool >
 {
 public:
-    Attribute(resource r) : Variant(r)
+    Attribute(resource r) : Variant(std::move(r))
```
ax3l, line 74: note: single argument constructors (converting constructors) should also be `explicit` in clean code. but this breaks to much in our usage for now.
https://en.cppreference.com/w/cpp/language/converting_constructor

PR #668, opened 2020-01-25
Series Test: FileBased Independent

ax3l:
Make sure that fileBase encoded series also support the independent writes as in #554.

~~This is currently problematic with ADIOS1 (in CI) and HDF5 (not in CI since I need #569 for that).~~

~~This causes interesting hang situations in WarpX particle flushes if one rank only writes attributes and other ranks actually write particles on first open.~~

## To Do

- [x] add test case
- ~~is this an MPICH 3.3a2 specific problem? even though CI looks good here, my local WarpX tests also hang when build with OpenMPI 3.1.5~~ local MPI mix-up in deps

## Future

- [x] An additional test variant in which only some writers even *declare* a record would be nice if supported (probably depending on backend): #671

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -25,7 +25,7 @@ TEST_CASE( "none", "[parallel]" )
 #endif
 
 #if openPMD_HAVE_MPI
-void write_test_zero_extent( std::string file_ending, bool writeAllChunks )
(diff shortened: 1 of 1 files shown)


ax3l:
The problem seams to be a local MPI mixup. Dang, lost half a day with it.

PR #674, opened 2020-01-27
ADIOS1: Create Attribute-Only Files

ax3l:
Extend Test: Flush more exotic write patterns in serial write, such as attribute-only series.

Demonstrates and fixes another ADIOS1 backend bug (open files either if attributes or dataset are going to be created).

Implement a `backendName()` property in `IOHandler`s for development/debugging purposes.

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -673,6 +673,7 @@ void fileBased_write_test(const std::string & backend)
         std::shared_ptr< uint64_t > positionOffset_local_1(new uint64_t);
         e_1["positionOffset"]["x"].resetDataset(Dataset(determineDatatype(positionOffset_local_1), {4}));
 
+        //o.setOpenPMDextension(1); // this would be before the first flush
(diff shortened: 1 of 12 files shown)


test/SerialIOTest.cpp
```diff
@@ -723,6 +724,10 @@ void fileBased_write_test(const std::string & backend)
 
         o.setOpenPMDextension(1);
         o.iterations[3].setTime(static_cast< double >(3));
+        o.iterations[4].setTime(static_cast< double >(4));
+        if( o.backend() != "ADIOS1" )  // FIXME fails with ADIOS1
```
ax3l: Error message:
```
ERROR: MPI open failed for ../samples/subdir/serial_fileBased_write00000004.bp: 'File not found: ../samples/subdir/serial_fileBased_write00000004.bp'
ERROR: File open failed: ../samples/subdir/serial_fileBased_write00000004.bp
  what():  [ADIOS1] Internal error: Failed to open_read ADIOS file
```

PR #675, opened 2020-01-27
Travis: Add ADIOS1 to Sanitizers

ax3l:
And add suppression file as needed.


ax3l:
travis CI removed #823

PR #679, opened 2020-01-29
added accesstype check before adding read operations

guj:
this is to avoid have read operations showing up in (multiple) write flushes
for the same iteration

### DIFF
--- a/.rodare.json
+++ b/.rodare.json
@@ -14,6 +14,12 @@
             "name": "Huebl, Axel",
             "affiliation": "Lawrence Berkeley National Laboratory",
             "orcid": "0000-0003-1943-7141"
+        },
+        {
+            "affiliation": "Lawrence Berkeley National Laboratory",
+            "name": "Gu, Junmin",
+            "orcid": "0000-0002-1521-8534",
+            "type": "Other"
         }
     ],
     "contributors": [
@@ -79,12 +85,6 @@
             "affiliation": "Helmholtz-Zentrum Dresden-Rossendorf",
             "name": "Pausch, Richard",
             "orcid": "0000-0001-7990-9564",
-            "type": "Other"
-        },
-        {
-            "affiliation": "Lawrence Berkeley National Laboratory",
-            "name": "Gu, Junmin",
-            "orcid": "0000-0002-1521-8534",
             "type": "Other"
         },
         {
--- a/include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
@@ -93,6 +93,10 @@ namespace openPMD
         std::unordered_map< std::shared_ptr< std::string >, ADIOS_FILE* > m_openReadFileHandles;
         std::unordered_map< ADIOS_FILE*, std::vector< ADIOS_SELECTION* > > m_scheduledReads;
         std::unordered_map< int64_t, std::unordered_map< std::string, Attribute > > m_attributeWrites;
+        /**
+         * Call this function to get adios file id for a Writable. Will create one if does not exist
+         * @return  returns an adios file id. 
+         */	  
         int64_t GetFileHandle(Writable*);
     }; // ADIOS1IOHandlerImpl
 #else
--- a/include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
@@ -93,6 +93,7 @@ namespace openPMD
         std::unordered_map< std::shared_ptr< std::string >, ADIOS_FILE* > m_openReadFileHandles;
         std::unordered_map< ADIOS_FILE*, std::vector< ADIOS_SELECTION* > > m_scheduledReads;
         std::unordered_map< int64_t, std::unordered_map< std::string, Attribute > > m_attributeWrites;
+        int64_t GetFileHandle(Writable*);
     }; // ADIOS1IOHandlerImpl
 #else
     class EXPORT ADIOS1IOHandlerImpl
--- a/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
+++ b/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
@@ -94,6 +94,7 @@ namespace openPMD
         std::unordered_map< std::shared_ptr< std::string >, ADIOS_FILE* > m_openReadFileHandles;
         std::unordered_map< ADIOS_FILE*, std::vector< ADIOS_SELECTION* > > m_scheduledReads;
         std::unordered_map< int64_t, std::unordered_map< std::string, Attribute > > m_attributeWrites;
+        int64_t GetFileHandle(Writable*);
         MPI_Comm m_mpiComm;
         MPI_Info m_mpiInfo;
     }; // ParallelADIOS1IOHandlerImpl
--- a/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
+++ b/include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
@@ -94,6 +94,10 @@ namespace openPMD
         std::unordered_map< std::shared_ptr< std::string >, ADIOS_FILE* > m_openReadFileHandles;
         std::unordered_map< ADIOS_FILE*, std::vector< ADIOS_SELECTION* > > m_scheduledReads;
         std::unordered_map< int64_t, std::unordered_map< std::string, Attribute > > m_attributeWrites;
+        /**
+         * Call this function to get adios file id for a Writable. Will create one if does not exist
+         * @return  returns an adios file id. 
+         */	  
         int64_t GetFileHandle(Writable*);
         MPI_Comm m_mpiComm;
         MPI_Info m_mpiInfo;
--- a/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
@@ -376,7 +376,7 @@ CommonADIOS1IOHandlerImpl::createFile(Writable* writable,
         /* defer actually opening the file handle until the first Operation::WRITE_DATASET occurs */
         m_existsOnDisk[m_filePaths[writable]] = false;
 
-	int64_t fd = GetFileHandle(writable);
+    int64_t fd = GetFileHandle(writable);
     }
 }
 
@@ -757,12 +757,12 @@ int64_t CommonADIOS1IOHandlerImpl::GetFileHandle(Writable* writable)
       // write all opening handles and close them
       {
         for( auto& group : m_attributeWrites )
-	  for( auto& att : group.second )
-	    flush_attribute(group.first, att.first, att.second);
-	
-	for( auto& f : m_openWriteFileHandles )
-	  close(f.second);
-	m_openWriteFileHandles.clear();
+      for( auto& att : group.second )
+        flush_attribute(group.first, att.first, att.second);
+    
+    for( auto& f : m_openWriteFileHandles )
+      close(f.second);
+    m_openWriteFileHandles.clear();
       }
 
       fd = open_write(writable);
--- a/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
@@ -759,7 +759,7 @@ int64_t CommonADIOS1IOHandlerImpl::GetFileHandle(Writable* writable)
         for( auto& group : m_attributeWrites )
       for( auto& att : group.second )
         flush_attribute(group.first, att.first, att.second);
-    
+
     for( auto& f : m_openWriteFileHandles )
       close(f.second);
     m_openWriteFileHandles.clear();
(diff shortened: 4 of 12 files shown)


ax3l:
I added a demonstrator of the bug that we are hunting in #680.

According to the failing tests, it is possible that some attributes are not written after the first `.flush()` or that their values are not read in after re-opening a series with a new `Series` object.

src/Iteration.cpp
```diff
@@ -118,6 +118,13 @@ Iteration::flushFileBased(std::string const& filename, uint64_t i)
         IOHandler->enqueue(IOTask(this, pCreate));
     } else
     {
+      if (IOHandler->accessTypeFrontend == AccessType::CREATE)
```
ax3l: if we can somehow determine that the "old" series has been read at least once, we can also try:
```suggestion
      if (IOHandler->accessTypeFrontend == AccessType::CREATE || IOHandler->accessTypeFrontend == AccessType::READ_WRITE)
```

I would generally be more comfortable if we can place this backend-specific work-around in the ADIOS1IOHandlerImpl file(s) instead of here, we could e.g. return the `OPEN_FILE` call that is triggered here.

(And probably have to add a work-around (skip-if-not-open) to the `OPEN_PATH`, too.)

guj: I think read_write should be allowed to read. That's why I did not include it. 

On the other hand, adios does have have both read/write mode

ax3l: you could add
```suggestion
      if (IOHandler->accessTypeFrontend == AccessType::CREATE && IOHandler->backend() == "MPI_ADIOS1")
```
as a really directed frontend hack (for a backend problem).

ax3l: As discussed on Slack, the most general solution to this problem is likely to open files directly in the backend instead of delaying their open.
That means `createFile` should open files already instead of waiting for `openFile`, `writeDataset`, an `writeAttribute`.
That would also reduce the massive open-close calls frequency.

guj: OK added

guj:
Found the cause of the problem.  Not a workaround anymore. 
Need help with travis,   times out for  Appleclang11.0

test/SerialIOTest.cpp
```diff
@@ -722,7 +722,7 @@ void fileBased_write_test(const std::string & backend)
             o.flush();
         }
 
-        o.setOpenPMDextension(1);
+        //o.setOpenPMDextension(1); // this will make iteration 1&2 have extension 0.
```
ax3l: With the root cause found, can we activate this "late" Series-global attribute again, please? The comment behind this was only true for ADIOS1, according to my tests.
If this is only necessary for ADIOS1, let's add another `o.backendName() == "ADIOS1"` if branch here.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -803,17 +809,11 @@ CommonADIOS1IOHandlerImpl::writeAttribute(Writable* writable,
...
     if( res == m_filePaths.end() )
         res = m_filePaths.find(writable->parent);
-    int64_t fd;
-    if( m_openWriteFileHandles.find(res->second) == m_openWriteFileHandles.end() )
-    {
-        fd = open_write(writable);
-        m_openWriteFileHandles[res->second] = fd;
-    } else
-        fd = m_openWriteFileHandles.at(res->second);
+    //int64_t fd = GetFileHandle(writable);
```
ax3l: Let's remove the dead code there:
```suggestion
```

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -749,24 +746,33 @@ CommonADIOS1IOHandlerImpl::deleteAttribute(Writable*,
...
+      fd = m_openWriteFileHandles.at(res->second);
+
+    return fd;
+}
+void
+CommonADIOS1IOHandlerImpl::writeDataset(Writable* writable,
+                                  Parameter< Operation::WRITE_DATASET > const& parameters)
+{
+    if( m_handler->accessTypeBackend == AccessType::READ_ONLY )
+        throw std::runtime_error("[ADIOS1] Writing into a dataset in a file opened as read only is not possible.");
```
ax3l: Typo?
```suggestion
        throw std::runtime_error("[ADIOS1] Writing into a dataset in a file opened as read-only is not possible.");
```

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -372,9 +372,11 @@ CommonADIOS1IOHandlerImpl::createFile(Writable* writable,
 
         /* our control flow allows for more than one open file handle
          * if multiple files are opened with the same group, data might be lost */
-        m_groups[m_filePaths[writable]] = initialize_group(name);
+
         /* defer actually opening the file handle until the first Operation::WRITE_DATASET occurs */
         m_existsOnDisk[m_filePaths[writable]] = false;
+
+        /*int64_t fd = */GetFileHandle(writable);
```
ax3l:
```suggestion
        GetFileHandle(writable);
```

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
```diff
@@ -94,6 +94,7 @@ namespace openPMD
         std::unordered_map< std::shared_ptr< std::string >, ADIOS_FILE* > m_openReadFileHandles;
         std::unordered_map< ADIOS_FILE*, std::vector< ADIOS_SELECTION* > > m_scheduledReads;
         std::unordered_map< int64_t, std::unordered_map< std::string, Attribute > > m_attributeWrites;
+        int64_t GetFileHandle(Writable*);
```
ax3l, line 101: Can you please add a doxygen string that describes the side-effects and return-value of this function?
We call this in several places without handling the return value, which might lead to confusion later on.

include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
```diff
@@ -93,6 +93,7 @@ namespace openPMD
         std::unordered_map< std::shared_ptr< std::string >, ADIOS_FILE* > m_openReadFileHandles;
         std::unordered_map< ADIOS_FILE*, std::vector< ADIOS_SELECTION* > > m_scheduledReads;
         std::unordered_map< int64_t, std::unordered_map< std::string, Attribute > > m_attributeWrites;
+        int64_t GetFileHandle(Writable*);
```
ax3l, line 100: Can you please add a doxygen string that describes the side-effects and return-value of this function?
We call this in several places without handling the return value, which might lead to confusion later on.

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -58,19 +58,26 @@ ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
...
+    // stored with different orders in different processors.
+    // which  caused trouble with  close(), which is collective
+    // so I just sort by file name to force  all processors  close
+    // all the fids in the same order
+    std::map< std::string, int64_t> allFiles;
         for( auto& f : m_openWriteFileHandles )
-            close(f.second);
+            allFiles[*(f.first)] = f.second;
+
+    for (auto p :  allFiles)
```
ax3l:
```suggestion
    for (auto& p :  allFiles)
```

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -58,19 +58,26 @@ ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
...
-        for( auto& f : m_filePaths )
-            if( m_openWriteFileHandles.find(f.second) == m_openWriteFileHandles.end() )
-                m_openWriteFileHandles[f.second] = open_write(f.first);
-
+    // unordered map caused the value of the same container
+    // stored with different orders in different processors.
+    // which  caused trouble with  close(), which is collective
+    // so I just sort by file name to force  all processors  close
+    // all the fids in the same order
+    std::map< std::string, int64_t> allFiles;
```
ax3l: The next three lines can probably be [shortened to](https://en.cppreference.com/w/cpp/container/map/map):
```suggestion
    std::map< std::string, int64_t> allFiles(
        m_openWriteFileHandles.begin(), m_openWriteFileHandles.end() );
```

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -58,19 +58,26 @@ ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
...
         for( auto& f : m_openWriteFileHandles )
-            close(f.second);
+            allFiles[*(f.first)] = f.second;
+
+    for (auto p :  allFiles)
+      {
+        auto fid =  p.second;
```
ax3l:
```suggestion
        auto const fid = p.second;
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -162,6 +162,10 @@ HDF5IOHandlerImpl::createPath(Writable* writable,
         groups.push(node_id);
         for( std::string const& folder : auxiliary::split(path, "/", false) )
         {
+            htri_t found = H5Lexists(groups.top(), folder.c_str(), H5P_DEFAULT);
```
ax3l: const-correctness and comment
```suggestion
            // avoid creation of paths that already exist
            htri_t const found = H5Lexists(groups.top(), folder.c_str(), H5P_DEFAULT);
```

src/Iteration.cpp
```diff
@@ -118,6 +118,14 @@ Iteration::flushFileBased(std::string const& filename, uint64_t i)
         IOHandler->enqueue(IOTask(this, pCreate));
     } else
     {
+      if ((IOHandler->accessTypeFrontend == AccessType::CREATE)
```
ax3l: is this work-around still needed?

guj: yes

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
guj: I would prefer to stay put.

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
```diff
@@ -58,19 +58,26 @@ ParallelADIOS1IOHandlerImpl::~ParallelADIOS1IOHandlerImpl()
...
+    // stored with different orders in different processors.
+    // which  caused trouble with  close(), which is collective
+    // so I just sort by file name to force  all processors  close
+    // all the fids in the same order
+    std::map< std::string, int64_t> allFiles;
         for( auto& f : m_openWriteFileHandles )
-            close(f.second);
+            allFiles[*(f.first)] = f.second;
+
+    for (auto& p :  allFiles)
```
ax3l:
```suggestion
    for (auto const & p : allFiles)
```

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -749,24 +746,33 @@ CommonADIOS1IOHandlerImpl::deleteAttribute(Writable*,
...
     auto res = m_filePaths.find(writable);
     if( res == m_filePaths.end() )
         res = m_filePaths.find(writable->parent);
     int64_t fd;
+
     if( m_openWriteFileHandles.find(res->second) == m_openWriteFileHandles.end() )
     {
-        fd = open_write(writable);
-        m_openWriteFileHandles[res->second] = fd;
+      std::string  name  = *(res->second);
```
ax3l:
```suggestion
      std::string const name = *(res->second);
```

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
guj: added

include/openPMD/IO/ADIOS/ADIOS1IOHandlerImpl.hpp
guj: added

test/SerialIOTest.cpp
guj: ok, found a solution.

src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
guj: changed

test/SerialIOTest.cpp
```diff
@@ -725,7 +724,7 @@ void fileBased_write_test(const std::string & backend)
         o.setOpenPMDextension(1);
         o.iterations[3].setTime(static_cast< double >(3));
         o.iterations[4].setTime(static_cast< double >(4));
-        o.flush();
+        //o.flush();
```
ax3l: removing that is confusing because it is a late flush. flushing here should be very similar to the destructor which is called next and flushes, too.

what I thought we chatted about was to set for ADIOS the `setOpenPMDextension` before the first iteration's flush occurs, because it seams to set it at that point and cannot be overwritten later on?

guj: Oh I thought I explained that the attributes in adios1  are written when the series is deleted, so it is not really an overwritten issue. 
I can put this flush back. Thought it is not needed. Is there a reason to have it there?

guj: OK I looked in the history flush was there. I put it back. Sorry

ax3l: Nothing to be sorry about, thanks a lot :)

ax3l:
This fixes #680, thank you so much!

.rodare.json
```diff
@@ -14,6 +14,12 @@
             "name": "Huebl, Axel",
             "affiliation": "Lawrence Berkeley National Laboratory",
             "orcid": "0000-0003-1943-7141"
+        },
+        {
+            "affiliation": "Lawrence Berkeley National Laboratory",
+            "name": "Gu, Junmin",
+            "orcid": "0000-0002-1521-8534",
+            "type": "Other"
```
ax3l, line 22: fixed in `dev` via 4ca1c9c

PR #680, opened 2020-01-29
ADIOS1: Skip with Iterations Test

ax3l:
Extend the skipped/partial-write test to include multiple iterations. This demonstrates an ADIOS1 backend bug seen in WarpX.

@guj We can cherry-pick this unit test on your PR in #679 to see if this fixes it :)


ax3l:
That's a useful test, I merged it (8757b90966fcac8ab8788667b3b7c4dd382ca912) with the failing ADIOS1 excluded (so we can keep it as quality-test for all other backends) and simplified this PR to just activate ADIOS1 in it again.

ax3l:
Added to #679, thank you @guj ! :rocket: :sparkles:

PR #690, opened 2020-02-28
Doc: Citation

ax3l:
Add a citation section. This section guides users to acknowledge openPMD and the openPMD-api.

### DIFF
--- a/docs/source/citation.rst
+++ b/docs/source/citation.rst
@@ -1,0 +1,93 @@
+.. _acknowledgement:
+
+Citation
+========
+
+openPMD (Open Standard for Particle-Mesh Data Files) is a community project with many people contributing to it.
+If you use openPMD and/or openPMD related software in your work, please credit it when publishing and/or presenting work performed with it in order to give back to the community.
+
+openPMD-standard
+----------------
+
+The central definition of **openPMD is the meta data schema** defined in `openPMD/openPMD-standard <https://github.com/openPMD/openPMD-standard>`_.
+The most general reference to openPMD is:
+
+.. tip::
+
+   Axel Huebl, Remi Lehe, Jean-Luc Vay, David P. Grote, Ivo F. Sbalzarini, Stephan Kuschel, David Sagan, Christopher Mayes, Frederic Perez, Fabian Koller, and Michael Bussmann.
+   *"openPMD: A meta data standard for particle and mesh based data,"*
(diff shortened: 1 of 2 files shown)


docs/source/citation.rst
```diff
@@ -0,0 +1,49 @@
...
+
+openPMD-api is a reference implementation of the openPMD-standard for popular data formats.
+
+openPMD-api is a scientific software and cited generally as
+
+    Fabian Koller, Franz Poeschel, Junmin Gu, and Axel Huebl.
+    "openPMD-api: C++ & Python API for Scientific I/O with openPMD"
+    `DOI:10.14278/rodare.27 <https://doi.org/10.14278/rodare.27>`_ (2018)
+
+As a software citation, you almost always want to refer to a specific version of openPMD-api in your works:
```
LDAmorim: This is a great PR @ax3l ! And it really helps users.
I was wondering if it would be interesting to explain why it is so useful to know the code versions (not just of openPMD-api, but of all those used). Feel free to not like my suggestion, but I'm thinking of something in the lines of:
```suggestion
It is good scientific practice to document all software versions in, e.g. a methods section, so you, reviewers and collaborators can reproduce your results.
In a software citation, you almost always want to refer to a specific version of openPMD-api in your work, as illustrated below.
```
What do you think?

docs/source/citation.rst
```diff
@@ -0,0 +1,49 @@
...
+
+`Release package page <https://github.com/openPMD/openPMD-api/releases>`_ ...
+Add the version number to the cited title, e.g.
+
+    ... (author list as above) ...
+    "openPMD-api 0.10.3: C++ & Python API for Scientific I/O with openPMD,"
+    `DOI:10.14278/rodare.209 <https://doi.org/10.14278/rodare.209>`_ (2019)
+
+Use the openPMD-api functionality ... to get and document software versions of dependent I/O backends.
+It is good scientific practice to document all software versions in, e.g. a methods section.
```
LDAmorim: I moved this line to before  this last section. Because, to me, it reads like the motivation for including the section in this documentation page. But I am not a good writer or anything, so feel free to disagree!

LDAmorim (changes requested):

docs/source/citation.rst
ax3l: That's a great point, thank you. I like giving reason with guidelines, yay!

docs/source/citation.rst
ax3l: Sure thing, thanks!

PR #706, opened 2020-03-26
Add particle info in serial example read

LDAmorim:
Hi all,
This mini PR is just to add the option of reading a scalar from a `ParticleSpecies Record` (as an example) in the files:

- **examples/2_read_serial.cpp**
- **examples/2_read_serial.py**

I think it is very useful to add it because it can really help the new users (such as myself) understand it better.

I am not sure if this is the more clean/elegant way to show how to do that to new users.
Or even if these changes should be added instead to the Docs first read entry. 
What do you think?

ps: I am having a minor issue testing the cpp file (hence the WIP).

Thanks again for openPMD!
Diana

### DIFF
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -56,8 +56,8 @@ int main()
         }
     }
 
-    std::pair<std::string,openPMD::ParticleSpecies> s_e = *i.particles.begin();
-    std::pair<std::string,openPMD::Record> s_c = *s_e.second.begin();
+    std::pair<std::string, openPMD::ParticleSpecies> s_e = *i.particles.begin();
+    std::pair<std::string, openPMD::Record> s_c = *s_e.second.begin();
     auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
     cout << "And first particle in particle species " << s_e.first
          << " has " << s_c.first << " = " << charge.get()[0];
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -59,7 +59,8 @@ int main()
     std::pair<std::string,openPMD::ParticleSpecies> s_e = *i.particles.begin();
     std::pair<std::string,openPMD::Record> s_c = *s_e.second.begin();
     auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
-    cout << "And " << s_c.first << " = " << charge;
+    cout << "And first particle in particle species " << s_e.first
+         << " has " << s_c.first << " = " << charge.get()[0];
     cout << '\n';
 
     MeshRecordComponent E_x = i.meshes["E"]["x"];
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -58,6 +58,7 @@ int main()
 
     openPMD::ParticleSpecies electrons = i.particles["electrons"];
     std::shared_ptr<double> charge = electrons["charge"][openPMD::RecordComponent::SCALAR].loadChunk<double>();
+    series.flush();
     cout << "And the first electron particle has a charge = " << charge.get()[0];
     cout << '\n';
 
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -55,7 +55,7 @@ int main()
             cout << '\n';
         }
     }
-    
+
     std::pair<std::string,openPMD::ParticleSpecies> s_e = *i.particles.begin();
     std::pair<std::string,openPMD::Record> s_c = *s_e.second.begin();
     auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -58,7 +58,7 @@ int main()
     
     std::pair<std::string,openPMD::ParticleSpecies> s_e = *i.particles.begin();
     std::pair<std::string,openPMD::Record> s_c = *s_e.second.begin();
-    auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<amrex::Real>().get()[0];
+    auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
     cout << "And " << s_c.first << " = " << charge;
     cout << '\n';
 
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -58,7 +58,7 @@ int main()
 
     std::pair<std::string, openPMD::ParticleSpecies> s_e = *i.particles.begin();
     std::pair<std::string, openPMD::Record> s_c = *s_e.second.begin();
-    auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
+    std::shared_ptr< double > charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>();
     cout << "And first particle in particle species " << s_e.first
          << " has " << s_c.first << " = " << charge.get()[0];
     cout << '\n';
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -50,7 +50,7 @@ int main()
     cout << "Iteration 100 contains " << i.particles.size() << " particle species:";
     for( auto const& ps : i.particles ) {
         cout << "\n\t" << ps.first;
-        for( auto const& r : ps.second ){
+        for( auto const& r : ps.second ) {
             cout << "\n\t" << r.first;
             cout << '\n';
         }
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -48,7 +48,7 @@ int main()
         cout << "\n\t" << m.first;
     cout << '\n';
     cout << "Iteration 100 contains " << i.particles.size() << " particle species:";
-    for( auto const& ps : i.particles ){
+    for( auto const& ps : i.particles ) {
         cout << "\n\t" << ps.first;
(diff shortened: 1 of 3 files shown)


examples/2_read_serial.py
```diff
@@ -29,6 +29,15 @@
...
     for ps in i.particles:
         print("\t {0}".format(ps))
+        print("With records:")
+        for r in i.particles[ps]:
+            print("\t {0}".format(r))
+
+    # printing a scalar value
+    s_e=[ps for ps in i.particles][0]
+    s_c=[r for r in i.particles[s_e]][0]
+    charge=i.particles[s_e][s_c][openpmd_api.Mesh_Record_Component.SCALAR][0]
```
ax3l: PEP8 style checker:
```suggestion
    s_e = [ps for ps in i.particles][0]
    s_c = [r for r in i.particles[s_e]][0]
    charge=i.particles[s_e][s_c][openpmd_api.Mesh_Record_Component.SCALAR][0]
```

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,18 @@ int main()
...
+        for( auto const& r : ps.second ){
+            cout << "\n\t" << r.first;
+            cout << '\n';
+        }
+    }
+
+    std::pair<std::string,openPMD::ParticleSpecies> s_e = *i.particles.begin();
+    std::pair<std::string,openPMD::Record> s_c = *s_e.second.begin();
+    auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
+    cout << "And " << s_c.first << " = " << charge;
```
ax3l:
```suggestion
    cout << "And first particle in particle species " << s_e.first
         << " has " << s_c.first << " = " << charge.get()[0];
```

examples/2_read_serial.py
```diff
@@ -29,6 +29,15 @@
...
+    print("And ", s_c, " = ", charge)
```
ax3l:
```suggestion
    print("And first particle in particle species {0} has {1} = {2}"
          .format(s_e, s_c, charge))
```

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,18 @@ int main()
         cout << "\n\t" << m.first;
     cout << '\n';
     cout << "Iteration 100 contains " << i.particles.size() << " particle species:";
-    for( auto const& ps : i.particles )
+    for( auto const& ps : i.particles ){
```
ax3l: style consistency:
```suggestion
    for( auto const& ps : i.particles ) {
```

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,18 @@ int main()
...
         cout << "\n\t" << ps.first;
+        for( auto const& r : ps.second ){
```
ax3l: style consistency:
```suggestion
        for( auto const& r : ps.second ) {
```

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,18 @@ int main()
...
+            cout << "\n\t" << r.first;
+            cout << '\n';
+        }
+    }
+
+    std::pair<std::string,openPMD::ParticleSpecies> s_e = *i.particles.begin();
+    std::pair<std::string,openPMD::Record> s_c = *s_e.second.begin();
```
ax3l: style consistency:
```suggestion
    std::pair<std::string, openPMD::ParticleSpecies> s_e = *i.particles.begin();
    std::pair<std::string, openPMD::Record> s_c = *s_e.second.begin();
```

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,18 @@ int main()
...
+    auto charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>().get()[0];
```
ax3l: Maybe this return type is useful to know:
```suggestion
    std::shared_ptr< double > charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>();
```

examples/2_read_serial.py
```diff
@@ -29,6 +29,15 @@
         len(i.particles)))
     for ps in i.particles:
         print("\t {0}".format(ps))
+        print("With records:")
+        for r in i.particles[ps]:
+            print("\t {0}".format(r))
+
+    # printing a scalar value
+    s_e=[ps for ps in i.particles][0]
+    s_c=[r for r in i.particles[s_e]][0]
```
ax3l: PEP8 style checker:
```suggestion
    s_e = [ps for ps in i.particles][0]
    s_c = [r for r in i.particles[s_e]][0]
```

ax3l (commented):
Thank you! :+1: 
Added a first review for style and more detailed output.

examples/2_read_serial.py
LDAmorim: This is really a cool way to print it. Thanks for this correction @ax3l !

examples/2_read_serial.cpp
LDAmorim: Thanks for changing this!

examples/2_read_serial.py
```diff
@@ -29,6 +29,16 @@
...
     for ps in i.particles:
         print("\t {0}".format(ps))
+        print("With records:")
+        for r in i.particles[ps]:
+            print("\t {0}".format(r))
+
+    # printing a scalar value
+    s_e = [ps for ps in i.particles][0]
+    s_c = [r for r in i.particles[s_e]][0]
+    charge=i.particles[s_e][s_c][openpmd_api.Mesh_Record_Component.SCALAR][0]
```
ax3l: PEP8 style check:
```suggestion
    charge = i.particles[s_e][s_c][openpmd_api.Mesh_Record_Component.SCALAR][0]
```

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,19 @@ int main()
...
+            cout << "\n\t" << r.first;
+            cout << '\n';
+        }
+    }
+
+    std::pair<std::string, openPMD::ParticleSpecies> s_e = *i.particles.begin();
+    std::pair<std::string, openPMD::Record> s_c = *s_e.second.begin();
+    std::shared_ptr< double > charge = s_c.second[openPMD::RecordComponent::SCALAR].loadChunk<double>();
+    cout << "And first particle in particle species " << s_e.first
+         << " has " << s_c.first << " = " << charge.get()[0];
```
ax3l: The containers we access here can actually be unsorted and therefore it is not clear that the first species is always an electron and its first record is charge.
But we know the input file, therefore:
```suggestion
    openPMD::ParticleSpecies electrons = i.particles["electrons"];
    std::shared_ptr<double> charge = electrons["charge"][openPMD::RecordComponent::SCALAR].loadChunk<double>();
    cout << "And the first electron particle has a charge = " << charge.get()[0];
```

examples/2_read_serial.py
```diff
@@ -29,6 +29,16 @@
...
+    print("And first particle in particle species {0} has {1} = {2}"
+          .format(s_e, s_c, charge))
```
ax3l: As above:
```suggestion
    electrons = i.particles["electrons"]
    charge = electrons["charge][openpmd_api.Mesh_Record_Component.SCALAR]
    print("And the first electron particle has a charge {}"
          .format(charge[0]))
```

examples/2_read_serial.cpp
LDAmorim: Oh good point! It is true that it is unordered. Do you think it would be more important to ask for "electrons" or a first species, regardless of which it is?

LDAmorim: I think you are right, it is nicer to not generalize this completely!
Thanks @ax3l !!

examples/2_read_serial.cpp
```diff
@@ -48,8 +48,17 @@ int main()
...
         cout << "\n\t" << ps.first;
+        for( auto const& r : ps.second ) {
+            cout << "\n\t" << r.first;
+            cout << '\n';
+        }
+    }
+
+    openPMD::ParticleSpecies electrons = i.particles["electrons"];
+    std::shared_ptr<double> charge = electrons["charge"][openPMD::RecordComponent::SCALAR].loadChunk<double>();
+    cout << "And the first electron particle has a charge = " << charge.get()[0];
```
ax3l, line 62: there is only a small bug left: we have to `flush()` before we can access data from `loadChunk`:
```suggestion
    series.flush();
    cout << "And the first electron particle has a charge = " << charge.get()[0];
```

examples/2_read_serial.py
```diff
@@ -29,6 +29,15 @@
...
     for ps in i.particles:
         print("\t {0}".format(ps))
+        print("With records:")
+        for r in i.particles[ps]:
+            print("\t {0}".format(r))
+
+    # printing a scalar value
+    electrons = i.particles["electrons"]
+    charge = electrons["charge"][openpmd_api.Mesh_Record_Component.SCALAR]
+    print("And the first electron particle has a charge {}"
```
ax3l, line 40: same here:
```suggestion
    series.flush()
    print("And the first electron particle has a charge {}"
```

ax3l (changes requested):

ax3l (approved):

examples/2_read_serial.py
```diff
@@ -29,6 +29,16 @@
         len(i.particles)))
     for ps in i.particles:
         print("\t {0}".format(ps))
+        print("With records:")
+        for r in i.particles[ps]:
+            print("\t {0}".format(r))
+
+    # printing a scalar value
+    electrons = i.particles["electrons"]
+    charge = electrons["charge"][openpmd_api.Mesh_Record_Component.SCALAR]
```
ax3l:
```suggestion
    charge = electrons["charge"][io.Mesh_Record_Component.SCALAR]
```

LDAmorim:
Thanks @ax3l !!

ax3l:
Thank you as well! :sparkles:

PR #709, opened 2020-03-27
Catch Throw in Series Ctor/Dtor

ax3l:
Destructors must not throw, which would abort e.g. if `flush()` reports a late problem in `~Series` or a backend's `~...IOHandler`.
Constructors can throw, but we must be able to catch this.

Typical problem: read-only opening of non-existent files throws in Ctor as it flushes all reads.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,8 @@ Features
 Bug Fixes
 """""""""
 
+- ``flush()`` exceptions in ``~Series``/``~..IOHandler`` do not abort anymore #709
+
 Other
 """""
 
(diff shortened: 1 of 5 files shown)


guj:
this does not not necessarily catch the exception from. ~ADIOS2IOHandler(){ flush();}

rather may have caught "fileBased output can not be written with no iterations" 
from Series::flushFileBased() first.

ax3l:
Thank you Junmin! Totally overlooked these.

guj:
Would it be possible to move the code in *.hpp to *.cpp?

ax3l:
For `ADIOS2IOHandler.hpp`? Yes, good idea.

ax3l:
Checked again: we build a defaulted destructor and no member vars in the case ADIOS2 is not enabled, so keeping this in the header is fine here.

PR #710, opened 2020-03-27
CMake: Update MPI Components

ax3l:
From the `FindMPI.cmake` [module documentation](https://github.com/Kitware/CMake/blob/v3.12.0/Modules/FindMPI.cmake#L15-L22):
```
The module exposes the components ``C``, ``CXX``, ``MPICXX`` and ``Fortran``.
Each of these controls the various MPI languages to search for.
The difference between ``CXX`` and ``MPICXX`` is that ``CXX`` refers to the
MPI C API being usable from C++, whereas ``MPICXX`` refers to the MPI-2 C++ API
that was removed again in MPI-3.
```

Clang+MPI: MPI::MPI_C target and `enable_language(C)` needed (basically as before).

macOS MPICH (with work-around flags for C) does NOT work with `MPI::MPI_C` linked because then it passes C compiler flags to CXX: https://github.com/Homebrew/homebrew-core/issues/80465


ax3l:
It looks like this could be worth to be added on Windows, because I can see the following warnings these days (MSVC & Clang):
```
 -- Could NOT find MPI_C (missing: MPI_C_LIB_NAMES MPI_C_HEADER_DIR MPI_C_WORKS) 
-- Could NOT find MPI_CXX (missing: MPI_CXX_LIB_NAMES MPI_CXX_HEADER_DIR MPI_CXX_WORKS) 
-- Could NOT find MPI (missing: MPI_C_FOUND MPI_CXX_FOUND) 
    Reason given by package: MPI component 'Fortran' was requested, but language Fortran is not enabled.
```

Update: ah, actually the reason given is just odd, but everything else is also missing.
Fix in https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5801

ax3l:
Merging to fix Homebrew MPICH builds with custom C flags.

PR #711, opened 2020-03-30
Record(Component) scalar/constant/empty

ax3l:
Add public methods to check if a:
- `Record`: only contains a `scalar` component
- `RecordComponent`: is `constant` check
- `RecordComponent`: is extent `empty` check

### DIFF
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -105,11 +105,19 @@ public:
     uint8_t getDimensionality() const;
     Extent getExtent() const;
 
+    /** Create a dataset with regular extent and constant value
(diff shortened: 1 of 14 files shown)


LDAmorim (approved):

include/openPMD/RecordComponent.hpp
```diff
@@ -313,9 +330,9 @@ template< typename T >
 inline void
 RecordComponent::storeChunk(std::shared_ptr<T> data, Offset o, Extent e)
 {
-    if( *m_isConstant )
+    if( constant() )
```
ax3l, line 333: uff, this broke WarpX.

PR #717, opened 2020-04-14
CD: Add Travis Architectures

ax3l:
Adds wheel builds for:
- ~~Linux aarch64~~: https://github.com/ornladios/ADIOS2/issues/2137
- ~~Linux ppc64le~~: https://github.com/pypa/auditwheel/issues/36 https://github.com/pypa/auditwheel/pull/213
- ~~Linux s390x~~: https://github.com/GTkorvo/dill/issues/15
- macOS 10.13 (XCode 9.4 ~2017)

https://docs.travis-ci.com/user/reference/osx/#using-macos

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
         python -m cibuildwheel --output-dir ../wheelhouse
 
     - name: Publish on pypi.org
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' && github.repository == 'openPMD/openPMD-api' && github.ref == 'wheels'
       env:
         TWINE_USERNAME: __token__
         TWINE_PASSWORD: ${{ secrets.pypa_gh_action_upload }}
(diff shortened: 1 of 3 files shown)


chuckatkins:
While the error should probably be fixed I think it's safe to assume that s390x, i.e. IBM z series mainframe, is not a priority architecture for ADIOS.

ax3l:
And `aarch64`? :)

ax3l:
ADIOS1 compiles on s390x!!111 ;-)

chuckatkins:
Aarch64 needs to be addressed, and I assume the issues are one in the same.  My point really though is that s390 specific issues though can be put on the back burner unless it's low hanging fruit.

ax3l:
Fully agree, just looked like a low-hanging fruit to me to make the build for ADIOS2 more robust. For example, if `dill` does encounter an unknown architecture it should gracefully be disabled instead of failing the whole ADIOS2 build, I think.

chuckatkins:
That's certainly the intent.  I likely just borked some of the logic back when I did the build system overhaul 2-3y ago.

PR #733, opened 2020-05-08
Remove noisy ADIOS2 JSON warning

franzpoeschel:
ADIOS2 backend currently always issues a warning when not configured
explicitly (which is the default). Remove that.


ax3l (approved):
Thanks a lot!

PR #740, opened 2020-05-19
Rename AccessType -> Access

ax3l:
Rename the user-facing `AccessType`/`Access_Type` enum classes to `Access`.

Rationale: "type" just like "action" are relatively meaningless words and do only add verbosity. One could have gone for "access_mode" or "mode" but `openPMD::Access::READ` is probably clear enough, especially in combination with the only usage in `openPMD::Series`.

The old names are still available as (now undocumented and) deprecated aliases and will be removed in a future version of the library.

### DIFF
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -71,14 +71,14 @@ public:
 #if openPMD_HAVE_MPI
     AbstractIOHandler(std::string path, Access at, MPI_Comm)
         : directory{std::move(path)},
-          accessTypeBackend{at},
-          accessTypeFrontend{at}
+          m_backendAccess{at},
+          m_frontendAccess{at}
     { }
 #endif
     AbstractIOHandler(std::string path, Access at)
         : directory{std::move(path)},
-          accessTypeBackend{at},
-          accessTypeFrontend{at}
+          m_backendAccess{at},
+          m_frontendAccess{at}
     { }
     virtual ~AbstractIOHandler() = default;
 
@@ -101,8 +101,8 @@ public:
     virtual std::string backendName() const = 0;
 
     std::string const directory;
-    Access const accessTypeBackend;
-    Access const accessTypeFrontend;
+    Access const m_backendAccess;
+    Access const m_frontendAccess;
     std::queue< IOTask > m_work;
 }; // AbstractIOHandler
 
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -21,7 +21,7 @@
 #pragma once
 
 #include "openPMD/config.hpp"
-#include "openPMD/IO/AccessType.hpp"
+#include "openPMD/IO/Access.hpp"
 #include "openPMD/IO/Format.hpp"
 #include "openPMD/IO/IOTask.hpp"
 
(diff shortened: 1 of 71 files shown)


include/openPMD/IO/AbstractIOHandlerHelper.hpp
```diff
@@ -60,8 +60,8 @@ createIOHandler(
  */
 std::shared_ptr< AbstractIOHandler >
 createIOHandler(
-    std::string path,
-    AccessType accessType,
-    Format format,
-    std::string const & options = "{}" );
+        std::string path,
+        Access accessType,
```
franzpoeschel: You probably meant to rename this one, too? `Access access`

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -117,26 +117,26 @@ class AbstractIOHandlerImpl
 
   /** Create a new file in physical storage, possibly overriding an existing file.
    *
-   * The operation should fail if m_handler->accessType is AccessType::READ_ONLY.
```
franzpoeschel, line 120: The documentation here has been outdated since the day we started distinguishing `accessTypeBackend` from `accessTypeFrontend`. Maybe now is a good time to fix that?

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -101,8 +101,8 @@ class AbstractIOHandler
     virtual std::string backendName() const = 0;
 
     std::string const directory;
-    AccessType const accessTypeBackend;
-    AccessType const accessTypeFrontend;
```
franzpoeschel, line 105: Is it on purpose that you did not rename those? `accessBackend` could me misleading, sounds like we're accessing the backend ;) Maybe `backendAccess`? Or just leave it

franzpoeschel (dismissed):
I'm fine with less verbosity. I think there are some files in the ADIOS1 backend that are not updated yet :)

include/openPMD/IO/AbstractIOHandlerHelper.hpp
ax3l: they were not user-facing so I did not bother too much :) but can do.

franzpoeschel: The parameter name in the parallel overload of this is changed, so it's probably good to keep it consistent ;)

ax3l:
@franzpoeschel general question: do you think instead of `Access` the term `Mode` would be better?

Do we use this control for stage/streams, too?
(Or should we use it?)

franzpoeschel:
The former `AccessType` is used in streaming contexts to specify source-sink relations. `Mode` is what ADIOS2 is using, so maybe it's good for consistency, so users of the one library will quickly know their way around the other?

ax3l:
Good info, thanks! Since we abstract the backends intentionally away, it does not matter too much how backend libs name things.

Thinking about it, "mode" is quite a generic term as well and "access" is slightly more specific.

include/openPMD/IO/AbstractIOHandlerHelper.hpp
ax3l: Serious though: shall I rename them or will this lead to more conflicts with your streaming branch? Those internals touch quite a few places.

franzpoeschel: Are you referring to `accessTypeBackend` and `…Frontend`? Think it's no problem if you rename those, there are no streaming-specific usages in the ADIOS2 backend.

include/openPMD/IO/Access.hpp
```diff
@@ -20,15 +20,20 @@
...
+    enum class Access
     {
         READ_ONLY,  //!< open series as read-only, fails if series is not found
         READ_WRITE, //!< open existing series as writable
         CREATE      //!< create new series and truncate existing (files)
-    }; // AccessType
+    }; // Access
+
+
+    using Access_Type OPENPMDAPI_DEPRECATED("Access_Type is deprecated, use Access instead.") = Access;
```
ax3l, line 43: MSVC 19.0.24241.7:
`openPMD/IO/Access.hpp(38): error C2059: syntax error: ''`

ax3l:
@franzpoeschel thx, review comments added :)

PR #746, opened 2020-05-26
Iteration::close

franzpoeschel:
While the openPMD API currently has a data layout (file-based layout) that creates new files for each iteration of data, no explicit method to close such a file currently exists. This PR adds a call `Iteration::close`.

TODO:
- [x] Implement for backends other than ADIOS2
- [x] `Iteration::close` is deferred, like essentially all other calls in the API so far, the file will only be closed upon `flush`ing. 1. I need to document that fact better, 2. should we keep it that way? Maybe something like `close(bool flush = true)` could be helpful?

### DIFF
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -132,7 +132,7 @@ template<>
 struct OPENPMDAPI_EXPORT Parameter< Operation::CLOSE_FILE > : public AbstractParameter
 {
     Parameter() = default;
-    Parameter( Parameter const & p ) : AbstractParameter(), name( p.name ){}
+    Parameter( Parameter const & ) : AbstractParameter() {}
 
     std::unique_ptr< AbstractParameter >
     clone() const override
@@ -140,8 +140,6 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CLOSE_FILE > : public AbstractPar
         return std::unique_ptr< AbstractParameter >(
             new Parameter< Operation::CLOSE_FILE >( *this ) );
     }
-
-    std::string name = "";
 };
 
 template<>
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -132,7 +132,7 @@ template<>
 struct OPENPMDAPI_EXPORT Parameter< Operation::CLOSE_FILE > : public AbstractParameter
 {
     Parameter() = default;
-    Parameter( Parameter const & p ) : AbstractParameter(), name( p.name ){};
+    Parameter( Parameter const & p ) : AbstractParameter(), name( p.name ){}
 
     std::unique_ptr< AbstractParameter >
     clone() const override
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -46,6 +46,7 @@ OPENPMDAPI_EXPORT_ENUM_CLASS(Operation)
 {
     CREATE_FILE,
     OPEN_FILE,
+    CLOSE_FILE,
     DELETE_FILE,
 
     CREATE_PATH,
@@ -122,6 +123,22 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::OPEN_FILE > : public AbstractPara
     {
         return std::unique_ptr< AbstractParameter >(
             new Parameter< Operation::OPEN_FILE >(*this));
+    }
+
+    std::string name = "";
+};
+
+template<>
+struct OPENPMDAPI_EXPORT Parameter< Operation::CLOSE_FILE > : public AbstractParameter
+{
+    Parameter() = default;
+    Parameter( Parameter const & p ) : AbstractParameter(), name( p.name ){};
+
+    std::unique_ptr< AbstractParameter >
+    clone() const override
+    {
+        return std::unique_ptr< AbstractParameter >(
+            new Parameter< Operation::CLOSE_FILE >( *this ) );
     }
 
     std::string name = "";
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -92,7 +92,8 @@ public:
      * alterations will be made to this iteration.
      * @return Reference to modified iteration.
      */
-    Iteration& close( );
+    Iteration &
+    close( bool flush = true );
 
     bool
     closed() const;
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -111,8 +111,7 @@ private:
     void flushFileBased(std::string const&, uint64_t);
     void flushGroupBased(uint64_t);
     void flush();
-    void
-    read();
+    void read();
 
     enum class CloseStatus
     {
@@ -130,9 +129,6 @@ private:
     std::shared_ptr< CloseStatus > m_closed =
         std::make_shared< CloseStatus >( CloseStatus::Open );
 
-    virtual void
-    linkHierarchy( std::shared_ptr< Writable > const & w );
-
     /**
      * @brief Verify that a closed iteration has not been wrongly accessed.
      *
@@ -141,9 +137,12 @@ private:
      */
     bool
     verifyClosed() const;
-}; // Iteration
 
-extern template float
+    virtual void linkHierarchy(std::shared_ptr< Writable > const& w);
+};  // Iteration
+
+extern template
+float
 Iteration::time< float >() const;
 
 extern template
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -98,25 +98,35 @@ public:
     bool
     closed() const;
 
+    bool
+    closedByWriter() const;
+
     Container< Mesh > meshes;
-    Container< ParticleSpecies > particles; //particleSpecies?
+    Container< ParticleSpecies > particles; // particleSpecies?
 
     virtual ~Iteration() = default;
+
 private:
     Iteration();
 
     void flushFileBased(std::string const&, uint64_t);
     void flushGroupBased(uint64_t);
     void flush();
-    void read();
+    void
+    read();
+
+    /*
+     * Indicates that an iteration has been logically closed.
+     * Will be physically closed upon next flush.
+     */
+    std::shared_ptr< bool > m_closed = std::make_shared< bool >( false );
 
     /*
      * Once an iteration is declared finalized, we do not flush it any longer
      * to avoid reading undefined data if a backend implements optimizations
      * based on this information.
      */
-    std::shared_ptr< bool > skipFlush =
-        std::make_shared< bool >( false );
+    std::shared_ptr< bool > skipFlush = std::make_shared< bool >( false );
 
     virtual void linkHierarchy(std::shared_ptr< Writable > const& w);
 };  //Iteration
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -114,19 +114,21 @@ private:
     void
     read();
 
-    /*
-     * Indicates that an iteration has been logically closed in the frontend,
-     * but not necessarily yet in the backend.
-     * Will be propagated to the backend upon next flush.
-     */
-    std::shared_ptr< bool > m_closedInFrontend =
-        std::make_shared< bool >( false );
+    enum class CloseStatus
+    {
+        Open,
+        ClosedInFrontend,
+        ClosedInBackend
+    };
 
     /*
-     * Whether this iteration has been closed in the backend.
+     * An iteration may be logically closed in the frontend,
+     * but not necessarily yet in the backend.
+     * Will be propagated to the backend upon next flush.
+     * Store the current status.
      */
-    std::shared_ptr< bool > m_closedInBackend =
-        std::make_shared< bool >( false );
+    std::shared_ptr< CloseStatus > m_closed =
+        std::make_shared< CloseStatus >( CloseStatus::Open );
 
     virtual void
     linkHierarchy( std::shared_ptr< Writable > const & w );
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -127,11 +127,20 @@ private:
      */
     std::shared_ptr< bool > skipFlush = std::make_shared< bool >( false );
 
-    virtual void linkHierarchy(std::shared_ptr< Writable > const& w);
-};  //Iteration
+    virtual void
+    linkHierarchy( std::shared_ptr< Writable > const & w );
 
-extern template
-float
+    /**
+     * @brief Verify that a closed iteration has not been wrongly accessed.
+     *
+     * @return true If closed iteration had no wrong accesses.
+     * @return false Otherwise.
+     */
+    bool
+    verifyClosed() const;
+}; // Iteration
+
+extern template float
 Iteration::time< float >() const;
 
 extern template
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -87,6 +87,16 @@ public:
      */
     Iteration& setTimeUnitSI(double newTimeUnitSI);
 
+    /**
+     * @brief Declare an iteration finalized. Indicates that no further
+     * alterations will be made to this iteration.
+     * @return Reference to modified iteration.
+     */
+    Iteration& close( );
+
+    bool
+    closed() const;
+
     Container< Mesh > meshes;
     Container< ParticleSpecies > particles; //particleSpecies?
 
@@ -98,6 +108,14 @@ private:
     void flushGroupBased(uint64_t);
     void flush();
     void read();
+
+    /*
+     * Once an iteration is declared finalized, we do not flush it any longer
+     * to avoid reading undefined data if a backend implements optimizations
+     * based on this information.
+     */
+    std::shared_ptr< bool > skipFlush =
+        std::make_shared< bool >( false );
 
     virtual void linkHierarchy(std::shared_ptr< Writable > const& w);
 };  //Iteration
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -115,17 +115,18 @@ private:
     read();
 
     /*
-     * Indicates that an iteration has been logically closed.
-     * Will be physically closed upon next flush.
+     * Indicates that an iteration has been logically closed in the frontend,
+     * but not necessarily yet in the backend.
+     * Will be propagated to the backend upon next flush.
      */
-    std::shared_ptr< bool > m_closed = std::make_shared< bool >( false );
+    std::shared_ptr< bool > m_closedInFrontend =
+        std::make_shared< bool >( false );
 
     /*
-     * Once an iteration is declared finalized, we do not flush it any longer
-     * to avoid reading undefined data if a backend implements optimizations
-     * based on this information.
+     * Whether this iteration has been closed in the backend.
      */
-    std::shared_ptr< bool > skipFlush = std::make_shared< bool >( false );
+    std::shared_ptr< bool > m_closedInBackend =
+        std::make_shared< bool >( false );
 
     virtual void
     linkHierarchy( std::shared_ptr< Writable > const & w );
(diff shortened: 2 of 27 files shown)


ax3l:
Thanks a lot for the PR! I think going for now with `close(bool flush = true)` would be good! Maybe we get #484 at some point for more global control...

include/openPMD/IO/IOTask.hpp
```diff
@@ -127,6 +128,22 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::OPEN_FILE > : public AbstractPara
     std::string name = "";
 };
 
+template<>
+struct OPENPMDAPI_EXPORT Parameter< Operation::CLOSE_FILE > : public AbstractParameter
+{
+    Parameter() = default;
+    Parameter( Parameter const & p ) : AbstractParameter(), name( p.name ){};
```
```suggestion
    Parameter( Parameter const & p ) : AbstractParameter(), name( p.name ){}
```

src/Iteration.cpp
```diff
@@ -97,6 +97,32 @@ Iteration::setTimeUnitSI(double newTimeUnitSI)
     return *this;
 }
 
+Iteration &
+Iteration::close( bool flush )
+{
+    setAttribute( "closed", true );
```
ax3l: do we need an attribute, e.g. for streaming reasons, or would a private member be sufficient?

franzpoeschel: > e.g. for streaming reasons

That's exactly the reason, yes. I use this in the data sink to know which iterations I need not parse any more.

ax3l: Sounds good then :)

franzpoeschel:
Does ADIOS1 have any issues with having multiple instances (i.e. a writer and a reader) interfere with one another? See the comment in the [latest commit](https://github.com/openPMD/openPMD-api/pull/746/commits/29e5251b34ca3a4897a5b66b862edff8336d4638).

franzpoeschel:
HDF5 seems to be working, ADIOS1 is still being a bit moody atm..

ax3l:
Could be a double free.

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -540,8 +540,68 @@ CommonADIOS1IOHandlerImpl::openFile(Writable* writable,
...
+
+            for( auto & sel : scheduled->second )
+                adios_selection_delete( sel );
+            m_scheduledReads.erase( scheduled );
+        }
+        close( handle_read->second );
+        m_openReadFileHandles.erase( handle_read );
+    }
+    m_filePaths.erase( myFile );
+    m_existsOnDisk.erase( myFile->second );
```
ax3l: Looks more logical/safe to me:
```suggestion
    m_existsOnDisk.erase( myFile->second );
    m_filePaths.erase( myFile );
```

test/SerialIOTest.cpp
```diff
@@ -91,8 +91,80 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+        it0.close( /* flush = */ false );
+    }
+    write.flush();
+    #endif
+    // }
+
+    /*
+     * This block will run fine if commented in
+     * But the following block will fail with a segfault
+     * upon adios_select_method ????
```
ax3l: that's because you cannot initialize ADIOS1 twice in the same process.
ADIOS1 uses some nasty static states.

So: no two series that are active at the same time in a process and use each an ADIOS1 backend.
See: https://openpmd-api.readthedocs.io/en/latest/backends/adios1.html#limitations

franzpoeschel: That explains things, was already suspecting something in this direction. So, the basic "close file" functionality should be working anyway for ADIOS1, but we should probably just run a simplified test, such as checking whether the file is on disk?
Alternatively, running an MPI test with two processes could circumvent that, but I'm feeling that would be overkill...

ax3l: Good idea, for ADIOS1 we can use our helper methods to check we found the files on disk from a single rank instead of reading them back early. Let's keep the logic for the other backends as is, it looks good to me.

franzpoeschel:
Writing attributes to indicate files having been closed has been problematic for read mode. I used a boolean flag on `Iteration` now to circumvent that.
I left the attribute in, since it will be necessary for streaming mode later on anyway.
Also fixed the ADIOS1 test.

include/openPMD/Iteration.hpp
```diff
@@ -97,7 +111,21 @@ class Iteration : public Attributable
...
     void flush();
-    void read();
+    void
+    read();
+
+    /*
+     * Indicates that an iteration has been logically closed.
+     * Will be physically closed upon next flush.
+     */
+    std::shared_ptr< bool > m_closed = std::make_shared< bool >( false );
```
ax3l: Can you quickly remind why this cannot be a member? Is some lifetime of the series/iteration preventing this?

include/openPMD/Iteration.hpp
```diff
@@ -97,7 +111,21 @@ class Iteration : public Attributable
...
+
+    /*
+     * Once an iteration is declared finalized, we do not flush it any longer
+     * to avoid reading undefined data if a backend implements optimizations
+     * based on this information.
+     */
+    std::shared_ptr< bool > skipFlush = std::make_shared< bool >( false );
```
ax3l: why do we need both `m_closed` and `[m_]skipFlush`?

src/Series.cpp
```diff
@@ -564,6 +564,10 @@ Series::flushFileBased()
         bool allDirty = dirty;
         for( auto& i : iterations )
         {
+            if ( *i.second.skipFlush )
+            {
+                continue;
```
ax3l, line 608: shouldn't we throw on this in case a users continues to issue API calls on an (accidentally) closed iteration?

franzpoeschel: The flushing procedures haven't been written with the possibility in mind that an iteration might not be present any longer. Flushing an iteration will e.g. always open the corresponding file, irregardless of whether it is still there or not.

include/openPMD/Iteration.hpp
franzpoeschel: Documentation here is outdated, will fix – also maybe the naming should be redone, something like `logicallyClosed` and `physicallyClosed`. `logicallyClosed` means that the iteration has been closed, but the corresponding `flush` has not run yet.

include/openPMD/Iteration.hpp
franzpoeschel: `class Iteration` is designed as a handle, all relevant information should be shared between copies.
```cpp
auto iteration0 = series.iterations[0];
iteration0.close();
```
If the flag were not behind a `shared_ptr`, we would then get:
```cpp
iteration0.m_closed == true
// but
series.iterations[0].m_closed == false
```
(This touches a broader point that we should keep in mind, if we renovate the frontend some day: While most classes in the openPMD API are handles, this is implemented rather in an ad-hoc way currently. I would prefer a design similar to ADIOS2, i.e. have some none-copyable resource like `internal::Iteration` and a user-facing handle `external::Iteration` that acts like a `shared_ptr` to `internal::Iteration`)

include/openPMD/Iteration.hpp
ax3l: then I would just call these states `m_closed` and `m_flushed` or so. (we have no control when something really is done transferring, that is up to the backend we talk to.)

src/Series.cpp
ax3l: Probably the suggested renaming above will already help a bit. Anyway, this looks like something that will be very hard to debug.

Maybe I missed this: When would a closed and flushed iteration be flushed again? When we change global parameters of an overall series such as extension at a late point, maybe? Sounds legit to ignore that and we should maybe write some warnings to stderr when this happens.

include/openPMD/Iteration.hpp
franzpoeschel: Hm, I think `m_flushed` does not sound very clear about what it's doing?
I mean, while you can certainly argue that `m_physicallyClosed` is not *technically* correct, it is "physically closed" from the frontend's perspective. Anything from now on is the backend's problem which the frontend needs not and cannot interfere with any longer. I'd say that's close enough? :D 

Alternatively, what about `m_closedInFrontend` and `m_closedInBackend`?

src/Series.cpp
franzpoeschel: >When would a closed and flushed iteration be flushed again?

The idea behind continuing with the next iteration in that case is exactly to *not* flush it again. I've added sanity checks now to check whether the iteration hasnt been wrongly accessed and we can safely skip it.

include/openPMD/Iteration.hpp
franzpoeschel: Changed this thing into an enumeration that can take the values `Open`, `ClosedInFrontend` and `ClosedInBackend` now.

ax3l:
@franzpoeschel please rebase against latest `dev` (or merge it in) for latest CI updates.

ax3l:
Ah wait, I actually see a memory leak is caught. Is that from ADIOS2? In case it's not wrong usage on our side that causes this we can report it upstream and add a temporary ignore in `.github/ci/sanitizer/clang/Leak.supp`

franzpoeschel:
I intentionally leak memory in the serial IO tests to avoid throwing in a destructor [1](https://github.com/openPMD/openPMD-api/blob/773578af6334a44c0681e513eca19696ab87b4b1/test/SerialIOTest.cpp#L110), [2](https://github.com/openPMD/openPMD-api/blob/773578af6334a44c0681e513eca19696ab87b4b1/test/SerialIOTest.cpp#L163).

Looking at https://github.com/openPMD/openPMD-api/pull/709 again, I should check again my implementation to see whether there is actually a throwing destructor somewhere (the issue which I fixed in here only came up further down the line in the streaming branch, so we might be fine just letting the destructor run normally). Will look into it tomorrow.

franzpoeschel:
I don't know exactly where you saw the memory leak – is it gone now? I reverted the leak in the tests now.

include/openPMD/Iteration.hpp
```diff
@@ -87,6 +87,43 @@ class Iteration : public Attributable
...
+     * @brief Has the iteration been closed?
+     *        A closed iteration may not (yet) be reopened.
+     *
+     * @return Whether the iteration has been closed.
+     */
+    bool
+    closed() const;
+
+    /**
+     * @brief Has the iteration been closed by the writer?
```
ax3l, line 115: from the description alone I do not understand the difference between `closed()` and `closedByWriter()`.

Do you need to add an additional sentence that states that the *reader* will probe this in streaming context?

include/openPMD/Iteration.hpp
```diff
@@ -87,6 +87,43 @@ class Iteration : public Attributable
...
+     *        Background: Upon calling Iteration::close(), the openPMD API
+     *        will add metadata to the iteration in form of an attribute,
+     *        indicating that the iteration has indeed been closed.
+     *        Useful mainly in streaming context.
```
ax3l, line 119: from the description alone I do not understand the difference between `closed()` and `closedByWriter()`.

Do you need to add an additional sentence that states that the *reader* will probe this in a streaming context?
```suggestion
     *        Useful mainly in streaming context for readers.
```

ax3l: from the description alone I do not understand the difference between `closed()` and `closedByWriter()`.

Do you need to add an additional sentence that states that the *reader* will probe this in a streaming context?
```suggestion
     *        Useful mainly in streaming context when a reader
     *        inquires from a writer that it is done writing.
```

include/openPMD/Iteration.hpp
```diff
@@ -99,8 +136,33 @@ class Iteration : public Attributable
     void flush();
     void read();
 
+    enum class CloseStatus
```
ax3l, line 144: Please see https://github.com/openPMD/openPMD-api/blob/beea53937ec1b4774ab087ae24f529d8df0c4cc1/include/openPMD/IO/Access.hpp#L28-L35 for an enum class doxygen string.

include/openPMD/Iteration.hpp
```diff
@@ -99,8 +136,33 @@ class Iteration : public Attributable
...
+        std::make_shared< CloseStatus >( CloseStatus::Open );
+
+    /**
+     * @brief Verify that a closed iteration has not been wrongly accessed.
+     *
+     * @return true If closed iteration had no wrong accesses.
+     * @return false Otherwise.
+     */
+    bool
+    verifyClosed() const;
```
ax3l: I would rename this to `closeViolated()` or `stillClosed()` or similar, `verifyClosed()` is not clear. What does this return if an iteration was never closed? The name would also depend on this.

include/openPMD/Mesh.hpp
```diff
@@ -182,6 +182,14 @@ class Mesh : public BaseRecord< MeshRecordComponent >
 
     void flush_impl(std::string const&) override;
     void read() override;
+
+    /**
```
ax3l: Writing a doxygen string without an element that captures it will propagate the block to the next location where it can "dock". This will lead to undesirable results.
```suggestion
    /*
```

include/openPMD/Mesh.hpp
```diff
@@ -182,6 +182,14 @@ class Mesh : public BaseRecord< MeshRecordComponent >
...
+     * @brief Verify that a mesh in a closed iteration has not
+     *        been wrongly accessed.
+     *
+     * @return true If closed iteration had no wrong accesses.
+     * @return false Otherwise.
+     */
```
ax3l:
```suggestion
     * @todo needs to be implemented
     */
```

ax3l: is it intentional that particles already have the method but meshes, don't? I think this is missing a declaration here.

include/openPMD/ParticleSpecies.hpp
```diff
@@ -45,6 +45,16 @@ class ParticleSpecies : public Container< Record >
...
+
+    /**
+     * @brief Verify that a particle species in a closed iteration has not
+     *        been wrongly accessed.
+     *
+     * @return true If closed iteration had no wrong accesses.
+     * @return false Otherwise.
+     */
+    bool
+    verifyClosed() const;
```
ax3l: see comment above about naming and behavior when called on a still-open `ParticleSpecies`

include/openPMD/RecordComponent.hpp
```diff
@@ -198,6 +198,16 @@ class RecordComponent : public BaseRecordComponent
...
+
+    /**
+     * @brief Verify that a record component in a closed iteration has not
+     *        been wrongly accessed.
+     *
+     * @return true If closed iteration had no wrong accesses.
+     * @return false Otherwise.
+     */
+    bool
+    verifyClosed() const;
```
ax3l: see comment above please

include/openPMD/backend/BaseRecord.hpp
```diff
@@ -97,7 +97,17 @@ class BaseRecord : public Container< T_elem >
...
+
+    /**
+     * @brief Verify that a base record in a closed iteration has not
+     *        been wrongly accessed.
+     *
+     * @return true If closed iteration had no wrong accesses.
+     * @return false Otherwise.
+     */
+    bool
+    verifyClosed() const;
```
ax3l: please see above

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -399,8 +399,46 @@ HDF5IOHandlerImpl::openFile(Writable* writable,
...
+    {
+        return;
+    }
+    hid_t fileID = fileID_it->second;
+    H5Fclose( fileID );
+    m_openFileIDs.erase( fileID );
+    m_fileIDs.erase( fileID_it );
+
+    /*
+     * std::unordered_map::erase:
```
```suggestion
    /* std::unordered_map::erase:
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -399,8 +399,46 @@ HDF5IOHandlerImpl::openFile(Writable* writable,
...
-HDF5IOHandlerImpl::openPath(Writable* writable,
-                            Parameter< Operation::OPEN_PATH > const& parameters)
+HDF5IOHandlerImpl::closeFile(
+    Writable * writable,
+    Parameter< Operation::CLOSE_FILE > const & )
+{
+    auto fileID_it = m_fileIDs.find( writable );
+    if( fileID_it == m_fileIDs.end() )
+    {
+        return;
```
ax3l, line 412: Should that warn? Did you already see that happen? I fear such things will make other problems hard to debug, so just double-checking.

src/Iteration.cpp
```diff
@@ -19,6 +19,9 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 #include "openPMD/Iteration.hpp"
+
+#include <tuple>
```
ax3l: please include std libs last

src/Iteration.cpp
```diff
@@ -97,6 +102,72 @@ Iteration::setTimeUnitSI(double newTimeUnitSI)
...
+{
+    using bool_type = unsigned char;
+    if( this->IOHandler->m_frontendAccess != Access::READ_ONLY )
+    {
+        setAttribute< bool_type >( "closed", 1u );
+    }
+    *m_closed = CloseStatus::ClosedInFrontend;
+    if( _flush )
+    {
+        Series * s = dynamic_cast< Series * >(
```
ax3l: please validate the `dynamic_cast` is not returning a `nullptr`: #745

test/SerialIOTest.cpp
```diff
@@ -91,8 +91,95 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+        Iteration it0 = write.iterations[ 0 ];
+        auto E_x = it0.meshes[ "E" ][ "x" ];
+        E_x.resetDataset( { Datatype::INT, { 2, 2 } } );
+        E_x.storeChunk( data, { 0, 0 }, { 2, 2 } );
+        it0.close( /* flush = */ false );
+    }
+    write.flush();
+    // }
+
+    if (isAdios1)
```
ax3l, line 113: There is a new `Series::backend()` aka `AbstractIOHandler::backendName()` that you can query instead.

src/Iteration.cpp
```diff
@@ -359,6 +430,30 @@ Iteration::read()
     readAttributes();
 }
 
+bool
+Iteration::verifyClosed() const
+{
+    if( dirty )
+    {
+        return false;
```
ax3l: Can you please add three little test cases, e.g. for a single backend, that trigger and assert the `false` logic from an iteration, particle and mesh?

include/openPMD/Iteration.hpp
```diff
@@ -99,8 +136,33 @@ class Iteration : public Attributable
...
+     * An iteration may be logically closed in the frontend,
+     * but not necessarily yet in the backend.
+     * Will be propagated to the backend upon next flush.
+     * Store the current status.
+     */
+    std::shared_ptr< CloseStatus > m_closed =
+        std::make_shared< CloseStatus >( CloseStatus::Open );
+
+    /**
+     * @brief Verify that a closed iteration has not been wrongly accessed.
```
ax3l: "wrongly accessed" is too vague, better be concrete what is not allowed:
```suggestion
     * @brief Verify that a closed iteration has not been accessed and/or manipulated further.
```

include/openPMD/Iteration.hpp
```diff
@@ -99,8 +136,33 @@ class Iteration : public Attributable
...
+     *
+     * @return true If closed iteration had no wrong accesses.
```
ax3l:
```suggestion
     * @return true If closed iteration had not further been accessed by the user.
```

include/openPMD/Mesh.hpp
franzpoeschel: I implemented the method on the base class instead (i.e. `BaseRecord`) since it also occurs somewhere in the hierarchy for particles. This is dead code, I removed it now.

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: I think we can let that throw instead. Closing a file should require that the file is present in the first place.?

franzpoeschel:
I think I added everything from your review now, and additionally Python bindings

ax3l (approved):
Awesome, thanks a lot! :rocket: :sparkles:

PR #753, opened 2020-06-08
Python Container: Redundant Move

ax3l:
Remove a redundant move warning (`-Wredundant-move`) seen with Cygwin/GNU 9.3.0

### DIFF
--- a/src/binding/python/Container.cpp
+++ b/src/binding/python/Container.cpp
@@ -63,7 +63,8 @@ namespace detail
     >
     py::class_<
         Map,
-        holder_type
+        holder_type,
+        Attributable
     > bind_container(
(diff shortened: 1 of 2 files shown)


src/binding/python/Container.cpp
```diff
@@ -195,7 +195,7 @@ namespace detail
             }
         );
 
-        return std::move(cl);
+        return cl;
```
ax3l, line 199: Oh, now clang 7 & 8 say:
```
local variable 'cl' will be copied despite being returned by name
      [-Wreturn-std-move]
```

ax3l: -> adjusted return type now :)

PR #769, opened 2020-09-05
Fix Iteration/Attributable Assign

### LINKED ISSUE #765: Copy assignment operator of Attributable does not copy all necessary values
[Copy assignment operator definition](https://github.com/openPMD/openPMD-api/blob/66472782e121d2d233dc8733183003668b91150e/src/backend/Attributable.cpp#L50):
```cpp
Attributable&
Attributable::operator=(Attributable const& a)
{
    if( this != &a )
    {
        Attributable tmp(a);
        std::swap(m_attributes, tmp.m_attributes);
    }
    return *this;
}
```
But the class also has the [following members](https://github.com/openPMD/openPMD-api/blob/66472782e121d2d233dc8733183003668b91150e/include/openPMD/backend/Attributable.hpp#L191):
```cpp
    std::shared_ptr< Writable > m_writable;
    /* views into the resources held by m_writable
     * purely for convenience so code that uses these does not have to go through m_wriable-> */
    AbstractFilePosition* abstractFilePosition;
    AbstractIOHandler* IOHandler;
    Writable* parent;
    bool& dirty;
    bool& written;
```
This means that all code accessing these members, will write to the wrong iteration. The following code triggers the error:
```cpp

#include <cstdlib>
#include <iostream>
#include <memory>
#include <numeric>


using std::cout;
using namespace openPMD;

int
main()
{
    using position_t = double;

    // open file for writing
    Series series( "electrons.bp", Access::CREATE );

    Datatype datatype = determineDatatype< position_t >();
    constexpr unsigned long length = 10ul;
    Extent global_extent = { length };
    Dataset dataset = Dataset( datatype, global_extent );
    std::shared_ptr< position_t > local_data(
        new position_t[ length ],
        []( position_t const * ptr ) { delete[] ptr; } );


    std::unique_ptr< Iteration > iteration_ptr;
    for( size_t i = 0; i < 100; ++i )
    {
        std::cout << "starting to write iteration " << i << std::endl;
        if( iteration_ptr )
        {
            *iteration_ptr = series.iterations[ i ];
        }
        else
        {
            // use copy constructor
            iteration_ptr =
                std::make_unique< Iteration >( series.iterations[ i ] );
        }
        Record electronPositions =
            iteration_ptr->particles[ "e" ][ "position" ];

        std::iota( local_data.get(), local_data.get() + length, i * length );
        for( auto const & dim : { "x", "y", "z" } )
        {
            RecordComponent pos = electronPositions[ dim ];
            pos.resetDataset( dataset );
            pos.storeChunk( local_data, Offset{ 0 }, global_extent );
        }
        iteration_ptr->close();
        // force re-flush of previous iterations
        series.flush();
    }

    return 0;
}
```
Output: 
```
starting to write iteration 0
starting to write iteration 1
  what():  [Series] Closed iteration has not been written. This is an internal error.
```

ax3l:
Bug fix for `Attributable` (and `Iteration` et al) to be assignable. Fix #765

- [x] add reproducer to tests
- [x] fix Attributable: make assignable

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -190,12 +190,12 @@ OPENPMD_protected:
 
     std::shared_ptr< Writable > m_writable;
     /* views into the resources held by m_writable
-     * purely for convenience so code that uses these does not have to go through m_wriable-> */
+     * purely for convenience so code that uses these does not have to go through m_writable-> */
     AbstractFilePosition* abstractFilePosition;
     AbstractIOHandler* IOHandler;
     Writable* parent;
-    bool& dirty;
-    bool& written;
+    bool& dirty() const { return m_writable->dirty; }
+    bool& written() const { return m_writable->written; }
 
 private:
     virtual void linkHierarchy(std::shared_ptr< Writable > const& w);
@@ -217,7 +217,7 @@ Attributable::setAttribute(std::string const& key, T const& value)
         throw no_such_attribute_error(out_of_range_msg(key));
     }
 
-    dirty = true;
+    dirty() = true;
     auto it = m_attributes->lower_bound(key);
     if( it != m_attributes->end() && !m_attributes->key_comp()(key, it->first) )
     {
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -184,7 +184,6 @@ close_and_copy_attributable_test( std::string file_ending )
     std::unique_ptr< Iteration > iteration_ptr;
     for( size_t i = 0; i < 100; ++i )
     {
-        std::cout << "starting to write iteration " << i << std::endl;
         if( iteration_ptr )
         {
             *iteration_ptr = series.iterations[ i ];
@@ -199,6 +198,9 @@ close_and_copy_attributable_test( std::string file_ending )
         }
         Record electronPositions =
                 iteration_ptr->particles[ "e" ][ "position" ];
+        // TODO set this automatically to zero if not provided
+        Record electronPositionsOffset =
+                iteration_ptr->particles[ "e" ][ "positionOffset" ];
 
         std::iota( local_data.get(), local_data.get() + length, i * length );
         for( auto const & dim : { "x", "y", "z" } )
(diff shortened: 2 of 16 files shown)


include/openPMD/backend/Attributable.hpp
```diff
@@ -190,12 +190,10 @@ class Attributable
 
     std::shared_ptr< Writable > m_writable;
     /* views into the resources held by m_writable
-     * purely for convenience so code that uses these does not have to go through m_wriable-> */
+     * purely for convenience so code that uses these does not have to go through m_writable-> */
     AbstractFilePosition* abstractFilePosition;
     AbstractIOHandler* IOHandler;
     Writable* parent;
-    bool& dirty;
-    bool& written;
```
ax3l, line 198: Those are the main problem which make the class non-assignable (at least not with valid member init).
Alternative approaches:
- remove members: https://github.com/ax3l/openPMD-api/tree/fix-copyIterationRemoveMember (backup of initial draft)
- expose as public member function (could save us a couple of new `friend` declarations in `Writable`; now in this PR)
- expose as public pointer member (like the rest of those convenience members one has to keep those in a consistent state with the writable)

ax3l:
@franzpoeschel no sure if that is new, but `./SerialIOTests  close_iteration_throws_test` writes
```
[~Series] An error occurred: [Series] Illegal access to iteration 0 that has been closed previously.
```

and `./SerialIOTests  close_iteration_test` writes
```
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
```
to stderr.

Could be due to the intentional `REQUIRE_THROWS` tests, just want to double-check that its ok.

test/SerialIOTest.cpp
```diff
@@ -26,6 +26,15 @@
...
 using namespace openPMD;
 
+#if __cplusplus < 201402L
+namespace {
+    template<typename T, typename... Args>
+    std::unique_ptr<T> make_unique(Args&&... args) {
+        return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+    }
+}
+#endif
```
```suggestion
namespace {
    // only needed until we require C++14 and newer (201402L+)
    template<typename T, typename... Args>
    std::unique_ptr<T> my_make_unique(Args&&... args) {
        return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
    }
}
```

test/SerialIOTest.cpp
```diff
@@ -156,8 +165,71 @@ TEST_CASE( "close_iteration_test", "[serial]" )
...
+        {
+            *iteration_ptr = series.iterations[ i ];
+        }
+        else
+        {
+            using namespace std;
+
+            // use copy constructor
+            iteration_ptr =
+                    make_unique< Iteration >( series.iterations[ i ] );
```
```suggestion
            // use copy constructor
            iteration_ptr =
                    my_make_unique< Iteration >( series.iterations[ i ] );
```

franzpoeschel:
Yep, this is to be expected. This test leaves the `Series` in a state where the `Series::flush` call during `~Series` will throw.

ax3l:
Ok, thanks for double-checking :relaxed:

PR #776, opened 2020-09-08
Fix Windows: CLI.py.help.ls Path

ax3l:
Fix test issue on conda-forge by adding the exact same hints as for other python tests.

Ref.: https://github.com/conda-forge/openpmd-api-feedstock/pull/50


ax3l:
temporarily moved into #840

PR #785, opened 2020-09-29
Add missing CLOSE_FILE IO task to Parallel ADIOS1backend

franzpoeschel:
I missed putting this there, leading to crashes when using `Iteration.close()` with parallel ADIOS1. Implementation is in the shared ADIOS1 implementation, this just adds the missing call.

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -1,9 +1,9 @@
 /* Running this test in parallel with MPI requires MPI::Init.
  * To guarantee a correct call to Init, launch the tests manually.
  */
-#include "openPMD/openPMD.hpp"
 #include "openPMD/auxiliary/Environment.hpp"
-
+#include "openPMD/auxiliary/Filesystem.hpp"
+#include "openPMD/openPMD.hpp"
 #include <catch2/catch.hpp>
 
 #if openPMD_HAVE_MPI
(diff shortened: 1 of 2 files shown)


ax3l:
@franzpoeschel thanks for catching this!
Hm, I am surprised there was not CI test failing on this... There is an entry
```
CXXSPEC="%clang@5.0.0" USE_MPI=ON USE_PYTHON=OFF USE_HDF5=ON USE_ADIOS1=ON USE_ADIOS2=ON OPENPMD_BP_BACKEND=ADIOS1 USE_SAMPLES=ON
```
in our CI matrix that should exactly trigger this, no?

Do we lack a parallel I/O test for this?

franzpoeschel:
If I'm understanding things correctly, compiling the openPMD API with parallel MPI doesn't mean that the parallel ADIOS1 IOHandler is always chosen, it still depends whether the user uses the parallel or non-parallel Series constructor (see [here](https://github.com/openPMD/openPMD-api/blob/1b30ead8c9c2f2c94ad8dced9118b0f908fafecb/src/IO/AbstractIOHandlerHelper.cpp#L78)). So, the test didn't cover the parallel ADIOS1 implementation.

Should we just duplicate the test to the parallel tests?

Side note: I've been trying to figure out why the clang-sanitizer-enabled test won't terminate on the streaming PR – apparently it doesn't terminate on [this PR](https://github.com/openPMD/openPMD-api/pull/785/checks?check_run_id=1183699283) either?

ax3l:
> Should we just duplicate the test to the parallel tests?

Sounds like a lack of coverage, which test would you duplicate exactly? It seems we are not testing the serial constructor for ADIOS sufficiently (SerialIOTests); so we need to add one in `SerialIOTests`; as soon as we add one the current CI matrix should already cover it.

> Side note: I've been trying to figure out why the clang-sanitizer-enabled test won't terminate

Yes, they are slightly too long currently for Travis-CI's hard 50min limit. I have to get creative, something got slower in the setup of that runner. Just ignore it for now, sorry for that (and do a local UBSAN/ASAN test when in doubt, please :) ).

franzpoeschel:
>  It seems we are not testing the serial constructor for ADIOS sufficiently (SerialIOTests)

Isn't the offending code in the parallel IO handler only reached when using a parallel constructor? I don't think we will ever trigger this error within the serial tests, no matter how complete the test matrix is? If we want to, [this test](https://github.com/franzpoeschel/openPMD-api/blob/dcbc6751609ac5bb30e2d8333fc79c14816a26f4/test/SerialIOTest.cpp#L159) will need to be run as a parallel test too, this time calling the parallel `Series` constructors.

ax3l:
Yes, do you like to add such a test? We should cover this case (parallel I/O with `::close()`), since it will be used by all our users in production (in fact, I already added it to WarpX :)).

franzpoeschel:
Added.

test/ParallelIOTest.cpp
```diff
@@ -382,4 +382,93 @@ TEST_CASE( "hzdr_adios_sample_content_test", "[parallel][adios1]" )
...
 }
+
+void
+close_iteration_test( std::string file_ending )
+{
+    int __mpi_rank{ -1 }, __mpi_size{ -1 };
+    MPI_Comm_rank( MPI_COMM_WORLD, &__mpi_rank );
+    MPI_Comm_size( MPI_COMM_WORLD, &__mpi_size );
+    unsigned mpi_rank{ static_cast< unsigned >( __mpi_rank ) },
+        mpi_size{ static_cast< unsigned >( __mpi_size ) };
```
ax3l: Double `__` prefixes are reserved for the compiler & stdlib:
```suggestion
    int i_mpi_rank{ -1 }, i_mpi_size{ -1 };
    MPI_Comm_rank( MPI_COMM_WORLD, &i_mpi_rank );
    MPI_Comm_size( MPI_COMM_WORLD, &i_mpi_size );
    unsigned mpi_rank{ static_cast< unsigned >( i_mpi_rank ) },
        mpi_size{ static_cast< unsigned >( i_mpi_size ) };
```

ax3l (approved):

PR #789, opened 2020-10-01
Iteration: Fix Segfault with ICC during Write

ax3l:
This fixes segfaults with ICC (tested: 19.1 and 21.1).

Trying older releases, this segfault is in here at least since release 0.10.3 (did not go further than that).

### DIFF
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -27,6 +27,7 @@
 #include "openPMD/auxiliary/StringManip.hpp"
 #include "openPMD/backend/Writable.hpp"
 
+#include <exception>
 #include <tuple>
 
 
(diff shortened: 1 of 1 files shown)


src/Iteration.cpp
```diff
@@ -172,10 +173,21 @@ Iteration::closedByWriter() const
 void
 Iteration::flushFileBased(std::string const& filename, uint64_t i)
 {
+    if (parent == nullptr)
+        throw std::runtime_error("[Iteration::flushFileBased] parent is a nullptr");
+    if (parent->attributable == nullptr)
+        throw std::runtime_error("[Iteration::flushFileBased] parent->attributable is a nullptr");
+    if (parent->attributable->parent == nullptr)
+        throw std::runtime_error("[Iteration::flushFileBased] parent->attributable->parent is a nullptr");
+    if (parent->attributable->parent->attributable == nullptr)
```
ax3l: This access provokes the segfault.

PR #790, opened 2020-10-01
Clang-Tidy: Rule of Five

ax3l:
Add a clang-tidy check that makes sure the rule of zero/five is obeyed.

Core Rule guideline C.21
(Note sure if this clang-tidy check also covers C.20... aka "rule of zero")

https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-special-member-functions.html


.clang-tidy
```diff
@@ -1,5 +1,8 @@
+HeaderFilterRegex: '((^(?!\/share\/openPMD\/).*)*include\/openPMD\/.+\.hpp|src\/^(?!binding).+\.cpp$)'
```
ax3l, line 1: Hm, do we filter too many files here?
My IDE shows more locations than the examples...

Issue #792, opened 2020-10-01
cmake build creates only lib64 not lib 


PrometheusPi:
**Description of the documentation bug**
If installing openPMD-api as described in the docs, a `lib64` directory is created not a `lib` directory as stated in the [documentation line 180](https://github.com/openPMD/openPMD-api/blob/dev/docs/source/install/install.rst#L180).  

**How to reproduce it**
When installing openPMD-api via:
```bash
mkdir -p ~/src ~/lib
cd ~/src
git clone https://github.com/openPMD/openPMD-api.git
cd openPMD-api
mkdir build && cd build
cmake .. -DopenPMD_USE_MPI=ON -DCMAKE_INSTALL_PREFIX=~/lib/openPMD-api
make -j $(nproc) install
```
The path `$HOME/lib/openPMD-api/lib` is not created but only `$HOME/lib/openPMD-api/lib64` - thus can be reproduced on summit/ORNL and hemera/HZDR. 

**Expected behavior**
I would have expected the `lib` path to be created or to be a symbolic link to `lib64`. 

**Software Environment**
 - version of openPMD-api: current dev
 - installed openPMD-api via:  from source
 - operating system: 
hemera:
```
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

summit:
```
NAME="Red Hat Enterprise Linux Server"
VERSION="7.6 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.6"
PRETTY_NAME="Red Hat Enterprise Linux Server 7.6 (Maipo)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.6:GA:server"
HOME_URL="https://www.redhat.com/"
[...]

PR #793, opened 2020-10-02
add comment on lib64

PrometheusPi:
This pull request adds a brief comment in the installation documentation that informs the user that on a 64-bit system, the library path might be named `lib64`. This solves the issue #792.

### DIFF
--- a/docs/source/install/install.rst
+++ b/docs/source/install/install.rst
@@ -178,6 +178,7 @@ CMake will summarize the install paths for you before the build step.
    # install prefix         |------------|
    export CMAKE_PREFIX_PATH=$HOME/somepath:$CMAKE_PREFIX_PATH
    export LD_LIBRARY_PATH=$HOME/somepath/lib:$LD_LIBRARY_PATH
+   # Please be aware that on a 64-bit system `.../lib` might need to be replaced by `.../lib64`.
 
    #                change path to your python MAJOR.MINOR version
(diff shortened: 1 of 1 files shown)


PrometheusPi:
FYI: the travis-ci test timed out and thus failed.

docs/source/install/install.rst
```diff
@@ -178,6 +178,7 @@ CMake will summarize the install paths for you before the build step.
    # install prefix         |------------|
    export CMAKE_PREFIX_PATH=$HOME/somepath:$CMAKE_PREFIX_PATH
    export LD_LIBRARY_PATH=$HOME/somepath/lib:$LD_LIBRARY_PATH
+   # Please be aware that on a 64-bit system `.../lib` might need to be replaced by `.../lib64`.
```
ax3l: Just an idea to keep the line short:
```suggestion
   # Note that one some systems, /lib might need to be replaced with /lib64.
```

docs/source/install/install.rst
```diff
@@ -178,6 +178,7 @@ CMake will summarize the install paths for you before the build step.
...
 
    #                change path to your python MAJOR.MINOR version
    export PYTHONPATH=$HOME/somepath/lib/python3.5/site-packages:$PYTHONPATH
```
ax3l: Let's update this while we are on it:
```suggestion
   export PYTHONPATH=$HOME/somepath/lib/python3.8/site-packages:$PYTHONPATH
```

ax3l (approved):
Thank you! :sparkles:

Issue #794, opened 2020-10-08
First Write Example - series.set_machine("str") is not working


ghost:
**Describe the bug**
When following the **First Write** example on openPMD (https://openpmd-api.readthedocs.io/en/0.12.0-alpha/usage/firstwrite.html ) via python / Jupyter
series.set_machine throws error (auto fill can't see the option) 

**To Reproduce**
Compile-able/executable code example to reproduce the problem:

Python:
```py
import openpmd_api as io
import numpy as np

series = io.Series("myOutput/data_001.h5", io.Access.create)

i = series.iterations[42]

series.set_author(
    "i Lomi <i.lomi@smth.de>")
series.set_machine(                      #This is not working
    "Some Machine")
series.set_attribute(
    "Foot", "Schnitzel")
i.set_attribute(
    "hot", True)
```

**Expected behavior**

AttributeError: 'openpmd_api.openpmd_api_cxx.Series' object has no attribute 'set_machine'

**Software Environment**
 - version of openPMD-api: [0.12.0-api]
 - installed openPMD-api via: [conda-forge]
 - operating system: [Ubuntu 18.04.5 LTS]
 - machine: [Not a Public Cluster]
 - name and version of Python implementation: [Python 3.6.9]
 - version of HDF5: [e.g. 1.12.0]
 - version of ADIOS1: [ 1.13.1]
 - version of ADIOS2: [ 2.6.0  ]
 - name and version of MPI: [ (openMPI) 2.1.2  ]

**Additional context**
@PrometheusPi

PrometheusPi:
I could reproduce this locally using `0.12.0-alpha` and `0.8.0-alpha`. 

`series.set_machine()` appears to not exist.

ax3l:
Thanks for the report and proposed PR to expose this functionality in Python #796.
This indeed slipped, let's add this.

The temporary work-around for this is:
```py
series.set_attribute("machine", "machine-site")
```

ax3l:
fixed with #796

PR #796, opened 2020-10-13
add missing python interface (read/write) for machine

PrometheusPi:
This pull request fixes the issue reported in #794. The getter and setter for `machine` were not defined in the python interface (they work properly in c++).

### DIFF
--- a/src/binding/python/Series.cpp
+++ b/src/binding/python/Series.cpp
@@ -131,8 +131,10 @@ void init_Series(py::module &m) {
         .def("set_particles_path", &Series::setParticlesPath)
         .def_property_readonly("author", &Series::author)
         .def("set_author", &Series::setAuthor)
-        .def_property_readonly("machine", &Series::machine)
-        .def("set_machine", &Series::setMachine)
+        .def_property("machine",
+            &Series::machine,
+            &Series::setMachine,
+            "Indicate the machine or relevant hardware that created the file.")
         .def_property_readonly("software", &Series::software)
         .def("set_software", &Series::setSoftware,
             py::arg("name"), py::arg("version") = std::string("unspecified"))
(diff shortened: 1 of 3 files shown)


src/binding/python/Series.cpp
```diff
@@ -131,6 +131,8 @@ void init_Series(py::module &m) {
         .def("set_particles_path", &Series::setParticlesPath)
         .def_property_readonly("author", &Series::author)
         .def("set_author", &Series::setAuthor)
+        .def_property_readonly("machine", &Series::machine)
+        .def("set_machine", &Series::setMachine)
```
ax3l: Very nice, thanks a lot!
I try to transition to read-write properties, because that's a bit more Pythonic (sorry for the dated existing properties in Series).

Can we please set this as
```suggestion
        .def_property("machine",
            &Series::machine,
            &Series::setMachine,
            "Indicate the machine or relevant hardware that created the file.")
```

With that, we are able to read and write them as this:
```py
s.machine = "hemera-hzdr"
print(s.machine)
```

As a last step, please pick a read-write test in `test/python/API/APITest.py` and add two lines (one writing to a series, one reading and asserting the value in the read) to check if writing and reading this works as expected.

Thanks a lot! :tada:

ax3l (commented):
Thank you for the PR, @PrometheusPi, this is great :tada: 

I added a small suggestion inline and please let us add two lines in the Python unit tests to make sure the new functionality is covered.

Please don't forget to update the `docs/source/usage/firstwrite.rst` that uses `machine` as well.
I will refactor further setters in Python's `Series` in a follow-up PR.

src/binding/python/Series.cpp
PrometheusPi: @ax3l I will apply your suggestion. I think this will require an update of the docs as well.

ax3l: Yep, did only put that in the summary: the file `docs/source/usage/firstwrite.rst` needs a slight update.

ax3l (approved):
Thank you, looks great! :rocket: :sparkles:

PR #801, opened 2020-10-19
Be less verbose about missing helper attributes for boolean attributes in ADIOS2 

### LINKED ISSUE #784: ADIOS2: closed attribute read warning
**Describe the bug**
Reading back ADIOS2 data, I see warnings of the type:
```
[ADIOS2] Warning: Attribute with name __is_boolean__/data/0/closed has no type in backend.
[ADIOS2] Warning: Attribute with name __is_boolean__/data/100/closed has no type in backend.
[ADIOS2] Warning: Attribute with name __is_boolean__/data/500/closed has no type in backend.
[ADIOS2] Warning: Attribute with name __is_boolean__/data/300/closed has no type in backend.
[ADIOS2] Warning: Attribute with name __is_boolean__/data/200/closed has no type in backend.
[ADIOS2] Warning: Attribute with name __is_boolean__/data/400/closed has no type in backend.
```

in my logs.

**To Reproduce**

Read-back of an ADIOS2 BP4 file that closed iterations with `::close()` during write.

**Expected behavior**

Maybe we can use an attribute type that we can re-identify easily (`int`) or fix read-back of our self-made `bool` types in ADIOS2? 

**Software Environment**
 - version of openPMD-api: 0.12.0
 - installed openPMD-api via: conda-forge
 - operating system: Ubuntu 18.04 (64bit) on x86
 - machine: my laptop
 - name and version of Python implementation: CPython 3.8.5
 - version of HDF5: 1.10.6
 - version of ADIOS1: 1.13.1
 - version of ADIOS2: 2.6.0
 - name and version of MPI: w/o MPI

**Additional context**

Saw this while trying to read-back ZFP compressed data sets.

franzpoeschel:
Fixes #784.

EDIT: Since boolean types are neither really supported by ADIOS or HDF5.. what about just kicking out `openPMD::Datatype::BOOL`?

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -805,8 +805,15 @@ namespace detail
         ( std::is_same< T, rep >::value )
         {
             std::string metaAttr = "__is_boolean__" + name;
-            auto type = attributeInfo( IO, "__is_boolean__" + name );
-            if ( type == determineDatatype<rep>() )
+            /*
+             * In verbose mode, attributeInfo will yield a warning if not
+             * finding the requested attribute. Since we expect the attribute
+             * not to be present in many cases (i.e. when it is actually not
+             * a boolean), let's tell attributeInfo to be quiet.
+             */
+            auto type = attributeInfo(
+                IO, "__is_boolean__" + name, /* verbose = */ false );
+            if( type == determineDatatype< rep >() )
             {
                 auto attr = IO.InquireAttribute< rep >( metaAttr );
                 if (attr.Data().size() == 1 && attr.Data()[0] == 1)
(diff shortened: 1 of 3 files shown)


ax3l:
I would like to, but @PrometheusPi uses them for flags in PIConGPU in production.

One could alternatively use an `int` for the same cost, but I started to like `bool` because it makes me feel so C++-y.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -805,8 +805,15 @@ namespace detail
         ( std::is_same< T, rep >::value )
```
ax3l, line 805: Just to clarify my mind, when is this true again? For each attribute read that is scalar?
What do we use under the hood for bool? `int` or `char`?
I think to remember we used `signed char` in ADIOS1 and did not implement this as user-facing type in turn.

ax3l (approved):

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: The ADIOS2 backend uses `unsigned char` for booleans, so it is true when trying to read any attribute of that type.

ax3l: I just checked, we also use `unsigned char` in ADIOS1: https://github.com/openPMD/openPMD-api/blob/0.12.0-alpha/src/IO/ADIOS/CommonADIOS1IOHandler.cpp#L48-L49

Issue #406, opened 2019-01-03
HDF5: Chunking


ax3l:
Our HDF5 backend does not yet implement proper chunking.

Note: we should stay with *fixed size* data spaces nonetheless for performance.

References:
- https://support.hdfgroup.org/HDF5/doc/Advanced/Chunking/
- https://support.hdfgroup.org/HDF5/faq/parallel.html

Pitfalls:
- https://github.com/ComputationalRadiationPhysics/libSplash/issues/133
- https://github.com/ComputationalRadiationPhysics/picongpu/issues/2841

libSplash implementation:
- https://github.com/ComputationalRadiationPhysics/libSplash/blob/master/src/DCDataSet.cpp#L169-L170
- https://github.com/ComputationalRadiationPhysics/libSplash/blob/master/src/include/splash/core/DCHelper.hpp#L87-L104

PR #802, opened 2020-10-20
Add ::availableChunks call to Record Component types

franzpoeschel:
This allows users to inquire chunks withing datasets that are ready to be loaded. This helps in several situations:
* If a dataset does not carry data at all indices, inquire the positions where data is held.
* Allow for a more efficient loading of data by considering the write pattern in determining a read pattern.

Todo:

- [x] Implement for backends other than ADIOS2
- [x] Documentation
- [x] Cleanup


franzpoeschel:
The only thing that I really find on this for HDF5 are [dataset chunks](https://docs.h5py.org/en/stable/high/dataset.html?highlight=chunk#chunked-storage), which seems to be related only by name?
If HDF5 had something like this, I would suspect to find it [here](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5S.html#Dataspace-ExtentDims), but there's nothing that looks useful…

ax3l:
You could return the information we have from particle patches (we don't currently store this explicitly for fields).

HDF5 chunks are relatively small but similar in thought, yes! We currently don't write in chunks but should do so later on #406 

Logic one could do (top-to-bottom fall-though):
- particle patches present: return chunks stored in particle patches
- HDF5: return chunking or full extent if no chunking present; ADIOS return PGs/blocks
- return one large chunk

The general fallback for now could be the last option; if we can already return HDF5 chunks that would be cool, too.

include/openPMD/IO/ADIOS/ParallelADIOS1IOHandlerImpl.hpp
```diff
@@ -58,7 +58,7 @@ namespace openPMD
         void extendDataset(Writable*, Parameter< Operation::EXTEND_DATASET > const&) override;
         void openFile(Writable*, Parameter< Operation::OPEN_FILE > const&) override;
         void closeFile(Writable*, Parameter< Operation::CLOSE_FILE > const&) override;
-        void openPath(Writable*, Parameter< Operation::OPEN_PATH > const&) override;
+        void availableChunks(Writable*, Parameter< Operation::AVAILABLE_CHUNKS > &) override;void openPath(Writable*, Parameter< Operation::OPEN_PATH > const&) override;
```
ax3l:
```suggestion
        void availableChunks(Writable*, Parameter< Operation::AVAILABLE_CHUNKS > &) override;
        void openPath(Writable*, Parameter< Operation::OPEN_PATH > const&) override;
```

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -127,6 +130,15 @@ class AbstractIOHandlerImpl
    */
   virtual void
   closeFile( Writable *, Parameter< Operation::CLOSE_FILE > const & ) = 0;
+  /** Report chunks that are available for loading from the dataset represented
```
ax3l, line 134:
```suggestion

  /** Report chunks that are available for loading from the dataset represented
```

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -127,6 +130,15 @@ class AbstractIOHandlerImpl
...
+   *  by this writable.
+   *
+   * The resulting chunks should be stored into parameters.chunks.
+   *
+   */
+  virtual void
+  availableChunks( Writable *, Parameter< Operation::AVAILABLE_CHUNKS > & )
+  {}
   /** Create a new file in physical storage, possibly overriding an existing file.
```
ax3l, line 143:
```suggestion

  /** Create a new file in physical storage, possibly overriding an existing file.
```

franzpoeschel:
Just putting this link here since it took me 15 minutes to find any documentation on any of the chunk querying functions: https://hdf5.io/develop/group___h5_d.html#gaccff213d3e0765b86f66d08dd9959807

franzpoeschel:
> You could return the information we have from particle patches (we don't currently store this explicitly for fields).
> 
> HDF5 chunks are relatively small but similar in thought, yes! We currently don't write in chunks but should do so later on #406
> 
> Logic one could do (top-to-bottom fall-though):
> 
>     * particle patches present: return chunks stored in particle patches

`RecordComponent::availableChunks` is information carried by the backends independently of the openPMD structure whereas particle patches are explicitly stored by the user within terms defined by the openPMD standard. I think we should keep both concepts separate.
>     * HDF5: return chunking or full extent if no chunking present; ADIOS return PGs/blocks

I've added detection on whether a dataset is chunked or not. Since we don't actually write chunked datasets in HDF5 (the corresponding code in `::writeDataset` is commented out), I've only implemented your last suggestion below, and use the chunking information solely to print different warnings.
>     * return one large chunk
> 
> 
> The general fallback for now could be the last option; if we can already return HDF5 chunks that would be cool, too.

ax3l:
We discussed offline, so I am just quickly summarizing a selected point here:

> `RecordComponent::availableChunks` is information carried by the backends independently of the openPMD structure whereas particle patches are explicitly stored by the user within terms defined by the openPMD standard. I think we should keep both concepts separate.

The whole idea about particle patches in the standard is that we find again chunks to read :-) That's how I used them in the past and are the motivating example for the standardization. The currently implementation does not make us of that yet, but we can potentially put a high-level layer over the current API to make it so in the future.

>> * return one large chunk
>
> The general fallback for now could be the last option; if we can already return HDF5 chunks that would be cool, too.

@franzpoeschel will as follow-up provide a function that receives (large) chunks and can chunk them into finer (or N>1) chunks for processing, with user-provided controls (e.g. maximum individual chunk extent).

include/openPMD/Dataset.hpp
```diff
@@ -52,4 +52,27 @@ class Dataset
...
+/**
+ * A chunk consists of its offset, its extent
+ * and the rank from which it was written.
+ * If not specified explicitly, the rank will be assumed to be 0.
+ */
+struct Chunk
+{
+    Offset offset;
+    Extent extent;
+    unsigned int rank = 0;
```
ax3l: Can you please document these fundamental members, i.e. do we really need to store the (MPI?) rank here?
I think a fundamental offset/extent Chunk class without those info would be better. Containers around it could then connect it to a rank.

```suggestion
    Offset offset;  //!< origin of the chunk
    Extent extent;  //!< size of the chunk
    unsigned int rank = 0;  //!< ... what is a rank here?
```

General note: best write `mpi_rank` when you mean MPI ranks, because for 2D-arrays, people can mismatch it with [the rank of a matrix](https://en.wikipedia.org/wiki/Rank_(linear_algebra)) in English or with the [HDF5 rank](https://support.hdfgroup.org/HDF5/doc/H5.intro.html) (aka dimension/ndim).

include/openPMD/backend/BaseRecordComponent.hpp
```diff
@@ -59,8 +59,25 @@ class BaseRecordComponent : public Attributable
...
+     * * The ADIOS backends (versions 1 and 2) will return those chunks that
+     *   the writer has originally written.
+     * * The JSON backend will reconstruct the chunks by iterating the dataset.
+     * * The HDF5 backend will return the whole dataset as one large chunk.
+     *
+     * The results depend solely on the backend and are independent of any
+     * openPMD-related information. Especially is this call explicitly unrelated
+     * to openPMD's concept of particle patches, which users may additionally
+     * wish to use to store user-defined, backend-independent chunking
+     * information on particle datasets.
```
ax3l: Let's change this, because the motiviation of particlePatches is exactly to feed a function like `availableChunks`.
It just was not implemented in openPMD-api that way, yet.

src/Dataset.cpp
```diff
@@ -82,4 +82,23 @@ Dataset::setCustomTransform(std::string const& parameter)
     transform = parameter;
     return *this;
 }
-} // openPMD
+
+Chunk::Chunk( Offset _offset, Extent _extent, int _rank )
```
ax3l: besides the mentioned limitations, leading underscores are uncommon.
Somewhat common in Google is a single trailing _ for member variables.

I would probably prefer to transition to `m_` prefixes for member variables instead, but since the ones here are public maybe just call the parameters here `offset_in`, `extent_in`, `rank_in`?

include/openPMD/Dataset.hpp
```diff
@@ -52,4 +52,27 @@ class Dataset
...
     std::string transform;
 };
-} // openPMD
+
+/**
+ * A chunk consists of its offset, its extent
+ * and the rank from which it was written.
+ * If not specified explicitly, the rank will be assumed to be 0.
+ */
+struct Chunk
```
ax3l: I think this should be it's own file. Please implement the rule of zero/five, especially since it has public members and is quite fundamental (we might want to use it elsewhere to group and offset and extent :) ).

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1130,6 +1155,36 @@ namespace detail
     }
 
     template < typename T >
+    void DatasetHelper<
+        T, typename std::enable_if< DatasetTypes< T >::validType >::type >::
+        blocksInfo(
+            Parameter< Operation::AVAILABLE_CHUNKS > & params,
+            adios2::IO IO,
+            adios2::Engine engine,
```
ax3l: do you intentionally pass both by value?

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -380,8 +380,77 @@ HDF5IOHandlerImpl::extendDataset(Writable* writable,
...
+    int chunkDimensionality =
+        H5Pget_chunk( propertyList, ndims, chunkExtent.data() );
+
+    if( chunkDimensionality < 0 )
+    {
+        std::cerr << R"END(
+[HDF5] Warning: HDF5 does not carry information on available chunks.
+       Returning instead the whole dataset as one chunk.
+       Some positions in the chunk may hold undefined data.
+)END" << std::endl;
```
ax3l: I don't thing this justifies a warning: no-chunks in a file can be silently interpreted as one chunk.

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -380,8 +380,77 @@ HDF5IOHandlerImpl::extendDataset(Writable* writable,
...
+    {
+        // so, the dataset indeed has chunks
+        // alas, this backend doesn't write chunks, so for now, reading them is
+        // unimplemented
+        std::cerr << R"END(
+[HDF5] Warning: HDF5 dataset has chunked layout. Since the openPMD API does not
+       support writing chunked HDF5 datasets, the whole dataset will be
+       returned as one chunk.
+       Some positions in the chunk may hold undefined data.
+)END" << std::endl;
```
ax3l: I don't understand this warning.
A user _reads_ a file from elsewhere with openPMD-api and we say: "we found chunks, but we don't tell you, since we cannot write chunks." That makes no sense to me.
I am also not sure about this statement:
> Some positions in the chunk may hold undefined data.

where did you find that?

src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -271,6 +272,221 @@ namespace openPMD
...
+        mergeChunks( ChunkTable & table )
+        {
+            bool stillChanging;
+            do
+            {
+                stillChanging = false;
+                auto innerLoops = [ &table ]() {
+                    /*
+                     * Iterate over pairs of chunks in the table.
+                     * If finding a pair that can be merged, merge it,
```
ax3l:
```suggestion
                     * When a pair that can be merged is found, merge it;
```

test/SerialIOTest.cpp
```diff
@@ -1528,6 +1596,51 @@ void optional_paths_110_test(const std::string & backend)
...
+
+    {
+        Series read( name, Access::READ_ONLY );
+        Iteration it0 = read.iterations[ 0 ];
+        auto E_x = it0.meshes[ "E" ][ "x" ];
+        ChunkTable table = E_x.availableChunks();
+        REQUIRE( table.size() == 1 );
+        /*
+         * Explicitly convert things to bool, so Catch doesn't get the splendid
+         * idea to print the Chunk struct.
```
ax3l, line 1630: What does it need for that, an `ostream operator<<` implementation for Chunk? Could be generally useful for that type:
https://github.com/catchorg/Catch2/blob/devel/docs/tostring.md

src/Dataset.cpp
franzpoeschel: Since I don't really know about any other conventions for this use case in C++, sounds fine with me :)

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: > I don't understand this warning.
> A user reads a file from elsewhere with openPMD-api and we say: "we found chunks, but we don't tell you, since we cannot write chunks." That makes no sense to me.

Simple reason: I can't really test an implementation of that easily at the moment, since I would need to create a chunked dataset for that, which the openPMD API doesn't currently do. If the openPMD API doesn't support writing chunks yet, it's fair that reading them is unsupported as well?

> Some positions in the chunk may hold undefined data.

Undefined as in "not defined explicitly by the writer". In ADIOS, if the writer writes data only to a subset of the available positions in the dataset, `::availableChunks` will only return positions in the dataset that have been written to. In HDF5, no such checks are done. Returning the whole dataset as one chunks means that some positions in the chunk may not hold sensible data if the dataset has been written partially only.

test/SerialIOTest.cpp
franzpoeschel: I don't find the failing builds any more, think they got lost in rebasing. It looked more like a linker bug in Catch that only came up in MSVC. g++ and clang had no issues with that.

include/openPMD/Dataset.hpp
franzpoeschel: The MPI rank is good to know for data distribution in streaming setups, see also [this issue](https://github.com/ornladios/ADIOS2/issues/1679) in ADIOS2 which resulted in the addition of [`Variable<T>::Info::WriterID`](https://github.com/ornladios/ADIOS2/blob/4b24a85077bd43d2eced60f963c9493aa1d97014/bindings/CXX11/adios2/cxx11/Variable.h#L345). In a way, this `Chunk` struct is supposed to be openPMD's variant of `Variable<T>::Info`.

I agree that having a separate, more basic `Chunk` class might be helpful. What about defining a basic `Chunk` class and deriving from it a `WrittenChunk` or something?

include/openPMD/backend/BaseRecordComponent.hpp
franzpoeschel: The description is fitting for the way that things are implemented right now. So, if changing this, should we also change the implementation? To what?

We could maybe think about doing a more high-level approach in a fallthrough manner:

1. Query available chunks from backend
2. If the backend doesn't know, query particle patches (ordering? this first?)
3. If none of this works, return whole dataset as one chunk.

include/openPMD/Dataset.hpp
franzpoeschel: > I think this should be it's own file. 

Done.

> Please implement the rule of zero

Don't I? I don't custom-declare/define any of the 5?

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: Not really. Doesn't really matter at this point, but I've changed it to references.

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: This one has more to do with what I wrote below: In other backends, the chunks returned by `::availableChunks` are chunks that have actually been written to. In HDF5 on the other hand, calling this function will give you one large chunk, possibly containing data regions that the writer never filled.

Maybe instead make this more explicit in the documentation of `::availableChunks` and drop the warning?

include/openPMD/Dataset.hpp
ax3l: Sounds great, agreed :)

include/openPMD/backend/BaseRecordComponent.hpp
ax3l: I meant that the formulation `Especially is this call explicitly unrelated to openPMD's concept of particle patches` can be changed to `Note that this call currently does not take into account the openPMD concept of particle patches...`.

include/openPMD/Dataset.hpp
ax3l: rule of 5/0: oh, wasn't looking carefully. looks good!

src/IO/HDF5/HDF5IOHandler.cpp
ax3l: I think there might be a misunderstanding. HDF5's concept of chunks is not related to what ADIOS does with blocks/PGs.

HDF5 chunks are just low-level memory blocks, similar to FS block sizes. They are not variable over the data set and are mainly a performance tuning / compression-enabling aspect of low-level HDF5 tuning and they can also be used to define a variable size (extent) on an HDF5 data set.

Maybe we should have called openPMD-api "chunks" "boxes" or "blocks" to avoid the confusion.

> Maybe instead make this more explicit in the documentation of ::availableChunks and drop the warning?

Yes, would say so. This is generally true for incompletely written data sets, even without the new function.

ax3l:
General "Chunk" question:
I think naming the new meta-data class "Chunk" could be confusing: users might ask - is a chunk the "*data* chunk" (`loadChunk`/`storeChunk`) or the *meta-data* (box with start & size) that describes the chunk. What do you think? `Box` or `Selection` could be a simpler name for the new class, maybe.

src/IO/HDF5/HDF5IOHandler.cpp
ax3l: I think there is a confusion on what HDF5 chunks actually are - please see explanation above. They are not as flexible/useful as ADIOS blocks/PGs, but rather a low-level storage detail, i.e. for compression and (in size) extensible data sets.

test/SerialIOTest.cpp
ax3l: Ok, let's try to remove this work-around then and let us compare the vars directly.

franzpoeschel: You mean, comparing the members of `Chunk`?

include/openPMD/backend/BaseRecordComponent.hpp
franzpoeschel: > currently does not take into account the openPMD concept of particle patches

Alright, that's an easier fix for sure. 

But the question still stands whether it would be wise to mix both notions one day. What if both particle patches and backend-specific chunks are available on a dataset? Then..
1. … both information may contradict each other. Especially given the clear semantical gap between chunks in HDF5 and particle patches, they're not even trying to achieve the same thing.
2. … we would need to pick one of both to return to the user. It's not really our place to make that decision because depending on the use case, either of both may be more interesting:
     * If optimizing for efficiency, throughput and data distribution, loading data in a way suggested by the backend (and with knowledge on writing ranks) is to be preferred. 
     * If interested in physical quantities, such as the hyperrectangle containing a particle path, loading the particle patches is to be preferred.

This is not saying that I am totally against it, but I think that those points should be considered. Rough draft:
1. If a user explicitly defines particle patches, this should count as explicitly overwriting the chunk information given implicitly by the backend. So *if* we decide to prioritize one over the other, we should probably prioritize the particle patches over the backend-dependent info.
2. If we prioritize particle patches, they should in general carry information at least as useful as the backend-dependent chunk info. So, standardize some kind of `writerID` attribute.
3. Since both notions of chunks may and will differ from one another, allow users to explicitly prefer the backend-dependent information if they wish. Something like `::availableChunks( ChunksInfoSource = ChunksInfoSource::ParticlePatches )` could do.

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: > I think there is a confusion on what HDF5 chunks actually are 

They are certainly different in meaning from chunks in ADIOS, which should be noted somewhere a user will see it.

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: I would then suggest, dropping both warnings and putting a note in the doxy string of `::availableChunks`, that HDF5 chunks are currently completely ignored. Implementing HDF5 chunking for reading and writing is better as a follow-up PR imo.

franzpoeschel:
I've renamed things as `ChunkInfo` and `WrittenChunkInfo` now. That fine? Still need to update the filenames.

include/openPMD/backend/BaseRecordComponent.hpp
ax3l: I concur with your analysis. Let's cross that bridge when we come to this, in practice we need particle patches for checkpoint-restart workflows and selection workflows to avoid iterating the whole particle table in massively parallel context. It also served as a good work-around for distributed HDF5 data on parallel filesystems, but that's not a primary concern if we have backend info, i.e. in ADIOS2.

Can we discuss this and questions on `writerID` in a VC? I have some experiences on how useful/not useful that would be in practice and need to fully understand how you want to use it.

src/IO/HDF5/HDF5IOHandler.cpp
ax3l: Perfect, yes. #406

test/SerialIOTest.cpp
ax3l: I mean this should probably read:
```c++
        REQUIRE(
            table[ 0 ] == WrittenChunkInfo( { 0, 0 }, { height, 4 } ) );
```

src/Chunk.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+}
+
+bool
+WrittenChunkInfo::operator==( WrittenChunkInfo const & other ) const
+{
+    return this->mpi_rank == other.mpi_rank &&
+        this->ChunkInfo::operator==( other );
+}
+} // namespace openPMD
```
ax3l: Missing EOF newline :)

test/name_tests.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
ax3l: Was this file accidentally committed?

test/SerialIOTest.cpp
```diff
@@ -83,6 +83,74 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+TEST_CASE( "available_chunks_test_json", "[serial][json]" )
+{
+    /*
+     * This test is JSON specific
+     * Our JSON backend does not store chunks explicitly,
+     * so the JSON backend will simply go through the multidimensional array
+     * and gather the data items into chunks
+     * Example dataset:
+     *
+     *      0123
```
ax3l, line 87: I think I have not yet seen a JSON file with multiple `storeChunk` parts in it - can you please provide me with an example so I can follow a bit better? :)

ax3l:
> I've renamed things as ChunkInfo and WrittenChunkInfo now. That fine? Still need to update the filenames.

Ok, I think that works, although `Info` is a bit of a generic qualifier. `ChunkBox` or `ChunkSelection`, maybe?

test/SerialIOTest.cpp
franzpoeschel: That's what I had originally, and I've re-run it to trigger the same error again: [MSVC](https://ci.appveyor.com/project/ax3l/openpmd-api/builds/36300565) doesn't like that.
```
SerialIOTest.obj : error LNK2019: unresolved external symbol "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > const Catch::Detail::unprintableString" (?unprintableString@Detail@Catch@@3V?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@B) referenced in function "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Catch::Detail::convertUnstreamable<struct openPMD::WrittenChunkInfo>(struct openPMD::WrittenChunkInfo const &)" (??$convertUnstreamable@UWrittenChunkInfo@openPMD@@@Detail@Catch@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@ABUWrittenChunkInfo@openPMD@@@Z) [C:\projects\openpmd-api\build\SerialIOTests.vcxproj]
C:\projects\openpmd-api\build\bin\Release\SerialIOTests.exe : fatal error LNK1120: 1 unresolved externals [C:\projects\openpmd-api\build\SerialIOTests.vcxproj]
```
For some reason, MSVC won't find [this symbol](https://github.com/openPMD/openPMD-api/blob/116f7ed29c254c7de8c28b95e3fc9f18a07b68ca/share/openPMD/thirdParty/catch2/include/catch2/catch.hpp#L12109), and I don't really know if it's really worth hunting down the reason for this error within a 15k line file of inlined code..?

test/name_tests.sh
franzpoeschel: There goes my attempt at smuggling ẞ into the codebase... Will remove

test/SerialIOTest.cpp
franzpoeschel: Uhm.. aren't you looking at an example? Or what would you like to see additionally?

franzpoeschel:
I can't really shake the feeling that ChunkBox sounds a bit.. double? But I think we can go with it. I like it more than ChunkSelection, mainly because a derived class WrittenChunkSelection sounds like I'm back at writing Java.

Maybe ChunkMeta if that's not too generic, too? At least, that expresses directly that we're not talking about the actual chunk.

test/SerialIOTest.cpp
franzpoeschel: If it's the JSON dataset that you'd like to see:
```json
          "x": {     
            "attributes": {
              "position": {
                "datatype": "VEC_DOUBLE",
                "value": [
                  0
                ]    
              }, 
              "unitSI": {
                "datatype": "DOUBLE",
                "value": 1
              }      
            },
            "data": [                                                                                   
              [                                                                                         
                null,                                                                                   
                null,                                                                                   
                null,                                                                                   
                null                                                                                    
              ],                                                                                        
              [                                                                                         
                null,                                                                                   
                null,                                                                                   
                null,                                                                                   
                null                                                                                    
              ],                                                                                        
              [                                                                                         
                2,                                                                                      
                4,                                                                                      
                6,                                                                                      
                8                                                                                       
              ],                                                                                        
              [                                                                                         
                2,                                                                                      
                4,                                                                                      
                6,                                                                                      
                8                                                                                       
              ],                                                                                        
              [                                                                                         
                2,                                                                                      
                4,                                                                                      
                6,                                                                                      
                8
              ],
              [
                2,
                4,
                6,
                8
              ],
              [
                2,
                4,
                6,
                8
              ],
              [
                2,
                4,
                null,
                null
              ],
              [
                2,
                4,
                null,
                2
              ],
              [
                null,
                null,
                null,
                4
              ]
            ],
            "datatype": "INT"
          }
```
So, you get nulls in unwritten positions. Nulls don't get "folded up" if a whole dimension is null, since the JSON backend takes care to fill every position in the n-d dataset at least with nulls.

include/openPMD/backend/BaseRecordComponent.hpp
franzpoeschel: Linking PR #824 here for an example on how I am intending to use this. Tldr: `writerID` by itself is only an identifier that we need to make sense of by adding meta information, such as e.g. hostnames, or similar information that we may wish to rely upon for chunk distribution strategies.

include/openPMD/Chunk.hpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    Offset offset; //!< origin of the chunk
+    Extent extent; //!< size of the chunk
+    unsigned int mpi_rank = 0; //!< the MPI rank of the writing process
+
+    /*
+     * If rank is smaller than zero, will be converted to zero.
+     */
+    explicit ChunkInfo() = default;
+    ChunkInfo( Offset, Extent, int mpi_rank );
```
ax3l: rm me :)

include/openPMD/Chunk.hpp
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Represents the meta info around a chunk in a dataset.
+ *
+ * A chunk consists of its offset and its extent
+ */
+struct ChunkInfo
+{
+    Offset offset; //!< origin of the chunk
+    Extent extent; //!< size of the chunk
+    unsigned int mpi_rank = 0; //!< the MPI rank of the writing process
```
ax3l: rm me :)

test/SerialIOTest.cpp
ax3l: Discussed today: seems this is indeed more challenging than implementing an `operator<<`, so we stay with the prior work-around.

@franzpoeschel can you please update? :)

test/SerialIOTest.cpp
ax3l: Ok, that's what I am familiar with. I somehow don't get the json "chunks" implementation then. I guess there is always only one chunk? That's what I would expect at least and what seems reasonable.

franzpoeschel: Well, that would be a possibility. But in JSON, it's possible to reconstruct chunks such that exactly those positions are covered that are non-null, so I figured I'd just do that.

ax3l: Ah, got it now - smart! :tada:

test/ParallelIOTest.cpp
```diff
@@ -287,6 +287,58 @@ TEST_CASE( "no_parallel_hdf5", "[parallel][hdf5]" )
...
+// this one works for both ADIOS1 and ADIOS2
+#if (openPMD_HAVE_ADIOS1 || openPMD_HAVE_ADIOS2) && openPMD_HAVE_MPI
+void
+available_chunks_test( std::string file_ending )
+{
+    int __mpi_rank{ -1 }, __mpi_size{ -1 };
+    MPI_Comm_rank( MPI_COMM_WORLD, &__mpi_rank );
+    MPI_Comm_size( MPI_COMM_WORLD, &__mpi_size );
+    unsigned mpi_rank{ static_cast< unsigned >( __mpi_rank ) },
+        mpi_size{ static_cast< unsigned >( __mpi_size ) };
```
ax3l: Slipped through: reserved naming.

Idea (r for read or so) - did not check all the member vars and hope it does not shadow anything this way:
```suggestion
    int r_mpi_rank{ -1 }, r_mpi_size{ -1 };
    MPI_Comm_rank( MPI_COMM_WORLD, &r_mpi_rank );
    MPI_Comm_size( MPI_COMM_WORLD, &r_mpi_size );
    unsigned mpi_rank{ static_cast< unsigned >( r_mpi_rank ) },
        mpi_size{ static_cast< unsigned >(r_mpi_size ) };
```

src/binding/python/Chunk.cpp
```diff
@@ -0,0 +1,67 @@
...
+            py::arg("offset"), py::arg("extent"))
+        .def("__repr__",
+            [](const ChunkInfo & c) {
+                return "<openPMD.ChunkInfo of dimensionality "
+                    + std::to_string(c.offset.size()) + "'>";
+            }
+        )
+        // .def_property_readonly("offset", [](Chunk & c){ return c.offset;   })
+        // .def_property_readonly("extent", [](Chunk & c){ return c.extent;   })
+        // .def_property_readonly("rank",   [](Chunk & c){ return c.mpi_rank; })
```
ax3l: Dead code: `property_` is only needed if one wants to wrap a getter/setter method into a direct property. Super useful, but we expose the whole field directly here since it is already public in C++.
```suggestion
```

src/binding/python/Chunk.cpp
```diff
@@ -0,0 +1,67 @@
...
+            py::arg("offset"), py::arg("extent"), py::arg("rank"))
+        .def("__repr__",
+            [](const WrittenChunkInfo & c) {
+                return "<openPMD.WrittenChunkInfo of dimensionality "
+                    + std::to_string(c.offset.size()) + "'>";
+            }
+        )
+        // .def_property_readonly("offset", [](Chunk & c){ return c.offset;   })
+        // .def_property_readonly("extent", [](Chunk & c){ return c.extent;   })
+        // .def_property_readonly("rank",   [](Chunk & c){ return c.mpi_rank; })
```
ax3l:
```suggestion
```

test/SerialIOTest.cpp
```diff
@@ -75,6 +75,74 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+     *    8 **_*
+     *    9 ___*
+     *
+     * Will be read as three chunks:
+     * 1. (2,0) -- (5,4) (offset -- extent)
+     * 2. (7,0) -- (2,2) (offset -- extent)
+     * 3. (8,3) -- (2,1) (offset -- extent)
+     *
+     * Since the chunks are reconstructed, they need not necessarily
+     * correspond with the way that the chunks are written.
```
```suggestion
     * Since the chunks are reconstructed, they won't necessarily
     * correspond with the way that the chunks were written.
```

ax3l:
> I've renamed things as `ChunkInfo` and `WrittenChunkInfo` now. That fine? Still need to update the filenames.

Looks good to me, let's not forget to rename the file(s), too.

test/ParallelIOTest.cpp
franzpoeschel: Yeah, that works

src/binding/python/Chunk.cpp
franzpoeschel: Ah, thanks for the explanation. I left it commented out for now so we could check which one we would prefer :)

ax3l (approved):

ax3l:
Thank you, great work! :sparkles:

PR #804, opened 2020-10-23
Add tests for frontendbugs – Making copies of the Series object will result in strange behavior

franzpoeschel:
This adds tests (currently failing) for the issues discussed in #534.

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -75,6 +75,53 @@ TEST_CASE( "multi_series_test", "[serial]" )
     allSeries.clear();
 }
 
+TEST_CASE( "multiple_series_handles_test", "[serial]" )
+{
+    /*
+     * First test: No premature flushes through destructor when another copy
+     * is still around
+     */
+    {
+        std::unique_ptr< openPMD::Series > series_ptr;
+        {
+            openPMD::Series series(
+                "sample%T.json", openPMD::AccessType::CREATE );
+            series_ptr = std::make_unique< openPMD::Series >( series );
+            /*
+             * we have two handles for the same Series instance now:
+             * series and series_ptr
+             * series goes out of scope before series_ptr
+             * destructor should only do a flush after both of them go out of
+             * scope, but it will currently run directly after this comment
+             * since no iteration has been written yet, an error will be thrown
+             */
+        }
+        series_ptr->iterations[ 0 ].meshes[ "E" ][ "x" ].makeEmpty< int >( 1 );
+    }
+    /*
+     * Second test: A Series handle should remain accessible even if the
+     * original handle is destroyed
+     */
+    {
+        std::unique_ptr< openPMD::Series > series_ptr;
+        {
+            openPMD::Series series(
+                "sample%T.json", openPMD::AccessType::CREATE );
+            series_ptr = std::make_unique< openPMD::Series >( series );
+            series_ptr->iterations[ 0 ].meshes[ "E" ][ "x" ].makeEmpty< int >(
+                1 );
+        }
+        /*
+         * series_ptr is still in scope, but the original Series instance
+         * has been destroyed
+         * since internal pointers actually refer to the original *handle*,
+         * doing anything with series_ptr now (such as flushing it) yields
+         * nullpointer accesses
+         */
+        series_ptr->flush();
+    }
+}
+
 void
 close_iteration_test( std::string file_ending )
 {


test/SerialIOTest.cpp
```diff
@@ -83,6 +83,54 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+    /*
+     * First test: No premature flushes through destructor when another copy
+     * is still around
+     */
+    {
+        std::unique_ptr< openPMD::Series > series_ptr;
+        {
+            openPMD::Series series(
+                "sample%T.json", openPMD::AccessType::CREATE );
+            series_ptr = std::unique_ptr< openPMD::Series >(
```
ax3l: just for modern C++, `new`-free style:
```suggestion
            series_ptr = std::make_unique< openPMD::Series >( series );
```

test/SerialIOTest.cpp
```diff
@@ -83,6 +83,54 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+     * Second test: A Series handle should remain accessible even if the
+     * original handle is destroyed
+     */
+    {
+        std::unique_ptr< openPMD::Series > series_ptr;
+        {
+            openPMD::Series series(
+                "sample%T.json", openPMD::AccessType::CREATE );
+            series_ptr = std::unique_ptr< openPMD::Series >(
+                new openPMD::Series( series ) );
```
ax3l: just for modern C++, `new`-free style:
```suggestion
            series_ptr = std::make_unique< openPMD::Series >( series );
```

test/SerialIOTest.cpp
```diff
@@ -83,6 +83,54 @@ TEST_CASE( "multi_series_test", "[serial]" )
...
+            /*
+             * we have two handles for the same Series instance now:
+             * series and series_ptr
+             * series goes out of scope before series_ptr
+             * destructor should only do a flush after both of them go out of
+             * scope, but it will currently run directly after this comment
+             * since no iteration has been written yet, an error will be thrown
```
ax3l, line 96: perfect description, thank you! :sparkles: 
looks like we should add reference counting *or* forbid the `Series` copy constructor...

Not sure how useful an alias for a `Series` in form of a handle object truly is. I would rather expect that if I copy a series that I can do things like `bp` to `h5` conversion or so. We could implement such things via a `.copy("new_name.ending")` method later on that then returns a truly new `Series` object.

What do you think?

franzpoeschel: I'm still liking the idea of having a clear separation of:
1. Non-copyable shared internal resources
2. Copyable handles that are exposed to user code and that have shared pointers to the internal resources.

This way, we could put the `flush`ing destructor on the internal `Series` and things would be fine.

Currently, all of our objects just hide any member variable that need to be shared between instances behind shared pointers, which results in bugs like this.
Also, having *actual* internal classes for non-copyable resources would allow us to have proper upwards pointers through the openPMD hierarchies. Because, currently `Writable::parent` just goes to a copyable handle, and that concerns not only `Series` but *all* of our objects inheriting from `Attributable`. The issue is fortunately hard enough to trigger, but I should maybe try to find a way to make it fail for classes other than `Series` too...

> Not sure how useful an alias for a Series in form of a handle object

All other classes that represent entities in the openPMD hierarchies are designed as handles, too, (sharing similar issues as argued above), so it's a question of consistency imo

test/SerialIOTest.cpp
franzpoeschel: Oo, I keep forgetting we're allowed to use C++14 now :)

test/SerialIOTest.cpp
franzpoeschel: >  or forbid the Series copy constructor...

Until we get to it, this would likely be a sane thing to do. (This is what I temporarily did when writing the openPMD plugin for PIConGPU in order to avoid this problem). Breaking change tho

test/SerialIOTest.cpp
ax3l: Oh, we did not officially switch yet - dang I always forget this is missing in C++11.
Soon - how far is PIConGPU in `dev`?

test/SerialIOTest.cpp
ax3l: I agree with your rationale above.

Yes, deleting the copy constructor would just be a quick-fix, because it is in fact broken and will likely catch bugs if someone used this on two alive instances of the same Series. For deleting this, we also need to define a Move constructor in `Series` and `Attributable` #790

test/SerialIOTest.cpp
franzpoeschel: Ah right, I confused both of them. PIConGPU is now running on C++14. So I'm keeping the raw constructors for now?

ax3l: Problem solved - you can now use C++14 #825 :)

franzpoeschel: Nice, updated

franzpoeschel:
I've cherry-picked this test into #886 now. The tests are passing there.

PR #807, opened 2020-10-29
CI: Replace set-env in Cygwin

ax3l:
Use the GitHub Action `env:` keyword instead of the deprecated `set-env` command:
* https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
* https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path
  * https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files
* https://github.community/t/migration-from-deprecated-add-path-on-windows/136265

### DIFF
--- a/.github/workflows/cygwin.yml
+++ b/.github/workflows/cygwin.yml
@@ -1,0 +1,73 @@
+name: Cygwin
+
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+    branches:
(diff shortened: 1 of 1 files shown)


ax3l:
MSYS2 migration needs #821

PR #814, opened 2020-11-03
[WIP] Series: Non-Copyable

ax3l:
Delete the Copy-constructor in Series and add a move-constructor to Series and Attributable. This is a quick-fix for currently inconsistent handle-creation during copy.

Reproducer and long-term strategy as handle: #804

## To Do

- [ ] tests for move constructor in `Attributable` and `Series`
- [ ] entry in `NEWS`
- [ ] Python also affected?

### DIFF
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -69,6 +69,12 @@ public:
         std::string const & filepath,
         Access at,
         std::string const & options = "{}" );
+
+    Series(Series &&) = default;
+
+    /** The meaning of copying a Series is not (yet) defined */
+    Series(Series const &) = delete;
+
     ~Series();
 
     /**
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -82,7 +82,7 @@ class Attributable
 public:
     Attributable();
     Attributable(Attributable const&);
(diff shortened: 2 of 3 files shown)


franzpoeschel:
Reason for the segfault:

In [this line](https://github.com/openPMD/openPMD-api/blob/c3d981c064aa9c2bf57dee84f8168aa473e9bf94/src/Iteration.cpp#L274), pointer `s` goes not to the move-constructed new `Series` object `copy`, but to the old (and now invalid) instance `o0`. Essentially, the `Series` class is currently (via some detours) self-referential and even moving it will break things.

The old test only really ever ran fine by accident: Flushing from the `copy` was ok, since `o0` was still around to be used. Now that it's moved from, the test breaks.

So, for a quick fix, I'd suppose that move operations must be disallowed too. For a long-term fix, pointers and references must only go to (internal) classes without copy and move constructors.

Also, I think that `Series::~Series` would currently fail on a moved-from `Series` object, likely on a null-pointer dereference of a moved-from [`Container::m_container`](https://github.com/openPMD/openPMD-api/blob/c3d981c064aa9c2bf57dee84f8168aa473e9bf94/include/openPMD/backend/Container.hpp#L281).

ax3l:
Fixed with #886 #936 #955

PR #816, opened 2020-11-10
8a/8b benchmarks renamed and added 3D support

guj:
This benchmark is intended to provide measurements to read and write particle in mesh data using openPMD-api. 

There are two parts, a write benchmark, and a read benchmark. 

The writer benchmark writes one dataset per iteration. Each dataset contains both mesh and particle variables. One can configure parameters like mesh size, number of iteration through a input file. 

The reader benchmark reads in datasets and is able to perform different read patterns, for example, metadata only,  or different partial read patterns.


docs/source/usage/benchmarks.rst
```diff
@@ -0,0 +1,90 @@
...
+The `Message Passing Interface (MPI) <https://www.mpi-forum.org/>`_ is an open communication standard for scientific computing.
+MPI is used on clusters, e.g. large-scale supercomputers, to communicate between nodes and provides parallel I/O primitives.
+
+Writing
+^^^^^^^
+
+*Source*: ``examples/8a_benchmark_write_parallel.cpp``
+
+This benchmark writes a few meshes and particles,
+either 1D, 2D or 3D.
```
```suggestion
This benchmark writes a few meshes and particles, either 1D, 2D or 3D.
```

docs/source/usage/benchmarks.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The meshes are viewed as grid of mini blocks.
+As an example, we assume the mini blocks dimension are [16, 32, 32].
+
+
+Next we define the grid based on the mini block.
+say, [32, 32, 16]. Then our actual mesh size is [16x32, 32x32, 32x16].
+
+
+Here is a sample input file:
```
ax3l: After reading this page a couple of times, I have to admit I do not know how to:
- call the executable(s)
- what arguments to pass
- how to name and where to place / how to pass the input file

Can you please clarify? :) Maybe write an example call.
Ideally, people can just "blindly" build the project and then read this page to use the benchmark executables (without reading their source code).

ax3l (approved):
Thank you for the update! :sparkles: 
I'll merge this  for now and we improve docs and command line further as we need it :)

PR #817, opened 2020-11-10
Docs: Franz with CASUS

ax3l:
Franz transitioned to [CASUS](https://www.casus.science/). Updates primary affiliation and funding.


franzpoeschel:
Thanks for thinking of this! Don't think I have any further comments on the PR.

PR #818, opened 2020-11-12
JSON configuration on the dataset level

### LINKED ISSUE #688: Backend- and dataset-specific configuration (e.g. compression)
Related: [backend-specific configuration via JSON](https://github.com/openPMD/openPMD-api/pull/569) and [passing of advanced backend parameters](https://github.com/openPMD/openPMD-api/issues/150).

In issue #150, @C0nsultant proposes to implement backend-specific configuration in two places: for `openPMD::Series` and `openPMD::Dataset`. So far, PR #569 implements backend-specific configuration for `openPMD::Series`:
```cpp
    Series(
        std::string const & filepath,
        AccessType at,
        std::string const & options = "{}" );
```
…, but not for `openPMD::Dataset`. This class, however, has two fields, that control a backend-dependent configuration of compression and transformation:
```cpp
class Dataset
{
    // ...
public:
    Dataset(Datatype, Extent);

    //...

    std::string compression;
    std::string transform;
};
```

The HDF5 backend implements neither of these, the ADIOS1 backend implements transform, the ADIOS2 backend implements compression.

After an offline discussion with @psychocoderHPC, we have come up with the following proposal: The fields `compression` and `transform` are deprecated/removed from the API (which should hopefully not be a huge issue due to the very incomplete implementation status, as mentioned above) and replaced through a JSON-formatted string with the exact same structure as for `openPMD::Series`. This would then  separate backend-specific configuration more clearly.

Backends should behave such that all configuration options can be set globally in the `openPMD::Series` configuration, but dataset-specific options (including `compression` and `transform`) can be overridden on a per-dataset level.

Some options do not make sense to be overriden on a per-dataset level, such as the chosen engine in ADIOS2 (e.g. file engines or streaming engines). The [`TracingJSON`](https://github.com/openPMD/openPMD-api/blob/938b6722c05c985598d354c43eda1a492dfae8de/include/openPMD/auxiliary/JSON.hpp) helper class may prove helpful to issue a warning if such an option has been specified for a dataset.

Example for ADIOS2, global configuration (in `openPMD::Series`):
```
{
  "adios2": {
    "engine": {
      "type": "sst",
      "parameters": ["BufferGrowthFactor": "2.0", "QueueLimit": "2"]
    },
    "dataset": {
      "compression": "bzip2"
    }
  }
}
```
…which may be overridden by a dataset-specific configuration:
```
{
  "adios2": {
    "dataset": {
      "compression": "sz"
    }
  }
}
```

franzpoeschel:
Adds a field `std::string openPMD::Dataset::options = "{}"` which may be used to pass options that are specific to backend as well as dataset. Close #688, see there for further description.
Implemented for ADIOS2 compression.

TODO:
- [x] Documentation.

### DIFF
--- a/docs/source/details/adios2.json
+++ b/docs/source/details/adios2.json
@@ -8,10 +8,13 @@
       }
     },
     "dataset": {
-      "operators": [ 
+      "operators": [
         {
-          "type": "bzip2",
-          "parameters": {}
+          "type": "blosc",
+          "parameters": {
+              "clevel": "1",
+              "doshuffle": "BLOSC_BITSHUFFLE"
+          }
         }
       ]
     }
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -19,7 +19,6 @@ Generally, keys of the configuration are *lower case*.
 Parameters that are directly passed through to an external library and not interpreted within openPMD API (e.g. ``adios2.engine.parameters``) are unaffected by this and follow the respective library's conventions.
 
 The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
-*A configuration per dataset is currently not yet implemented.*
 This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
 
 For a consistent user interface, backends shall follow the following rules:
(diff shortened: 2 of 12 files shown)


franzpoeschel:
For a follow-up PR: What about having something like `options = "@/some/path/to/a/json/file.json"` so users can tell the openPMD API to load things from a file automatically? This would be similar to how [curl does it](https://stackoverflow.com/questions/6408904/send-request-to-curl-with-post-data-sourced-from-a-file) and no valid JSON string starts with an `@` anyway.

ax3l:
@franzpoeschel little conflict to solve :)

ax3l (approved):
Thank you, great work! :sparkles:

PR #824, opened 2020-11-23
Chunk distribution strategies

franzpoeschel:
~~Based on [topic-available-chunks](https://github.com/franzpoeschel/openPMD-api/tree/topic-available-chunks), see #802.~~
Based on #1043

Adds strategies for guiding applications to efficient parallel chunk loading patterns.
Idea: PR #802 lets reading applications inquire the chunks that are available for loading in the backend. This PR adds a function `chunk_assignment::assignChunks`, taking such a `ChunkTable` as well as a strategy as input and produces a further `ChunkTable` that may be used as a load pattern. 

Implemented strategies:
- Round Robin: Assign chunks to reading processes in turn.
- by hostname: Assign chunks to processes on the same host. Parameterized by the interior distribution strategy.
- by cuboid slice: Assign chunks according to a slicing into `n_ranks` hyperslabs of the data space.
- by binpacking: Assign chunks in an approximatively balanced manner via an approximation algorithm for the NP complete binpacking problem.

Also, add an attribute to the global `Series` object (currently called `rankMetaInfo`, I'm sure we'll change this) that can be used to write string-formatted meta information for the single writing processes and assign meaning to them (e.g. hostnames). Used by the hostname-based strategy.

TODO:

- [x] Python bindings
- [x] Testing, mostly done manually so far
- [x] Cleanup
- [x] Documentation
- [x] ~~This PR adds many MPI-based methods to the Series class. It would probably be good to first separate MPI and non-MPI headers, and then to add this functionality to the MPI headers.~~ Still a goal, but no longer relevant for this particular PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -422,7 +422,7 @@ set(CORE_SOURCE
         src/backend/PatchRecord.cpp
         src/backend/PatchRecordComponent.cpp
         src/backend/Writable.cpp
-        src/benchmark/mpi/OneDimensionalBlockSlicer.cpp
+        src/auxiliary/OneDimensionalBlockSlicer.cpp
         src/helper/list_series.cpp
         src/snapshots/ContainerImpls.cpp
         src/snapshots/ContainerTraits.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1346,7 +1346,7 @@ if(openPMD_BUILD_TESTING)
                 )
                 add_test(NAME CLI.pipe.py
                     COMMAND sh -c
-                        "${MPI_TEST_EXE} ${Python_EXECUTABLE}                      \
+                        "${MPI_TEST_EXE} -n 2 ${Python_EXECUTABLE}                 \
                             ${openPMD_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe       \
                             --infile ../samples/git-sample/data%T.h5               \
                             --outfile ../samples/git-sample/data%T.bp &&           \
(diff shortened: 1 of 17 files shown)


franzpoeschel:
@ax3l ping

franzpoeschel:
I've added Python bindings now, including usage of one distribution strategy in openpmd-pipe as an integrated usage example.

franzpoeschel:
One issue of this approach is that the attribute `rankMetaInfo` is a vector of as many strings as there are parallel writers which I suspect is part of the reason why we saw scaling issues in some benchmarks.
Instead, a parallel dataset should preferrably be written.

EDIT: I've transformed this to a parallel dataset now. Needs some more checking for char portability. Also, we should probably buffer these values on reading, since the table can only be read in steps in which it was written. In short: Make this less like an attribute that can be set and inquired, and rather more automatic during Series construction.

franzpoeschel:
I am done cleaning this up @ax3l. Could you have a look, please? :)

CMakeLists.txt
```diff
@@ -1336,7 +1336,7 @@ if(openPMD_BUILD_TESTING)
                 )
                 add_test(NAME CLI.pipe.py
                     COMMAND sh -c
-                        "${MPI_TEST_EXE} ${Python_EXECUTABLE}                      \
+                        "${MPI_TEST_EXE} -n 2 ${Python_EXECUTABLE}                 \
```
ax3l: Can you instead comment in the line 1335 pls?

ax3l (approved):
Nice!

Please add the small CMake update and then it can go in.

PR #829, opened 2020-11-26
Resizable Datasets

franzpoeschel:
~~Based on [topic-available-chunks](https://github.com/openPMD/openPMD-api/pull/802) since that one will be merged soon anyway.~~
Based on #927 since this requires ADIOS 2.7.0, see [here](https://github.com/ax3l/openPMD-api/compare/adios-2.7.0...franzpoeschel:topic-resizable-datasets) for the diff.

Motivation: For some simulations, it is impossible to know at the beginning of writing an iteration's data e.g. the number of particles to expect until the end of the iteration. Hence, an iteration's datasets need to be resized while writing the iteration.

Hence, allow calling `RecordComponent::resetDataset` multiple times with multiple dimensions.

- This does not (yet) allow turning (1) empty (2) constant (3) array-form datasets into one another. This would be much trickier to implement and is not really necessary for the use cases above. 
    EDIT: empty and constant record components can now be turned into one another and extending the extents of any of those two is fully supported.
- [x]  →Throw exceptions if empty/constant datasets are turned into an array-formed one. Other directions are already tested.
- The frontend does not compare the new extent to the old extent in any form whatsoever. Supporting (1) shrinking or (2) changing the dimensionality is the decision of the backend. Growing should generally be supported.

TODO:
- [x] Documentation
- [x] Implementation in the frontend.
- [x] ADIOS2: ~~Apparently, the ADIOS2 semantics for resizing datasets are targeted for a different use case: `Variable<T>::SetShape` does nothing to a variable within the same ADIOS step. So, the ADIOS semantics for resizing datasets are for a use case that we have been supporting for a long time anyway (by creating new datasets for each iteration): Having different-sized datasets across steps/iterations. Unless I'm missing something, implementing this in ADIOS2 is currently not possible. Within the same step, ADIOS2 will currently fully ignore the resizing and any write operation to any of the extended places.~~ https://github.com/ornladios/ADIOS2/pull/2545 fixes this. This means, we'll need to wait for ADIOS 2.7.0 to merge this PR. **This is also the reason for the failing tests.**
- [x] JSON
- [x] HDF5. ~~There is already an old implementation of resizing datasets in there, it's just never been used and I'll need to check whether it works.~~ HDF5 only allows this when writing with chunked storage #406. So I suppose that this one is postponed until we implement writing chunked storage in HDF5. `HDF5IOHandlerImpl::extendDataset` will now throw an exception that gives notice of this.

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -371,6 +371,63 @@ TEST_CASE( "available_chunks_test", "[parallel][adios]" )
     available_chunks_test( "bp" );
 }
 
+void
+extendDataset( std::string const & ext )
+{
+    std::string filename = "../samples/parallelExtendDataset." + ext;
+    int r_mpi_rank{ -1 }, r_mpi_size{ -1 };
+    MPI_Comm_rank( MPI_COMM_WORLD, &r_mpi_rank );
+    MPI_Comm_size( MPI_COMM_WORLD, &r_mpi_size );
+    unsigned mpi_rank{ static_cast< unsigned >( r_mpi_rank ) },
+        mpi_size{ static_cast< unsigned >( r_mpi_size ) };
+    std::vector< int > data1( 25 );
+    std::vector< int > data2( 25 );
+    std::iota( data1.begin(), data1.end(), 0 );
+    std::iota( data2.begin(), data2.end(), 25 );
+    {
+        Series write( filename, Access::CREATE, MPI_COMM_WORLD );
+        if( ext == "bp" && write.backend() != "ADIOS2" )
+        {
+            // dataset resizing unsupported in ADIOS1
+            return;
+        }
+        Dataset ds1{ Datatype::INT, { mpi_size, 25 } };
+        Dataset ds2{ { mpi_size, 50 } };
+
+        // array record component -> array record component
+        // should work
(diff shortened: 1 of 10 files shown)


franzpoeschel:
Note to self: Rebase this back onto #927 once that one is rebased on the current dev.

ax3l (approved):

test/ParallelIOTest.cpp
```diff
@@ -371,6 +371,63 @@ TEST_CASE( "available_chunks_test", "[parallel][adios]" )
...
+    MPI_Comm_size( MPI_COMM_WORLD, &r_mpi_size );
+    unsigned mpi_rank{ static_cast< unsigned >( r_mpi_rank ) },
+        mpi_size{ static_cast< unsigned >( r_mpi_size ) };
+    std::vector< int > data1( 25 );
+    std::vector< int > data2( 25 );
+    std::iota( data1.begin(), data1.end(), 0 );
+    std::iota( data2.begin(), data2.end(), 25 );
+    {
+        Series write( filename, Access::CREATE, MPI_COMM_WORLD );
+        if( ext == "bp" && write.backend() != "ADIOS2" )
```
ax3l, line 389: Oh no, this does not execute. The backend name is `MPI_ADIOS2`.
Correct syntax: `if( ext == "bp" && write.backend() == "MPI_ADIOS1" )`

#1126

PR #830, opened 2020-12-02
added env option OPENPMD_HDF5_ALIGNMENT

guj:
Added env option `OPENPMD_HDF5_ALIGNMENT`, when assigned to a positive `<bytes>`,
the `H5Pset_alignment(m_fileAccessProperty, 0, bytes)` will be called for Parallel HDF5 calls.

### DIFF
--- a/src/IO/HDF5/ParallelHDF5IOHandler.cpp
+++ b/src/IO/HDF5/ParallelHDF5IOHandler.cpp
@@ -74,6 +74,14 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(AbstractIOHandler* handler,
     herr_t status;
     status = H5Pset_dxpl_mpio(m_datasetTransferProperty, xfer_mode);
 
+    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "0" );       
+    std::stringstream sstream(strByte);
+    hsize_t bytes;
+    sstream >> bytes;
+
+    if ( bytes > 0 ) 
+         H5Pset_alignment(m_fileAccessProperty, 0, bytes);
+
     VERIFY(status >= 0, "[HDF5] Internal error: Failed to set HDF5 dataset transfer property");
     status = H5Pset_fapl_mpio(m_fileAccessProperty, m_mpiComm, m_mpiInfo);
     VERIFY(status >= 0, "[HDF5] Internal error: Failed to set HDF5 file access property");
--- a/src/IO/HDF5/ParallelHDF5IOHandler.cpp
+++ b/src/IO/HDF5/ParallelHDF5IOHandler.cpp
@@ -74,12 +74,12 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(AbstractIOHandler* handler,
     herr_t status;
     status = H5Pset_dxpl_mpio(m_datasetTransferProperty, xfer_mode);
 
-    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "0" );
+    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "1" );
     std::stringstream sstream(strByte);
     hsize_t bytes;
     sstream >> bytes;
 
-    if ( bytes > 0 )
+    if ( bytes > 1 )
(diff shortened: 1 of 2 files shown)


ax3l:
Let's also document this in the table in `docs/source/backends/hdf5.rst`

Do you think we can add a reasonable default for this? [[1]](https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_alignment.htm)
> For MPI IO and other parallel systems, choose an alignment which is a multiple of the disk block size. 

On Cori [[2]](https://www.nersc.gov/users/training/online-tutorials/introduction-to-scientific-i-o/?start=5) the recommendation seems to be to align to Lustre stripes.

src/IO/HDF5/ParallelHDF5IOHandler.cpp
```diff
@@ -74,6 +74,14 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(AbstractIOHandler* handler,
     herr_t status;
     status = H5Pset_dxpl_mpio(m_datasetTransferProperty, xfer_mode);
 
+    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "0" );
```
ax3l: The default value is `1` according to the [docs](https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_alignment.htm):
```suggestion
    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "1" );
```

src/IO/HDF5/ParallelHDF5IOHandler.cpp
```diff
@@ -74,6 +74,14 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(AbstractIOHandler* handler,
...
+    std::stringstream sstream(strByte);
+    hsize_t bytes;
+    sstream >> bytes;
+
+    if ( bytes > 0 )
```
ax3l:
```suggestion
    if ( bytes > 1 )
```

guj:
> Let's also document this in the table in `docs/source/backends/hdf5.rst`
> 
> Do you think we can add a reasonable default for this? [[1]](https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_alignment.htm)
> 
> > For MPI IO and other parallel systems, choose an alignment which is a multiple of the disk block size.
> 
> On Cori [[2]](https://www.nersc.gov/users/training/online-tutorials/introduction-to-scientific-i-o/?start=5) the recommendation seems to be to align to Lustre stripes.

I saw. that you changed the documentation. 
I would like to add that 1 is not a good default.  It means 1 byte..   

My suggestion is that by default this option shall not be setup,  and let HDF5 deal with its default behavior.  The two known cases to spell out this option are: 
(1) on Lustre systems,  alignment with stripe size is suggested to be explicitly spelled out through this option.  (2) on Summit, recommendatation is to use this option with  16MB 

Does it sound clear? I can make the change in the doc, or you can do it.

ax3l:
Thanks for taking a look!
I read in the docs: https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_alignment.htm
> Default values for threshold and alignment are one, implying no alignment

Agreed:
> My suggestion is that by default this option shall not be setup, ...

The logic does not set it up by default now either (see `> 1`)

Feel free to push to the PR once more to change it though.

guj:
> Thanks for taking a look!
> I read in the docs: https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_alignment.htm
> 
> > Default values for threshold and alignment are one, implying no alignment
> 
> 
 I meant no need for openPMD-api to set it again. It would make sense to set a default if this function will always to called by openPMD-api with a specific value (other than 1). 

> Agreed:
> 
> > My suggestion is that by default this option shall not be setup, ...
> 
> The logic does not set it up by default now either (see `> 1`)
> 
> Feel free to push to the PR once more to change it though.

OK, will update doc

ax3l (approved):

ax3l:
I think we can also safely increase the alignment sizes to at least 4MB, I used quite large numbers in the past to match the filesystem block size and also increase the collective buffer sizes to few 100MB: https://github.com/ComputationalRadiationPhysics/libSplash/issues/108

PR #831, opened 2020-12-02
Turn off Statistics (Min/Max) in ADIOS2 backend

franzpoeschel:
As documented [here](https://adios2.readthedocs.io/en/latest/engines/engines.html), ADIOS2 will produce statistics for variables by default. This is a time-intensive procedure and we don't use the information anyway, so we override this and turn it off.

If a user really wants these, they can be turned on again via JSON. The statistics cannot be accessed via openPMD, but some people may be loading data manually via ADIOS2.

Some little time measurements: I ran a single-threaded simulation with PIConGPU, producing per data dump between 14 and 17GB of data (dumping every 50 time steps for 400 steps in total).
 *  With statistics: 6 minutes, 31 seconds
 *  Without: 6 minutes, 4 seconds

That is a speedup of nearly half a minute.

Also, replace a dead link in the documentation while I'm at it.

### DIFF
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -33,11 +33,15 @@ environment variable                  default    description
 ``OPENPMD_BP_BACKEND``                ``ADIOS2`` Chose preferred ``.bp`` file backend if ``ADIOS1`` and ``ADIOS2`` are available.
 ===================================== ========== ================================================================================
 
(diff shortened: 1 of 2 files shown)


ax3l (approved):
Great catch.
I remember we turned them off by default in ADIOS1 as well, and making it a configuration option is splendid! :sparkles:

ax3l:
cc @guj just FYI, this might be visible in the next benchmarks as an improvement.

PR #834, opened 2020-12-04
added option to assign workload on rank

guj:
Added an command line option to assign workload per rank. This is relevant to ADIOS2 to determine how data is distributed in subfiles.


examples/8a_benchmark_write_parallel.cpp
```diff
@@ -460,15 +465,19 @@ int parseArgs( int argc, char *argv[], TestInput& input )
...
         input.m_Steps = atoi( argv[4] );
 
-
     if (argc >= 6)
       input.m_Dim = atoi( argv[5] );
 
-    //if ( (dataDim > 3) || (dataDim < 0) )  {
-    //return -1;
-    //}
+    if (argc >= 7) {
```
ax3l, line 482: Please don't forget to document the new option in `docs/source/usage/benchmarks.rst` so people know how to use it :)

examples/8a_benchmark_write_parallel.cpp
```diff
@@ -310,9 +310,16 @@ class TestInput
 
   int m_MPISize = 1; //!< MPI size
   int m_MPIRank = 0; //!< MPI rank
+
   unsigned long m_XBulk = 64ul; //!< min num of elements at X dimension
   unsigned long m_YBulk = 32ul;  //!< min num of elements at Y dimension
   unsigned long m_ZBulk = 32ul;
+
+  /**  relative expansion of min grid(m_XBulk, m_YBulk, m_ZBulk)
+   *   to form a max grid. by fault max:min=1
```
ax3l: I don't fully understand this part yet, but maybe the additional doc will solve that.

examples/8a_benchmark_write_parallel.cpp
guj: This option is making sense for bp users. The benchmark right now has fast track options for H5 users.  Will need to add another section later.

examples/8a_benchmark_write_parallel.cpp
guj: Added more comments there. Mainly this option is to direct range of workload per rank. Maybe the name is not fitting really well.

ax3l:
@guj I rebased your branch because it was outdated and thus hard to review.

Please make sure to always update to the latest `dev` branch when starting new branches :)

ax3l (approved):
Thank you, Junmin! :christmas_tree:

PR #822, opened 2020-11-20
Allow reading a file-based series with many iterations without crashing the number of file handles

franzpoeschel:
Issue first brought up [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3434#issuecomment-729555373) by @cbontoiu.

The openPMD API already allows *writing* a file-based Series with arbitrarily many iterations by using `Iteration::close()` after finishing the write operation. So far, *reading* such a Series was not possible on a system with limited number of allowed file handles per process: The openPMD API would during parsing open all files, but never close a single file handle. This PR closes a parsed iteration directly after parsing. After reading an iteration, a user will need to `Iteration::close()` the iteration again in order to give the file handle back to the system.

Some reorganization was necessary for this in the HDF5 backend, since that backend previously stored the explicit HDF5 file handle for each `Writable` object. Since closing and reopening a file changes the handle, the HDF5 backend will now only store file names for each such object and keep a map from file names to active file handles.

This is a quick fix and more sophisticated methods such as the following are **not** implemented:
* Keeping a stack of open file handles and automatically closing old handles if a certain threshold is reached.
* Lazy parsing of iterations: Only parse an iteration upon `Series.iterations.operator[]()`. This would reduce the number of file accesses while opening a `Series` and drastically speed up this process in general.
* Allow reopening an iteration that a user has explicitly `Iteration::close()`d. Currently, the openPMD API is allowed to close and reopen iterations (which is, what this PR implements). Users are not (yet).

EDIT: Do we have sufficiently fine-grained control over the CI containers to allow setting ulimits to a low number? I don't really want to go overboard with the tests for this one…

PR #837, opened 2020-12-07
Explicitly record close status if temporarily closing

franzpoeschel:
PR #822 introduced logic to close an iteration temporarily in file-based iteration layout.
I noticed two issues:
* The method `Series::flushFileBased` will not close iterations that are not dirty. Fix that.
* Iterations did not store whether they were temporarily closed. This led to issues on the streaming branch with double closing. Let's make that explicit in `Iteration::m_closed`.


ax3l (approved):
Thank you! :sparkles:

PR #838, opened 2020-12-07
Determinize destruction order of files in ADIOS2

franzpoeschel:
See https://github.com/ornladios/ADIOS2/issues/2533.

Interestingly, ADIOS 2.6.0 had no issue when closing different engines in different orders across different ranks.
Current master has issues with that, so this PR determinizes destruction order.


ax3l (approved):
Good catch, thank you!

PR #842, opened 2020-12-10
Test: HiPACE I/O Pattern

ax3l:
Add a new test to ensure the output pattern of collective, intependent I/O with interleaved write works, as perfomed by HiPACE.

cc @MaxThevenet @SeverinDiederichs @AlexanderSinn @RemiLehe

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -555,6 +555,8 @@ hipace_like_write( std::string file_ending )
 
     // open a parallel series
     Series series( name, Access::CREATE, MPI_COMM_WORLD );
+    series.setIterationEncoding( IterationEncoding::groupBased );
+    series.flush();
 
     // in HiPACE, ranks write one-by-one to a "swiped" step, overlapping
     // each other in time;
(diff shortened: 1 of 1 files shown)


test/ParallelIOTest.cpp
```diff
@@ -523,4 +524,106 @@ TEST_CASE( "close_iteration_test", "[parallel]" )
...
+            io::determineDatatype< precision >( ),
+            {global_Nz, global_Nx});
+        E_x.resetDataset(dataset);
+
+        Offset chunk_offset = {local_Nz * mpi_rank, 0};
+        Extent chunk_extent = {local_Nz, global_Nx};
+        E_x.storeChunk(
+            io::shareRaw(E_x_data),
+            chunk_offset, chunk_extent);
+        series.flush();
```
ax3l, line 630: flush is [MPI collective](https://openpmd-api.readthedocs.io/en/0.12.0-alpha/details/mpi.html) and hangs in the app at this point.
Why does that test not hang?

ax3l: Ah jeeze, I forgot that of course there is other, app-related collective/blocking I/O in this loop...
(Adding an MPI_Barrier before `continue` demonstrates the hang.)

PR #843, opened 2020-12-10
clang-tidy: no modernize-use-auto

ax3l:
This flag adds a bit too much context and would enforce that we write
auto in calls such as
```C++
src/IO/ADIOS/ADIOS2IOHandler.cpp:1994:9: warning:
  use auto when initializing with a template cast to avoid
  duplicating the type name [modernize-use-auto]
        Datatype ret = switchType< Datatype >(
        ^~~~~~~~
        auto
```

which is a bit too smart (those calls a re not really a cast and it can be useful to keep them explicit.

Spotted by @franzpoeschel

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,5 +2,5 @@
 # FIXME: all performance-* reports
 # FIXME: all cert-* reports
 # FIXME: all bugprone-* reports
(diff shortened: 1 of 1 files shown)


franzpoeschel:
Thanks for this!
Should we also take the chance to deactivate `modernize-return-braced-init-list`?
Example:
```
/home/runner/work/openPMD-api/openPMD-api/src/Series.cpp:378:12: warning: avoid repeating the return type from the declaration; use a braced initializer list instead [modernize-return-braced-init-list]
    return ReadIterations( this );
           ^~~~~~~~~~~~~~~      ~
           {                    }
```

PR #847, opened 2020-12-15
Explicitly set SubStreams for available_chunks parallel test

franzpoeschel:
ADIOS2 BP engines assign a chunk's `writerID` by its containing subfile. Since ADIOS 2.7 will no longer use one substream per process by default, this breaks our parallel `available_chunks_test`. This prepares the openPMD API for ADIOS 2.7.

Given that the writer ID apparently has only an indirect relation to the writing MPI rank, we should maybe rename `WrittenChunkInfo::mpi_rank` to something more generic?

Also, should we maybe expose some rudimentary JSON editing helpers? Passing this parameter is rather ugly atm..

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -299,9 +299,27 @@ available_chunks_test( std::string file_ending )
         mpi_size{ static_cast< unsigned >( r_mpi_size ) };
     std::string name = "../samples/available_chunks." + file_ending;
 
+    std::stringstream parameters;
+    parameters << R"END(
+{
+    "adios2":
+    {
+        "engine":
+        {
+            "type": "bp4",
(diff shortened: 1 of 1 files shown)


ax3l:
Thank you for anticipating the change!

> Given that the writer ID apparently has only an indirect relation to the writing MPI rank, we should maybe rename WrittenChunkInfo::mpi_rank to something more generic?

Yes, I think that's a good id. Maybe `chunkID` or `index` or so?

> Also, should we maybe expose some rudimentary JSON editing helpers? Passing this parameter is rather ugly atm..

Wonder if we want to expose an alias namespace and include for JSON that one could use on the user-facing API as helper...

ax3l (approved):

PR #813, opened 2020-11-02
Use ADIOS variables for openPMD attributes

franzpoeschel:
~~Note: Don't get scared at the diff, this is based upon my [topic-streaming](https://github.com/franzpoeschel/openPMD-api/tree/topic-streaming) branch, since many of the issues leading to this PR became obvious only upon implementing streaming mode.
For comparing: https://github.com/franzpoeschel/openPMD-api/compare/topic-streaming...franzpoeschel:topic-adios-variables-for-attributes~~

It has turned out that ADIOS attributes are too restricted to use them for modeling openPMD attributes. ADIOS attributes are independent of ADIOS steps, whereas in openPMD we create new sets of attributes for each iteration (which in our Streaming API are equivalent to ADIOS steps). This is not the preferred way to model scientific data in ADIOS. ADIOS prefers metadata to be setup once only and reuse that  metadata along steps. The mid-term goal might be to investigate a data layout in openPMD that agrees with this concept better. Up until then, the short-term solution is to use ADIOS variables for openPMD attributes.

ADIOS provides a zero-dimensional variable shape called *global single value* for storing scalars, intended for metadata that changes over steps. Since our metadata is not exclusively scalar, we cannot fully rely on those, but need to use regular *global array* variables as well.

In order to distinguish openPMD datasets from openPMD attributes, openPMD datasets will from now on be suffixed by `/__data__`. As an example, an E field created by PIConGPU:
```
int8_t    /data/0/fields/E/axisLabels                                      {3, 2}                                                                           
  string    /data/0/fields/E/dataOrder                                       scalar                                                                           
  string    /data/0/fields/E/fieldSmoothing                                  scalar                                                                           
  string    /data/0/fields/E/geometry                                        scalar                                                                           
  double    /data/0/fields/E/gridGlobalOffset                                {3}    
  float     /data/0/fields/E/gridSpacing                                     {3}   
  double    /data/0/fields/E/gridUnitSI                                      scalar      
  float     /data/0/fields/E/timeOffset                                      scalar
  double    /data/0/fields/E/unitDimension                                   {7}   
  float     /data/0/fields/E/x/__data__                                      {56, 56, 56}
  float     /data/0/fields/E/x/position                                      {3}    
  double    /data/0/fields/E/x/unitSI                                        scalar
  float     /data/0/fields/E/y/__data__                                      {56, 56, 56}
  float     /data/0/fields/E/y/position                                      {3}   
  double    /data/0/fields/E/y/unitSI                                        scalar 
[...]

PR #855, opened 2020-12-21
Variable-Based iteration layout

franzpoeschel:
~~Based on [topic-adios-variables-for-attributes](https://github.com/franzpoeschel/openPMD-api/tree/topic-adios-variables-for-attributes) since "variable attributes" are needed for this.
For comparing the changes: https://github.com/franzpoeschel/openPMD-api/compare/topic-adios-variables-for-attributes...franzpoeschel:topic-stepbased~~

This adds a third iteration layout to the openPMD API: Next to file-based and group-based, a step-based layout.

Idea: ADIOS prefers consuming data in steps. Until now, the openPMD API created entirely new datasets for each iteration, while ADIOS(2) actually allows to reuse datasets across steps. A simple dataset may now look like this in ADIOS2:

```
  string    /basePath                        scalar
  uint64_t  /data/__step__                   10*scalar
  uint8_t   /data/closed                     10*scalar
  double    /data/dt                         10*scalar
  int8_t    /data/meshes/E/axisLabels        10*{1, 2}
  string    /data/meshes/E/dataOrder         10*scalar
  string    /data/meshes/E/geometry          10*scalar
  double    /data/meshes/E/gridGlobalOffset  10*{1}
  double    /data/meshes/E/gridSpacing       10*{1}
  double    /data/meshes/E/gridUnitSI        10*scalar
  float     /data/meshes/E/timeOffset        10*scalar
  double    /data/meshes/E/unitDimension     10*{7}
  int32_t   /data/meshes/E/x/__data__        10*{1000}
  double    /data/meshes/E/x/position        10*{1}
  double    /data/meshes/E/x/unitSI          10*scalar
  int32_t   /data/meshes/E/y/__data__        10*{__}
  double    /data/meshes/E/y/position        10*{1}
  double    /data/meshes/E/y/unitSI          10*scalar
  double    /data/time                       10*scalar
  double    /data/timeUnitSI                 10*scalar
  string    /date                            scalar
  string    /iterationEncoding               scalar
  string    /iterationFormat                 scalar
  string    /meshesPath                      scalar
  string    /openPMD                         scalar
  uint32_t  /openPMDextension                scalar
  string    /software                        scalar
  string    /softwareVersion                 scalar
```

This helps controlling the amount of metadata by reusing attributes and variables across steps.

TODO:

- [x] Remove `currentIteration` group and merge contents with layer above.
      EDIT: See comment below, there are pros and cons and this should be discussed first.
- [x] Merge [topic-adios-variables-for-attributes](https://github.com/franzpoeschel/openPMD-api/tree/topic-adios-variables-for-attributes) before this.
- [x] Merge #938 before this
- [x] CI failures are because I made the use of steps the default behavior for the new schema and there are bugs in ADIOS2 2.6.0 that prevent that. So, merge this only after bumping the required version to ADIOS2 2.7.0.
- [x] Documentation
- [x] More thorough testing
- [x] Formalize this in the openPMD standard: https://github.com/openPMD/openPMD-standard/pull/250
- [x] `iterationFormat` and similar attributes
- [x] Use something like `data%V.bp` for automatic recognition of the iteration layout?
- [x] test that changing extents (shapes) over time work for meshes & particles
    Update: Doesn't work at the moment. I know the reason, need to find a way to fix it.
    Update: I've reordered the PRs, this is now based on #938 (lazy parsing). In lazy parsing mode, this now works. In eager parsing, parsing is done before any step is opened and datasets are not (yet) re-parsed. Fix that.
    Update: Both should be working now.
- [x] Naming: rename `__step__` to `snapshot` (see https://github.com/openPMD/openPMD-standard/issues/148)
- [ ] Expose iteration encoding as JSON option – maybe postpone to a more thorough JSON PR
- [x] Guarantee removal of stale content when reparsing

### DIFF
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -233,6 +233,7 @@ Iteration::flushFileBased(std::string const& filename, uint64_t i)
         {
             Parameter< Operation::OPEN_FILE > fOpen;
             fOpen.name = filename;
+            fOpen.encoding = IterationEncoding::fileBased;
             IOHandler()->enqueue(IOTask(s, fOpen));
             flush();
 
@@ -262,6 +263,21 @@ Iteration::flushGroupBased(uint64_t i)
 }
 
 void
+Iteration::flushVariableBased( uint64_t i )
+{
+    if( !written() )
+    {
+        /* create iteration path */
+        Parameter< Operation::OPEN_PATH > pOpen;
+        pOpen.path = "";
+        IOHandler()->enqueue( IOTask( this, pOpen ) );
+        this->setAttribute( "__step__", i );
+    }
+
+    flush();
+}
+
+void
 Iteration::flush()
 {
     if(IOHandler()->m_frontendAccess == Access::READ_ONLY )
@@ -279,8 +295,10 @@ Iteration::flush()
         if( !meshes.empty() || s->containsAttribute("meshesPath") )
         {
             if( !s->containsAttribute("meshesPath") )
+            {
                 s->setMeshesPath("meshes/");
-            s->flushMeshesPath();
+                s->flushMeshesPath();
+            }
             meshes.flush(s->meshesPath());
             for( auto& m : meshes )
                 m.second.flush(m.first);
@@ -293,8 +311,10 @@ Iteration::flush()
         if( !particles.empty() || s->containsAttribute("particlesPath") )
         {
             if( !s->containsAttribute("particlesPath") )
+            {
                 s->setParticlesPath("particles/");
-            s->flushParticlesPath();
+                s->flushParticlesPath();
+            }
             particles.flush(s->particlesPath());
             for( auto& species : particles )
                 species.second.flush(species.first);
@@ -323,11 +343,11 @@ void Iteration::read()
     auto const & deferred = m_deferredParseAccess->get();
     if( deferred.fileBased )
     {
-        readFileBased( deferred.filename, deferred.index );
+        readFileBased( deferred.filename, deferred.path );
     }
     else
     {
-        readGroupBased( deferred.index );
+        readGroupBased( deferred.path );
     }
     // reset this thing
     *m_deferredParseAccess = auxiliary::Option< DeferredParseAccess >();
@@ -521,6 +541,7 @@ Iteration::beginStep()
             file = m_attributableData.get();
             break;
         case IE::groupBased:
+        case IE::variableBased:
             file = &series;
             break;
     }
@@ -532,7 +553,8 @@ Iteration::beginStep()
     }
 
     // re-read -> new datasets might be available
-    if( series.iterationEncoding() == IE::groupBased &&
+    if( ( series.iterationEncoding() == IE::groupBased ||
+          series.iterationEncoding() == IE::variableBased ) &&
         ( this->IOHandler()->m_frontendAccess == Access::READ_ONLY ||
           this->IOHandler()->m_frontendAccess == Access::READ_WRITE ) )
     {
@@ -542,7 +564,7 @@ Iteration::beginStep()
         auto newType =
             const_cast< Access * >( &this->IOHandler()->m_frontendAccess );
         *newType = Access::READ_WRITE;
-        series.readGroupBased( false );
+        series.readGorVBased( false );
         *newType = oldType;
         series.iterations.written() = previous;
     }
@@ -564,6 +586,7 @@ Iteration::endStep()
             file = m_attributableData.get();
             break;
         case IE::groupBased:
+        case IE::variableBased:
             file = &series;
             break;
     }
@@ -582,6 +605,7 @@ Iteration::getStepStatus()
         case IE::fileBased:
             return *this->m_stepStatus;
         case IE::groupBased:
+        case IE::variableBased:
             return s->m_stepStatus;
         default:
             throw std::runtime_error( "[Iteration] unreachable" );
@@ -599,6 +623,7 @@ Iteration::setStepStatus( StepStatus status )
             *this->m_stepStatus = status;
             break;
         case IE::groupBased:
+        case IE::variableBased:
             s->m_stepStatus = status;
             break;
         default:
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -445,7 +445,7 @@ void Iteration::read_impl( std::string const & groupPath )
         pOpen.path = s->meshesPath();
         IOHandler()->enqueue(IOTask(&meshes, pOpen));
 
-        meshes.readAttributes();
+        meshes.readAttributes( ReadMode::FullyReread );
 
         auto map = meshes.eraseStaleEntries();
 
@@ -510,7 +510,7 @@ void Iteration::read_impl( std::string const & groupPath )
         pOpen.path = s->particlesPath();
         IOHandler()->enqueue(IOTask(&particles, pOpen));
 
-        particles.readAttributes();
+        particles.readAttributes( ReadMode::FullyReread );
 
         /* obtain all particle species */
         pList.paths->clear();
@@ -532,7 +532,7 @@ void Iteration::read_impl( std::string const & groupPath )
         particles.dirty() = false;
     }
 
-    readAttributes();
+    readAttributes( ReadMode::FullyReread );
 }
 
 AdvanceStatus
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -334,13 +334,8 @@ void Iteration::deferParseAccess( DeferredParseAccess dr )
         auxiliary::makeOption< DeferredParseAccess >( std::move( dr ) );
 }
 
-void Iteration::read( std::string path, bool reread )
+void Iteration::read()
 {
-    if( reread )
-    {
-        read_impl( path );
-        return;
-    }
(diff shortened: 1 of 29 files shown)


franzpoeschel:
The [latest commit](https://github.com/openPMD/openPMD-api/pull/855/commits/60b0394ca2a7b882e035a0cfbcb5fc5e72c61e7f) removes the `currentIteration` group:
```
  string    /basePath                        scalar
  uint8_t   /data/closed                     10*scalar
  double    /data/dt                         10*scalar
  uint64_t  /data/iterationIndex             10*scalar
  int8_t    /data/meshes/E/axisLabels        10*{1, 2}
  string    /data/meshes/E/dataOrder         10*scalar
  string    /data/meshes/E/geometry          10*scalar
  double    /data/meshes/E/gridGlobalOffset  10*{1}
  double    /data/meshes/E/gridSpacing       10*{1}
  double    /data/meshes/E/gridUnitSI        10*scalar
  float     /data/meshes/E/timeOffset        10*scalar
  double    /data/meshes/E/unitDimension     10*{7}
  int32_t   /data/meshes/E/x/__data__        10*{1000}
  double    /data/meshes/E/x/position        10*{1}
  double    /data/meshes/E/x/unitSI          10*scalar
  double    /data/time                       10*scalar
  double    /data/timeUnitSI                 10*scalar
  string    /date                            scalar
  string    /iterationEncoding               scalar
  string    /iterationFormat                 scalar
  string    /meshesPath                      scalar
  string    /openPMD                         scalar
  uint32_t  /openPMDextension                scalar
  string    /software                        scalar
  string    /softwareVersion                 scalar
```
 I'm not sure how to feel about this.
**Pro:** Looks nicer, is simpler
**Con:** It's no longer possible to distinguish attributes written to `series.iterations` and `series.iterations[0]`.

Would this lead to problems? Are attributes commonly written to `series.iterations`?
Also if we go with this:
- [x] Even if the other backends don't support step-based iteration layout, make sure they deal correctly with `openPath(path="")`. The ADIOS2 backend didn't.

ax3l:
Hi @franzpoeschel, can you please rebase this PR for review? :)

> https://github.com/openPMD/openPMD-api/pull/855#issuecomment-749482899
>
>  Pro: Looks nicer, is simpler
> Con: It's no longer possible to distinguish attributes written to series.iterations and series.iterations[0].
>
> Would this lead to problems? Are attributes commonly written to series.iterations?

The layout without the extra group is exactly what we want. Attributes that were on each `/data/<iteration>/` should not clash with anything in `/data` by design of openPMD-standard, so we are good to go. This is similar to how we handle scalar and vector records already.

I would rename `iterationIndex` to `__steps__` as we already do for `__data__`.

For tests, I would say we can add as we do the "ext"(ension) argument for many steps another "encoding" argument and just increase write/read coverage by running many more unit tests through it.

franzpoeschel:
> For tests, I would say we can add as we do the "ext"(ension) argument for many steps another "encoding" argument and just increase write/read coverage by running many more unit tests through it.

This iteration layout requires use of the streaming API, so the number of tests we can hijack is limited by that. But testing can surely be extended there.
Currently, picking this iteration layout is supported only via `Series::setIterationLayout`. I think we should add some kind of pattern for this as well, so applications won't have to explicitly add support for this layout. I propose something like `series%V.bp`. That would also simplify hijacking tests.

franzpoeschel:
> Hi @franzpoeschel, can you please rebase this PR for review? :)

Done
> 
> > [#855 (comment)](https://github.com/openPMD/openPMD-api/pull/855#issuecomment-749482899)
> > Pro: Looks nicer, is simpler
> > Con: It's no longer possible to distinguish attributes written to series.iterations and series.iterations[0].
> > Would this lead to problems? Are attributes commonly written to series.iterations?
> 
> The layout without the extra group is exactly what we want. Attributes that were on each `/data/<iteration>/` should not clash with anything in `/data` by design of openPMD-standard, so we are good to go. This is similar to how we handle scalar and vector records already.

It's also a bit helpful for implementation. This way, I can inquire `Iteration::getAttribute("__step__")` also via `Series::iterations::getAttribute("__step__")`, so we know which iteration to open in the first place. Otherwise, things get a bit hen-and-egg ;)
So, since that one is currently implemented, I'm keeping it.
> 
> I would rename `iterationIndex` to `__steps__` as we already do for `__data__`.

Done

ax3l:
Should we default to the new attribute layout #813 with `stepBased` from the beginning?

Maybe I overlook it in the review, but I cannot spot where this is exactly triggered in this PR if it is.

src/Iteration.cpp
```diff
@@ -259,6 +260,21 @@ Iteration::flushGroupBased(uint64_t i)
...
+void
+Iteration::flushVariableBased( uint64_t i )
+{
+    if( !written() )
+    {
+        /* create iteration path */
+        Parameter< Operation::OPEN_PATH > pOpen;
+        pOpen.path = "";
+        IOHandler()->enqueue( IOTask( this, pOpen ) );
+        this->setAttribute( "__step__", i );
```
ax3l: Just double-checking: this attribute is now, by virtue of our backend implementation, an ADIOS variable?
Let's just update the quoted layout in the PR description with the final layout when we merge it now :)

ax3l (approved):

test/SerialIOTest.cpp
```diff
@@ -3454,6 +3455,151 @@ TEST_CASE( "serial_iterator", "[serial][adios2]" )
...
+            REQUIRE( E_y.getExtent() == changingExtent );
+            last_iteration_index = iteration.iterationIndex;
+        }
+        REQUIRE( last_iteration_index == 9 );
+    }
+}
+
+TEST_CASE( "variableBasedSeries", "[serial][adios2]" )
+{
+    variableBasedSeries( "../samples/variableBasedSeries%V.bp" );
```
ax3l: Discussed offline: let's move this to JSON options

franzpoeschel: I suggest moving this to a separate PR, since this will also be used for the other iteration layouts and there are further refinements to the JSON configs that I would like to do apart from this, including:
* pick backend via JSON
* caps-insensitivity
* clearly define and document precedences (environment variables > JSON options > automatic detection from filenames)
* extended patterns as in PIConGPU (that one maybe as a further separate PR)

src/Iteration.cpp
franzpoeschel: Yes, it is now an ADIOS variable. PR https://github.com/openPMD/openPMD-api/pull/949 refines this notion a bit further by distinguishing attributes that change over steps from constant ones `bool changesOverSteps = false;`.
If the backend doesn't support this, the attribute should not be written at all, and likewise the frontend should have a fallback logic if the attribute is not found in the backend.

franzpoeschel:
Think I found an ancient bug: ADIOS1 reports the datatype of `unitDimension` as `VEC_DOUBLE`. This never came up since the `unitDimension` attributes are default-initialized and the defaults were not overwritten by what was actually in the file. They now are.

src/Series.cpp
```diff
@@ -996,6 +1044,33 @@ SeriesImpl::readGroupBased( bool do_init )
...
+    {
+    case IterationEncoding::groupBased:
+    /*
+     * Sic! This happens when a file-based Series is opened in group-based mode.
+     */
+    case IterationEncoding::fileBased:
+        for( auto const & it : *pList.paths )
+        {
+            uint64_t index = std::stoull( it );
+            readSingleIteration( index, it, true );
```
franzpoeschel, line 1079: Memo to self: Look at this again

franzpoeschel:
> Should we default to the new attribute layout #813 with `stepBased` from the beginning?
> 
> Maybe I overlook it in the review, but I cannot spot where this is exactly triggered in this PR if it is.

The ADIOS2 backend will now automatically switch to the new layout when selecting step-based iteration encoding.

Otherwise, this is ready for review from my side :) @ax3l

ax3l:
Awesome, can you please rebase out the little merge conflicts from the last merge and I'll get to it before I am on PTO :-)

franzpoeschel:
Passing apart from a CI run that fails to initialize. I think this one is ready @ax3l

ax3l:
Awesome! Addressing the `brew` macOS CI thingy in #970

ax3l (approved):
Awesome! :rocket: :sparkles:

ax3l:
@franzpoeschel I just realized: did we intentionally call this now `variableBased`? We designed this as `stepBased`:
https://github.com/openPMD/openPMD-standard/pull/250 I can change this in the standards as well, I just cannot recall if we intentionally renamed it.

franzpoeschel:
I initially named the new encoding "step-based", but in an offline discussion you suggested the use of "variable-based" encoding to stress the fact that we are using variables, i.e. datasets with version numbers, to encode openPMD iterations. @ax3l

ax3l:
@franzpoeschel Thanks. I looked at this again from an openPMD perspective, where we don't define the term "variable".
We currently have a "Series of Iterations", which will likely be called "Series of Snapshots" in 2.0 https://github.com/openPMD/openPMD-standard/issues/148

It's generally ok-ish to go with `variableBased` since it's ADIOS specific, but as I said it introduces a new term (`stepBased` would do so as well). Have no better idea at the moment either, though :)

Issue #856, opened 2020-12-21
Read from one (out of many) ranks. 


guj:
**Describe the bug**
Unable to open a  file-based series in an MPI program, then read from one of the rank. E.g . rank 0. 

**To Reproduce**
Compile-able/executable code example to reproduce the problem:

```C++
#include <openPMD/openPMD.hpp>
#include <mpi.h>

#include <iostream>
#include <memory>
#include <cstddef>


using std::cout;
using namespace openPMD;

int main(int argc, char *argv[])
{
    MPI_Init(&argc, &argv);

    int mpi_size;
    int mpi_rank;

    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);
 {
        Series series = Series(
            "../samples/8a_parallel_3Db_%T.bp",
            Access::READ_ONLY, MPI_COMM_WORLD);            

        MeshRecordComponent E_x = series.iterations[1].meshes["B"]["x"];

        Offset chunk_offset = {
          static_cast< long unsigned int >(mpi_rank) + 1, 1, 1
        };
        Extent chunk_extent = {2, 2, 1};

        if (0 == mpi_rank) 
          {
            auto chunk_data = E_x.loadChunk<double>(chunk_offset, chunk_extent);
            series.flush();        
          }
    }

    MPI_Finalize();

    return 0;
}

// ...
```

**Expected behavior**
when running with 1 rank,  read is successful. 
when running with 2 ranks,  hangs.  
Reason: 
when  Series is opened for reading , Series.cpp's readFileBased() is called (by all ranks).  This functions issues fClose upon finishes.  When Rank 0 wants to read a variable, readFileBased is invoked by rank 0 upon flush().  Which then tries to open the iteration again.  However because Rank 1 is inactive, open, a collective call,  can not proceed successfully. 

**Software Environment**
 current with 
 - version of ADIOS2: [e.g. 2.6.0]
 (did not check with H5)

franzpoeschel:
Some background:
 In ADIOS, opening a new engine (i.e. file) is a collective operation. This explains why the first access to an iteration in file-based iteration layout is collective, too. This is true for readers as well as for writers (a similarly structured writer as in Junmin's example would fail for writers, too).

[...]

PR #861, opened 2020-12-23
Release: 0.13.0

ax3l:
### Streaming Support, Python, Benchmarks

This release adds first support for streaming I/O via ADIOS2's SST engine. More I/O benchmarks have been added with realistic application load patterns. Many Python properties for openPMD attributes have been modernized, with slight breaking changes in Iteration and Mesh data order. This release requires C++14 and adds support for Python 3.9. With this release, we leave the "alpha" phase of the software and declare "beta" status.

Thanks to @ax3l, @franzpoeschel, @guj, @PrometheusPi, and @sbastrakov for contributing to this release!

### Python

```python
import openpmd_api
```

### C++

```C++
#include <iostream>

int main() {
    // query compile-time macros:

    return 0;
}
```

### CLI

```bash
openpmd-ls --version
```


ax3l:
We will address one more bug (#856/#862) before cutting the release.

CHANGELOG.rst
```diff
@@ -3,34 +3,102 @@
 Changelog
 =========
 
-0.13.0-beta
------------
-**Date:** TBA
+0.13.0
+------
+**Date:** 2020-12-23
```
ax3l: Update me once ready

PR #862, opened 2020-12-23
Iteration::open()

### LINKED ISSUE #856: Read from one (out of many) ranks.

ax3l:
Fix #856

### DIFF
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -23,6 +23,7 @@
 #include "openPMD/Datatype.hpp"
 #include "openPMD/Series.hpp"
 #include "openPMD/auxiliary/DerefDynamicCast.hpp"
+#include "openPMD/auxiliary/Filesystem.hpp"
 #include "openPMD/auxiliary/StringManip.hpp"
 #include "openPMD/backend/Writable.hpp"
 
@@ -171,6 +172,24 @@ Iteration::close( bool _flush )
                                       "unimplemented in auto-stepping mode." );
         }
     }
+    return *this;
+}
+
+Iteration &
+Iteration::open()
+{
+    Series * s = &auxiliary::deref_dynamic_cast< Series >(
+        parent->attributable->parent->attributable );
+    // figure out my iteration number
+    auto begin = s->indexOf( *this );
+    //auto end = begin;
+    //++end;
+    s->openIteration( begin->first, *this );
+    //s->flush_impl( begin, end ); // does not work as expected
+    //s->flush(); // does not work as expected
+    this->flush();
+    IOHandler->flush();
+
     return *this;
 }
 
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -182,13 +182,12 @@ Iteration::open()
         parent->attributable->parent->attributable );
     // figure out my iteration number
     auto begin = s->indexOf( *this );
-    //auto end = begin;
-    //++end;
-    s->openIteration( begin->first, *this );
-    //s->flush_impl( begin, end ); // does not work as expected
-    //s->flush(); // does not work as expected
-    this->flush();
-    IOHandler->flush();
+    auto end = begin;
+    ++end;
+    // set dirty, so Series::flush will open the file
+    this->dirty() = true;
+    s->flush_impl( begin, end );
+    this->dirty() = false;
 
     return *this;
 }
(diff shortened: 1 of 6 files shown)


src/Iteration.cpp
```diff
@@ -174,6 +175,24 @@ Iteration::close( bool _flush )
...
+Iteration &
+Iteration::open()
+{
+    Series * s = &auxiliary::deref_dynamic_cast< Series >(
+        parent->attributable->parent->attributable );
+    // figure out my iteration number
+    auto begin = s->indexOf( *this );
+    //auto end = begin;
+    //++end;
+    s->openIteration( begin->first, *this );
```
ax3l: Something is still wrong here, because it still hangs in the test.

test/ParallelIOTest.cpp
```diff
@@ -547,6 +547,102 @@ TEST_CASE( "close_iteration_test", "[parallel]" )
...
+    // check non-collective, parallel read
+    {
+        Series read( name, Access::READ_ONLY, MPI_COMM_WORLD );
+        Iteration it = read.iterations[ 30 ];
+        it.open(); // collective
+        if( mpi_rank == 0 ) // non-collective branch
+        {
+            auto E_x = it.meshes["E"]["x"];
+            auto data = E_x.loadChunk< double >();
+            read.flush();
```
ax3l, line 634: expected hang here that we try to fix with `it.open()`

src/Iteration.cpp
```diff
@@ -174,6 +175,24 @@ Iteration::close( bool _flush )
...
+    //s->flush_impl( begin, end ); // does not work as expected
+    //s->flush(); // does not work as expected
+    this->flush();
+    IOHandler->flush();
```
franzpoeschel:
```suggestion
    // figure out my iteration number
    auto begin = s->indexOf( *this );
    auto end = begin;
    ++end;
    // set dirty, so Series::flush will open the file
    this->dirty() = true;
    s->flush_impl( begin, end );
    this->dirty() = false;
```

`Series::openIteration` is for opening an iteration's file in file-based layout, so that will likely break if using group-based iteration layout.

src/Iteration.cpp
franzpoeschel: The hang is at the backends' fault i think. I've pushed a quick fix for ADIOS2, but I think ADIOS1 needs to be fixed too.

franzpoeschel (dismissed):

src/Iteration.cpp
franzpoeschel: The issue is with the backends. I've pushed a fix for ADIOS2, still need to look into the other ones

src/Iteration.cpp
ax3l: Oh, thanks!
Maybe we need to note that in the doxygen string of `openIteration`

ax3l:
I'll merge this for now, since ADIOS2 & HDF5 work, which is sufficient to unblock the 0.13.0 release #861 .

I will open a follow-up PR for ADIOS1 where we can fix that backend as well, which can go into 0.13.1/0.14.0.

PR #864, opened 2021-01-03
Fix ADIOS1: Iteration::open()

ax3l:
Prepare the support for `Iteration::open()` for following non-collective, parallel reads for ADIOS1.

Follow-up to #862 fixing ADIOS1 in #856


ax3l:
Thank you, @franzpoeschel! Great catch! I printed around that open but did not notice the double-open :)

PR #871, opened 2021-01-04
Adapt sourceID test to nondeterministic subfile order

### LINKED ISSUE #867: source ID changes in parallel I/O test
**Describe the bug**

On conda-forge, I see the following assert in ParallelIOTests:
```
$SRC_DIR/test/ParallelIOTest.cpp:346: FAILED:
  REQUIRE( chunk.offset == Offset{ static_cast< unsigned >( chunk.sourceID ), 0 } )

ParallelIOTests is a Catch v2.6.1 host application.
Run with -? for options

available_chunks_test
$SRC_DIR/test/ParallelIOTest.cpp:358

$SRC_DIR/test/ParallelIOTest.cpp:346: FAILED:
  REQUIRE( chunk.offset == Offset{ static_cast< unsigned >( chunk.sourceID ), 0 } )
with expansion:
  { 0, 0 } == { 1, 0 }

with expansion:
  { 0, 0 } == { 1, 0 }
```

**To Reproduce**
This occurs on macOS with OpenMPI. MPICH is unaffected.

Just running the tests as usual with `ctest`.

See: https://github.com/conda-forge/openpmd-api-feedstock/pull/59

**Expected behavior**
Should not fail or not check an attribute that is runtime dependent.

**Software Environment**
 - version of openPMD-api: 0.13.0
 - installed openPMD-api via: conda-forge (curing release workflow)
 - operating system: macOS 10.15 with Clang 11.0.0 (from LLVM, not AppleClang)
 - machine: Azure Pipelines
 - name and version of Python implementation: CPython 3.9 (all affected)
 - version of HDF5: 1.10.6
 - version of ADIOS1: 1.13.1
 - version of ADIOS2: 2.6.0
 - name and version of MPI: OpenMPI 4.0.5

**Additional context**
Occurred during release of openPMD-api 0.13.0 on conda-forge.

franzpoeschel:
The order of Subfiles in the ADIOS2 BP engines appears to be nondeterministic. Relying on that order in our tests is not safe, avoid that. Fix #867.

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -337,16 +337,26 @@ available_chunks_test( std::string file_ending )
         Iteration it0 = read.iterations[ 0 ];
         auto E_x = it0.meshes[ "E" ][ "x" ];
         ChunkTable table = E_x.availableChunks();
+        std::sort(
+            table.begin(),
+            table.end(),
+            []( auto const & lhs, auto const & rhs ) {
+                return lhs.offset[ 0 ] < rhs.offset[ 0 ];
+            } );
         std::vector< int > ranks;
         ranks.reserve( table.size() );
-        for( auto const & chunk : table )
+        for( size_t i = 0; i < ranks.size(); ++i )
         {
-            REQUIRE(
-                chunk.offset ==
-                Offset{ static_cast< unsigned >( chunk.sourceID ), 0 } );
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you! This looks good to me, will check it with the next release :)
(CI might have covered this but maybe we were just not hitting it yet - it's probably an unordered map or random order from parallel write in ADIOS2 that causes this)

PR #878, opened 2021-01-05
Ensure that a Series is not dirtyRecursive() after reading

franzpoeschel:
Reading routines (i.e. anything triggered by `Series::read()` should mark `dirty() = false` on all objects in the openPMD hierarchy. Re-reading a Series in streaming mode takes some shortcuts, and I previously forgot ensuring that any object is not dirty when taking such a shortcut.

@ax3l first noticed this while using `openpmd-ls` on the output of `3_write_serial.cpp`.
```
> openpmd-ls ../samples/3_write_serial.h5
[…]
number of iterations: 1 (groupBased)
  all iterations: 1 [~Series] An error occurred: [Series] Illegal access to iteration 1 that has been closed previously.
An error occurred while opening the specified openPMD series!
[Series] Illegal access to iteration 1 that has been closed previously.
```
This should fix that.

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -333,6 +333,7 @@ Mesh::read()
             MeshRecordComponent& rc = (*this)[component];
             if ( *rc.hasBeenRead )
             {
+                dirty() = false;
                 continue;
             }
             pOpen.path = component;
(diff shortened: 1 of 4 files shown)


ax3l (approved):
Thank you! :sparkles:

PR #882, opened 2021-01-05
Fix: listSeries & Iteration::close()

ax3l:
Fix calls of `listSeries` on files that used `Iteration::close()` during creation.


ax3l:
Thank you, @franzpoeschel :sparkles:

PR #883, opened 2021-01-05
Fix ADIOS1: listSeries

ax3l:
Fix ADIOS1 issue on listSeries seen with two tests.

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -1112,8 +1112,7 @@ void test_complex(const std::string & backend) {
     if( backend != "json" ) //! @todo clarify that complex data is not N+1 data in JSON
     {
         Series list{ "../samples/serial_write_complex." + backend, Access::READ_ONLY };
-        if( list.backend() != "ADIOS1" ) // FIXME: "Internal error: Failed to inquire about ADIOS variable during dataset opening"
-            helper::listSeries( list );
+        helper::listSeries( list );
     }
 }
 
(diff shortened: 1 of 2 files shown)


test/SerialIOTest.cpp
```diff
@@ -1112,8 +1112,7 @@ void test_complex(const std::string & backend) {
     if( backend != "json" ) //! @todo clarify that complex data is not N+1 data in JSON
     {
         Series list{ "../samples/serial_write_complex." + backend, Access::READ_ONLY };
-        if( list.backend() != "ADIOS1" ) // FIXME: "Internal error: Failed to inquire about ADIOS variable during dataset opening"
-            helper::listSeries( list );
+        helper::listSeries( list );
```
ax3l, line 1115:
```
[ADIOS1] Internal error: Failed to inquire about ADIOS variable
    '/data/0/meshes/CfltCflt' during dataset opening
```

Curious, since it should only have a problem with `Cldbl`, which we don't write.
Looks like it doubled the variable name in the path: `CfltCflt` instead of `Cflt`?

test/SerialIOTest.cpp
```diff
@@ -1536,8 +1535,7 @@ TEST_CASE( "patch_test", "[serial]" )
         if( t != "json" ) // FIXME: "[JSON] No such attribute in the given location."
         {
             Series list{"../samples/serial_patch." + t, Access::READ_ONLY};
-            if( list.backend() != "ADIOS1" ) // FIXME: "Internal error: Failed to inquire about ADIOS variable during dataset opening"
-                helper::listSeries(list);
+            helper::listSeries(list);
```
ax3l: Same curious variable-name doubling here (`xx` should read `x`):
```
[ADIOS1] Internal error: Failed to inquire about ADIOS variable
    '/data/1/particles/e/particlePatches/offset/xx' during dataset opening
```

franzpoeschel:
This seems to be essentially the same issue as in #884. Apparently, our tests didn't yet cover workflows where a dataset is opened twice and the JSON and ADIOS1 backends did not deal with that correctly.

ax3l:
Epic, thank you! :sparkles:

PR #884, opened 2021-01-05
Fix JSON: listSeries

ax3l:
Fix a JSON issue on listSeries seen with a test.

PR #885, opened 2021-01-05
JSON Complex Datatype Read

ax3l:
Add support to read back complex data back with the JSON backend.
Follow-up to #639

- [x] add recognition in the JSON backend that `COMPLEX` datatypes in JSON are not N+1 data


franzpoeschel:
Needed to rebase since the fix to #884 was relevant here, too

ax3l (commented):
Yay, thank you a lot! :rocket: :sparkles:

PR #889, opened 2021-01-08
Read JSON options from file

franzpoeschel:
Allow reading JSON options from a file. Triggered by prepending the filename with a '@'.

TODO

- [x] Documentation
- [x] Should we trip leading/suffix whitespaces? Currently, something like `Series("filename.bp", Access::Create, "    @config.json"` would fail. (Axel: yes)
- [x] Python bindings
- [x] Only read files for Series


src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -27,10 +27,39 @@
...
+{
+    nlohmann::json
+    parseOptions( std::string const & options )
+    {
+        if( options.at( 0 ) == '@' )
+        {
+            std::fstream handle;
+            std::string filename = options;
+            filename.erase( 0, 1 );
+            handle.open( filename, std::ios_base::in );
```
ax3l: I think this is reasonable, but in practice (read: e.g. an ADIOS run at scale) libs like Ascent and AMReX with similar handlers of option files rather read from one rank and MPI-Broadcast the result to all ranks - avoiding to hammer the filesystem with many meta-data operations opening the same file.

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -27,10 +27,39 @@
...
 {
+namespace
+{
+    nlohmann::json
+    parseOptions( std::string const & options )
+    {
+        if( options.at( 0 ) == '@' )
+        {
+            std::fstream handle;
+            std::string filename = options;
```
ax3l, line 44: Yes, I think implementing a `std::string trim( std::string )` method here is a good idea.

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -27,10 +27,39 @@
...
 #include <nlohmann/json.hpp>
 
 namespace openPMD
 {
+namespace
+{
+    nlohmann::json
+    parseOptions( std::string const & options )
+    {
+        if( options.at( 0 ) == '@' )
```
ax3l: Yes, I think implementing a `std::string trim( std::string )` method here is a good idea.

src/IO/AbstractIOHandlerHelper.cpp
franzpoeschel: Ah, good catch. Will do.

src/IO/AbstractIOHandlerHelper.cpp
franzpoeschel: Alright, then I'll add that.

src/IO/AbstractIOHandlerHelper.cpp
franzpoeschel: Done

docs/source/details/backendconfig.rst
```diff
@@ -21,6 +21,16 @@ Parameters that are directly passed through to an external library and not inter
 The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
 This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
 
+A JSON configuration may either be specified as a regular string that can be parsed as a JSON object, or alternatively as a path to a JSON-formatted text file.
+In the latter case, the file path must be prepended by an ampersand ``@``.
+
+.. note::
+
+   Reading text files in parallel openPMD is implemented in MPI-collective manner.
+   The file is read by rank 0 and its contents are then broadcast to all other ranks.
```
```suggestion
   The file is read by rank 0 and its contents are then broadcasted to all other ranks.
```

docs/source/details/backendconfig.rst
```diff
@@ -21,6 +21,16 @@ Parameters that are directly passed through to an external library and not inter
...
+   This happens at construction of the ``openPMD::Series`` object and when calling ``openPMD::RecordComponent::resetDataset()``, making those calls collective.
```
ax3l: Please add `RecordComponent::resetDataset()` to the table in `docs/source/details/mpi.rst`.

docs/source/details/backendconfig.rst
```diff
@@ -21,6 +21,16 @@ Parameters that are directly passed through to an external library and not inter
...
+   In order to avoid repeated (collective or non-collective) file access, the file contents for a dataset configuration may be buffered by calling ``openPMD::Dataset::resolveOptions()``.
```
ax3l: I am not sure how useful `resolveOptions` is or if it complicates the API.

In practice, when reading from a file, we don't change the options while a Series is already open. So for that case we can read in and broadcast on `Series()` and have all options available when we need them.

I think that we need additional string-based control on a per-dataset level, but I don't think we need a per-dataset control as a dataset-specific file - I think in practice a per-dataset control would be in the file that the `Series()` call already reads.

Following those assumptions, I think we can simplify this here a bit. Please tell me if discussing this in more detail in a VC can help :)

franzpoeschel: > In practice, when reading from a file, we don't change the options while a Series is already open.

Exactly this is supposed to be the use case. Otherwise, the file would be read upon each `RecordComponent::resetDataset()`. This method allows to read the file once and be done with it.

> I think that we need additional string-based control on a per-dataset level, but I don't think we need a per-dataset control as a dataset-specific file

True somewhat. Major reason I added this is for consistency.

If we allow reading JSON configs from files for datasets, I think some functionality to control MPI-collective operations explicitly like `resolveOptions()` is needed. Otherwise, we can drop it alltogether.

Sure, we can discuss this in a VC :)

docs/source/details/backendconfig.rst
franzpoeschel: Hm, according to [Wiktionary](https://en.wiktionary.org/wiki/broadcasted#English), both are correct forms but the short form is more common?

docs/source/details/backendconfig.rst
```diff
@@ -21,6 +21,9 @@ Parameters that are directly passed through to an external library and not inter
 The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
 This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
 
+A JSON configuration may either be specified as a regular string that can be parsed as a JSON object, or in the constructor of ``openPMD::Series`` alternatively as a path to a JSON-formatted text file.
+In the latter case, the file path must be prepended by an ampersand ``@``.
```
ax3l: and ampersand is a `&` :)
```suggestion
In the latter case, the file path must be prepended by an at-sign ``@``.
```

include/openPMD/auxiliary/JSON.hpp
```diff
@@ -23,11 +23,16 @@
...
-
 #include <memory>  // std::shared_ptr
+#include <nlohmann/json.hpp>
 #include <utility> // std::forward
 
+#if openPMD_HAVE_MPI
+
+#include <mpi.h>
+
+#endif
```
ax3l: Can you revert the include order for `<nlohmann/json.hpp>`? The one before was right (stdlibs last)

Also please move `mpi.h` before the stdlibs:
```suggestion
#include <nlohmann/json.hpp>

#if openPMD_HAVE_MPI
#   include <mpi.h>
#endif

#include <memory>  // std::shared_ptr
#include <utility> // std::forward
```

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -20,14 +20,14 @@
...
 #include "openPMD/IO/AbstractIOHandlerHelper.hpp"
 
+#include "openPMD/auxiliary/JSON.hpp"
 #include "openPMD/IO/ADIOS/ADIOS1IOHandler.hpp"
 #include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
 #include "openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp"
 #include "openPMD/IO/DummyIOHandler.hpp"
 #include "openPMD/IO/HDF5/HDF5IOHandler.hpp"
 #include "openPMD/IO/HDF5/ParallelHDF5IOHandler.hpp"
 #include "openPMD/IO/JSON/JSONIOHandler.hpp"
```
ax3l: I would continue to include this last, so we catch headers that forget to include used auxiliary functionality:
```suggestion
#include "openPMD/IO/ADIOS/ADIOS1IOHandler.hpp"
#include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
#include "openPMD/IO/ADIOS/ParallelADIOS1IOHandler.hpp"
#include "openPMD/IO/DummyIOHandler.hpp"
#include "openPMD/IO/HDF5/HDF5IOHandler.hpp"
#include "openPMD/IO/HDF5/ParallelHDF5IOHandler.hpp"
#include "openPMD/IO/JSON/JSONIOHandler.hpp"
#include "openPMD/auxiliary/JSON.hpp"
```

src/auxiliary/Filesystem.cpp
```diff
@@ -30,7 +30,10 @@
 #   include <dirent.h>
 #endif
 
+#include <fstream>
+#include <iostream>
 #include <stdexcept>
+
```
ax3l:
```suggestion
```

src/auxiliary/Filesystem.cpp
```diff
@@ -168,5 +171,92 @@ remove_file( std::string const& path )
...
+    struct MPI_Types< unsigned >
+    {
+        static MPI_Datatype const value;
+    };
+
+    /*
+     * Only some of these are actually instanciated,
+     * so suppress warnings for the others.
+     */
+    __attribute__((unused))
```
ax3l: This is GNU-ish known but a compiler-extension to std C++.
Please guard as usual via helpers like this:
https://github.com/openPMD/openPMD-api/blob/0.12.0-alpha/include/openPMD/auxiliary/Deprecated.hpp

Ref.: https://stackoverflow.com/a/31909713/2719194

src/auxiliary/JSON.cpp
```diff
@@ -21,10 +21,15 @@
...
-#include "openPMD/config.hpp"
-
+#include <cctype> // std::isspace
+#include <fstream>
+#include <sstream>
 #include <vector>
 
+#include "openPMD/auxiliary/Filesystem.hpp"
+#include "openPMD/auxiliary/Option.hpp"
+#include "openPMD/auxiliary/StringManip.hpp"
```
ax3l: Please move the three includes before the stdlib includes

test/ParallelIOTest.cpp
```diff
@@ -7,16 +7,16 @@
...
-#   include <list>
-#   include <memory>
-#   include <numeric>
-#   include <tuple>
+#    include <algorithm>
+#    include <fstream>
+#    include <iostream>
+#    include <list>
+#    include <memory>
+#    include <mpi.h>
```
ax3l: Please keep `mpi.h` before the stdlib includes

docs/source/details/backendconfig.rst
franzpoeschel: Will not do, since the change is undone.

src/auxiliary/Filesystem.cpp
```diff
@@ -168,5 +171,92 @@ remove_file( std::string const& path )
...
+                "Failed reading JSON config from file " + path + "." );
+        }
+        stringLength = res.size() + 1;
+    }
+    MPI_Datatype datatype = MPI_Types< size_t >::value;
+    int err = MPI_Bcast( &stringLength, 1, datatype, 0, comm );
+    if( err )
+    {
+        throw std::runtime_error(
+            "[collective_file_read] MPI_Bcast failure." );
```
ax3l:
```suggestion
            "[collective_file_read] MPI_Bcast stringLength failure." );
```

src/auxiliary/Filesystem.cpp
```diff
@@ -168,5 +171,92 @@ remove_file( std::string const& path )
...
+    std::vector< char > recvbuf( stringLength, 0 );
+    if(rank == 0)
+    {
+        std::copy_n(res.c_str(), stringLength, recvbuf.data());
+    }
+    err = MPI_Bcast( recvbuf.data(), stringLength, MPI_CHAR, 0, comm );
+    if( err )
+    {
+        throw std::runtime_error(
+            "[collective_file_read] MPI_Bcast failure." );
```
ax3l:
```suggestion
            "[collective_file_read] MPI_Bcast file content failure." );
```

ax3l (approved):
Looks great!
Will just make the error message unique.

PR #895, opened 2021-01-11
Support Null engine in ADIOS2

franzpoeschel:
The ADIOS2 backend only allows ADIOS2 engines that are explicitly supported by us. Add the `null` engine to that list, @guj needs it for performance testing.


guj (approved):
Thanks Franz! That would work!

Just FYI, null engine will not generate any file. It does nothing, as the name suggests.

franzpoeschel:
> Just FYI, null engine will not generate any file. It does nothing, as the name suggests.

Yes, I am aware. I just added the filename extension in there for completeness.

ax3l:
Oh, didn't we already add this in #656?

ax3l (approved):
I see, the new logic compares against allowed values :)
Thank you! :sparkles:

Issue #578, opened 2019-10-23
Add T3PIO


ax3l:
As in

A. Huebl et al., DRBSD-1 - ISC'17 (2017),
DOI:10.1007/978-3-319-67630-2_2, [arXiv:1706.00522](https://arxiv.org/abs/1706.00522)

we should add optional support for [T3PIO](https://github.com/TACC/t3pio) again to automatically tune PHDF5 operations on Lustre.
In the reference above we tuned MPI-I/O hints according to best-practices at the time but T3PIO still gave slightly better performance.

### Reference

Robert  McLay,  Doug  James,  Si  Liu,  John  Cazes,  and  William  Barth. *A  user-friendly approach for tuning parallel file operations.* In Proceedings of the International Conference for High Performance Computing, Networking, Storage andAnalysis, SC ’14, pages 229–236. IEEE Press, 2014.  DOI: 10.1109/SC.2014.24.

ax3l:
Alternatively, we could also try to find good empirics for `H5Pset_alignment` (and `H5Pset_sieve_buf_size`):
https://github.com/ComputationalRadiationPhysics/libSplash/issues/108#issuecomment-38681837

T3PIO seems to have set such hints via MPI info flags alone:
https://github.com/ComputationalRadiationPhysics/libSplash/issues/108#issuecomment-46784819

PR #898, opened 2021-01-13
[WIP] Parallel HDF5: 4MB Alignment & Buffer

ax3l:
FS blocksize:
```
stat -fc %s .
```

Tried those options on Cori (Scratch and CFS): 8_benchmark case with `-w`, KNL partition, WarpX-like MPI-rank placement.
modules: ... `darshan/3.1.7` `gcc/8.3.0` `cray-mpich/7.7.10` `cray-hdf5-parallel/1.10.5.2` ...

Scratch: 1MB recommended blocksize (confusingly, `stat -fc %s <dir>` reports 4KiB)
CFS: 16 MB blocksize (with 4MiB subblocks)

Support quote:
> blocksize is a quirky parameter for parallel file systems because between your compute node and the actual block devices are a bunch of network and RAID layers that have their own magic sizes.  Some arcane knowledge is required

Sets medium [striping](https://docs.nersc.gov/performance/io/lustre/).
**Note:** for proper ADIOS2 timings, keep the small default striping (it creates subfiles that should not be heavily striped); for proper HDF5 timings, enable striping (single output file that should be heavily striped).

For HDF5, we can also try [T3PIO](https://github.com/ComputationalRadiationPhysics/libSplash/issues/108) `MPI_Info` hints again.

[cori.sbatch.txt](https://github.com/openPMD/openPMD-api/files/5817775/cori.sbatch.txt)

### Cori: Darshan Logs

```bash
# MPICH statistics collection
export MPICH_MPIIO_STATS=1
export MPICH_MPIIO_HINTS_DISPLAY=1
export MPICH_MPIIO_TIMERS=1

# Darshan extended trace (dxt) logs
export DARSHAN_DISABLE_SHARED_REDUCTION=1
export DXT_ENABLE_IO_TRACE=4

# work-around needed
export LD_PRELOAD=/global/common/cori_cle7/software/darshan/3.1.7/lib/libdarshan.so

// srun

# disable work-around
unset LD_PRELOAD
```

### DIFF
--- a/docs/source/backends/hdf5.rst
+++ b/docs/source/backends/hdf5.rst
@@ -25,7 +25,7 @@ The following environment variables control HDF5 I/O behavior at runtime.
 environment variable                  default   description
 ===================================== ========= ====================================================================================
 ``OPENPMD_HDF5_INDEPENDENT``          ``ON``    Sets the MPI-parallel transfer mode to collective (``OFF``) or independent (``ON``).
-``OPENPMD_HDF5_ALIGNMENT``            ``1``     Tuning parameter for parallel I/O, choose an alignment which is a multiple of the disk block size.
+``OPENPMD_HDF5_ALIGNMENT``            ``ABC``   Tuning parameter for parallel I/O, choose an alignment which is a multiple of the disk block size.
 ``OPENPMD_HDF5_CHUNKS``               ``auto``  Defaults for ``H5Pset_chunk``: ``"auto"`` (heuristic) or ``"none"`` (no chunking).
 ``H5_COLL_API_SANITY_CHECK``          unset     Set to ``1`` to perform an ``MPI_Barrier`` inside each meta-data operation.
 ===================================== ========= ====================================================================================
--- a/docs/source/backends/hdf5.rst
+++ b/docs/source/backends/hdf5.rst
@@ -21,14 +21,14 @@ Backend-Specific Controls
 
 The following environment variables control HDF5 I/O behavior at runtime.
 
-===================================== ========= ====================================================================================
-environment variable                  default   description
-===================================== ========= ====================================================================================
-``OPENPMD_HDF5_INDEPENDENT``          ``ON``    Sets the MPI-parallel transfer mode to collective (``OFF``) or independent (``ON``).
-``OPENPMD_HDF5_ALIGNMENT``            ``ABC``   Tuning parameter for parallel I/O, choose an alignment which is a multiple of the disk block size.
-``OPENPMD_HDF5_CHUNKS``               ``auto``  Defaults for ``H5Pset_chunk``: ``"auto"`` (heuristic) or ``"none"`` (no chunking).
-``H5_COLL_API_SANITY_CHECK``          unset     Set to ``1`` to perform an ``MPI_Barrier`` inside each meta-data operation.
-===================================== ========= ====================================================================================
+===================================== =========== ====================================================================================
+environment variable                  default     description
+===================================== =========== ====================================================================================
+``OPENPMD_HDF5_INDEPENDENT``          ``ON``      Sets the MPI-parallel transfer mode to collective (``OFF``) or independent (``ON``).
+``OPENPMD_HDF5_ALIGNMENT``            ``4194304`` Tuning parameter for parallel I/O, choose an alignment which is a multiple of the disk block size.
+``OPENPMD_HDF5_CHUNKS``               ``auto``    Defaults for ``H5Pset_chunk``: ``"auto"`` (heuristic) or ``"none"`` (no chunking).
+``H5_COLL_API_SANITY_CHECK``          unset       Set to ``1`` to perform an ``MPI_Barrier`` inside each meta-data operation.
+===================================== =========== ====================================================================================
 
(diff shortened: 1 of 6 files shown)


src/IO/HDF5/ParallelHDF5IOHandler.cpp
```diff
@@ -73,16 +73,65 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(AbstractIOHandler* handler,
 
     herr_t status;
     status = H5Pset_dxpl_mpio(m_datasetTransferProperty, xfer_mode);
+    VERIFY(status >= 0, "[HDF5] Internal error: Failed to set H5Pset_dxpl_mpio");
 
-    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "1" );
+    auto const strByte = auxiliary::getEnvString( "OPENPMD_HDF5_ALIGNMENT", "4194304" );
```
ax3l: Sample & bin directory `du -hs` with:
| `OPENPMD_HDF5_ALIGNMENT` | size |
|--|--|
| 1 | 7.8M + 1.4G |
| 4194304 | 7.8M + 1.4G |

on my laptop (4KiB blocksize).

`ls` output (less reliable) also ok for parallel files (not padded to multiples of 4MiB). So either this is cleverly compacted or has no influence...

src/IO/HDF5/ParallelHDF5IOHandler.cpp
```diff
@@ -73,16 +73,65 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(AbstractIOHandler* handler,
...
+     * file drivers, all calls to H5Dread and H5Dwrite will access the disk directly,
+     * and H5Pset_cache will have no effect on performance."
+     */
+    int metaCacheElements = 0;
+    size_t rawCacheElements = 0;
+    size_t rawCacheSize = 0;
+    double policy = 0.0;
+    status = H5Pget_cache(m_fileAccessProperty, &metaCacheElements, &rawCacheElements, &rawCacheSize, &policy);
+    VERIFY(status >= 0, "[HDF5] Internal error: Failed to set H5Pget_cache");
+    rawCacheSize = bytes * 4; // default: 1 MiB per dataset
```
ax3l: should be guarded so we don't accidentally set 1Byte cache if a user provides `1`

ax3l:
We can run these tests again after #916 was merged, maybe we see some improvement when setting striping with chunked data sets

ax3l:
Should also set  `H5Pset_alignment` (and `H5Pset_sieve_buf_size`) here:
https://github.com/openPMD/openPMD-api/issues/578#issuecomment-865377511

docs/source/backends/hdf5.rst
```diff
@@ -25,7 +25,7 @@ The following environment variables control HDF5 I/O behavior at runtime.
 environment variable                  default   description
 ===================================== ========= ====================================================================================
 ``OPENPMD_HDF5_INDEPENDENT``          ``ON``    Sets the MPI-parallel transfer mode to collective (``OFF``) or independent (``ON``).
-``OPENPMD_HDF5_ALIGNMENT``            ``1``     Tuning parameter for parallel I/O, choose an alignment which is a multiple of the disk block size.
+``OPENPMD_HDF5_ALIGNMENT``            ``ABC``   Tuning parameter for parallel I/O, choose an alignment which is a multiple of the disk block size.
```
ax3l: todo with final value

ax3l:
Next measurements we should try on Cori (Suren Byna):

## Option 1

- Set the alignment to 8 MB (in the `H5Pset_alignment()` call, threshold of `0` and alignment of `8MB`)
- Set striping on the directory where the data is being written. 
- Stripe count: 40
- Stripe size: 8 MB

Just in case, here’s the command to set the stripe on a directory.
```bash
lfs setstripe --stripe_count 40 --stripe_size 8m ./benchmarks
```

##  Option 2

- Alignment of 16 MB
- Stripe count : 40
- Stripe size: 16 MB

## Jan/Feb tests

We tried various sizes in Jan/February with the job script linked above in the PR description. We saw no improvement on Cori at the time.

Since then, we implementing chunking #406 and changed the benchmark from then from 4D to 3D: #1010
Also, we have new parallel benchmarks now (8a, 8b).

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -86,11 +86,37 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(
...
+            }
+            if( alignmentConfig.json().contains( "threshold" ) )
+            {
+                auto const strThreshold = alignmentConfig[ "threshold" ].json().get< std::string >();
+                std::stringstream sstream(strThreshold);
+                sstream >> m_threshold;
+            }
+            // align all (no threshold) if only alignment is set
+            if( m_alignment > 1 && m_threshold == 1 )
+                m_threshold = 0;
```
ax3l, line 118: This needs to be moved behind the whole `if( config.contains( "hdf5" ) )` block, otherwise the env var `OPENPMD_HDF5_ALIGNMENT` will not imply `m_threshold = 0` for values `>1`

PR #899, opened 2021-01-13
Weak scaling in benchmark example

franzpoeschel:
So far, the `8_benchmark_parallel.cpp` examples writes a fixed-size dataset of ~3/4GB independent of MPI size. This PR multiplies that size by the `MPI_size` to achieve weak scaling.

TODO:
- [ ] Let users choose the size per rank via command line arguments?

### DIFF
--- a/docs/source/usage/benchmarks.rst
+++ b/docs/source/usage/benchmarks.rst
@@ -6,14 +6,14 @@ Benchmarks
 Parallel benchmark 8
 --------------------
 
-Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`_, this benchmark executes a simple parallel read-write test.
+Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`, this benchmark executes a simple parallel read-write test.
 
-In particular, this test case writes and reads a 4D array of type ``long``, sliced 1D along the first dimension.
+In particular, this test case writes and reads a 4D array of type ``uint64_t``, sliced 1D along the first dimension.
 
 .. code-block:: cpp
 
    openPMD::Extent total{
-       100 * scale_up, // slices along or this axis?
+       100 * scale_up, // sliced along this axis over MPI ranks
        100,
        100,
        10
--- a/docs/source/usage/benchmarks.rst
+++ b/docs/source/usage/benchmarks.rst
@@ -8,19 +8,19 @@ Parallel benchmark 8
 
 Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`_, this benchmark executes a simple parallel read-write test.
 
-In particular, this test case writes and reads a 4D array of type ..., sliced 1D along the ... dimension.
+In particular, this test case writes and reads a 4D array of type ``long``, sliced 1D along the first dimension.
 
 .. code-block:: cpp
 
    openPMD::Extent total{
-       100 * scale_up, // slices along or this...?
+       100 * scale_up, // slices along or this axis?
        100,
        100,
-       10 // or this axis?
+       10
    };
 
-That means in the strong-scaling case, always ... GB of data are produced.
-In the weak-scaling case, the data scales as :math:`N * ... \mathrm{GiB}` with :math:`N` as the number of participating MPI ranks.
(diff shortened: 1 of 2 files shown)


guj (commented):
From the way it was, this test is a strong scaling case to write a few variables of size {100,100,100,10},  while the work load is sliced along the first dimension.  File size is about 700MB 

And multiple the first dimension with mpi size, then it becomes a weak scaling case. 

I like the idea to have a parameter for the first dimension size,  So we can control the file size. 
Would you like another parameter for weak/strong scaling option (default can be either).   So this test is able to do both if needed.

ax3l:
Let's also update the docs in `docs/source/usage/benchmarks.rst` on how to use it :)

docs/source/usage/benchmarks.rst
```diff
@@ -3,6 +3,28 @@
 Benchmarks
 ==========
 
+Parallel benchmark 8
+--------------------
+
+Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`_, this benchmark executes a simple parallel read-write test.
+
+In particular, this test case writes and reads a 4D array of type ..., sliced 1D along the ... dimension.
```
ax3l: @franzpoeschel can you potentially fill in the `...`s? :)

docs/source/usage/benchmarks.rst
```diff
@@ -3,6 +3,28 @@
...
+.. code-block:: cpp
+
+   openPMD::Extent total{
+       100 * scale_up, // slices along or this...?
+       100,
+       100,
+       10 // or this axis?
+   };
+
+That means in the strong-scaling case, always ... GB of data are produced.
```
ax3l: Here as well please :)

ax3l (commented):
I added CLI options and started the docs :)

@franzpoeschel can you please review my changes and fill in the last free spots in the docs? :)

docs/source/usage/benchmarks.rst
```diff
@@ -3,6 +3,28 @@
...
+--------------------
+
+Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`_, this benchmark executes a simple parallel read-write test.
+
+In particular, this test case writes and reads a 4D array of type ..., sliced 1D along the ... dimension.
+
+.. code-block:: cpp
+
+   openPMD::Extent total{
+       100 * scale_up, // slices along or this...?
```
ax3l: let's place the `// sliced along this axis over MPI ranks` comment on the right axis

franzpoeschel:
I've filled in the gaps. Should I add another option to control the amount of data per rank? Or stick with what we have now?

docs/source/usage/benchmarks.rst
```diff
@@ -3,6 +3,28 @@
...
+--------------------
+
+Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`_, this benchmark executes a simple parallel read-write test.
+
+In particular, this test case writes and reads a 4D array of type ``long``, sliced 1D along the first dimension.
+
+.. code-block:: cpp
+
+   openPMD::Extent total{
+       100 * scale_up, // slices along or this axis?
```
```suggestion
       100 * scale_up, // sliced along this axis over MPI ranks
```

docs/source/usage/benchmarks.rst
```diff
@@ -3,6 +3,28 @@
 Benchmarks
 ==========
 
+Parallel benchmark 8
+--------------------
+
+Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`_, this benchmark executes a simple parallel read-write test.
```
```suggestion
Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`, this benchmark executes a simple parallel read-write test.
```

docs/source/usage/benchmarks.rst
```diff
@@ -3,6 +3,28 @@
...
+
+In particular, this test case writes and reads a 4D array of type ``long``, sliced 1D along the first dimension.
```
```suggestion
In particular, this test case writes and reads a 4D array of type ``uint64_t``, sliced 1D along the first dimension.
```

ax3l (commented):
Thank you, that's great.
I think the current version is good, we can add a control later on if we need it.

ax3l (approved):

PR #901, opened 2021-01-14
Expose internal buffers to writers

franzpoeschel:
~~Based on #904 (see below for the reason), for comparison: https://github.com/franzpoeschel/openPMD-api/compare/topic-pipe-executable...franzpoeschel:topic-span~~

This makes the functionality provided by `adios2::Variable::Span<T>` available to users of the openPMD API.

Use case: If in the writing application, write buffers are newly allocated right before `RecordComponent::storeChunk()` (as opposed to passing buffers to `storeChunk()` that already exist in the application), this PR allows the buffer to be allocated by the openPMD backend (e.g. in ADIOS2: provide a view into the serialization buffers). In such a case, this may avoid memcopies, ideally cutting memory usage in half.

Possible benefitting applications:
* PIConGPU: Before storing record components, the openPMD plugin in PIConGPU converts data from the PIConGPU-internal AoS-style representation into the openPMD-style SoA-style representation. Using this PR, data can then be converted directly into backend buffers.
    UPDATE: I've implemented this in a PIConGPU branch, [this diff](https://github.com/franzpoeschel/picongpu/compare/38bd5db0f2ab63a67e2bd928048cf61abff24995...franzpoeschel:topic-span) gives an example on usage.
* [openPMD-pipe](https://github.com/openPMD/openPMD-api/pull/904): Data can be directly loaded from the reading backend into the writing backend, avoiding the detour through user-space buffers.

TODO:
- [x] Parallel testing, more testing. Note: Our use of this in openpmd-pipe very exhaustively tests this in parallel.
- [x] Generic fallback implementation for backends that do not support this mode
- [x] ADIOS2 implementation
- [x] ~~[Failing test](https://github.com/openPMD/openPMD-api/pull/901/checks?check_run_id=1704324106): I can reproduce this with ADIOS 2.6.0, ADIOS 2.7.0 does not have this issue. Hence, this needs to wait until we bump the required ADIOS2 version.~~ nah, that one was on me actually
- [x] There seems to be some failure in ADIOS2 for larger datasets (`openpmd-pipe.py`ing the HDF5 git sample crashes)
- [x] Add IOTask to ADIOS1 queues
- [x] Python bindings
- [x] Documentation, cleanup
- [x] Fix handling of scalar record components
- [x] Fix intermittent resizing of serialization buffer in ADIOS2
- [x] Make old spans reusable?
- [x] Merge #904 first
- [x] This one reshuffles parts of our flushing logic. Properly document and check that things are fine.
- [x] Allow deferred calls to avoid buffer reallocations in ADIOS2.
- [x] I've rebased this onto #945 to ease benchmarking with `openpmd-pipe`. So, merge that first.
- [x] Avoid buffer reallocations in ADIOS – impossible without explicit support from ADIOS
- [x] Should we update the [`Span` interface](https://github.com/openPMD/openPMD-api/blob/fe52a2fef91dc06ae237553fd619548a62310f9f/include/openPMD/Span.hpp#L36) in C++ to be more similar to the [one in Python](https://github.com/openPMD/openPMD-api/blob/fe52a2fef91dc06ae237553fd619548a62310f9f/src/binding/python/RecordComponent.cpp#L689)? Maybe this silent buffer reallocation thing is a bit too much to hide it automatically behind a Span interface.. ?


franzpoeschel:
I've benchmarked this for a PIConGPU simulation that dumped 124GB of data in 4 IO passes from 1 GPU (¹).
Note that PIConGPU already saves memory by reusing store buffers and flushing them to the backend for each single dataset anew, reducing the amount of improvement attainable by using the span-based API.

The current memory usage, as profiled by KDE Heaptrack, peaks at 87.1GB:
Bildschirmfoto von 2021-02-23 14-02-57
Simulation times for the same run, but without Heaptrack:
```
PIConGPUVerbose INPUT_OUTPUT(32) | openPMD: IO plugin ran for 32sec 737msec (average: 35sec 126msec)
calculation  simulation time:  3min 43sec 439msec = 223 sec
full simulation time:  4min 54sec 746msec = 294 sec
```

After using this PR in PIConGPU, it peaks at 85.4GB, saving 1.7 ~ 1.8GB of heap memory:
Bildschirmfoto von 2021-02-23 14-01-58
Simulation times for the same run, but without Heaptrack:
```
PIConGPUVerbose INPUT_OUTPUT(32) | openPMD: IO plugin ran for 28sec 372msec (average: 31sec 969msec)
calculation  simulation time:  3min 30sec 703msec = 210 sec
full simulation time:  5min 24sec 709msec = 324 sec
```
So, we saved ~3 seconds of memory allocation time per run of the IO plugin.

The remaining memory spikes stem from the DoubleBuffer allocation strategy in the openPMD plugin.
I expect starker differences in our little `openpmd-pipe` tool, will come back to that.

[1] KelvinHelmholtz example called with `picongpu -s 150 -d 1 1 1 -g 256 512 128 --openPMD.period 50 --openPMD.file dump --openPMD.infix NULL --openPMD.ext bp --openPMD.json '{ "adios2": { "engine": { "usesteps": true, "parameters": { "InitialBufferSize": "34Gb", "Profile": "On" } }, "dataset": { "operators": [ ] } } } ' --versionOnce`

include/openPMD/RecordComponent.hpp
```diff
@@ -75,6 +75,57 @@ struct IsContiguousContainer< std::array< T_Value, N > >
 };
 } // namespace traits
 
+/**
+ * @brief Subset of C++20 std::span class template.
+ *
+ * Any existing member behaves equivalently to those documented here:
+ * https://en.cppreference.com/w/cpp/container/span
+ */
+template< typename T >
```
ax3l, line 78: Let's put this in its own header file, please :)

include/openPMD/RecordComponent.hpp
```diff
@@ -75,6 +75,57 @@ struct IsContiguousContainer< std::array< T_Value, N > >
...
+        return static_cast< T * >( m_param.out->ptr );
+    }
+
+    T &operator[]( size_t i ) const
+    {
+        return data()[ i ];
+    }
+};
+
+template class Span< int >;
```
ax3l: Maybe in a `Span.cpp` file?

include/openPMD/RecordComponent.hpp
```diff
@@ -430,4 +525,103 @@ RecordComponent::storeChunk(T_ContiguousContainer & data, Offset o, Extent e)
 
     storeChunk(shareRaw(data), offset, extent);
 }
+
+template< typename T, typename F >
+inline Span< T >
+RecordComponent::storeChunk( Offset o, Extent e, F && createBuffer )
```
ax3l: Is it worth adding a doxygen string to this overload that shows a usage example or so?

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -56,6 +56,28 @@ class unsupported_data_error : public std::runtime_error
     virtual ~unsupported_data_error() { }
 };
 
+/**
+ * @brief Determine what items should be flushed upon Series::flush()
+ *
+ */
+enum class FlushLevel : unsigned char
+{
+    UserFlush,
```
ax3l, line 70: Can you also add a `UserFlush` doc string, please?

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -56,6 +56,28 @@ class unsupported_data_error : public std::runtime_error
...
+    /**
+     * Default mode, flush everything that can be flushed
+     * Does not need to uphold user-level guarantees about clearing and filling
```
ax3l: I don't understand the 2nd sentence yet:
Are you saying that this cannot check the API contract we have with the user, aka that the user must have given us valid buffers and they are ready to use at this point?

test/SerialIOTest.cpp
```diff
@@ -3310,11 +3310,49 @@ iterate_nonstreaming_series( std::string const & file )
...
-            // so let's do the switcharoo for this test
+            E_x.storeChunk( data, { 0, 0 }, { 1, extent } );
+            bool taskSupportedByBackend = true;
+            Span< int > span = E_x.storeChunk< int >(
+                { 1, 0 },
+                { 1, extent },
+                /*
+                 * Hijack the functor that is called for buffer creation if the
+                 * backend doesn't support the task to see whether the backend
+                 * did support it or not.
```
ax3l: Can you please split this into two sentences? It's a bit hard to parse for me :)

ax3l: Should we make this lambda a helper function? Looks like a generally useful implementation that people would use?

include/openPMD/RecordComponent.hpp
```diff
@@ -75,6 +75,57 @@ struct IsContiguousContainer< std::array< T_Value, N > >
...
+class Span
+{
+    friend class RecordComponent;
+
+private:
+    using param_t = Parameter< Operation::GET_BUFFER_VIEW >;
+    param_t m_param;
+    size_t m_size;
+    // @todo make this safe
+    Writable * m_writable;
```
ax3l: todo :)

include/openPMD/IO/AbstractIOHandler.hpp
franzpoeschel: I need to rename some of those anyway, planned to do that then.

include/openPMD/RecordComponent.hpp
franzpoeschel: done

include/openPMD/RecordComponent.hpp
franzpoeschel: Ah, that one was just for quickly instantiating to get feedback from the compiler. Removed it again.

include/openPMD/RecordComponent.hpp
franzpoeschel: done

include/openPMD/RecordComponent.hpp
franzpoeschel: A usage example might be better in form of an actual executable example? Should definitely add that to demonstrate usage.
EDIT: Added it

franzpoeschel:
The last commit adds documentation on the envisioned `flush()` workflow. I think I'll need some feedback on whether this is the way to go forward @ax3l

include/openPMD/IO/AbstractIOHandler.hpp
franzpoeschel: I've clarified documentation now by introducing a concept of flush points. Difference between both modes is that UserFlush defines a flush point and ~~FlushEverything~~ InternalFlush does not.

test/SerialIOTest.cpp
franzpoeschel: > Can you please split this into two sentences? It's a bit hard to parse for me :)

Will do

> Should we make this lambda a helper function? Looks like a generally useful implementation that people would use?

I'm not sure about it.. This sidechannels the return value (the boolean) as a value caught-by-reference. Can we write this in a way that would actually be a good API?

franzpoeschel:
While benchmarking `openpmd-pipe` (old storeChunk() vs. the new span-based one), I noticed that ADIOS2 supports no deferred span-based put. The old `openpmd-pipe` does not run into the resizing issues described in [this bug report](https://github.com/ornladios/ADIOS2/issues/2629) because it works in three phases:
1. Schedule all storeChunk()s of an iteration without flushing anything
2. Fill all the buffers that have been scheduled for writing
3. Close the iteration, thus flushing the chunks.

This allows the ADIOS2 backend to use the clear separation into "declaration" and "definition" that ADIOS2 prefers. The ADIOS2 serialization buffer can be set to the right size before the first flush even happens. 
Using the Span-based API, this line however will allocate its required memory immediately:
```cpp
        typename adios2::Variable< T >::Span span = engine.Put( variable );
```
At first, this seems necessary in order to actually return a memory view in form of a span. But one single `Span` does not point to the same memory location during its lifetime anyway since resize operations may happen in between (e.g. call `span.data()` on the same object twice and you might get two different pointers). Hence, a deferred `Engine::Put()` would make sense.

Reproducer:
```cpp
#include <adios2.h>
#include <vector>

int main( int argsc, char ** argsv )
{
    std::string engine_type = "bp4";

    adios2::ADIOS adios;
    adios2::IO IO = adios.DeclareIO( "IO" );
    IO.SetEngine( engine_type );
    adios2::Engine engine = IO.Open( "span.bp", adios2::Mode::Write );

    // write 6GB of data in 1GB pieces
    using datatype = char;
    size_t extent = 1024 * 1024 * 1024;

    std::vector< adios2::detail::Span< datatype > > spans;
    spans.reserve( 6 );

    // first phase: get spans, but don't write anything yet
    for( size_t i = 0; i < 6; ++i )
    {
        std::string variableName = "variable" + std::to_string( i );
        auto variable = IO.DefineVariable< datatype >(
            variableName, { extent }, { 0 }, { extent } );
        // this call doesn't technically need to allocate memory yet
        // but it does
        // Norbert feedback: it does need to allocate, because it returns a pointer
        // to us
        spans.emplace_back( engine.Put( variable ) );
    }

    // second phase: fill spans with user data
    for( size_t i = 0; i < 6; ++i )
    {
        auto & span = spans[ i ];
        std::fill( span.begin(), span.end(), i );
    }

    // third phase: let the engine read what we wrote
    engine.Close();
}
```
KDE Heaptrack visualizes the resizing behavior:
Bildschirmfoto von 2021-03-15 14-57-49

If not setting the `InitialBuffersize` correctly, the new `openpmd-pipe` shows a similar behavior while the old one doesn't (see upcoming post).

**Is this important enough a use case for us that I should open a feature request for this?**

franzpoeschel:
The benchmark results for `openpmd-pipe` can be seen below, row by row (first picture: general statistics for the 4 benchmarks, row by row; second picture: memory consumption over time for the 4 benchmarks, row by row). Except for the 4th benchmark, I use a file-based iteration layout in openPMD, meaning that I open a new ADIOS2 file for each iteration, forcing ADIOS2 to re-initialize its internal buffers. The first iteration is the largest at 1.8GB, the last one the smallest at 42MB.
1. The old `openpmd-pipe` behavior. Solid, but not outstanding performance. Memory consumption peaks at 4.1GB, the runtime is 38 seconds.
2. The new `openpmd-pipe` behavior, without specifying any ADIOS2 options such as the initial buffer size. The graph shows the repeated buffer resizing in small steps. Memory consumption peaks at 3.7GB, slightly lower than (1), but the runtime is dominated by resizing, 90 seconds.
3. Specifying the `InitialBufferSize` as 2GB brings the memory consumption down to 2.3GB, cutting it nearly in half. This shows that we do not any longer duplicate data, but write it directly into ADIOS2 buffers instead. The runtime is 43 seconds.
4. Benchmark (3) is slower than (1) since the last iterations are very small and buffers need not be initialized to 2GB, making buffer allocation in (1) more efficient than in (3). A trick to avoid this is to use group-based iteration layout instead and activate ADIOS2 steps, making it possible for ADIOS2 to reuse buffers across steps. The memory line is hence very flat (but still peaks at 2.3GB), saving some 15 seconds in runtime and bringing it down to 28 seconds.

**Conclusions**:
* This PR allows avoiding data duplications, cutting memory usage in half in extreme cases.
* While previously, `openpmd-pipe` was already reasonably efficient at the default configuration, the `--outconfig` now needs to be specified if a performance degradation is to be avoided.

Bildschirmfoto von 2021-03-15 15-37-40
Bildschirmfoto von 2021-03-15 15-37-53

test/SerialIOTest.cpp
franzpoeschel: > Will do

Done

docs/source/usage/workflow.rst
```diff
@@ -0,0 +1,30 @@
...
+IO operations are in general not performed by the openPMD API immediately after calling the corresponding API function.
+Rather, operations are enqueued internally and performed at so-called *flush points*.
+A flush point is a point within an application's sequential control flow where the openPMD API must uphold the following guarantees:
+
+*   In write mode, any change made to a user buffer whose data shall be stored in a dataset up to the flush point must be found in the written dataset.
+*   In write mode, no change made to a user buffer whose data shall be stored in a dataset after the flush point must be found in the written dataset.
+*   In read mode, a buffer into which data from a dataset should be filled, must not be altered by the openPMD API before the flush point.
+*   In read mode, a buffer into which data from a dataset should be filled, must have been filled with the requested data after the flush point.
+
+In short: ``storeChunk()`` and ``loadChunk()`` operations must happen exactly at flush points.
```
ax3l: Maybe:
```suggestion
In short: operations requested by ``storeChunk()`` and ``loadChunk()`` must happen exactly at flush points.
```
?

ax3l: We thought for some time, that we would allow these operations in the API generally between `store/loadChunk` and the `flush()`, but as we discussed recently, we do no want to keep that (do you remember the discussion @franzpoeschel? We should have taken minutes.)

I think the main reason was that we return the spans in those operations and then fill them before flush from the user side (for write).

docs/source/usage/workflow.rst
```diff
@@ -0,0 +1,30 @@
...
+Flush points are triggered by:
+
+*   Calling ``Series::flush()``.
+*   Calling ``Iteration::flush( flush=true )``.
+    Flush point guarantees affect only the corresponding iteration.
+*   The streaming API (i.e. ``Series.readIterations()`` and ``Series.writeIteration()``) automatically before accessing the next iteration.
+
+Attributes are (currently) unaffected by this:
+
+*   In writing, attributes are stored by value and can afterwards not be aliased by the user.
```
ax3l: What is meant with aliased? :sweat_smile:

ax3l: (Does anything change for ADIOS2 where attributes become vars in the new layout? Probably not, just double-checking.)

ax3l:
We discussed the comment https://github.com/openPMD/openPMD-api/pull/901#issuecomment-799460797 and it would be good to transfer it to ADIOS2 as a motivating example for BP5 to keep track of. One of the solutions in BP5 could be to allow similar to openPMD-api to pass buffers of individual (large) variables from the user-side to ADIOS2 to keep track of (basically `PutDeferred`).

The return of a span currently cannot easily keep track of the proper span pointers if resizes occur. One could introduce additional phases to declare when all local size declaration (puts) are done, in both APIs (openPMD-api, ADIOS2), and keep track of the sizes. But maybe that is too complicated in workflows and API communication to users?

During the meeting we had another idea today: ADIOS2 could keep a reference of its `span`s (or shared buffers) and change the data pointers (in `.data()`/`.get()`). The API contract would then be that users need to access a fresh pointer after each `Put()` in case. This makes it a bit more convenient during resizes, but does not solve the resizes are still occurring generally. (Reminder: during resizes, already written data behind a span is copied over, so there is no problem with that at least.)

For BP4, it looks like the only work-around is to pre-allocate (request) once form the user-side for a sufficiently large buffer and store into that buffer via the provided spans. For BP5, @pnorbert considers dropping span for, e.g. a `shared_ptr` interface of user-side buffers as we do (for large variables). Or a serialize in ADIOS2 that is chunked in a more sophisticated way for resizes: (see also https://github.com/openPMD/openPMD-api/issues/937#issuecomment-795641082) See also https://github.com/ornladios/ADIOS2/issues/2649 for memory control (drain/flush) in ADIOS2. For now, we need `span`s because they are the only way to get an "allocate-once" workflow implemented.

also cc-ing @williamfgc on this in case he has thoughts on this :)

docs/source/usage/workflow.rst
```diff
@@ -0,0 +1,30 @@
+.. _workflow:
+
+Usual workflow
```
ax3l: I would even call this "Deferred Data API Contract" or so #484

docs/source/usage/workflow.rst
franzpoeschel: > (do you remember the discussion @franzpoeschel? We should have taken minutes.)

Yes, I do. The result of it was that I would document the logic in this PR, so that discussion is the reason why I added this :)

> I think the main reason was that we return the spans in those operations and then fill them before flush from the user side (for write).

Also, the `openpmd-pipe` tool already assumes that `storeChunk` works this way, even without spans.

docs/source/usage/workflow.rst
franzpoeschel: > What is meant with aliased? sweat_smile

After `defineAttribute`, the attribute values are store some place internal that a user has no access to, so a user cannot have a pointer to the value or something. "aliased" in the same meaning as C `noalias`. The openPMD API can deal with attributes whatever way it wishes to do, since it's entirely transparent to the user.

> (Does anything change for ADIOS2 where attributes become vars in the new layout? Probably not, just double-checking.)

No, since these arguments are entirely about frontend things. Attributes to be written in the new layout are buffered in `struct BufferedAttributeWrite`.

franzpoeschel:
> We discussed the comment [#901 (comment)](https://github.com/openPMD/openPMD-api/pull/901#issuecomment-799460797) and it would be good to transfer it to ADIOS2 as a motivating example for BP5 to keep track of. 

The reproducer that you linked is pure ADIOS2 code, so that's already done I think?

> One of the solutions in BP5 could be to allow similar to openPMD-api to pass buffers of individual (large) variables from the user-side to ADIOS2 to keep track of (basically `PutDeferred`).

Do I understand it correctly, this would mean that a buffer content given to ADIOS would not be copied into ADIOS buffers but used directly? If that would work, spans would not be necessary any longer. (This PR would still stay relevant because it is an important preparation for using the Inline Engine some day)
> 
> The return of a span currently cannot easily keep track of the proper span pointers if resizes occur.

It does, though? You have to be careful as a user not to use a stale pointer after a resize, but call `span.data()` again, but it works. Consequence of that is of course that openPMD spans will work much the same way.

> One could introduce additional phases to declare when all local size declaration (puts) are done, in both APIs (openPMD-api, ADIOS2), and keep track of the sizes. But maybe that is too complicated in workflows and API communication to users?

It does sound somewhat complicated. Would that workflow stay relevant once the new serializer lands? I'm hesitant to introduce major temporary workarounds (which this PR seems to partially turn out to be already).
> 
> During the meeting we had another idea today: ADIOS2 could keep a reference of its `span`s (or shared buffers) and change the data pointers (in `.data()`/`.get()`). The API contract would then be that users need to access a fresh pointer after each `Put()` in case. This makes it a bit more convenient during resizes, but does not solve the resizes are still occurring generally. (Reminder: during resizes, already written data behind a span is copied over, so there is no problem with that at least.)

According to my tests, that already happens? If I call `span.data()` after a resize, I get the correct new address. I looked into the ADIOS source and `span.data()` seems to calculate the address by adding its offset to the base pointer of the serialization vector, and since the serialization pointer keeps track of its resizes (it's a STL vector), that trickles down to the span.
The only issue in my tests was that the serialization vector is resized right after doing `engine.Put()` which is not necessary *because the user needs to call `span.data()` anew anyway*.
> 
> For BP4, it looks like the only work-around is to pre-allocate (request) once form the user-side for a sufficiently large buffer and store into that buffer via the provided spans. For BP5, \at pnorbert considers dropping span for, e.g. a `shared_ptr` interface of user-side buffers as we do (for large variables). 

Just a note that we should be careful because most writers don't actually have their data allocated in a `shared_ptr` and do some tricks like `shared_ptr<T>(ptr, [](auto const*){})`. At least I know that PIConGPU does.

> Or a serialize in ADIOS2 that is chunked in a more sophisticated way for resizes: (see also [#937 (comment)](https://github.com/openPMD/openPMD-api/issues/937#issuecomment-795641082)) See also [ornladios/ADIOS2#2649](https://github.com/ornladios/ADIOS2/issues/2649) for memory control (drain/flush) in ADIOS2. For now, we need `span`s because they are the only way to get an "allocate-once" workflow implemented.

The chunking would address the resizing, the `shared_ptr` thing would address the double allocation, right? So they address different challenges.

ax3l:
> > We discussed the comment #901 (comment) and it would be good to transfer it to ADIOS2 as a motivating example for BP5 to keep track of.
>
> The reproducer that you linked is pure ADIOS2 code, so that's already done I think?

Sorry, I meant we should document this also as motivating example issue in their repo :)


>> The return of a span currently cannot easily keep track of the proper span pointers if resizes occur.
>
> It does, though? You have to be careful as a user not to use a stale pointer after a resize, but call span.data() again, but it works. Consequence of that is of course that openPMD spans will work much the same way.

Of perfect, that's good to know that it already works (if `.data()` is called directly before usage).

franzpoeschel:
~~Should we update the [`Span` interface](https://github.com/openPMD/openPMD-api/blob/fe52a2fef91dc06ae237553fd619548a62310f9f/include/openPMD/Span.hpp#L36) in C++ to be more similar to the [one in Python](https://github.com/openPMD/openPMD-api/blob/fe52a2fef91dc06ae237553fd619548a62310f9f/src/binding/python/RecordComponent.cpp#L689)? Maybe this silent buffer reallocation thing is a bit too much to hide it automatically behind a Span interface.. ?~~
Edit: This is now a TODO

franzpoeschel:
If this runs green, then it's ready for review :) @ax3l

src/binding/python/RecordComponent.cpp
```diff
@@ -352,6 +354,219 @@ store_chunk(RecordComponent & r, py::array & a, py::tuple const & slices)
...
+pybind11::memoryview
+GetCurrentView::operator()< std::string >( PythonDynamicMemoryView const & )
+{
+    throw std::runtime_error( "[DynamicMemoryView] Only PODs allowed." );
+}
+} // namespace
+
+pybind11::memoryview PythonDynamicMemoryView::currentView() const
+{
+    static GetCurrentView cv;
```
ax3l: A static variable here looks like this could be problematic if a user writes multiple record components behind each other, e.g. defining "E/x", "E/y", "E/z" and then writing each of them. Is there a lifetime question you tried to work-around with this?

src/binding/python/Attributable.cpp
```diff
@@ -365,7 +365,7 @@ void init_Attributable(py::module &m) {
                 return "<openPMD.Attributable with '" + std::to_string(attr.numAttributes()) + "' attributes>";
             }
         )
-        .def("series_flush", &Attributable::seriesFlush)
+        .def("series_flush", (void (Attributable::*)()) &Attributable::seriesFlush)
```
ax3l: Do you try to select the non-argument overload? :)
```suggestion
        .def("series_flush",
             py::overload_cast< >(&Attributable::seriesFlush))
```
https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods

src/binding/python/RecordComponent.cpp
franzpoeschel: It's an empty struct just used for the switchType template. But thinking about this again, `static` isn't really needed here, but I can make it `const`.
Maybe we should make `switchType` accept its "functor" parameter only as a template parameter instead of an actual function parameter to avoid this? I think there is only one place where this struct is non-empty and that can be easily rewritten.
I.e. something like
```cpp
template< typename Action, typename... Args >
auto switchType( Datatype dt, Args &&... args ) -> […]
```
instead of 
```cpp
template< typename Action, typename... Args >
auto switchType( Datatype dt, Action action, Args &&... args ) -> […]
```

src/binding/python/Attributable.cpp
franzpoeschel: Huh, didn't know that one.

src/binding/python/RecordComponent.cpp
ax3l: I think the current way (with the parameter) is more general since you can always pass in an argument `dt, Action{}, ...` to get the first case.

But if we know that our switch type actions are always stateless then we can also rework it that way.

src/binding/python/Attributable.cpp
ax3l: Jau, saves some C style function cast magic :)

src/binding/python/RecordComponent.cpp
franzpoeschel: If you need the state, you can always define it as one of the `Args&&...` parameters, so having the `action` as a parameter is a duplication of some sort.

ax3l: True, agreed :)

ax3l:
@franzpoeschel can you please rebase this one? :)

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -303,6 +306,28 @@ class AbstractIOHandlerImpl
...
+   * The operation should fail if chunk positions parameters.offset+parameters.extent do not reside inside the dataset.
+   * The dataset should match the dataype parameters.dtype.
+   * The buffer should be stored as a cast-to-char pointer to a flattened version of the backend buffer in parameters.out->ptr. The chunk is stored row-major.
+   * The buffer's content should be written to storage not before the next call to AbstractIOHandler::flush where AbstractIOHandler::m_flushLevel == FlushLevel::InternalFlush.
+   * The precise time of data consumption is defined by the backend:
+   * * Data written to the returned buffer should be consumed not earlier than the next call to AbstractIOHandler::flush where AbstractIOHandler::m_flushLevel == FlushLevel::InternalFlush.
+   * * Data should be consumed not later than the next Operation::ADVANCE task where parameter.mode == AdvanceMode::ENDSTEP.
+   *
+   * This task is optional and should either (1) not be implemented by a backend at all or (2) be implemented as indicated above and set parameters.out->taskSupportedByBackend = true.
```
ax3l: I am not sure if "task" is the right naming here. Because of our `IOTasks`? I think they are a bit of an implementation detail and we give this naming here out to the user.

I think we mean "backend has explicit support for buffer creation" - maybe for the variable the name "backendManagedBuffer = true/false" is better?

docs/source/usage/workflow.rst
```diff
@@ -0,0 +1,30 @@
...
+*   In read mode, a buffer into which data from a dataset should be filled, must have been filled with the requested data after the flush point.
+
+In short: operations requested by ``storeChunk()`` and ``loadChunk()`` must happen exactly at flush points.
+
+Flush points are triggered by:
+
+*   Calling ``Series::flush()``.
+*   Calling ``Iteration::flush( flush=true )``.
+    Flush point guarantees affect only the corresponding iteration.
+*   The streaming API (i.e. ``Series.readIterations()`` and ``Series.writeIteration()``) automatically before accessing the next iteration.
```
ax3l, line 26:
```suggestion
*   Calling ``Writable::seriesFlush()`` or ``Attributable::seriesFlush()``.
*   The streaming API (i.e. ``Series.readIterations()`` and ``Series.writeIteration()``) automatically before accessing the next iteration.
```

docs/source/usage/workflow.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Attributes are (currently) unaffected by this:
+
+*   In writing, attributes are stored by value and can afterwards not be aliased by the user (i.e. they are stored internally in the openPMD API and are not accessible to users).
```
ax3l: I don't get "aliased", let's clarify :)

docs/source/usage/workflow.rst
```diff
@@ -0,0 +1,30 @@
...
+*   In reading, attributes are parsed upon opening the Series / an iteration and are available to read right-away without performing any IO.
```
ax3l: Just to avoid confusion:
```suggestion
*   In reading, attributes are parsed upon opening the Series / an iteration and are available to read right-away.
```

examples/12_span_write.cpp
```diff
@@ -0,0 +1,100 @@
+#include <openPMD/openPMD.hpp>
```
ax3l, line 1: Don't forget to link the two new examples in the docs page: https://openpmd-api.readthedocs.io/en/latest/usage/examples.html
(I also forgot to link a few new ones :sweat_smile: )

examples/12_span_write.py
```diff
@@ -0,0 +1,35 @@
+import openpmd_api as io
+import numpy as np
+
+
+def span_write(filename):
+    series = io.Series(filename, io.Access_Type.create)
+
+    datatype = np.dtype("double")
+    length = 10
+    extent = [10]
```
ax3l:
```suggestion
    extent = [length]
```

examples/12_span_write.py
```diff
@@ -0,0 +1,35 @@
...
+    dataset = io.Dataset(datatype, extent)
+
+    iterations = series.write_iterations()
+    for i in range(10):
```
ax3l: Maybe another number than `10` to avoid confusion:
```suggestion
    for i in range(12):
```

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -1041,7 +1068,9 @@ namespace detail
         std::vector< std::unique_ptr< BufferedAction > > m_buffer;
         std::map< std::string, BufferedAttributeWrite > m_attributeWrites;
         std::vector< BufferedAttributeRead > m_attributeReads;
+        std::vector< std::unique_ptr< BufferedAction > > m_alreadyEnqueued;
```
ax3l, line 1096: Should we add further doxygen strings to these member variables?
The amount of member variables indicates we are doing quite involved things here :)

include/openPMD/RecordComponent.tpp
```diff
@@ -0,0 +1,330 @@
+/* Copyright 2017-2021 Fabian Koller and Franz Poeschel
```
ax3l: Great you already added this to the install logic, so it does not break installs :+1: 
https://github.com/openPMD/openPMD-api/blob/6fd39365ff4bf6e350faec66a5ebac2f3804e8dd/CMakeLists.txt#L908-L911

ax3l: Feel free to add you & me in that `.hpp` and `.tpp` by name; I always forget to add myself when editing files :sweat_smile:

include/openPMD/Span.hpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+    inline reverse_iterator rend() const
+    {
+        return reverse_iterator{ data() };
+    }
+};
+
+/**
+ * @brief A view into a buffer that might be reallocated at some points and
+ *      have changing base pointers over time.
```
ax3l:
```suggestion
 *      thus has changing base pointers over time.
```

src/Series.cpp
```diff
@@ -475,20 +478,33 @@ SeriesImpl::initDefaults()
...
 std::future< void >
-SeriesImpl::flush_impl( iterations_iterator begin, iterations_iterator end )
+SeriesImpl::flush_impl(
+    iterations_iterator begin,
+    iterations_iterator end,
+    FlushLevel level )
 {
-    switch( iterationEncoding() )
+    IOHandler()->m_flushLevel = level;
+    try
```
ax3l, line 487: Here and below: why do we use exception handling at this point and rethrow another exception?
I am just concerned that we absorb legit exceptions from backends, e.g. if a low-level ADIOS/HDF5 operation fails.

The try-catch pattern on this high level looks a bit duck-tape-y to me.

ax3l (commented):
Looks already in great shape! Notes inline :)

src/Series.cpp
franzpoeschel: The purpose of this is to emulate the same thing that the Python or Java `finally` keywords do. This does *not* absorb exceptions from a layer below, this construct does some cleanup and passes the exception on:
> Rethrows the currently handled exception. Abandons the execution of the current catch block and passes control to the next matching exception handler (but not to another catch clause after the same try block: its compound-statement is considered to have been 'exited'), reusing the existing exception object: no new objects are made.

(https://en.cppreference.com/w/cpp/language/throw)

The purpose is to ensure that the base state is restored *even if an exception is thrown*.

docs/source/usage/workflow.rst
franzpoeschel: Alright, will do

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: Good point, will do

include/openPMD/IO/AbstractIOHandlerImpl.hpp
franzpoeschel: > I think they are a bit of an implementation detail and we give this naming here out to the user.

This documentation is for the methods of `AbstractIOHandlerImpl` – a class that users do not deal with. This file is not even included anywhere in our frontend. So I think, documenting in terms of implementation details should be fine here?

> I think we mean "backend has explicit support for buffer creation" - maybe for the variable the name "backendManagedBuffer = true/false" is better?

I can rename the variable, this name sounds good

include/openPMD/RecordComponent.tpp
franzpoeschel: Will do
> I always forget to add myself when editing files

same

include/openPMD/IO/AbstractIOHandlerImpl.hpp
ax3l: Yes, my comment on first thought was more about the user-facing variable name.

Can we add IO before "task" to make the relation clearer? :)
```suggestion
   * This IOTask is optional and should either (1) not be implemented by a backend at all or (2) be implemented as indicated above and set parameters.out->backendManagedBuffer = true.
```

examples/12_span_write.cpp
ax3l: Added, thank you! :)

src/Series.cpp
ax3l: Perfect, thanks!

ax3l (approved):
Thank you for the great PR, this is ready to go! :rocket: :sparkles:

PR #903, opened 2021-01-15
Activate streaming examples in CI

franzpoeschel:
So far, the streaming examples are commented out because our CI couldn't deal with that. Writers and Readers need to be executed asynchronously. If executing them sequentially, like our CI does, one of them will wait for the other to show up indefinitely.

Activate them.

TODO
- [x] Fix things in the CI so things don't hang.
- [x] CI runs fine on Windows, but ADIOS2 SST engine doesn't. Changing extension to `bp` makes the examples work.
    EDIT: SST [unsupported on MSVC](https://github.com/ornladios/ADIOS2/blob/5948ca8a85e05eaf2ff07c6b64fa049fe0c4f9bb/cmake/DetectOptions.cmake#L320). So, deactivate Streaming examples on Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1034,7 +1034,10 @@ if(openPMD_BUILD_TESTING)
         if(openPMD_HAVE_HDF5)
             if(EXAMPLE_DATA_FOUND)
                 foreach(examplename ${openPMD_EXAMPLE_NAMES})
-                    if(${examplename} MATCHES "^.*_parallel$")
+                    message(STATUS "Example ${examplename}")
+                    if(${examplename} MATCHES "^10.*$")
+                        # streaming examples are done separately
+                    elseif(${examplename} MATCHES "^.*_parallel$")
                         if(openPMD_HAVE_MPI)
                             add_test(NAME MPI.${examplename}
                                     COMMAND ${MPI_TEST_EXE} ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/${examplename}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1053,6 +1053,11 @@ if(openPMD_BUILD_TESTING)
                 endforeach()
             endif()
         endif()
+        if(openPMD_HAVE_ADIOS2)
+            add_test(NAME Asynchronous.10_streaming
+                     COMMAND sh -c "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/10_streaming_write & ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/10_streaming_read"
+                     WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY})
+        endif()
     endif()
 
     # Command Line Tools
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1034,7 +1034,6 @@ if(openPMD_BUILD_TESTING)
         if(openPMD_HAVE_HDF5)
             if(EXAMPLE_DATA_FOUND)
                 foreach(examplename ${openPMD_EXAMPLE_NAMES})
-                    message(STATUS "Example ${examplename}")
                     if(${examplename} MATCHES "^10.*$")
                         # streaming examples are done separately
                     elseif(${examplename} MATCHES "^.*_parallel$")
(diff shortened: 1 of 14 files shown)


ax3l:
@franzpoeschel can you please add, after a double white-space at the end-of-the-line, a `# lgtm [py/unreachable-statement]` to silence the Linux LGTM warnings?

examples/10_streaming_write.cpp
```diff
@@ -1,4 +1,9 @@
-#define BUILD_STREAMING_EXAMPLE false
+#ifdef _WIN32
```
ax3l: Is this a general windows issue or a problem with MSVC?
(Note: we also ship and often prefer Clang on Windows)

CMakeLists.txt
```diff
@@ -1050,6 +1052,11 @@ if(openPMD_BUILD_TESTING)
                 endforeach()
             endif()
         endif()
+        if(openPMD_HAVE_ADIOS2)
```
ax3l, line 1055: If the issue is only in execution, you could skip the `add_test` here for Windows or try to search for `sh` and skip it conditionally if the command is not found:
https://github.com/ECP-WarpX/WarpX/blob/21.01/cmake/WarpXFunctions.cmake#L1-L12

examples/10_streaming_write.cpp
franzpoeschel: [Just with MSVC](https://github.com/ornladios/ADIOS2/blob/5948ca8a85e05eaf2ff07c6b64fa049fe0c4f9bb/cmake/DetectOptions.cmake#L320). Should I just drop the check and do a better check in the ADIOS2 backend?

CMakeLists.txt
franzpoeschel: `sh` on Windows is no issue actually. I didn't expect it, but it seems to work.

examples/10_streaming_write.cpp
ax3l: Yes, that would be better I think.
Let's also add a link to that place as a comment.

franzpoeschel: done

examples/10_streaming_write.cpp
```diff
@@ -49,7 +52,4 @@ main()
               << std::endl;
     return 0;
 #endif
-}
-#else
-int main(){ return 0; }
-#endif
+}
```
ax3l, line 57:
```suggestion
}

```

src/binding/python/Series.cpp
```diff
@@ -169,6 +169,7 @@ void init_Series(py::module &m) {
         .def("flush", &Series::flush)
 
         .def_property_readonly("backend", &Series::backend)
+        .def("backend_property", &Series::backendProperty)
```
ax3l: Just wondering if we need a new `series` property. We already have `openPMD::getFileExtensions()`/`openPMD_api.file_extensions` where we still need to add `.sst`, I think.

Do you have concrete additional plans to use this new property for?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2042,8 +2042,25 @@ ADIOS2IOHandler::ADIOS2IOHandler(
...
 }
 
-std::future< void >
-ADIOS2IOHandler::flush()
+std::string
+ADIOS2IOHandler::backendProperty( std::string const &property ) const
+{
+    if( property == "sst" )
+    {
+#if defined( __GNUG__ ) || defined( __clang__ )
```
ax3l: This looks a bit empiric - is there an ADIOS2 property/define that we could query for SST capability?

franzpoeschel: I have found no such thing unfortunately. Can ask this tomorrow in the VC though.

src/binding/python/Series.cpp
franzpoeschel: Didn't think of `getFileExtensions()`, I can use that one too.

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l: Sounds good! Yes, one challenge is also that SST support in ADIOS2 depends on the binary variant we build for ADIOS2. So you can get a fully valid ADIOS2 install in some cases that has SST disabled. We might need a runtime (or at least compile-time) query for that from ADIOS2.

ax3l (approved):
Looking good, thank you! :sparkles:

PR #904, opened 2021-01-15
Add openpmd-pipe.py command line tool

franzpoeschel:
This adds a little Python-based command line tool that can pipe data from a source to a sink.
Also adds handling of attributes by explicit datatype to the Python API to make generic attribute writing safe.

Possible uses:
* Converting openPMD-based data from one backend to another one.
* Multiplexing in staging setups

### TODO
- [x] Documentation
- [x] Cleanup
- [x] Testing: Resulting files must again be convertible using `openpmd-pipe`
- [x] Integration into the build system
- [x] Switch to a serial mode if MPI is not configured
- [x] Particle Patches
- [x] rebase after #912 was merged

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -663,6 +663,9 @@ set(openPMD_TEST_NAMES
 set(openPMD_CLI_TOOL_NAMES
     ls
 )
+set(openPMD_PYTHON_CLI_TOOL_NAMES
+    pipe
+)
 set(openPMD_PYTHON_CLI_MODULE_NAMES ${openPMD_CLI_TOOL_NAMES})
 # examples
 set(openPMD_EXAMPLE_NAMES
@@ -881,6 +884,16 @@ if(openPMD_INSTALL)
         install(TARGETS openPMD.py
             DESTINATION ${CMAKE_INSTALL_PYTHONDIR}/openpmd_api
         )
+        if(openPMD_BUILD_CLI_TOOLS)
+            foreach(toolname ${openPMD_PYTHON_CLI_TOOL_NAMES})
+                install(
+                    FILES ${openPMD_SOURCE_DIR}/src/cli/${toolname}.py
+                    DESTINATION ${CMAKE_INSTALL_BINDIR}
+                    RENAME openpmd-${toolname}.py
+                    PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ
+                )
+            endforeach()
+        endif()
     endif()
     install(DIRECTORY "${openPMD_SOURCE_DIR}/include/openPMD"
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
@@ -1079,6 +1092,21 @@ if(openPMD_BUILD_TESTING)
         endif()
     endif()
 
+    function(test_set_pythonpath test_name)
+        if(WIN32)
+            set_property(TEST ${test_name}
+                PROPERTY ENVIRONMENT
+                    "PATH=${WIN_BUILD_BINDIR}\\${CMAKE_BUILD_TYPE}\;${WIN_PATH}\n"
+                    "PYTHONPATH=${WIN_BUILD_BASEDIR}\\${CMAKE_INSTALL_PYTHONDIR}\\${CMAKE_BUILD_TYPE}\;${WIN_PYTHONPATH}"
+            )
+        else()
+            set_tests_properties(${test_name}
+                PROPERTIES ENVIRONMENT
+                    "PYTHONPATH=${openPMD_BINARY_DIR}/${CMAKE_INSTALL_PYTHONDIR}:$ENV{PYTHONPATH}"
+            )
+        endif()
+    endfunction()
+
     # Python CLI Modules
     if(openPMD_HAVE_PYTHON)
         # (Note that during setuptools install, these are furthermore installed as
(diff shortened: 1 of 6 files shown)


src/cli/pipe.py
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env python3
+"""
+This file is part of the openPMD-api.
+
+This module provides functions that are wrapped into sys.exit(...()) calls by
+the setuptools (setup.py) "entry_points" -> "console_scripts" generator.
+
+Copyright 2020 openPMD contributors
```
ax3l:
```suggestion
Copyright 2021 openPMD contributors
```

ax3l:
Looks good!
Do you potentially want to implement this as a python module so people can use it with `python -m openpmd.pipe ...` and then add a console_script hook on top? See `openpmd-ls` for an example.

The reason behind this is that not all distributions add paths like `~/.local/bin` to `PATH` and thus some `pip` users will not be able to use it as a CLI tool but need to go via the module call.

franzpoeschel:
> Looks good!
> Do you potentially want to implement this as a python module so people can use it with `python -m openpmd.pipe ...` and then add a console_script hook on top? See `openpmd-ls` for an example.
> 
> The reason behind this is that not all distributions add paths like `~/.local/bin` to `PATH` and thus some `pip` users will not be able to use it as a CLI tool but need to go via the module call.

Can you check whether I've done this somewhat correctly?

franzpoeschel:
I've rebased this on to #912 to remove the GC workaround. Otherwise, this is ready for review from my side :)

ax3l:
Now that #912 is merged, this PR needs a little rebase against `dev` :)
(Don't forget to remove the `[WIP]` to signal it's ready for review & merge :) )

src/binding/python/Attributable.cpp
```diff
@@ -289,6 +317,18 @@ void init_Attributable(py::module &m) {
                 a
             );
         })
+        .def("set_attribute", [](
+                Attributable & attr,
+                std::string const& key,
+                py::object& obj,
+                std::string const& datatype )
```
ax3l: I usually try to use the `np.dtype` for all datatype options in the Python bindings

src/binding/python/Attributable.cpp
```diff
@@ -335,6 +375,10 @@ void init_Attributable(py::module &m) {
             return v.getResource();
             // TODO instead of returning lists, return all arrays (ndim > 0) as numpy arrays?
         })
+        .def("get_attribute_type", []( Attributable & attr, std::string const& key ) {
```
ax3l: Ideally we would also return the `numpy.dtype` here for consistency.
```suggestion
        .def("get_attribute_dtype", []( Attributable & attr, std::string const& key ) {
```

ax3l: oh, and we can make this a (read-only) `property` such as `attribute_dtype`

franzpoeschel: Wouldn't a property only work for something without parameters?

Recent commit uses numpy types. There are some workarounds involved, but it's working as long as noone gets the idea to define an attribute of type `array<double,7>` not named `unitDimension`.

I also took the freedom to add a version of `switchType` that only switches over non-vector types, I was annoyed of always having to define a hundred of extra structs. Also, I made two improvements to that version: (1) not having to give the return type any more and (2) making the `action.template operator()<0>(...)` (resp. 1000) optional.

ax3l: Oh, right - didn't see that here.

Probably a cleaner interface on the long term, as in h5py, would be `<an attributable>.attrs` as a key-value store and then we can ask the `.dtype` of the numpy scalars/arrays inside.

Switchtype updates are great, I have used this way too little in the Python bindings yet.

include/openPMD/Datatype.hpp
```diff
@@ -772,6 +771,182 @@ ReturnType switchType( Datatype dt, Action action, Args &&... args )
...
+                std::forward< Args >( args )... );
+    case Datatype::STRING:
+        return action.OPENPMD_TEMPLATE_OPERATOR()< std::string >(
+            std::forward< Args >( args )... );
+    case Datatype::BOOL:
+        return action.OPENPMD_TEMPLATE_OPERATOR()< bool >(
+            std::forward< Args >( args )... );
+    case Datatype::DATATYPE:
+        return detail::CallUndefinedDatatype<
+            1000, ReturnType, Action, void, Args &&... >::
```
ax3l: Can we express that magic number in other way?
(Last entry of the data type enum + 1, or a constexpr with a docstring, or so?)

Maybe the same for the `0` below, just to make things more clear by naming them (first/last/lowest/highest-something).

include/openPMD/Datatype.hpp
```diff
@@ -772,6 +771,182 @@ ReturnType switchType( Datatype dt, Action action, Args &&... args )
...
+/**
+ * Purpose of this struct is to detect at compile time whether
+ * Action::template operator()\<0\>() exists. If yes, call
+ * Action::template operator()\<n\>() with the passed arguments.
+ * If not, throw an error.
+ *
+ * @tparam n As in switchType().
+ * @tparam ReturnType As in switchType().
+ * @tparam Action As in switchType().
+ * @tparam Dummy For SFINAE, set to void.
```
ax3l: For [inclusive wording](https://developers.google.com/style/inclusive-documentation) we could replace _dummy_ with _placeholder_.

CMakeLists.txt
```diff
@@ -881,6 +884,16 @@ if(openPMD_INSTALL)
         install(TARGETS openPMD.py
             DESTINATION ${CMAKE_INSTALL_PYTHONDIR}/openpmd_api
         )
+        if(openPMD_BUILD_CLI_TOOLS)
+            foreach(toolname ${openPMD_PYTHON_CLI_TOOL_NAMES})
+                install(
+                    FILES ${openPMD_SOURCE_DIR}/src/cli/${toolname}.py
+                    DESTINATION ${CMAKE_INSTALL_BINDIR}
+                    RENAME openpmd-${toolname}.py
```
ax3l: I would call the tool just `openpmd-pipe` because for the user it will make no difference if the tool is a script or an executable:
```suggestion
                    RENAME openpmd-${toolname}
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+        # all tools must provide a "--help"
+        # The custom target ensures that the file is copied to the build dir
+        # anew if the source file changes (and not only if openPMD.py is rebuilt)
+        add_custom_target(openPMD.py.cliTools)
+        add_dependencies(openPMD.py openPMD.py.cliTools)
+        foreach(toolname ${openPMD_PYTHON_CLI_TOOL_NAMES})
+            add_custom_command(TARGET openPMD.py.cliTools POST_BUILD
+                COMMAND ${CMAKE_COMMAND} -E copy
+                        ${openPMD_SOURCE_DIR}/src/cli/${toolname}.py
+                        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-${toolname}.py
```
```suggestion
                        ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-${toolname}
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+            )
+            add_test(NAME CLI.help.${toolname}.py
+                COMMAND ${Python_EXECUTABLE}
+                    ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-${toolname}.py --help
```
```suggestion
                    ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-${toolname} --help
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+            if( openPMD_HAVE_MPI )
+                set(MPI_TEST_EXE
+                    ${MPIEXEC_EXECUTABLE}
+                    ${MPI_ALLOW_ROOT}
+                    #${MPIEXEC_NUMPROC_FLAG} 2
+                )
+                add_test(NAME CLI.pipe.py
+                    COMMAND sh -c
+                        "${MPI_TEST_EXE} ${Python_EXECUTABLE}                      \
+                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe.py      \
```
```suggestion
                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe         \
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+                            --infile ../samples/git-sample/thetaMode/data%T.h5     \
+                            --outfile ../samples/git-sample/thetaMode/data%T.bp;   \
+                                                                                   \
+                            ${Python_EXECUTABLE}                                   \
+                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe.py      \
```
```suggestion
                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe         \
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+                    WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
                 )
             else()
-                set_tests_properties(CLI.py.help.${pymodulename}
-                    PROPERTIES ENVIRONMENT
-                        "PYTHONPATH=${openPMD_BINARY_DIR}/${CMAKE_INSTALL_PYTHONDIR}:$ENV{PYTHONPATH}"
+                add_test(NAME CLI.pipe.py
+                    COMMAND sh -c
+                        "${Python_EXECUTABLE}                                      \
+                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe.py      \
```
```suggestion
                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe         \
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+                            --infile ../samples/git-sample/thetaMode/data%T.h5     \
+                            --outfile ../samples/git-sample/thetaMode/data%T.bp;   \
+                                                                                   \
+                            ${Python_EXECUTABLE}                                   \
+                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe.py      \
```
```suggestion
                            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe         \
```

CMakeLists.txt
```diff
@@ -1088,19 +1116,74 @@ if(openPMD_BUILD_TESTING)
...
+            )
+            add_test(NAME CLI.help.${toolname}.py
+                COMMAND ${Python_EXECUTABLE}
+                    ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/openpmd-${toolname}.py --help
+                WORKING_DIRECTORY ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+            )
+            test_set_pythonpath(CLI.help.${toolname}.py)
+        endforeach()
+
+        # openpmd-pipe.py test
```
```suggestion
        # openpmd-pipe (python) test
```

ax3l (dismissed):
.

ax3l (commented):
Great SFINAE-magic :D Hope it's not too wild on the compile time.

Overall this looks great :sparkles: 
I have some minor suggestions inline.

franzpoeschel:
> Hope it's not too wild on the compile time.

I'm planning a second PR after this one to apply those changes to the old `switchType` too. Since those functions don't really need to be in public headers, I would take the chance to move them to a separate header, that should have that issue solved.

include/openPMD/Datatype.hpp
franzpoeschel: I did not yet apply that to the original switchType yet since I'm planning a separate PR for that one.

src/binding/python/Attributable.cpp
franzpoeschel: That would be a breaking interface change though, deprecating our current `Attributable.attributes` call, right?
For now, I've turned this into a readonly property `Attributable.attribute_dtypes` (returning a dictionary: name -> numpy type).

ax3l (approved):
:rocket: :sparkles:

PR #905, opened 2021-01-15
CMake 3.15.0: MSVC Runtime Lib (Static for Py)

ax3l:
Make sure the MSVC runtime libs are static for pybind modules.
https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ class CMakeBuild(build_ext):
             '-DopenPMD_BUILD_SHARED_LIBS:BOOL=' + BUILD_SHARED_LIBS,
             '-DHDF5_USE_STATIC_LIBRARIES:BOOL=' + HDF5_USE_STATIC_LIBRARIES,
             '-DADIOS_USE_STATIC_LIBS:BOOL=' + ADIOS_USE_STATIC_LIBS,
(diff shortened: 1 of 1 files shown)


ax3l:
@robertmaynard would this also be the right thing to add for Python modules?

robertmaynard:
Hmmm 

We need to wait on merging this. Something is cousin has the wrong runtime flags to be set.

ax3l:
> Something is cousin has the wrong runtime flags to be set.

@robertmaynard sorry, the autocorrect on cousin confuses me. Can you rephrase please? :)

robertmaynard:
> > Something is cousin has the wrong runtime flags to be set.
> 
> @robertmaynard sorry, the autocorrect on cousin confuses me. Can you rephrase please? :)

Should have read:

We need to wait on merging this. Something is causing the wrong runtime flags to be set.


Anyway, this response was meant for the AMReX merge request, as I had originally placed the policy update  at the wrong place ( it needs to go before the `project` call) .

ax3l:
I think we don't need this flag here, works well without.

PR #906, opened 2021-01-20
support null core instead of null

guj:
The NullCore engine is recommended in ADIOS, instead of Null Engine.


franzpoeschel:
We use the nullcore engine since it deals better with steps than the null engine. I've added a test that uses steps.

franzpoeschel (approved):

Issue #907, opened 2021-01-22
Reading `particlePatches["numParticles"]` fails


psychocoderHPC:
**Describe the bug**
I am testing https://github.com/openPMD/openPMD-api/tree/9e201df4e20f3cf41b75df55a19ae9a33f57cdae together with PIConGPU und realized an error when loading an old checkpoint.
After investing some hours in debugging the issue I tested the release `0.12.0-alpha` and all works fine.

The error is:
```
Unhandled exception of type 'St13runtime_error' with message 'Chunk does not reside inside dataset (Dimension on index 0. DS: 262144 - Chunk: 28299152)', terminating
```

**To Reproduce**
You can use this dataset [checkpoint_blosc_000100.bp.zip](https://github.com/openPMD/openPMD-api/files/5857933/checkpoint_blosc_000100.bp.zip) (the dataset is not compressed, please ignore that it is named blosc)
The dataset was produced with a single MPI rank.

The data is read with the C++ API: https://github.com/ComputationalRadiationPhysics/picongpu/blob/b17941d2b90afb83b17284cae1b70e62d33e088e/include/picongpu/plugins/openPMD/restart/LoadSpecies.hpp#L108-L110

ADIOS2 internal tools shows that the data in the file is correct:
```
bpls checkpoint_blosc_000100.bp --dump -a /data/100/particles/e/particlePatches/numParticles
  uint64_t  /data/100/particles/e/particlePatches/numParticles               {1}
    (0)    262144 
```

**Expected behavior**

**Software Environment**
 - version of openPMD-api: https://github.com/openPMD/openPMD-api/tree/9e201df4e20f3cf41b75df55a19ae9a33f57cdae (current openpmd-api dev branch)
 - installed openPMD-api via: self compiled
 - operating system: Red Hat Enterprise Linux release 8.2 (Ootpa)
 - machine: closed system
 - name and version of Python implementation: [e.g. CPython 3.8]
 - version of ADIOS2: https://github.com/ornladios/ADIOS2/tree/dd822e30103f04418637688bd5cda362be28b0c4
 - name and version of MPI: [e.g. OpenMPI 4.1.0]

**Additional context**
As I wrote above after switching from the current dev branch to `0.12.0-alpha` all works fine.
For unknown reason openpmd-api is reading `28299152` from the dataset instead of `262144` what is stored in the adios file.

psychocoderHPC:
[...]

PR #909, opened 2021-01-25
Check whether particlePatches are dirty

franzpoeschel:
Fix for #907. `::dirtyRecursive()` ignored particle patches, leading to skipped flushes.
We should do a minor release soon including this one since this makes restarting from openPMD impossible in PIConGPU.

TODO:
- [x] Add a test for this

### DIFF
--- a/include/openPMD/backend/PatchRecordComponent.hpp
+++ b/include/openPMD/backend/PatchRecordComponent.hpp
@@ -49,6 +49,7 @@ class PatchRecordComponent : public BaseRecordComponent
     friend
     class Container;
 
+    template< typename > friend class BaseRecord;
     friend class ParticlePatches;
     friend class PatchRecord;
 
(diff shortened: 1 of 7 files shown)


include/openPMD/backend/PatchRecordComponent.hpp
```diff
@@ -74,6 +75,18 @@ class PatchRecordComponent : public BaseRecordComponent
     void read();
 
     std::shared_ptr< std::queue< IOTask > > m_chunks;
+    std::shared_ptr< bool > hasBeenRead = std::make_shared< bool >( false );
+
+    /**
+     * @brief Check recursively whether this RecordComponent is dirty.
+     *        It is dirty if any attribute or dataset is read from or written to
+     *        the backend.
```
ax3l, line 83: Can we add one sentence here repeating what "dirty" means? :)

ax3l (commented):
Looks great! Also great doxygen string.

Just waiting for the test before merge.

psychocoderHPC:
I tested this PR together with PIConGPU and the crash during the restart is fixed.

ax3l (approved):

ax3l:
Wonderful, goes in! Yes, let's cut a `0.13.2` release with this fix very soon.

PR #911, opened 2021-01-26
[WIP] ADIOS: 2.7.0+

ax3l:
Increase support for ADIOS2 to require v2.7.0+ only.
This allows us to avoid some work-arounds and use new functionality unconditionally.

Note: this PR will stay open for a week or two before we will apply it.


franzpoeschel:
Is this a duplicate of #927?

ax3l:
Oh wow, yes

PR #912, opened 2021-01-26
Python: store_chunk(temporary)

### LINKED ISSUE #833: Python garbage collection: store_chunk does not keep data alive
Garbage collection in the Python API will not keep alive datasets inserted into `Record_Component.store_chunk()`.

**To Reproduce**
```python
import numpy as np
import openpmd_api as io

dtype = np.dtype('float')
extent = 1000

def write_good():
    series = io.Series("sample_good.bp", io.Access_Type.create)
    E_x = series.iterations[0].meshes["E"]["x"]
    E_x.reset_dataset(io.Dataset(dtype, [extent]))
    data = np.zeros(extent, dtype=dtype)     # variable `data` will keep the dataset alive
    E_x.store_chunk(data, [0], [extent])
    series.flush()

def write_bad():
    series = io.Series("sample_bad.bp", io.Access_Type.create)
    E_x = series.iterations[0].meshes["E"]["x"]
    E_x.reset_dataset(io.Dataset(dtype, [extent]))
    E_x.store_chunk(np.zeros(extent, dtype=dtype), [0], [extent])     # no variable to keep the dataset alive
    series.flush()

write_bad()
write_good()

```
Results:
```
> bpls sample_bad.bp/ -d | head -n 10
  double   /data/0/meshes/E/x  {1000}
    (  0)    0 4.65427e-310 4.65427e-310 4.65427e-310 0 8.74496e-322
    (  6)    6.91176e-310 4.65427e-310 0 0 4.65427e-310 4.94066e-323
    ( 12)    5.78208e+160 1.32113e-319 7.90505e-323 0 4.65427e-310 3.45846e-323
    ( 18)    8.74753e-308 0 0 0 0 0
    ( 24)    7.90505e-323 0 0 7.16395e-322 1.1363e-313 4.65427e-310
    ( 30)    4.94066e-323 5.78208e+160 1.32113e-319 7.90505e-323 4.65427e-310 3.45846e-323
    ( 36)    8.74753e-308 0 0 0 0 0
    ( 42)    7.90505e-323 0 0 2.42092e-322 4.65427e-310 4.65427e-310
    ( 48)    4.65427e-310 4.65427e-310 0 2.42092e-322 4.65427e-310 4.65427e-310

> bpls sample_good.bp/ -d | head -n 10
  double   /data/0/meshes/E/x  {1000}
    (  0)    0 0 0 0 0 0
    (  6)    0 0 0 0 0 0
    ( 12)    0 0 0 0 0 0
    ( 18)    0 0 0 0 0 0
    ( 24)    0 0 0 0 0 0
    ( 30)    0 0 0 0 0 0
    ( 36)    0 0 0 0 0 0
    ( 42)    0 0 0 0 0 0
    ( 48)    0 0 0 0 0 0
```
Garbage collection bugs being as unpredictable as they are, this bug may not be always triggered. For me, switching out the two function calls resulted in a usable dataset.

**Expected behavior**
Garbage collection should not claim a chunk as long as it is still needed in the backend.

**Software Environment**
 - version of openPMD-api: current dev, commit `8f153f487352b56837c69cc93048d31e14ab46b4`.
 - installed openPMD-api via: manually
 - operating system: Ubuntu 20.10
 - machine: own laptop
 - name and version of Python implementation: Conda-installed Python 3.8.3
 - version of ADIOS2: 2.6.0

**Additional information**
The reason for this error is relatively easily recognized by the use of `share_raw` in [here](https://github.com/openPMD/openPMD-api/blob/574269a39c75fa6844450876b3de957d457b0c73/src/binding/python/RecordComponent.cpp#L288). I'm not quite sure how to actually solve this, though.

The `.load_chunk` method shares the same issue from looking at the implementation, but it's probably less critical here since users will generally keep around the variable for later access anyway. But if we find a solution for `.store_chunk`, it's likely to be applicable for `.load_chunk` as well anyway.

ax3l:
Fix #833

- [x] Test: Store a temporary array that is garbage collected after the call. (note: does not yet trigger a fail...)
- [x] fix `store_chunk`

Note: there is no in-memory overload for `load_chunk` yet exposed in Python.

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -282,30 +282,17 @@ store_chunk(RecordComponent & r, py::array & a, Offset const & offset, Extent co
     // @todo in order to implement stride handling, one needs to
     //       loop over the input data strides during write below
 
-    // check: did the user pass a temporary buffer? In that case, we copy as we
-    //        do in the shared_ptr API for C++
-    // note: there might be a corner-case if the user passes a temporary via
-    //       additional wrapper function calls here. We can cover this by
-    //       increasing the Py_SET_REFCNT of a ourselves and reducing it after
-    //       flush(), but that's too complicated/exotic to handle for now.
-    // yeah, every call adds a +1
-    // so here it's 3 (temporary) or >=4 (regular external var)
-    const bool copy = a.ref_count() <= 3 ? true : false;
-
-    void* data = a.mutable_data();
-    auto store_data = [ &r, &data, &offset, &extent, &copy ]( auto cxxtype ){
+    // here, we increase a reference on the user-passed data so that
+    // temporary and lost-scope variables stay alive until we flush
+    // note: this does not yet prevent the user, as in C++, to build
+    //       a race condition by manipulating the data they passed
+    auto store_data = [ &r, &a, &offset, &extent ]( auto cxxtype ) {
         using CXXType = decltype(cxxtype);
-        if( copy ) {
-            size_t num_elements = 1u;
-            for( auto const& si : extent )
-                num_elements *= si;
-            std::shared_ptr< CXXType > data_copy( new CXXType[num_elements],
-                                                  [](CXXType const *p) { delete[] p; } );
-            std::memcpy( data_copy.get(), data, num_elements * sizeof(CXXType) );
-            r.storeChunk( data_copy, offset, extent );
-        }
-        else
-            r.storeChunk( shareRaw( (CXXType*)data ), offset, extent );
+        a.inc_ref();
+        void* data = a.mutable_data();
+        std::shared_ptr< CXXType > shared( ( CXXType * )data,
+                                           [ a ]( CXXType * ) { a.dec_ref(); } );
+        r.storeChunk( std::move( shared ), offset, extent );
     };
 
     // store
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -27,13 +27,14 @@
 #include "openPMD/auxiliary/ShareRaw.hpp"
 #include "openPMD/binding/python/Numpy.hpp"
 
-#include <complex>
-#include <string>
 #include <algorithm>
-#include <tuple>
-#include <iostream>
+#include <complex>
 #include <cstdint>
+#include <cstring>
 #include <exception>
+#include <iostream>
+#include <string>
+#include <tuple>
 
 namespace py = pybind11;
 using namespace openPMD;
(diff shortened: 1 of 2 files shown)


ax3l:
Hm @franzpoeschel I have a hard time building a reproducer for tests, although I think I do something as in the example you posted... Probably needs more detail here to reproduce first so we can be sure it is fixed.

src/binding/python/RecordComponent.cpp
```diff
@@ -279,44 +280,55 @@ store_chunk(RecordComponent & r, py::array & a, Offset const & offset, Extent co
         throw py::index_error(
             "strides in chunk are inefficient, not implemented!");
     // @todo in order to implement stride handling, one needs to
-    //       loop over the input data strides during write below,
-    //       also data might not be owned!
+    //       loop over the input data strides during write below
+
+    // check: did the user pass a temporary buffer? In that case, we copy as we
+    //        do in the shared_ptr API for C++
+    // note: there is a corner-case if the user passes a temporary via additional
```
ax3l: is->might be ... not sure if that would not decrease the refcount again once passed as an arg

src/binding/python/RecordComponent.cpp
```diff
@@ -279,44 +280,55 @@ store_chunk(RecordComponent & r, py::array & a, Offset const & offset, Extent co
...
+    //       wrapper function calls here. We can cover this by increasing the
+    //       Py_SET_REFCNT of a ourselves and reducing it after flush(), but that's
+    //       too complicated/exotic to handle for now.
+    // yeah, every call adds a +1
+    // so here it's 3 (temporary) or >=4 (regular external var)
+    const bool copy = a.ref_count() <= 3 ? true : false;
```
ax3l: even setting this to
```suggestion
    const bool copy = false; // a.ref_count() <= 3 ? true : false;
```
does not trigger the bug for me... something in the Python test design does not work yet as we think...

test/python/unittest/API/APITest.py
```diff
@@ -1598,10 +1599,10 @@ def makeAvailableChunksRoundTrip(self, ext):
         data = np.array(
             [[2, 4, 6, 8], [10, 12, 14, 16]], dtype=np.dtype("int"))
         E_x.store_chunk(data, [1, 0], [2, 4])
-        data = np.array([[2, 4], [6, 8], [10, 12]], dtype=np.dtype("int"))
-        E_x.store_chunk(data, [4, 2], [3, 2])
-        data = np.array([[2], [4], [6], [8]], dtype=np.dtype("int"))
-        E_x.store_chunk(data, [6, 0], [4, 1])
+        data2 = np.array([[2, 4], [6, 8], [10, 12]], dtype=np.dtype("int"))
+        E_x.store_chunk(data2, [4, 2], [3, 2])
+        data3 = np.array([[2], [4], [6], [8]])
```
ax3l: Ups (but unrelated)
```suggestion
        data3 = np.array([[2], [4], [6], [8]], dtype=np.dtype("int"))
```

franzpoeschel:
> Hm @franzpoeschel I have a hard time building a reproducer for tests, although I think I do something as in the example you posted... Probably needs more detail here to reproduce first so we can be sure it is fixed.

The first time that I noticed this issue was during the implementation of #904. You can try removing [this line](https://github.com/openPMD/openPMD-api/blob/b6afe55a7de59bc14006d4a62003da054e7bed1c/src/binding/python/openpmd_api/pipe/__init__.py#L214) to trigger the issue again.

ax3l:
Oh it fails on CI now, good enough :D (although I have the fix applied, lol)

franzpoeschel:
Hmm.. Doesn't this approach only push the issue to more extreme corner cases?
Let's assume, I change the Python test to:
```python
    def writeFromTemporaryStore(self, E_x):
        semi_temporary_array = \
            np.array([[1, 2, 3], [4, 5, 6]],
                     dtype=np.dtype("int"))
        E_x.store_chunk(semi_temporary_array)
```
Then, `E_x.store_chunk` should think that this is a non-temp array and simply share it. But it only lives until the end of the function, which does not include the flush point.

Might something like this work?
```cpp
    void *data = a.mutable_data();
    auto store_data =
        [ &r, &data, &offset, &extent, &copy, &a ]( auto cxxtype ) {
            using CXXType = decltype( cxxtype );
            a.inc_ref();
            std::shared_ptr< CXXType > shared_ptr{
                ( CXXType * )data, [ a ]( CXXType * ) { a.dec_ref(); } };
            r.storeChunk( std::move( shared_ptr ), offset, extent );
        };
```
Since `pybind11::array` derives from `pybind11::handle`, capturing this by copy into the inner lambda should not copy the actual array, right? So, we could just let the `shared_ptr` have one reference to the array, and as soon as C++ is done with the array, it can let Python know about this in `~shared_ptr()`.

ax3l:
I was wondering as well in the comments and was just afraid to change the ref count, but let's give it a try!

Will push an update.

franzpoeschel:
Alternatively, [this](https://stackoverflow.com/questions/64654945/increase-ref-count-of-pybind11-handle-in-c) looks like pybind11 might actually be able to handle shared pointers automatically? But yeah, let's try this one first.

ax3l:
Nice, works! I'll check that CI crashes now on the test and then push the fix again :)

test/python/unittest/API/APITest.py
```diff
@@ -1629,6 +1630,47 @@ def testAvailableChunks(self):
...
+
+        del write
+
+        read = io.Series(
+            name,
+            io.Access_Type.read_only
+        )
+
+        r_E_x = read.iterations[0].meshes["E"]["x"]
+        # self.assertEqual(len(r_E_x.available_chunks()), 2) # oddly, "1"
```
ax3l: @franzpoeschel this is interestingly raising `1` instead of `2`.

franzpoeschel: This is because the results from `.available_chunks()` are backend-dependent. The JSON backend only sees this dataset:
```json
            "data": [
              [
                1,
                2,
                3
              ],
              [
                4,
                5,
                6
              ]
            ],
```
… and reports one chunk going from [0,0] to [2,3].

ax3l: Makes sense! :)

franzpoeschel:
> Note: there is no in-memory overload for load_chunk yet exposed in Python.

I think it is still possible to trigger this bug in the `load_chunk` version. Notice that we create a new `py::array` on [this line](https://github.com/openPMD/openPMD-api/blob/400d2e7d30de89f4f9141a2293196a875980550e/src/binding/python/RecordComponent.cpp#L371) and pass it via `shareRaw` to the C++ API [a few lines later](https://github.com/openPMD/openPMD-api/blob/400d2e7d30de89f4f9141a2293196a875980550e/src/binding/python/RecordComponent.cpp#L374). If the returned array is garbage-collected before the next flush, the next flush will write to free'd memory. That won't really happen in any sensible workflow, but we should still try to make it somewhat impossible to trigger memory errors from Python, and this one can be fixed much the same way that we already fixed the write-side API in this PR.

ax3l:
That's a good point, a simple example would be `r.load_chunk()` without catching the returned object.

I try to address all `load_chunk` issues and missing overload in #913. Do you think regarding `store_chunk`, this PR is ready, too?

PR #913, opened 2021-01-29
Python: fix load_chunk to temporary

ax3l:
Keep a reference in `load_chunk` until data is not used anymore.

Follow-up to #912 and related to #833

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -180,6 +180,34 @@ parseTupleSlices(uint8_t const ndim, Extent const & full_extent, py::tuple const
     return std::make_tuple(offset, extent, flatten);
 }
 
+/** Check an array is a contiguous buffer
+ *
+ * Required are contiguous buffers for store and load
+ *
+ * - not strided with paddings
+ * - not a view in another buffer that results in striding
+ */
+inline void
+check_buffer_is_contiguous( py::array & a ) {
+
+    auto* view = new Py_buffer();
+    int flags = PyBUF_STRIDES | PyBUF_FORMAT;
+    if( PyObject_GetBuffer( a.ptr(), view, flags ) != 0 )
+    {
+        delete view;
+        throw py::error_already_set();
+    }
+    bool isContiguous = ( PyBuffer_IsContiguous( view, 'A' ) != 0 );
+    PyBuffer_Release( view );
+    delete view;
+
+    if( !isContiguous )
+        throw py::index_error(
+                "strides in chunk are inefficient, not implemented!");
+    // @todo in order to implement stride handling, one needs to
+    //       loop over the input data strides in store/load calls
+}
+
 /** Store Chunk
  *
  * Called with offset and extent that are already in the record component's
@@ -260,27 +288,7 @@ store_chunk(RecordComponent & r, py::array & a, Offset const & offset, Extent co
             );
     }
 
-    /* required are contiguous buffers
-     *
-     * - not strided with paddings
-     * - not a view in another buffer that results in striding
-     */
-    auto* view = new Py_buffer();
-    int flags = PyBUF_STRIDES | PyBUF_FORMAT;
-    if( PyObject_GetBuffer( a.ptr(), view, flags ) != 0 )
-    {
-        delete view;
-        throw py::error_already_set();
-    }
-    bool isContiguous = ( PyBuffer_IsContiguous( view, 'A' ) != 0 );
-    PyBuffer_Release( view );
-    delete view;
-
-    if( !isContiguous )
-        throw py::index_error(
-            "strides in chunk are inefficient, not implemented!");
-    // @todo in order to implement stride handling, one needs to
-    //       loop over the input data strides during write below
+    check_buffer_is_contiguous( a );
 
     // here, we increase a reference on the user-passed data so that
     // temporary and lost-scope variables stay alive until we flush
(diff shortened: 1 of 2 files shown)


franzpoeschel:
We might want to be careful not to do this twice ;) @ax3l 
https://github.com/openPMD/openPMD-api/pull/901

I needed to implement something very similar for that PR

ax3l:
Yes, that's the follow-up PR to address the additional issue you mentioned and add an overload to allow in-memory reads:

> > Note: there is no in-memory overload for load_chunk yet exposed in Python.
>
> I think it is still possible to trigger this bug in the `load_chunk` version. Notice that we create a new `py::array` on [this line](https://github.com/openPMD/openPMD-api/blob/400d2e7d30de89f4f9141a2293196a875980550e/src/binding/python/RecordComponent.cpp#L371) and pass it via `shareRaw` to the C++ API [a few lines later](https://github.com/openPMD/openPMD-api/blob/400d2e7d30de89f4f9141a2293196a875980550e/src/binding/python/RecordComponent.cpp#L374). If the returned array is garbage-collected before the next flush, the next flush will write to free'd memory. That won't really happen in any sensible workflow, but we should still try to make it somewhat impossible to trigger memory errors from Python, and this one can be fixed much the same way that we already fixed the write-side API in this PR.

src/binding/python/RecordComponent.cpp
```diff
@@ -425,7 +440,25 @@ load_chunk(RecordComponent & r, py::tuple const & slices)
...
+            std::begin(extent),
+            std::end(extent),
+            std::begin(shape),
+            [&maskIt](std::uint64_t){
+                return !*(maskIt++);
+            }
+    );
+
+    auto const dtype = dtype_to_numpy( r.getDatatype() );
+    auto a = py::array( dtype, shape );
```
ax3l, line 471: Note from [here](https://github.com/openPMD/openPMD-api/pull/912#issuecomment-769739567): in the original code, users can delete the returned object (or not catch it at all) which results in a use-after-free in flush.

Double-check that this becomes safe with this PR.

test/python/unittest/API/APITest.py
```diff
@@ -1629,6 +1630,59 @@ def testAvailableChunks(self):
...
+                        [1, 0])
+        data = np.array([[1, 2, 3]], dtype=np.dtype("int"))
+        E_x.store_chunk(data)
+
+    def loadToTemporaryStore(self, r_E_x):
+        #data = np.zeros((2, 3,), dtype=np.dtype("int"))
+        #r_E_x.load_chunk(data, [0, 0], [2, 3])
+        #data2 = np.zeros((2, 3,), dtype=np.dtype("int"))
+        #r_E_x.load_chunk(data2)
+        return
```
ax3l, line 1723:
```suggestion
        # not catching the return value shall not result in a use-after-free:
        # r_E_x.load_chunk()
        return
```

test/python/unittest/API/APITest.py
```diff
@@ -1636,6 +1636,11 @@ def writeFromTemporaryStore(self, E_x):
         data = np.array([[1, 2, 3]], dtype=np.dtype("int"))
         E_x.store_chunk(data)
 
+    def loadToTemporaryStore(self, r_E_x):
+        # not catching the return value shall not result in a use-after-free:
+        r_E_x.load_chunk()
```
ax3l, line 1719: @franzpoeschel it looks like these tests does not trigger the issue we were concerned about.

Maybe the GC zeroes out the data it frees or we are otherwise lucky... But then we should see this throw: https://github.com/openPMD/openPMD-api/blob/0.13.1/include/openPMD/RecordComponent.hpp#L334-L335

Do you have another Python snippet we could add in the test here that demonstrates the `load_chunk` issue?

src/binding/python/RecordComponent.cpp
```diff
@@ -349,64 +357,83 @@ store_chunk(RecordComponent & r, py::array & a, py::tuple const & slices)
...
+	    str_extent_shape.append(" ").append(std::to_string(si));
+    }
+    for( py::ssize_t d = 0; d < a.ndim(); ++d ) {
+	    s_array *= a.shape()[d];
+	    str_array_shape.append(" ").append(std::to_string(a.shape()[d]));
+    }
 
-    auto const dtype = dtype_to_numpy( r.getDatatype() );
+    /* we allow flattening of the result dimension
+    if( size_t(a.ndim()) != extent.size() )
```
ax3l: We could check this, since generally `<=` is allowed here
```suggestion
    if( size_t(a.ndim()) > extent.size() )
```

ax3l:
> We might want to be careful not to do this twice ;) @ax3l
> #901
>
> I needed to implement something very similar for that PR

@franzpoeschel sorry, there are too many LOC and feature editions in #901 that I did not see the partly overlap :see_no_evil: 

I splitted my PR in a bugfix (this PR) and a feature addition (#914) to make them concise, backportable and easy to review. Let's see how we mingle them together.

Is #901 getting smaller once its dependent PR is merged and it can be rebased? :)

franzpoeschel:
> Is #901 getting smaller once its dependent PR is merged and it can be rebased? :)

A bit, but not drastically. Since this PR is the more "precise" one, I'd say go forward with implementing things here. I'll deal with getting things compatible during rebasing afterwards.

test/python/unittest/API/APITest.py
franzpoeschel: I'd say this is likely a combination of (1) the rather small dataset that we test this on, (2) the fact that `py::array` initializes its data upon constructon and (3) that we don't allocate anything in between. So, maybe we could trigger this by making things go a bit wilder.

franzpoeschel: https://github.com/franzpoeschel/openPMD-api/tree/topic-pyStoreChunkInMem-increaseBuffersize
This branch triggers the issue on my machine and your implementation fixes it.

ax3l: Great hint, thank you - increasing the tested memory triggers the issue reliably :)

PR #914, opened 2021-01-29
[WIP] Python: in-memory version of load_chunk

ax3l:
Add an overload to store chunks in Python into pre-allocated memory.

- [x] rebase after #913 was merged

### DIFF
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -1863,7 +1863,7 @@ class APITest(unittest.TestCase):
                             [2, 0])
 
     def loadToTemporaryStore(self, r_E_x):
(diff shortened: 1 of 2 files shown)


ax3l:
I think we might have added this in #901 already...

test/python/unittest/API/APITest.py
```diff
@@ -1863,6 +1863,12 @@ def writeFromTemporaryStore(self, E_x):
                             [2, 0])
 
     def loadToTemporaryStore(self, r_E_x):
+        if found_numpy:
```
ax3l: Just another write-to-temp check that we probably do not support yet:
```suggestion
        if found_numpy and False:
```

PR #915, opened 2021-01-29
Python: Complex Types in store_chunk

ax3l:
Add missing support for complex types in the Python store chunk interfaces of `Record_Component`.

Follow-up to #639


test/python/unittest/API/APITest.py
```diff
@@ -525,14 +525,13 @@ def makeConstantRoundTrip(self, file_ending):
                             np.dtype('double'))
             self.assertTrue(ms["longdouble"][SCALAR].load_chunk(o, e).dtype
                             == np.dtype('longdouble'))
-            if file_ending != "json":
```
ax3l, line 528: works now since #885

Issue #510, opened 2019-03-29
Extensible datasets / Flexible Extent until Close


RemiLehe:
Does the openPMD-api support extensible datasets?
(e.g. dataset that can be resized with [H5Dset_extent](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5D.html#Dataset-SetExtent))

A typical example would be in order to write the properties of all the particles that cross a given plane, thoughout a simulation: 
in this case we do not know the number of particles that we need to write in advance, but we can resize the dataset every time that we need to dump this particle data. (in principle at every iteration, but in practice, this data can be cashed in RAM for several iterations, so that we only need to resize datasets and dump data to disk at every ~10 or ~100 iterations).

The above example is closely related to writing the particle in the lab frame, during a boosted-frame simulation, which is what I am really interested in, in the end :)

ax3l:
Not yet, but it can be enabled for some backends such HDF5 and ADIOS1/2.

Nevertheless, please be  aware that extensible datasets are horrible in performance in HDF5. We have done benchmarks with libSplash in the past (of course, we forgot to post the numbers, argh) and I would not use them for anything serious: https://github.com/ComputationalRadiationPhysics/libSplash/issues/133 https://github.com/ComputationalRadiationPhysics/libSplash/pull/135

If your workflow allows, avoid `H5F_UNLIMITED` at all costs (also plays into the nightmare of chunking in H5 #406), which will make your I/O faster and your choice of backends more flexible. I know it's unfortunate, but HDF5 today do not implement resizing/appending in an efficient way at scale.

In ADIOS(1/2), "extent" of a dataset is a more of loose meta concept and one could just write more and more chunks and store the final size in a variable. There it should work well already, but one should check. We probably should expose this "flexible" extent then in the frontend as well, otherwise the read API might crop the extent on what you initially gave it.

Those are not the latest slides from a few weeks ago, but from 2016. Still, the tutorial contains bounding boxes: https://users.nccs.gov/~pnorbert/adios-tutorial.pdf in p. 69+. We should be able to expose them that way :)

I personally would address such a task with chunks in openPMD-api and probably also the upcoming streaming API by @franzpoeschel, but it works regularly as well. Maybe we can add such a workflow as an example!

RemiLehe:
Thanks a lot for the above comments and advice! This makes a lot of sense!

I agree that resizable datasets should be avoided whenever possible. However, there are cases in which they seem (to me) to be hard to avoid.

[...]

PR #916, opened 2021-01-29
HDF5: Empiric for Optimal Chunk Size

### LINKED ISSUE #406: HDF5: Chunking

ax3l:
This ports a prior empirical algorithm from libSplash to determine an optimal (large) chunk size for an HDF5 dataset based on its datatype and global extent.

Original implementation by Felix Schmitt @f-schmitt (ZIH, TU Dresden) in [libSplash](https://github.com/ComputationalRadiationPhysics/libSplash).

Original source:
- https://github.com/ComputationalRadiationPhysics/libSplash/blob/v1.7.0/src/DCDataSet.cpp
- https://github.com/ComputationalRadiationPhysics/libSplash/blob/v1.7.0/src/include/splash/core/DCHelper.hpp

Close #406
Related to #898 (improve HDF5 baseline performance)
Required for #510: basis to extend resizable data sets (#829) to HDF5

### To Do

- [x] pass tests
- [x] measure performance ([h5chunks_cori.sbatch.txt](https://github.com/openPMD/openPMD-api/files/6640943/h5chunks_cori.sbatch.txt))
  - Cori seems to be still slow, also when playing further with `OPENPMD_HDF5_INDEPENDENT="OFF"` and `OPENPMD_HDF5_ALIGNMENT="1048576"` for our `8_benchmark_parallel -w` benchmark
- [x] measure filesize overhead

Sample & bin directory `du -hs bin samples` with:
| `OPENPMD_HDF5_CHUNKS` | size |
|--|--|
| "auto"/*unset* | 13M + 1.6G |
| "none" | 11M + 1.6G |

on my laptop (4KiB blocksize).

With `"auto"` the `MPI.8_benchmark_parallel` test is significantly slower. Changing the 4D test to 3D brings the difference down to about 20% slowdown #1010. (Maybe the 4th, 10-element dimension is sub-ideal for chunking?)

# Follow-Ups
- [ ] allow definition on a per-dataset basis
- [ ] allow passing user-defined per-dataset Byte arrays for chunks

### DIFF
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -74,6 +74,23 @@ Explanation of the single keys:
 Any setting specified under ``adios2.dataset`` is applicable globally as well as on a per-dataset level.
 Any setting under ``adios2.engine`` is applicable globally only.
 
+HDF5
+^^^^
+
+A full configuration of the HDF5 backend:
+
+.. literalinclude:: hdf5.json
+   :language: json
+
+All keys found under ``hdf5.dataset`` are applicable globally (future: as well as per dataset).
+Explanation of the single keys:
+
+* ``adios2.dataset.chunks``: This key contains options for data chunking via `H5Pset_chunk <https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_chunk.htm>`__.
+  The default is ``"auto"`` for a heuristic.
+  ``"none"`` can be used to disable chunking.
+  Chunking generally improves performance and only needs to be disabled in corner-cases, e.g. when heavily relying on independent, parallel I/O that non-collectively declares data records.
+
+
 Other backends
 ^^^^^^^^^^^^^^
 
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -51,7 +51,8 @@ namespace openPMD
 #       define VERIFY(CONDITION, TEXT) do{ (void)sizeof(CONDITION); } while( 0 )
 #   endif
 
-HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
+HDF5IOHandlerImpl::HDF5IOHandlerImpl(
+    AbstractIOHandler* handler, nlohmann::json config)
         : AbstractIOHandlerImpl(handler),
           m_datasetTransferProperty{H5P_DEFAULT},
           m_fileAccessProperty{H5P_DEFAULT},
@@ -82,6 +83,21 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
     H5Tinsert(m_H5T_CDOUBLE, "i", sizeof(double), H5T_NATIVE_DOUBLE);
     H5Tinsert(m_H5T_CLONG_DOUBLE, "r", 0, H5T_NATIVE_LDOUBLE);
     H5Tinsert(m_H5T_CLONG_DOUBLE, "i", sizeof(long double), H5T_NATIVE_LDOUBLE);
+
+    if( config.contains( "hdf5" ) )
+    {
+        m_config = std::move( config[ "hdf5" ] );
+        /*
+         * @todo Apply global configuration options from the JSON config here.
+         */
+        auto shadow = m_config.invertShadow();
+        if( shadow.size() > 0 )
+        {
+            std::cerr << "Warning: parts of the JSON configuration for "
+                         "HDF5 remain unused:\n"
+                      << shadow << std::endl;
+        }
+    }
 }
 
 HDF5IOHandlerImpl::~HDF5IOHandlerImpl()
@@ -238,6 +254,25 @@ HDF5IOHandlerImpl::createDataset(Writable* writable,
             name = auxiliary::replace_first(name, "/", "");
         if( auxiliary::ends_with(name, '/') )
             name = auxiliary::replace_last(name, "/", "");
+
+        auto config = nlohmann::json::parse( parameters.options );
+        if( config.contains( "hdf5" ) &&
+            config[ "hdf5" ].contains( "dataset" ) )
+        {
+            auxiliary::TracingJSON datasetConfig{
+                config[ "hdf5" ][ "dataset" ] };
+            /*
+             * @todo Read options from config here.
+             */
+            auto shadow = datasetConfig.invertShadow();
+            if( shadow.size() > 0 )
+            {
+                std::cerr << "Warning: parts of the JSON configuration for "
+                             "HDF5 dataset '"
+                          << name << "' remain unused:\n"
+                          << shadow << std::endl;
+            }
+        }
 
         /* Open H5Object to write into */
         auto res = getFile( writable );
@@ -1843,9 +1878,9 @@ HDF5IOHandlerImpl::getFile( Writable * writable )
 #endif
 
 #if openPMD_HAVE_HDF5
-HDF5IOHandler::HDF5IOHandler(std::string path, Access at)
+HDF5IOHandler::HDF5IOHandler(std::string path, Access at, nlohmann::json config)
         : AbstractIOHandler(std::move(path), at),
-          m_impl{new HDF5IOHandlerImpl(this)}
+          m_impl{new HDF5IOHandlerImpl(this, std::move(config))}
 { }
 
 HDF5IOHandler::~HDF5IOHandler() = default;
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -20,6 +20,7 @@
  */
 #include "openPMD/IO/HDF5/HDF5IOHandler.hpp"
 #include "openPMD/IO/HDF5/HDF5IOHandlerImpl.hpp"
+#include "openPMD/auxiliary/Environment.hpp"
 
 #if openPMD_HAVE_HDF5
 #   include "openPMD/Datatype.hpp"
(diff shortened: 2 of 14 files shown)


ax3l:
At the moment, a naïve port of the logic causes a deadlock in the MPI tests.

Stuck only in `hipace_like_write` with
```
dims: 300 20
optimal chunks found: 256 16
```

After some tests, it looks like adding chunking makes the dataset declaration `H5Dcreate` collective in that case, urgh... (It's actually declared as [always collective](https://support.hdfgroup.org/HDF5/doc/RM/CollectiveCalls.html).)

[Notes:](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5P.html#Property-SetChunk)
```
Chunk size cannot exceed the size of a fixed-size dataset. For example, a dataset consisting of a 5x4 fixed-size array cannot be defined with 10x10 chunks. 
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -255,21 +256,30 @@ HDF5IOHandlerImpl::createDataset(Writable* writable,
...
+        if( num_elements != 0u )
+        {
+            // get chunking dimensions
+            std::vector< hsize_t > chunk_dims = getOptimalChunkDims(dims, toBytes(d));
+
+            // TODO: allow overwrite with user-provided chunk size
+            //for( auto const& val : parameters.chunkSize )
+            //    chunk_dims.push_back(static_cast< hsize_t >(val));
+
+            herr_t status = H5Pset_chunk(datasetCreationProperty, chunk_dims.size(), chunk_dims.data());
```
ax3l, line 337: Fun fact: `H5Pset_chunk_opts` (HDF5 1.10.0+):
```
H5Pset_chunk_opts is used to specify storage options for chunks on the edge of a dataset’s dataspace. This capability allows the user to tune performance in cases where the dataset size may not be a multiple of the chunk size and the handling of partial edge chunks can impact performance. 
```

franzpoeschel:
I've added some scaffolding for JSON options in HDF5

ax3l:
Finished the global options (JSON & env) to disable chucking when needed (mainly for HiPACE's legacy pipeline + potential regressions).

ax3l:
It's a bit concerning that the clang sanitizer run parallel benchmark (8) runs into a time-out with the new patch:
```
15/32 Test #15: MPI.8_benchmark_parallel ...............***Timeout 1500.17 sec
```

But I don't see an immediately relation. 'll turn the chunking off for this one.

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -81,6 +84,42 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
...
+        if( m_config.json().contains( "dataset" ) )
+        {
+            auto datasetConfig = m_config[ "dataset" ];
+            if( datasetConfig.json().contains( "chunks" ) )
+            {
+                m_chunks = datasetConfig.json()[ "chunks" ];
+            }
+            else
+            {
+                m_chunks = auxiliary::getEnvString( "OPENPMD_HDF5_CHUNKS", "auto" );
```
ax3l, line 102: This is too deep inside to take effect if `! config.contains( "hdf5" )`

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -81,6 +84,42 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
     H5Tinsert(m_H5T_CDOUBLE, "i", sizeof(double), H5T_NATIVE_DOUBLE);
     H5Tinsert(m_H5T_CLONG_DOUBLE, "r", 0, H5T_NATIVE_LDOUBLE);
     H5Tinsert(m_H5T_CLONG_DOUBLE, "i", sizeof(long double), H5T_NATIVE_LDOUBLE);
+
+    if( config.contains( "hdf5" ) )
```
ax3l, line 90:
```suggestion
    m_chunks = auxiliary::getEnvString( "OPENPMD_HDF5_CHUNKS", "auto" );
    // JSON option can overwrite env option:
    if( config.contains( "hdf5" ) )
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -81,6 +84,42 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
...
+        {
+            auto datasetConfig = m_config[ "dataset" ];
+            if( datasetConfig.json().contains( "chunks" ) )
+            {
+                m_chunks = datasetConfig.json()[ "chunks" ];
+            }
+            else
+            {
+                m_chunks = auxiliary::getEnvString( "OPENPMD_HDF5_CHUNKS", "auto" );
+            }
```
ax3l: This is too deep inside to take effect if `! config.contains( "hdf5" )`
```suggestion
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -81,6 +84,40 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
...
+
+        // check for global dataset configs
+        if( m_config.json().contains( "dataset" ) )
+        {
+            auto datasetConfig = m_config[ "dataset" ];
+            if( datasetConfig.json().contains( "chunks" ) )
+            {
+                m_chunks = datasetConfig.json()[ "chunks" ];
+            }
+            datasetConfig.declareFullyRead();
```
franzpoeschel:
```suggestion
```

`TracingJSON` does not trace beyond JSON arrays since those are a bit tricky to handle and it didn't matter so far. The purpose of `declareFullyRead` (and I forgot to document that) is to declare an array as fully read. By declaring the `datasetConfig` object as fully read, you will ignore wrong keys that the user passed, so you'll want to drop that.

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -81,6 +84,40 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(AbstractIOHandler* handler)
...
+    {
+        m_config = std::move( config[ "hdf5" ] );
+
+        // check for global dataset configs
+        if( m_config.json().contains( "dataset" ) )
+        {
+            auto datasetConfig = m_config[ "dataset" ];
+            if( datasetConfig.json().contains( "chunks" ) )
+            {
+                m_chunks = datasetConfig.json()[ "chunks" ];
```
franzpoeschel:
```suggestion
                m_chunks = datasetConfig[ "chunks" ].json().get< std::string >();
```
Calling `operator[]()` on the `TracingJSON` object will trace the key. I think it's good to be explicit with the type that you want, I already ran in to trouble because I wasn't in the past.

franzpoeschel (changes requested):

franzpoeschel:
I tried out whether this PR alone already enables extensible datasets in HDF5, apparently not:
```
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: ../../../src/H5D.c line 906 in H5Dset_extent(): unable to set extend dataset
    major: Dataset
  #001: ../../../src/H5Dint.c line 2615 in H5D__set_extent(): unable to modify size of dataspace
    major: Dataset
  #002: ../../../src/H5S.c line 2005 in H5S_set_extent(): dimension cannot exceed the existing maximal size (new: 10 max: 5)
    major: Dataspace
    minor: Bad value
```

According to the [documentation](https://support.hdfgroup.org/HDF5/doc/RM/RM_H5D.html#Dataset-SetExtent), only certain kinds of datasets can have their extents extended:
1. A chunked dataset with unlimited dimensions
2. A chunked dataset with fixed dimensions if the new dimension sizes are less than the maximum sizes set with maxdims (see H5Screate_simple)

Our datasets seem to fall into the second category. I guess, for extensible datasets, we would have to create unlimited datasets from the beginning?

ax3l:
In order to make a dataset resizable, we need to pass `H5F_UNLIMITED` into the dimensions of `H5Screate_simple` that should be resizable. Yes, we need chunking (this PR) to be able to use this ([ref](https://github.com/ComputationalRadiationPhysics/libSplash/issues/133#issuecomment-44014726) / [PR](https://github.com/ComputationalRadiationPhysics/libSplash/pull/135/files)).

Setting `H5F_UNLIMITED` on dimensions can have a negative performance implication, so users need to declare that intent upfront and then we just make all dims of a data set resizable.

src/IO/HDF5/HDF5IOHandler.cpp
ax3l: Oh, ok! Can you add these doc details in a follow-up? :)

docs/source/details/backendconfig.rst
```diff
@@ -74,6 +74,23 @@ Explanation of the single keys:
 Any setting specified under ``adios2.dataset`` is applicable globally as well as on a per-dataset level.
 Any setting under ``adios2.engine`` is applicable globally only.
 
+HDF5
+^^^^
+
+A full configuration of the HDF5 backend:
+
+.. literalinclude:: hdf5.json
```
ax3l, line 82: @franzpoeschel I just realized I forgot to add a `hdf5.json` file here :zany_face:

docs/source/details/backendconfig.rst
```diff
@@ -74,6 +74,23 @@ Explanation of the single keys:
...
+   :language: json
+
+All keys found under ``hdf5.dataset`` are applicable globally (future: as well as per dataset).
+Explanation of the single keys:
+
+* ``adios2.dataset.chunks``: This key contains options for data chunking via `H5Pset_chunk <https://support.hdfgroup.org/HDF5/doc/RM/H5P/H5Pset_chunk.htm>`__.
```
ax3l, line 88: Ouch, that should read `hdf5.dataset.chunks`...

docs/source/details/backendconfig.rst
ax3l: Ah, I accidentally named it `json.json`

docs/source/details/backendconfig.rst
ax3l: -> #1169

docs/source/details/backendconfig.rst
ax3l: -> #1169

PR #917, opened 2021-02-01
HDF5 Auxiliary: .cpp file

ax3l:
Add a `.cpp` file for auxiliary HDF5 functions.


ax3l:
I'll add this already so we can rebase #916 and add the json control to HDF5 :)

PR #919, opened 2021-02-03
MSVC: Proper __cplusplus macro

ax3l:
Modern MSVC versions finally set the `__cplusplus` macro correctly, but we need to request this with an additional flag.

Refs.:
- https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-160
- https://gitlab.kitware.com/cmake/cmake/-/issues/18837

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -395,9 +395,14 @@ set_target_properties(openPMD PROPERTIES
     POSITION_INDEPENDENT_CODE ON
     WINDOWS_EXPORT_ALL_SYMBOLS ON
 )
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-    target_compile_options(openPMD PUBLIC "/bigobj")
-endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -395,9 +395,10 @@ set_target_properties(openPMD PROPERTIES
     POSITION_INDEPENDENT_CODE ON
     WINDOWS_EXPORT_ALL_SYMBOLS ON
 )
-if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
-    target_compile_options(openPMD PUBLIC "/bigobj")
-endif()
+set(_cxx_msvc   "$<AND:$<COMPILE_LANGUAGE:CXX>,$<CXX_COMPILER_ID:MSVC>>")
+set(_msvc_1914  "$<VERSION_GREATER_EQUAL:$<CXX_COMPILER_VERSION>,19.14>")
+set(_msvc_options /bigobj$<${_cxx_msvc}:$<${_msvc_1914}:;/Zc:__cplusplus>>)
```
ax3l: This builds a semicolon-separated list with a conditional 2nd argument

PR #923, opened 2021-02-05
ParticleSpecies: Read to pandas.DataFrame

ax3l:
Add a method tha reads a particle species into a `pandas.DataFrame`.

### To Do

- [x] rebase after #924 was merged

### DIFF
--- a/examples/11_particle_dataframe.py
+++ b/examples/11_particle_dataframe.py
@@ -8,20 +8,28 @@ License: LGPLv3+
 """
 import openpmd_api as io
 import numpy as np
-
+import sys
 try:
     import pandas as pd
-
-    s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
-    electrons = s.iterations[400].particles["electrons"]
-
-    # all particles
-    df = electrons.to_df()
-    print(df)
-
-    # only first 100 particles
-    df = electrons.to_df(np.s_[:100])
-    print(df)
-
 except ImportError:
     print("pandas NOT found. Install pandas to run this example.")
+    sys.exit()
+
+
+s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
+electrons = s.iterations[400].particles["electrons"]
+
+# all particles
+df = electrons.to_df()
+print(type(df) is pd.DataFrame)
+print(df)
+
+# only first 100 particles
+df = electrons.to_df(np.s_[:100])
+print(df)
+
+# get optimal chunks
+for chunk in electrons["position"]["x"].available_chunks():
+    stride = np.s_[chunk.offset[0]:chunk.extent[0]]
+    df = electrons.to_df()
+    print(df)
--- a/examples/11_particle_dataframe.py
+++ b/examples/11_particle_dataframe.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+"""
+This file is part of the openPMD-api.
+
+Copyright 2021 openPMD contributors
+Authors: Axel Huebl
+License: LGPLv3+
+"""
+import openpmd_api as io
+import numpy as np
+
+try:
+    import pandas as pd
+
+    s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
+    electrons = s.iterations[400].particles["electrons"]
+
+    # all particles
+    df = electrons.to_df()
+    print(df)
+
+    # only first 100 particles
+    df = electrons.to_df(np.s_[:100])
+    print(df)
+
+except ImportError:
+    print("pandas NOT found. Install pandas to run this example.")
--- a/src/binding/python/openpmd_api/DataFrame.py
+++ b/src/binding/python/openpmd_api/DataFrame.py
@@ -1,0 +1,64 @@
+"""
+This file is part of the openPMD-api.
+
+Copyright 2021 openPMD contributors
+Authors: Axel Huebl
+License: LGPLv3+
+"""
+import numpy as np
+try:
+    import pandas as pd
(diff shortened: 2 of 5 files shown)


src/ParticleSpecies.cpp
```diff
@@ -116,6 +118,31 @@ namespace
     }
 }
 
+void
+ParticleSpecies::seriesFlush()
+{
+    /** should look correctly somehow like this
```
ax3l, line 124: @franzpoeschel Do you have an idea how we can implement this safely here? :)

src/ParticleSpecies.cpp
```diff
@@ -116,6 +118,31 @@ namespace
     }
 }
 
+void
+ParticleSpecies::seriesFlush()
```
ax3l: @franzpoeschel Do you have an idea how we can implement this method safely here? :)
If we find a general solution, we could later on also add this to a more general class like `Writable`, I think

I also tried looping through `parent->attributable` until I reach it's root via a `nullptr` but that did not lead anywhere yet

The commented attempts segfault, so I guess I can either not call `flush()` or `flush_impl()` like this or I did not hit the right object depth.

franzpoeschel: [Little hint](https://github.com/openPMD/openPMD-api/blob/242cb1dbf5c508f6f8da3669f48fe26bbfc35b37/src/backend/Attributable.cpp#L151) ;) 
Needed to do the same trick for #901

I'm not quite sure though why your implementation didn't work, I don't see what I did different.

ax3l: Ah perfect, I was close. Will try that construct, let's see :)

ax3l: That works! :) Thank you :sparkles:

src/binding/python/openpmd_api/DataFrame.py
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+try:
+    import pandas as pd
+    found_pandas = True
+except ImportError:
+    print("pandas NOT found. Install pandas for DataFrame support.")
+    found_pandas = False
+
+
+def particles_to_dataframe(particle_species, slice=None):
```
ax3l, line 16: Potential extra arguments:
- `records = iterable`: a list of strings or regex expressions to load only specific records of the species
- maybe an inverse parameter to deselect specific records from loading

examples/11_particle_dataframe.py
```diff
@@ -0,0 +1,23 @@
...
+
+Copyright 2021 openPMD contributors
+Authors: Axel Huebl
+License: LGPLv3+
+"""
+import openpmd_api as io
+import numpy as np
+
+
+if __name__ == "__main__":
```
ax3l: The main check here is actually not needed / not pretty.

Instead, we could add a guard here and execute this only if pandas was found

src/binding/python/openpmd_api/DataFrame.py
```diff
@@ -0,0 +1,65 @@
...
+Copyright 2021 openPMD contributors
+Authors: Axel Huebl
+License: LGPLv3+
+"""
+import numpy as np
+try:
+    import pandas as pd
+    found_pandas = True
+except ImportError:
+    print("pandas NOT found. Install pandas for DataFrame support.")
```
dmitry-ganyushin: Do you have separation between critical errors and warnings? Why not to raise a custom exception here

examples/11_particle_dataframe.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
```
dmitry-ganyushin: Is it python 2 or python3?

src/binding/python/openpmd_api/DataFrame.py
```diff
@@ -0,0 +1,65 @@
...
+        for rc_name, rc in record.items():
+            if record.scalar:
+                column_name = record_name
+            else:
+                column_name = record_name + "_" + rc_name
+            columns[column_name] = rc[slice]
+            particle_species.series_flush()
+            columns[column_name] *= rc.unit_SI
+
+    return pd.DataFrame(columns)
```
dmitry-ganyushin, line 64: How large are particle species data? It looks like panda lib has capability to work with tables not fitting into memory.

dmitry-ganyushin (approved):
Ok as a frist version, but see comments.

src/binding/python/openpmd_api/DataFrame.py
ax3l: At this point I want to be able to not cause issues if pandas is not available but the function in this module is never called. I could also do this quietly, though, no reason to print here at all (throw is below).

examples/11_particle_dataframe.py
ax3l: All python3, could specify here but that is ensured anyway via packaging logic.

src/binding/python/openpmd_api/DataFrame.py
ax3l: Particle data can be anything between 0 and 1e13 as of today.

Can you point me to information on how to handle out-of-core data in pandas?

examples/11_particle_dataframe.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
+electrons = s.iterations[400].particles["electrons"]
+
+# all particles
+df = electrons.to_df()
+print(df)
+
+# only first 100 particles
```
ax3l, line 27: extend example:
```
l_chunks = electrons.available_chunks()
```

PR #933, opened 2021-02-19
bp4_steps test: actually use nullcore engine

franzpoeschel:
Previously, the nullcore configuration of that test forgot to set the nullcore engine. This fixes that.


ax3l (approved):
Good catch! :)

PR #935, opened 2021-02-25
ParticleSpecies: Read to dask.dataframe

ax3l:
Add a method that reads a particle species into a `dask.dataframe`.

Feel the power :fire:

Cheers to @dmitry-ganyushin for helping with this!

### DIFF
--- a/.rodare.json
+++ b/.rodare.json
@@ -90,6 +90,12 @@
             "affiliation": "Helmholtz-Zentrum Dresden-Rossendorf",
             "name": "Ordyna, Paweł",
             "type": "Other"
+        },
+        {
+            "affiliation": "Oak Ridge National Laboratory",
+            "name": "Ganyushin, Dmitry",
+            "orcid": "0000-0001-7337-2161",
+            "type": "Other"
         }
     ],
     "title": "C++ & Python API for Scientific I/O with openPMD",
--- a/examples/11_particle_dataframe.py
+++ b/examples/11_particle_dataframe.py
@@ -41,14 +41,17 @@ if found_dask:
     df = electrons.to_dask()
     print(df)
 
+    # check chunking of a variable
+    print("chunking={}".format(df["momentum_z"].to_dask_array()))
+
     # trigger a couple of compute actions
     #   example1: average momentum in z
     print("<momentum_z>={}".format(df["momentum_z"].mean().compute()))
 
     #   example2: momentum histogram
-    h, bins = da.histogram(df["momentum_z"], bins=50,
-                           range=[-8.0e-23, 8.0e-23])
-    #                      weights=df["weighting"]
+    h, bins = da.histogram(df["momentum_z"].to_dask_array(), bins=50,
+                           range=[-8.0e-23, 8.0e-23],
+                           weights=df["weighting"].to_dask_array())
     print(h.compute())
 
     #   example3: save all data data to parquet files
--- a/examples/11_particle_dataframe.py
+++ b/examples/11_particle_dataframe.py
@@ -3,7 +3,7 @@
 This file is part of the openPMD-api.
 
 Copyright 2021 openPMD contributors
-Authors: Axel Huebl
+Authors: Axel Huebl, Dmitry Ganyushin
 License: LGPLv3+
 """
 import openpmd_api as io
@@ -14,6 +14,13 @@ try:
 except ImportError:
     print("pandas NOT found. Install pandas to run this example.")
     sys.exit()
+found_dask = False
+try:
+    from dask.delayed import delayed
+    import dask.array as da
+    found_dask = True
+except ImportError:
+    print("dask NOT found. Install dask to run the 2nd example.")
 
 
 s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
@@ -28,8 +35,22 @@ print(df)
 df = electrons.to_df(np.s_[:100])
 print(df)
 
-# get optimal chunks
-for chunk in electrons["position"]["x"].available_chunks():
-    stride = np.s_[chunk.offset[0]:chunk.extent[0]]
-    df = electrons.to_df()
+
+# Dask example: process large datasets
+if found_dask:
+    df = electrons.to_dask()
     print(df)
+
+    # trigger a couple of compute actions
+    #   example1: average momentum in z
+    print("<momentum_z>={}".format(df["momentum_z"].mean().compute()))
+
+    #   example2: momentum histogram
+    h, bins = da.histogram(df["momentum_z"], bins=50,
+                           range=[-8.0e-23, 8.0e-23])
+    #                      weights=df["weighting"]
+    print(h.compute())
+
+    #   example3: save all data data to parquet files
+    delayed_save = delayed(df.to_parquet("electrons.parquet"))
+    delayed_save.compute()
--- a/src/binding/python/openpmd_api/DaskDataFrame.py
+++ b/src/binding/python/openpmd_api/DaskDataFrame.py
@@ -1,0 +1,83 @@
+"""
+This file is part of the openPMD-api.
+
+Copyright 2021 openPMD contributors
+Authors: Axel Huebl, Dmitry Ganyushin
+License: LGPLv3+
+"""
+import numpy as np
+try:
+    import dask.dataframe as dd
+    from dask.delayed import delayed
+    found_dask = True
+except ImportError:
+    found_dask = False
+try:
+    import pandas as pd
+    found_pandas = True
+except ImportError:
+    found_pandas = False
+
+
+def particles_to_daskdataframe(particle_species):
+    """
+    Load all records of a particle species into a Dask DataFrame.
+
+    Parameters
+    ----------
+    particle_species : openpmd_api.ParticleSpecies
+        A ParticleSpecies class in openPMD-api.
+
+    Returns
+    -------
+    dask.dataframe
+        A dask dataframe with particles as index and openPMD record
+        components of the particle_species as columns.
+
+    Raises
+    ------
+    ImportError
+        Raises an exception if dask or pandas are not installed
+
+    See Also
+    --------
+    openpmd_api.BaseRecordComponent.available_chunks : available chunks that
+        are used internally to parallelize particle processing
+    dask.dataframe : the central dataframe object created here
+    """
+    if not found_dask:
+        raise ImportError("dask NOT found. Install dask for Dask DataFrame "
+                          "support.")
+    if not found_pandas:
+        raise ImportError("pandas NOT found. Install pandas for DataFrame "
+                          "support.")
+
+    # get optimal chunks: query first non-constant record component and
+    #                     assume the same chunking applies for all of them
+    #                     in a particle species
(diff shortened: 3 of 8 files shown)


examples/11_particle_dataframe.py
```diff
@@ -28,8 +35,21 @@
...
+    df = electrons.to_dask()
     print(df)
+
+    # trigger a couple of compute actions
+    #   example1: average momentum in z
+    print("<momentum_z>={}".format(df["momentum_z"].mean().compute()))
+
+    #   example2: momentum histogram
+    h, bins = da.histogram(df["momentum_z"], bins=50, range=[-8.0e-23, 8.0e-23])
+    #                      weights=df["weighting"]
```
ax3l: Some issue if I pass this argument alone the lines "`Series` has no attribute `chunks`" deep inside Dask... Not sure if it refers to *our* series, though - I think not :sweat_smile: 

Also, let's ask  the RAPIDS team tomorrow if we can also generate 2D and ND histograms. That would be tremendously helpful, but I cannot spot such a function in the docs.
Update: opened as https://github.com/dask/dask/issues/7307

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -0,0 +1,83 @@
...
+        for k_rc, rc in r.items():
+            if not rc.constant:
+                chunks = rc.available_chunks()
+                break
+        if chunks:
+            break
+
+    # TODO: implement available_chunks for constant record components
+    #       and fall back to a single, big chunk here
+    if chunks is None:
```
ax3l: @franzpoeschel I tried to query `available_chunks` from a `constant` `BaseRecordComponent` and realized this throws a backend error.

Probably the cleanest way for us to handle this would be to check for `constant()` in the frontend and return the full extend as a single chunk in that case, what do you think?

.rodare.json
```diff
@@ -90,6 +90,11 @@
             "affiliation": "Helmholtz-Zentrum Dresden-Rossendorf",
             "name": "Ordyna, Paweł",
             "type": "Other"
+        },
+        {
+            "affiliation": "Oak Ridge National Laboratory",
+            "name": "Ganyushin, Dmitry",
+            "type": "Other"
```
ax3l, line 98: @dmitry-ganyushin, do you have an ORCID? :)
```suggestion
            "orcid": "0000-0001-7337-2161",
            "type": "Other"
```

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -0,0 +1,83 @@
...
+        raise ValueError("All records are constant. No dask chunking "
+                         "implemented, use pandas dataframes.")
+
+    def read_chunk(species, chunk):
+        stride = np.s_[chunk.offset[0]:chunk.extent[0]]
+        return species.to_df(stride)
+
+    # merge DataFrames
+    dfs = delayed(pd.concat(
```
dmitry-ganyushin: Would be nice to add a test for that

ax3l: Yep! Until then: examples like this are executed and tested by our CI, too.

---

To do: use dask partitions instead of an in-memory pandas here that will run out of memory on a single node.

We can potentially "just" pass the list to `delayed()`:
```suggestion
    dfs = delayed(
```

ax3l: Yep, until then: examples are executed and tested by our CI.

.rodare.json
dmitry-ganyushin: 0000-0001-7337-2161

src/binding/python/openpmd_api/DaskDataFrame.py
franzpoeschel: Ah good catch, yeah that's probably the best solution.

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -0,0 +1,83 @@
...
+                          "support.")
+
+    # get optimal chunks: query first non-constant record component and
+    #                     assume the same chunking applies for all of them
+    #                     in a particle species
+    chunks = None
+    for k_r, r in particle_species.items():
+        for k_rc, rc in r.items():
+            if not rc.constant:
+                chunks = rc.available_chunks()
```
franzpoeschel, line 62: This assumes that chunks are equal across components. What happens if they're not? Will things just be less efficient or will things not work? In the latter case, we should probably guard for this case and throw an error.

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -0,0 +1,83 @@
...
+        raise ValueError("All records are constant. No dask chunking "
+                         "implemented, use pandas dataframes.")
+
+    def read_chunk(species, chunk):
+        stride = np.s_[chunk.offset[0]:chunk.extent[0]]
+        return species.to_df(stride)
+
+    # merge DataFrames
+    dfs = delayed(pd.concat(
+        [(read_chunk)(particle_species, chunk) for chunk in chunks]
```
franzpoeschel: Are the parens around `read_chunk` necessary?

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -0,0 +1,83 @@
...
+            break
+
+    # TODO: implement available_chunks for constant record components
+    #       and fall back to a single, big chunk here
+    if chunks is None:
+        raise ValueError("All records are constant. No dask chunking "
+                         "implemented, use pandas dataframes.")
+
+    def read_chunk(species, chunk):
+        stride = np.s_[chunk.offset[0]:chunk.extent[0]]
```
franzpoeschel, line 78: Similarly, this assumes that we are dealing with particle data (and hence 1D). Is this checked? We do have a Python class `<openPMD.ParticleSpecies>`, so this could theoretically be guarded against.
(Or are those lines enough checking?):
```
ParticleSpecies.to_df = particles_to_dataframe  # noqa
ParticleSpecies.to_dask = particles_to_daskdataframe  # noqa
```

src/binding/python/openpmd_api/DaskDataFrame.py
ax3l: Yep, will just be less efficient.
(Also very unlikely.)

src/binding/python/openpmd_api/DaskDataFrame.py
ax3l: haha, right, they are not - doesn't it look nice and functional to write them? :D

src/binding/python/openpmd_api/DaskDataFrame.py
ax3l: Good idea to check the `chunk` to be 1D, yep

Since this is implemented as `species` and `ParticleSpecies`, there should be little chance to accidentally pass in a field. Particle arrays are always 1D.

src/binding/python/openpmd_api/DaskDataFrame.py
franzpoeschel: Got me for a minute I have to admit ;)

dmitry-ganyushin: Yes, now paratnesses there are redundant. That is because there was a "delayed" statement in the delayed collections example of dask, but the next function,  pd.concat, does not understand delayed objects. That is probably the reason why I did not observe any speed-up by reading.

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -0,0 +1,83 @@
...
+        return species.to_df(stride)
+
+    # merge DataFrames
+    dfs = delayed(pd.concat(
+        [(read_chunk)(particle_species, chunk) for chunk in chunks]
+    ))
```
ax3l: The example [here](https://docs.dask.org/en/latest/delayed-collections.html) reads like this:
```suggestion
    dfs = [
        delayed(read_chunk)(particle_species, chunk) for chunk in chunks
    ]
```

examples/11_particle_dataframe.py
```diff
@@ -28,8 +35,22 @@
...
+# Dask example: process large datasets
+if found_dask:
+    df = electrons.to_dask()
     print(df)
+
+    # trigger a couple of compute actions
+    #   example1: average momentum in z
+    print("<momentum_z>={}".format(df["momentum_z"].mean().compute()))
+
+    #   example2: momentum histogram
```
ax3l, line 51: 2D histograms in dask for a phase space example:
- https://github.com/dask/dask/pull/7387
- https://github.com/xgcm/xhistogram/

src/binding/python/openpmd_api/DaskDataFrame.py
ax3l: Implemented in #942 :tada:

examples/11_particle_dataframe.py
ax3l: It's a `dask.dataframe.core.Series` that does not have the chunks...

ax3l: Adding `.to_dask_array()` fixes this

ax3l:
I think I found what we need for meshes: https://docs.dask.org/en/latest/array-api.html?highlight=from_array#other-functions
```
dask.array.from_array(x, chunks='auto', name=None, lock=False, asarray=None, fancy=True, getitem=None, meta=None, inline_array=False)
```
```
Create dask array from something that looks like an array.

Input must have a .shape, .ndim, .dtype and support numpy-style slicing.
```

```
Parameters: |
       x : array_like
  chunks : int, tuple
    How to chunk the array. Must be one of the following forms:
    * A blocksize like 1000.
    * A blockshape like (1000, 1000).
    * Explicit sizes of all blocks along all dimensions like ((1000, 1000, 500), (400, 400)).
    * A size in bytes, like “100 MiB” which will choose a uniform block-like shape
    * The word “auto” which acts like the above, but uses a configuration value array.chunk-size for the chunk size

  -1 or None as a blocksize indicate the size of the corresponding dimension.
...
 asarray : bool, optional
    If True then call np.asarray on chunks to convert them to numpy arrays. If False then chunks are passed through unchanged. If None (default) then we use True if the __array_function__ method is undefined.
...
   fancy : bool, optional
     If x doesn’t support fancy indexing (e.g. indexing with lists or arrays) then set to False.
     Default is True.
...
```

Not entirely sure yet how to tell it that it needs to call our `.flush()` once it really reads those blocks, assuming that the read happens delayed... I think we need to write a `Record_Component` wrapper for Dask that calls `.flush()` on `.asarray()`...

src/binding/python/openpmd_api/DaskDataFrame.py
jakirkham: Would suggest moving this to a module level function instead of a closure. While using closures should work, there's extra overhead pickling closures vs. module level functions

ax3l: Thank you for the review & continued guidance!
Fixed in #951

PR #936, opened 2021-02-26
Follow-up of frontend redesign: Remove unnecessary pointer members

franzpoeschel:
This makes the members of `SeriesData` and `AttributableData` plain members instead of shared pointers.

TODO:
- [x] Member ordering?

Also, I've noticed that we could theoretically merge `Writable` with `AttributableData`, simplifying our class structure somewhat. `Writable m_writable` is now a plain member of `AttributableData`, it is neither movable nor copyable so it's not going anywhere and it has a pointer back to the `AttributableData`, so there's no real reason to keep them separate apart from the large diff that would be..


include/openPMD/Series.hpp
```diff
@@ -52,10 +52,54 @@
...
+    //! Index of the last opened iteration
+    std::shared_ptr< SharedResources > shared;
+
+public:
+    mapped_type &
+    operator[]( key_type const & key ) override;
+    mapped_type &
+    operator[]( key_type && key ) override;
+};
+
```
franzpoeschel: Needed to move to front now since `m_writeIterations` is now a plain member of `SeriesData`. We could use a unique pointer to avoid this.

ax3l: I think a plain member is quite good.
We could move the `WriteIterations` class in its own `.hpp`/`.cpp` files, though.

test/SerialIOTest.cpp
```diff
@@ -822,6 +822,7 @@ TEST_CASE( "particle_patches", "[serial]" )
 inline
 void dtype_test( const std::string & backend )
 {
+    std::cout << "SERIES HAS SIZE " << sizeof(internal::SeriesData) << std::endl;
```
ax3l: Leftover debug message:
```suggestion
```

ax3l (commented):
This is great!

I added two minor updates inline.

If you like to merge `Writable` and `AttributableData` I am ok in a follow-up. If you like we can also discuss it briefly on Tuesday first.

include/openPMD/ReadIterations.hpp
```diff
@@ -0,0 +1,125 @@
...
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "openPMD/Iteration.hpp"
+#include "openPMD/Series.hpp"
+
+/** Writing side of the streaming API.
```
ax3l: If you add a doxygen block like this here, it will "pop down" to the next best match, which is `namespace openPMD`.

Do you like to make this a block for the file itself? In that case do this:
```suggestion
/** @file
 *
 * Writing side of the streaming API.
```

include/openPMD/WriteIterations.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "openPMD/Iteration.hpp"
+#include "openPMD/auxiliary/Option.hpp"
+#include "openPMD/backend/Container.hpp"
+
+/** Writing side of the streaming API.
```
ax3l, line 32: If you add a doxygen block like this here, it will "pop down" to the next best match, which is `namespace openPMD`.

Do you like to make this a block for the file itself? In that case do this:
```suggestion
/** @file
 *
 * Writing side of the streaming API.
```

ax3l: But somehow this looks doubled, maybe another location was planned for this block?

src/ReadIterations.cpp
```diff
@@ -0,0 +1,157 @@
...
+ReadIterations::iterator_t ReadIterations::begin()
+{
+    return iterator_t{ m_series };
+}
+
+ReadIterations::iterator_t ReadIterations::end()
+{
+    return SeriesIterator::end();
+}
+}
```
ax3l:
```suggestion
} // namespace openPMD

```

src/WriteIterations.cpp
```diff
@@ -0,0 +1,78 @@
...
+    shared->currentlyOpen = key;
+    auto & res = shared->iterations[ std::move( key ) ];
+    if( res.getStepStatus() == StepStatus::NoStep )
+    {
+        res.beginStep();
+        res.setStepStatus( StepStatus::DuringStep );
+    }
+    return res;
+}
+}
```
ax3l: EOF newlines are good for:
- compiler robustness (pre C++11, this was even required by the C++ standard)
- better `git diff`s

I am adding a namespace comment here since that makes the `}}` more readable, since it has no indentation.
```suggestion
} // namespace openPMD

```

include/openPMD/ReadIterations.hpp
franzpoeschel: Copy paste error, that's not supposed to be there. Thanks for noticing.

include/openPMD/WriteIterations.hpp
franzpoeschel: Yep, moved it now

ax3l (approved):
:tada: very nice :)

PR #938, opened 2021-03-05
lazy parsing of iterations

franzpoeschel:
Opening a Series with many iterations is currently a time-intensive procedure due to eager parsing.
This adds an option to lazily parse a Series. If that option is chosen, `Iteration::open()` must be called before accessing. `ReadIterations` will do that automatically.

I previously attempted parsing iterations automatically upon accessing them via the `Container` interface, but there were too many edge cases and it's probably better to do such things explicitly in parallel situations.

TODO:
- [x] It's probably time to use a builder pattern for the Series or something alike, the number of optional parameters keeps growing.
- [x] Python bindings
- [x] Documentation
- [x] Cleanup, ~~including removal of failed first attempt~~
- [x] Support lazy parsing for group-based layout
- [x] Merge #941 first.
- [x] This PR is affected by the issue described in #953. Worked around it in here by using `auxiliary::Option` instead of NaNs to represent no scaling.


test/SerialIOTest.cpp
```diff
@@ -59,9 +59,15 @@ write_and_read_many_iterations( std::string const & ext ) {
         }
         // ~Series intentionally not yet called
 
-        Series read(filename, Access::READ_ONLY);
-        for (auto iteration : read.iterations) {
-            // std::cout << "Reading iteration " << iteration.first << std::endl;
+        Series read = SeriesBuilder()
```
ax3l: C++ builder pattern.

include/openPMD/SeriesBuilder.hpp
```diff
@@ -0,0 +1,94 @@
...
+     */
+    SeriesBuilder & parseEagerly();
+    /**
+     * @brief Sets the parseLazily flag in the Series constructor to true.
+     */
+    SeriesBuilder & parseLazily();
+#if openPMD_HAVE_MPI
+    /**
+     * @brief As in the Series constructor. Default is to construct a
+     *        non-parallel Series.Mit
```
ax3l: Is `Series.Mit` a typo? :)

ax3l (commented):
added inline :)

include/openPMD/Iteration.hpp
```diff
@@ -153,22 +154,48 @@ class Iteration : public LegacyAttributable
 private:
     Iteration();
 
+    struct DeferredRead
```
ax3l: Let's add a doxygen string here, just to clarify what this switch does.
Would `DeferredReadAccess` be more explicit? Not sure, ... maybe.

ax3l (commented):
The implementation of the lazy/deferred parsing reads well. One general comment, I would unify the naming for this in the code base, at the moment we use:
- lazy
- deferred read
- deferred access
- (not) yet accessed

To describe the same thing, which can be confusing as the same concept in the code is meant.

Should we skip the introduction of the term "lazy" and instead use "deferred (iteration) access" in all places?

---

Do we want to make this a Series constructor parameter or JSON option?

---

Should we introduce the builder pattern in a separate PR and/or use it more consistently throughout the code base & examples?

include/openPMD/SeriesBuilder.hpp
franzpoeschel: Guess I typed something in the wrong window heh
I've removed the builder thing from this PR now anyway though

include/openPMD/Iteration.hpp
franzpoeschel: If we rename this one anyway, I'd go for `DeferredParseAccess` to avoid confusion with deferred load_chunks

franzpoeschel (commented):
> The implementation of the lazy/deferred parsing reads well. One general comment, I would unify the naming for this in the code base, at the moment we use:
> 
>     * lazy
> 
>     * deferred read
> 
>     * deferred access
> 
>     * (not) yet accessed
> 
> 
> To describe the same thing, which can be confusing as the same concept in the code is meant.
> 
> Should we skip the introduction of the term "lazy" and instead use "deferred (iteration) access" in all places?

I've renamed things now:
```
parse_lazily -> defer_iteration_parsing
NotYetAccessed -> ParseAccessDeferred
Iteration::deferRead -> Iteration::deferParseAccess
DeferredRead -> DeferredParseAccess
m_deferredRead -> m_deferredParseAccess
parseLazily -> runDeferredParseAccess
```
> 
> Do we want to make this a Series constructor parameter or JSON option?

Good idea, done
> 
> Should we introduce the builder pattern in a separate PR and/or use it more consistently throughout the code base & examples?

I introduced the builder pattern since it made for a better way to have a constructor with more than one defaulted argument. Since we're now back at one defaulted argument, we don't necessarily need it anymore. I've isolated the builder pattern to a branch `franzpoeschel/topic-builder` now, should I do a PR?

Also, we can definitely use it more in the examples and tests. In the code base itself probably not so much since we seldomly create a `Series` object ourselves.

docs/source/details/backendconfig.rst
```diff
@@ -33,6 +33,18 @@ For a consistent user interface, backends shall follow the following rules:
...
 
+Backend-independent JSON configuration
+--------------------------------------
+
+The key ``defer_iteration_parsing`` can be used to optimize the process of opening an openPMD Series.
+By default, a Series is parsed eagerly, i.e. opening a Series implies reading all available iterations.
+Especially when a Series has many iterations, this can be a costly operation and users may wish to defer parsing of iterations to a later point adding ``{"defer_iteration_parsing": true}`` to their JSON configuration.
+
+When parsing non-eagerly, each iteration needs to be explicitly opened with ``Iteration::open()`` before accessing.
+(Notice that Iteration::open() is generally recommended to be used in parallel contexts to avoid parallel file accessing hazards).
```
```suggestion
(Notice that ``Iteration::open()`` is generally recommended to be used in parallel contexts to avoid parallel file accessing hazards).
```

docs/source/details/backendconfig.rst
```diff
@@ -33,6 +33,18 @@ For a consistent user interface, backends shall follow the following rules:
...
+Using the Streaming API (i.e. SeriesImpl::readIteration()) will do this automatically.
```
```suggestion
Using the Streaming API (i.e. ``SeriesImpl::readIteration()``) will do this automatically.
```

docs/source/details/backendconfig.rst
```diff
@@ -33,6 +33,18 @@ For a consistent user interface, backends shall follow the following rules:
...
+Parsing eagerly might be very expensive for a Series with many iterations, but will avoid bugs by forgotten calls to ``Iteration::open()``.
+
```
ax3l, line 48: Let's add this if true (we should also add CI for this if true):
```suggestion
In complex environments, calling ``Iteration::open()`` on an already open environment does no harm (and does not incur additional runtime cost for additional ``open()`` calls).

```

include/openPMD/Series.hpp
```diff
@@ -328,6 +331,9 @@ class SeriesImpl : public AttributableImpl
     }
 
     std::unique_ptr< ParsedInput > parseInput(std::string);
+    // use a template in order not to expose nlohmann_json to users
+    template< typename JSON >
+    void parseJsonOptions( JSON const & );
```
ax3l: Since you only call this in the `.cpp` file, you can define this helper function only in the `Series.cpp` file and in a `namespace {}` (as you already do) and avoid that it gets exposed as a function and symbol at all :)

test/python/unittest/API/APITest.py
```diff
@@ -1634,7 +1635,7 @@ def makeIteratorRoundTrip(self, backend, file_ending):
         read = io.Series(
             "../samples/unittest_serialIterator." + file_ending,
             io.Access_Type.read_only,
-            jsonConfig
+            jsonConfig,
```
ax3l:
```suggestion
            jsonConfig
```

test/python/unittest/API/APITest.py
```diff
@@ -1682,9 +1683,11 @@ def makeAvailableChunksRoundTrip(self, ext):
 
         read = io.Series(
             name,
-            io.Access_Type.read_only
+            io.Access_Type.read_only,
+            options="{\"defer_iteration_parsing\": true}"
```
ax3l: Probably not worth here, but for all larger JSON options in Python (e.g. `jsonConfig` above) you can create a pretty Python dictionary and make it JSON like this:
```py
import json

options = {"defer_iteration_parsing": True}
json_str = json.dumps(options)
```

e.g.
```suggestion
            options=json.dumps({"defer_iteration_parsing": True})
```

Simple for here:
```suggestion
            options='{"defer_iteration_parsing": true}'
```

src/Series.cpp
```diff
@@ -49,6 +50,20 @@ namespace
      */
     std::string cleanFilename(std::string const &filename, Format f);
 
+    struct Match
```
ax3l, line 54:
```suggestion
    /** Compound return type for regex matching of filenames */
    struct Match
```

src/Series.cpp
```diff
@@ -49,6 +50,20 @@ namespace
...
+    {
+        bool isContained{};
+        int padding{};
+        uint64_t iteration{};
+
+        // support for std::tie
+        operator std::tuple< bool &, int &, uint64_t & >()
```
ax3l, line 61: Clever :)

src/Series.cpp
```diff
@@ -49,6 +50,20 @@ namespace
      */
     std::string cleanFilename(std::string const &filename, Format f);
 
+    struct Match
+    {
+        bool isContained{};
+        int padding{};
+        uint64_t iteration{};
```
ax3l:
```suggestion
        bool isContained{}; //! pattern match successful
        int padding{}; //! number of zeros used for padding of iteration
        uint64_t iteration{}; //! iteration found in regex pattern (default: 0)
```
To avoid confusion, one could make iteration an `int64_t` and return `-1` by default (or so). But we already use `uint64_t` everywhere, so that's fine, too.

include/openPMD/Series.hpp
franzpoeschel: It's not that easy because that function accesses private members. But we can theoretically make all members of `SeriesData` public and inherit it privately in `SeriesInternal`. So, the data struct can be passed to non-friend functions for manipulation?
Edit: Pushed a commit that does just that.
Edit: Not so easy either, this [breaks dynamic casting](https://stackoverflow.com/questions/62430684/why-c-requires-public-inheritance-ignoring-friend-declarations-to-make-dynam). I've made it a public inheritance now, meaning that `SeriesInternal` has all members public now. `Series` still hides them.

test/python/unittest/API/APITest.py
franzpoeschel: Ah, good to know. I'll leave it at this for now, then.

src/Series.cpp
franzpoeschel: If we want to do this, I think the cleanest way would be by combining `isContained` and `iteration` into an `auxiliary::Option<uint64_t>`. Should we use something like this right away?

ax3l (approved):
Thank you, this is great!

PR #941, opened 2021-03-08
Use schema-based versioning in ADIOS2 backend

franzpoeschel:
This steals and isolates some commits of #855:
* Use YYYYMMDD format for the schema used in ADIOS2
* Automatically recognize the right schema when reading

TODO
- [x] Remove some unnecessary changes that are only relevant to #855
- [x] Documentation
- [x] Remove the `new_attribute_layout` from some leftover places
- [x] Merge #927 first
- [x] Merge #943 first
- [x] This PR also fixes two bugs that I encountered while working on it.  Should I separate those into another PR?
    1. Read-Write mode should inspect whether the file already exists and chose read-only or write-only mode accordingly. Properly supporting ADIOS Append mode is a Todo for the future though.
    2. Parallel IO tests check the wrong file name for existence.
- [x] Rename `__is_boolean__/(.*)` to `__openPMD_internal/is_boolean/\1` in new schema


ax3l:
@franzpoeschel `This PR also fixes two bugs that I encountered while working on it. Should I separate those into another PR?`

Yes, that would be great in case I need to patch & cherry-pick them around! Thanks!

franzpoeschel:
Ready for review (if CI goes green) :) @ax3l

ax3l (approved):
:tada:

PR #942, opened 2021-03-08
Fix available_chunks() for  constant components

franzpoeschel:
Bug found [here](https://github.com/openPMD/openPMD-api/pull/935#discussion_r583265666)


ax3l (approved):
Thank you! :rocket: :sparkles:

PR #943, opened 2021-03-11
Decay ReadWrite mode into adios2::Mode::Read if the file exists

franzpoeschel:
Bug detected while working on #855.
EDIT: Alright, found out why this bug was triggered by #855, but not previously. Until now, the ADIOS2 backend delayed the opening of the .bp file until an actual access happened. So [this test](https://github.com/openPMD/openPMD-api/blob/09cb592ca611c9a2361c6ee80e5d320242971ea8/test/SerialIOTest.cpp#L1529) never actually created the file for iteration 6, bypassing this bug. Now, with eager opening, this bug is exposed.

TODO:
- [x] Write test that exposes this
- [x] ~~The test exposed the same issue in HDF5. Fix that in this PR too?~~
    That was a different issue. I wrote a little mesh dataset to that iteration in the test originally. Now, that iteration had a defined `/meshesPath` attribute. While reading the file back in some lines further down, the openPMD API expected to find meshes in the other iterations, too, didn't find any and crashed. Is our workflow with the `meshesPath` really that good right now?

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -904,7 +904,8 @@ ADIOS2IOHandlerImpl::availableChunks(
         datatype, rbi, parameters, ba.m_IO, engine, varName );
 }
 
-adios2::Mode ADIOS2IOHandlerImpl::adios2AccessMode( )
+adios2::Mode
+ADIOS2IOHandlerImpl::adios2AccessMode( std::string const & fullPath )
 {
     switch ( m_handler->m_backendAccess )
     {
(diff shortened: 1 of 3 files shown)


test/SerialIOTest.cpp
```diff
@@ -1396,6 +1396,13 @@ void fileBased_write_test(const std::string & backend)
         REQUIRE(o.iterations.size() == 5);
         o.iterations[6];
         REQUIRE(o.iterations.size() == 6);
+        // write something to trigger opening of the file
+        // write something to trigger opening of the file
```
ax3l: Looks a bit doubly :)
```suggestion
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -913,10 +914,21 @@ adios2::Mode ADIOS2IOHandlerImpl::adios2AccessMode( )
...
+                         "(Append) mode. "
+                      << "Replacing with Read mode." << std::endl;
+            return adios2::Mode::Read;
+        }
+        else
+        {
+            std::cerr << "ADIOS2 does currently not yet implement ReadWrite "
+                         "(Append) mode. "
+                      << "Replacing with Write mode." << std::endl;
+            return adios2::Mode::Write;
```
ax3l, line 930: Is this safe to keep going instead of asking for another mode?

franzpoeschel: Choosing read mode is safe anyway and write mode is only chosen if the file does not exist, so it should be safe, too. I can imagine that there might be issues in parallel, I should test how ADIOS2 behaves for that and potentially add an MPI barrier.

franzpoeschel: According to some quick experiments I did, ADIOS does not create files on the disk if not all of the parallel writers have called `Engine::open()` yet. So we could leave it as it is or alternatively add an MPI Barrier for some extra safety.

ax3l (approved):

PR #945, opened 2021-03-12
Particle Patches: Do not emplace patch records if they don't exist in the file being read

franzpoeschel:
The `GenerationPolicy` for `ParticleSpecies` automatically fills the `particlePatches` container with the `numParticles` and `numParticlesOffset` components, meaning that a reading code has no way to know whether those records actually exist. This resulted in the `openpmd-pipe` tool failing to convert a dataset containing a particle species without patches (with extremely cryptic error messages from somewhere deep in the backend). This PR adds (1) a failing test and (2) a fix.


ax3l (approved):
Great catch and coverage, thank you! :rocket:

PR #946, opened 2021-03-16
Document the meaning of Writable::written()

franzpoeschel:
I think I finally figured out what that one means.


ax3l (approved):
Yay!! Thank you so much! :tada:

PR #947, opened 2021-03-18
Reintroduce Clang Sanitizer

franzpoeschel:
Re-add ASAN and UBSAN tests to CI.

### DIFF
--- a/.github/workflows/unix.yml
+++ b/.github/workflows/unix.yml
@@ -28,6 +28,37 @@ jobs:
         cat clang-tidy.log
         if [[ $(wc -m <clang-tidy.log) -gt 1 ]]; then exit 1; fi
 
+  clangsanitizer10_nopy_ompi_h5_ad1_ad2:
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v2
+    - name: Spack Cache
+      uses: actions/cache@v2
+      with: {path: /opt/spack, key: clang10_nopy_ompi_h5_ad1_ad2 }
+    - name: Install
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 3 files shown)


ax3l:
Since the tests are relatively long-running, let's keep the clang-tidy & sanitizer matrix entries separate

franzpoeschel:
Note: I noticed that the `clang10_py38_nompi_h5_ad1_ad2_libcpp` test does not build any ADIOS backends
EDIT: The added sanitizer test now does exactly what that test promises to do, but additionally with sanitization. So, I'd say we can just keep that one as a no-Adios-test and rename it accordingly?

.github/ci/spack-envs/clang10_nopy_ompi_h5_ad1_ad2_/spack.yaml
```diff
@@ -0,0 +1,65 @@
+# This is a Spack environment file.
```
ax3l: It looks like the directory name here ends with an `_`, which looks like a typo

.github/workflows/unix.yml
```diff
@@ -20,14 +20,44 @@ jobs:
       run: |
         eval $(spack env activate --sh .github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad1_ad2/)
         spack install
```
ax3l: Just reverting the white space change:
```suggestion
        spack install

```

ax3l (approved):
Wuhu, thanks for re-introducing this!

Safety first! :)

ax3l:
Wow, this still runs for an hour. But no worries, we will think about optimizing this if we need to.

Issue #15, opened 2018-01-15
Writing and Reading Non-openPMD Infos


ax3l:
It would be necessary to allow storing non-openPMD information as well.

In the simple case, this means adding additional, user-defined attributes to records, record-components and paths.

In the more complex case, application specific states such as RNG states, unqiue-number-generator states, etc. need to be stored as records outside of the openPMD-interpreted paths.

anokfireball:
The simple case is already possible, maybe not made explicit enough yet. Every "position" in an openPMD hierearchy (which in this APIs terms is an ```Attributable```) supports a templated operation [```setAttribute```](https://github.com/openPMD/openPMD-api/blob/0.9.0-alpha/include/openPMD/backend/Attributable.hpp#L91-L105), taking a string and a generic argument whose type has to map to a supported [```Datatype```](https://github.com/openPMD/openPMD-api/blob/0.9.0-alpha/include/openPMD/Datatype.hpp#L36-L66)

[Here](https://openpmd-api.readthedocs.io/en/0.9.0-alpha/usage/firstwrite.html#attributes)'s an example demonstrating said feature by annotating the root group with a std::string.

ax3l:
That's already a great start, thx! :)

The latter case is quite relevant for PIConGPU checkpoints.

ax3l:
cc @ejcjason you also need this, right?

JunCEEE:
> cc @ejcjason you also need this, right?

I cannot access the link "here", but I guess the suggested solution is to change the attribute of "position" under one particle group into something we want. That's good. But what if we want to add one group, say "observable", in the iteration group (i.e., in the same level of particles)? 
- data/
    - 1/
        - observable/
        - particles/
            - Cu

It seems that the above solution doesn't work in this case, because there is no command like `i.meshes[xx]` or `i.particles[xx]` to create the `observable`. Maybe it can be achieved if we have something to change the group name of `particles` or `meshes`?

ax3l:
Links updated with permanently accessible examples. `setAttribute` examples is now also [in the manual](https://openpmd-api.readthedocs.io/en/0.9.0-alpha/usage/firstwrite.html#attributes).

Yes, what you want is to side-channel to the low-level write API to write groups and datasets outside of openPMD. That's also what I meant.

> Maybe it can be achieved if we have something to change the group name of particles or meshes?

[...]

Issue #12, opened 2017-10-03
python API: Dask Compatible


ax3l:
Idea for  the future:

When we read data in the python API, we should make sure this is compatible and possible to be distributed for compute tasks inside [dask](http://dask.pydata.org). See the hdf5 read example within. (Similar to [pyDive](https://github.com/ComputationalRadiationPhysics/pyDive), but more powerful.)

ax3l:
Part of 0.14.0+ :tada: 

X-ref: #952 #958 #959 #951 #963 et al. :)

Issue #11, opened 2017-07-03
Parallel: Offset Namings


ax3l:
@C0nsultant namings parameters of offsets of MPI-ranks in parallel writes can be tricky.

Ideally, stick as close as possible to the openPMD namings in parameters.
If anything is missing, [here is how we call offsets in PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu/wiki/PIConGPU-domain-definitions) to fill the rest. (The old libSplash namings are outdated and inconsistent, don't read those please.)

anokfireball:
I am not sure where in the code this would apply. The only place where the term ```Offset``` is used (but *not* enforced by openPMD) is the alias of ```std::vector``` used during dataset IO:

https://github.com/openPMD/openPMD-api/blob/b8e8de9187e7cc179589c974d042de9cf3ae92b4/include/openPMD/Dataset.hpp#L34

Across the whole codebase, that ```Offset``` is used to signal the offset in the overall dataset on persistent storage (i.e. the left edge of a chunk inside a dataset). It is always referred to simply as ```Offset offset``` or ```Offset o```.
No concepts from the parallel domain are incorporated at the moment. Every position (i.e. every ```Offset```) is treated as global.

PR #949, opened 2021-03-23
Mapping between ADIOS steps and openPMD iterations

franzpoeschel:
Background: Until now, our Streaming API [assumes](https://github.com/openPMD/openPMD-api/blob/15a98c7605f353fae4fa7e8f2209de650e3f8f5a/src/ReadIterations.cpp#L94) that each ADIOS step corresponds with exactly one openPMD iteration and that those iterations are found in ascending order. Once we expose the ADIOS2 Append access mode, this will not necessarily hold true any longer, so this PR explores more flexible alternatives.

Scenario: Run a simulation with data output all 50 steps, checkpoints all 500 steps, use step-based iteration layout (or group-based iteration layout and activate ADIOS steps). Crash at step 750, restart from 500. Data output then needs to be appended to the (single file!) `output.bp`. From the first run, we have the following:
> 0 50 100 150 200 250 300 350 400 450 500 550 600 650 700 750 

When appending, we *cannot* remove any old steps, just append new ones. So, our file will look like:
> 0 50 100 150 200 250 300 350 400 450 500 550 600 650 700 750  500 550 600 650 700 …

Goal: Be able to read that. 

First step (useful independent of this issue): Annotate for each ADIOS step the openPMD iteration defined by it.

My current approach is to use the ~~`/data/__step__`~~ `/data/snapshot` attribute introduced by #855 and use it to store the openPMD iteration(s) stored in the current ADIOS step. Afterwards, the reading procedures can inquire that attribute and see which iteration they should return to the user. Fallback to the old solution if the attribute isn't found.

TODO:
- [ ] reading: as long as we have no ADIOS append-and-truncate for writes, we need to pick the "last" key in `/data/snapshot` when reading `series.iteration[key]`
    fixing this is for a follow-up PR
- [x] Testing, documentation, cleanup, edge cases
- [x] Merge #1007 first
- [x] Add some further iterations to appending test
- [x] Fix todos (see in-line comments)
- [x] Check: snapshot attribute in RW mode (don't set for existing iterations)
- [x] update standard: WIP in https://github.com/ax3l/openPMD-standard/pull/1, https://github.com/openPMD/openPMD-standard/pull/250:
- [x] See latest bugfix in topic-read-leniently
- [x] merge #1218 first 
- [x] merge #1302 first
- [x] see https://github.com/openPMD/openPMD-standard/pull/250/files#r949996254, https://github.com/openPMD/openPMD-standard/pull/250/files#r949999743

### DIFF
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -121,6 +121,28 @@ namespace internal
     FlushParams const defaultFlushParams{};
 
     struct ParsedFlushParams;
+
+    /**
+     * Some parts of the openPMD object model are read-only when accessing
+     * a Series in Access::READ_ONLY mode, notably Containers and Attributes.
+     * They are filled at parse time and not modified afterwards.
+     * Such state-changing operations are hence allowed under either of two
+     * conditions:
+     * 1) The Series is opened in an open mode that allows writing in any way.
+     *    (Currently any but Access::READ_ONLY).
+     * 2) The Series is in Parsing state. This way, modifying the open mode
+     *    during parsing can be avoided.
+     */
+    enum class SeriesStatus : unsigned char
+    {
+        Default, ///< Mutability of objects in the openPMD object model is
+                 ///< determined by the open mode (Access enum), normal state in
+                 ///< which the user interacts with the Series.
+        Parsing ///< All objects in the openPMD object model are temporarily
+                ///< mutable to allow inserting newly-parsed data.
+                ///< Special state only active while internal routines are
+                ///< running.
+    };
 } // namespace internal
 
 /** Interface for communicating between logical and physically persistent data.
@@ -192,6 +214,7 @@ public:
     // why do these need to be separate?
     Access const m_backendAccess;
     Access const m_frontendAccess;
+    internal::SeriesStatus m_seriesStatus = internal::SeriesStatus::Default;
     std::queue<IOTask> m_work;
 }; // AbstractIOHandler
 
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -537,6 +537,7 @@ struct OPENPMDAPI_EXPORT Parameter<Operation::WRITE_ATT>
         : AbstractParameter()
         , name(p.name)
         , dtype(p.dtype)
+        , changesOverSteps(p.changesOverSteps)
         , resource(p.resource)
     {}
 
@@ -548,6 +549,13 @@ struct OPENPMDAPI_EXPORT Parameter<Operation::WRITE_ATT>
 
     std::string name = "";
     Datatype dtype = Datatype::UNDEFINED;
+    /*
+     * If true, this attribute changes across IO steps.
+     * It should only be written in backends that support IO steps,
+     * otherwise writing should be skipped.
+     * The frontend is responsible for handling both situations.
+     */
+    bool changesOverSteps = false;
     Attribute::resource resource;
 };
 
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -28,7 +28,10 @@
 #include "openPMD/backend/Attributable.hpp"
 #include "openPMD/backend/Container.hpp"
 
+#include <cstdint>
+#include <deque>
 #include <optional>
+#include <tuple>
 
 namespace openPMD
 {
@@ -283,13 +286,56 @@ private:
     void read_impl(std::string const &groupPath);
 
     /**
+     * Status after beginning an IO step. Currently includes:
+     * * The advance status (OK, OVER, RANDOMACCESS)
+     * * The opened iterations, in case the snapshot attribute is found
+     */
+    struct BeginStepStatus
+    {
+        using AvailableIterations_t = std::optional<std::deque<uint64_t> >;
+
+        AdvanceStatus stepStatus{};
+        /*
+         * If the iteration attribute `snapshot` is present, the value of that
+         * attribute. Otherwise empty.
+         */
+        AvailableIterations_t iterationsInOpenedStep;
+
+        /*
+         * Most of the time, the AdvanceStatus part of this struct is what we
+         * need, so let's make it easy to access.
+         */
+        inline operator AdvanceStatus() const
+        {
+            return stepStatus;
+        }
+
+        /*
+         * Support for std::tie()
+         */
+        inline operator std::tuple<AdvanceStatus &, AvailableIterations_t &>()
+        {
+            return std::tuple<AdvanceStatus &, AvailableIterations_t &>{
+                stepStatus, iterationsInOpenedStep};
+        }
+    };
+
+    /**
      * @brief Begin an IO step on the IO file (or file-like object)
      *        containing this iteration. In case of group-based iteration
      *        layout, this will be the complete Series.
      *
-     * @return AdvanceStatus
+     * @return BeginStepStatus
      */
-    AdvanceStatus beginStep(bool reread);
+    BeginStepStatus beginStep(bool reread);
+
+    /*
+     * Iteration-independent variant for beginStep().
+     * Useful in group-based iteration encoding where the Iteration will only
+     * be known after opening the step.
+     */
+    static BeginStepStatus
+    beginStep(std::optional<Iteration> thisObject, Series &series, bool reread);
 
     /**
      * @brief End an IO step on the IO file (or file-like object)
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -38,6 +38,7 @@
 #endif
 
 #include <cstdint>
+#include <deque>
 #include <map>
 #include <optional>
 #include <set>
@@ -584,8 +585,10 @@ OPENPMD_private
      * Note on re-parsing of a Series:
      * If init == false, the parsing process will seek for new
      * Iterations/Records/Record Components etc.
+     * If series.iterations contains the attribute `snapshot`, returns its
+     * value.
      */
-    void readGorVBased(bool init = true);
+    std::optional<std::deque<uint64_t> > readGorVBased(bool init = true);
     void readBase();
     std::string iterationFilename(uint64_t i);
 
@@ -636,6 +639,8 @@ OPENPMD_private
         iterations_iterator it,
         Iteration &iteration);
 
+    AdvanceStatus advance(AdvanceMode mode);
+
     /**
      * @brief Called at the end of an IO step to store the iterations defined
      *        in the IO step to the snapshot attribute.
@@ -643,6 +648,12 @@ OPENPMD_private
      * @param doFlush If true, flush the IO handler.
      */
     void flushStep(bool doFlush);
+
+    /*
+     * Returns the current content of the /data/snapshot attribute.
+     * (We could also add this to the public API some time)
+     */
+    std::optional<std::vector<uint64_t> > currentSnapshot() const;
 }; // Series
 } // namespace openPMD
 
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -37,8 +37,10 @@
 #include <mpi.h>
 #endif
 
+#include <cstdint>
 #include <map>
 #include <optional>
+#include <set>
 #include <string>
 
 // expose private and protected members for invasive testing
@@ -82,6 +84,12 @@ namespace internal
          * the same instance.
          */
         std::optional<WriteIterations> m_writeIterations;
+        /**
+         * For writing: Remember which iterations have been written in the
+         * currently active output step. Use this later when writing the
+         * snapshot attribute.
+         */
+        std::set<uint64_t> m_currentlyActiveIterations;
         /**
          * Needed if reading a single iteration of a file-based series.
          * Users may specify the concrete filename of one iteration instead of
@@ -627,6 +635,14 @@ OPENPMD_private
         internal::AttributableData &file,
         iterations_iterator it,
         Iteration &iteration);
+
+    /**
+     * @brief Called at the end of an IO step to store the iterations defined
+     *        in the IO step to the snapshot attribute.
+     *
+     * @param doFlush If true, flush the IO handler.
+     */
+    void flushStep(bool doFlush);
 }; // Series
 } // namespace openPMD
 
--- a/include/openPMD/Streaming.hpp
+++ b/include/openPMD/Streaming.hpp
@@ -19,8 +19,9 @@ namespace openPMD
  */
 enum class AdvanceStatus : unsigned char
 {
-    OK, /* stream goes on */
-    OVER /* stream is over */
+    OK, ///< stream goes on
+    OVER, ///< stream is over
+    RANDOMACCESS ///< there is no stream, it will never be over
 };
 
 /**
--- a/src/Iteration.cpp
+++ b/src/Iteration.cpp
@@ -573,49 +573,92 @@ void Iteration::read_impl(std::string const &groupPath)
     readAttributes(ReadMode::FullyReread);
 }
 
-AdvanceStatus Iteration::beginStep(bool reread)
+auto Iteration::beginStep(bool reread) -> BeginStepStatus
 {
-    using IE = IterationEncoding;
+    BeginStepStatus res;
     auto series = retrieveSeries();
+    return beginStep({*this}, series, reread);
+}
+
+auto Iteration::beginStep(
+    std::optional<Iteration> thisObject, Series &series, bool reread)
+    -> BeginStepStatus
+{
+    BeginStepStatus res;
+    using IE = IterationEncoding;
     // Initialize file with this to quiet warnings
     // The following switch is comprehensive
     internal::AttributableData *file = nullptr;
     switch (series.iterationEncoding())
     {
     case IE::fileBased:
-        file = &Attributable::get();
+        if (thisObject.has_value())
+        {
+            file = &static_cast<Attributable &>(*thisObject).get();
+        }
+        else
+        {
+            throw error::Internal(
+                "Advancing a step in file-based iteration encoding is "
+                "iteration-specific.");
+        }
         break;
     case IE::groupBased:
     case IE::variableBased:
         file = &series.get();
         break;
     }
-    AdvanceStatus status = series.advance(
-        AdvanceMode::BEGINSTEP, *file, series.indexOf(*this), *this);
-    if (status != AdvanceStatus::OK)
+
+    AdvanceStatus status;
+    if (thisObject.has_value())
     {
-        return status;
+        status = series.advance(
+            AdvanceMode::BEGINSTEP,
+            *file,
+            series.indexOf(*thisObject),
+            *thisObject);
+    }
+    else
+    {
+        status = series.advance(AdvanceMode::BEGINSTEP);
+    }
+
+    switch (status)
+    {
+    case AdvanceStatus::OVER:
+        res.stepStatus = status;
+        return res;
+    case AdvanceStatus::OK:
+    case AdvanceStatus::RANDOMACCESS:
+        break;
     }
 
     // re-read -> new datasets might be available
-    if (reread &&
+    auto IOHandl = series.IOHandler();
+    if (reread && status != AdvanceStatus::RANDOMACCESS &&
         (series.iterationEncoding() == IE::groupBased ||
          series.iterationEncoding() == IE::variableBased) &&
-        (this->IOHandler()->m_frontendAccess == Access::READ_ONLY ||
-         this->IOHandler()->m_frontendAccess == Access::READ_WRITE))
+        (IOHandl->m_frontendAccess == Access::READ_ONLY ||
+         IOHandl->m_frontendAccess == Access::READ_WRITE))
     {
-        switch (IOHandler()->m_frontendAccess)
+        switch (IOHandl->m_frontendAccess)
         {
         case Access::READ_ONLY:
         case Access::READ_WRITE: {
             bool previous = series.iterations.written();
             series.iterations.written() = false;
-            auto oldType = this->IOHandler()->m_frontendAccess;
-            auto newType =
-                const_cast<Access *>(&this->IOHandler()->m_frontendAccess);
-            *newType = Access::READ_WRITE;
-            series.readGorVBased(false);
-            *newType = oldType;
+            auto oldStatus = IOHandl->m_seriesStatus;
+            IOHandl->m_seriesStatus = internal::SeriesStatus::Parsing;
+            try
+            {
+                res.iterationsInOpenedStep = series.readGorVBased(false);
+            }
+            catch (...)
+            {
+                IOHandl->m_seriesStatus = oldStatus;
+                throw;
+            }
+            IOHandl->m_seriesStatus = oldStatus;
             series.iterations.written() = previous;
             break;
         }
@@ -626,7 +669,8 @@ AdvanceStatus Iteration::beginStep(bool reread)
         }
     }
 
-    return status;
+    res.stepStatus = status;
+    return res;
 }
 
 void Iteration::endStep()
@@ -648,6 +692,7 @@ void Iteration::endStep()
     }
     // @todo filebased check
     series.advance(AdvanceMode::ENDSTEP, *file, series.indexOf(*this), *this);
+    series.get().m_currentlyActiveIterations.clear();
 }
 
 StepStatus Iteration::getStepStatus()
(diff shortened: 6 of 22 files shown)


guj:
For the group/variable based files, Is there an option to not write an iteration if it already exists and is valid?

franzpoeschel:
> For the group/variable based files, Is there an option to not write an iteration if it already exists and is valid?

Do you mean when appending to an existing Series? If yes, that's a bit challenging, as ADIOS Append mode does not give any read access and openPMD has no handling for redundantly defined iterations yet.
So short answer: No, but I want to look at that specific situation again once #1007 and this PR are merged

guj:
> > For the group/variable based files, Is there an option to not write an iteration if it already exists and is valid?
> 
> Do you mean when appending to an existing Series? If yes, that's a bit challenging, as ADIOS Append mode does not give any read access and openPMD has no handling for redundantly defined iterations yet. So short answer: No, but I want to look at that specific situation again once #1007 and this PR are merged

Oh, I was referring to in your example, restarting at check point 500, which is a few steps before the latest iteration 750.   Guess it is a bit of a work  to  read contents  from the existing file first  and then ask adios to append at the right place.  

Maybe in the future an alternative is to  consider to have one file per checkpoint.  This way there is no need to append.  Always start a new file at checkpoint.

franzpoeschel:
> Oh, I was referring to in your example, restarting at check point 500, which is a few steps before the latest iteration 750. Guess it is a bit of a work to read contents from the existing file first and then ask adios to append at the right place.

In that case, it would be better to overwrite the old data with the new one
For that, we will either have to truncate the old file at write time, or to skip redundant iterations at read time. Either option is not entirely trivial and might require further support from Adios

> 
> Maybe in the future an alternative is to consider to have one file per checkpoint. This way there is no need to append. Always start a new file at checkpoint.

That would still give you redundantly defined iterations which need to be handled at read time somehow, while adding the additional complexity of needing to handle several files. I'm not sure there would be any benefit to that approach?

guj:
> That would still give you redundantly defined iterations which need to be handled at read time somehow, while adding the additional complexity of needing to handle several files. I'm not sure there would be any benefit to that approach?

If every checkpoint has its own file, there is no append needed. At restart always overwrites, so we shall not see redundant iterations.   e.g.  in your example,  file0_500.bp, file550_1000.bp, etc.  If crash happened at step 750, restart at step 500. and rewrite file550_1000.bp. 

Yes it needs to add new support to read this set of files.  

ADIOS is not likely to support remove/update functions as far as I can see.  Just my two cents to work around it.

franzpoeschel:
> > That would still give you redundantly defined iterations which need to be handled at read time somehow, while adding the additional complexity of needing to handle several files. I'm not sure there would be any benefit to that approach?
> 
> If every checkpoint has its own file, there is no append needed. At restart always overwrites, so we shall not see redundant iterations. e.g. in your example, file0_500.bp, file550_1000.bp, etc. If crash happened at step 750, restart at step 500. and rewrite file550_1000.bp.

This is *not* about checkpoints. It's about what happens to regular data output when restarting from a checkpoint. Checkpoints already usually work the way you describe and there's no reason to change that. 

But when restarting from a checkpoint, you get an "overlap zone" where output steps are written a second time. This is tricky to handle in group/variable-based iteration encodings. This PR is a first step toward solving that, though it is not yet a solution.
> 
> Yes it needs to add new support to read this set of files.
> 
> ADIOS is not likely to support remove/update functions as far as I can see. Just my two cents to work around it.

Norbert did suggest a truncate option for appending once. Alternatively, we can eliminate duplicate iterations on our own at read time.

franzpoeschel:
Note: ~~Please don't review this just yet, I'm currently rebasing this upon #1007 because this needs additional logic to deal with files created via Append mode.~~

Ok, things are fixed now, the other PR should still go first anyway

src/Series.cpp
```diff
@@ -701,11 +701,16 @@ void Series::flushFileBased(
...
             {
                 using IO = IterationOpened;
+            case IO::RemainsClosed:
+                // we might need to proceed further if the close status is
+                // ClosedInFrontend
+                // hence no continue here
+                // otherwise, we might forget to close files physically
+                break;
             case IO::HasBeenOpened:
+                // continue below
```
franzpoeschel, line 748: todo: add iteration to `m_currentlyActiveIterations` here too
8cb746f8 Writing changes: Write current step(s) to snapshot attribute

franzpoeschel:
Note: [BP5](https://adios2.readthedocs.io/en/latest/engines/engines.html#bp5) now has a feature for truncation upon appending.
It should already be possible to manually use this e.g. by specifying `{"adios": {"engine": {"parameters": {"AppendAfterSteps": "-3"}}}}`. 

Possible improvement for a follow-up PR: Handle datasets with duplicate iterations more gracefully, e.g. by skipping any but the last instance of the iteration (currently: any but the first is skipped)

src/Series.cpp
franzpoeschel: No idea what I meant here, this is the read mode, there's no need for this in read mode

test/python/unittest/API/APITest.py
```diff
@@ -1066,8 +1066,9 @@ def testListSeries(self):
         series = self.__series
         self.assertRaises(TypeError, io.list_series)
         io.list_series(series)
-        io.list_series(series, False)
-        io.list_series(series, True)
+        # @todo make list_series callable repeatedly
```
ax3l, line 1070: fixme? :)

franzpoeschel: To be fair, I'm not quite sure why this even worked so far. We are using the streaming API in `list_series()`, and once the `Series` is drained in streaming mode, then it's over. This todo is more for a future addition, either:

1. Extend `list_series()` to allow switching between streaming and random-access mode
2. Allow replaying a Series in the streaming API if the backend allows it

Going with solution 1. would be easier, but I would still do it in a separate PR since it changes the usage of `openpmd-ls`.

franzpoeschel:
Note: This PR solves two forms of dataset layouts:

1. Datasets in which iterations are written in non-chronological order (e.g. iterations are written in order 1,2,4,3,5)
2.  With the Append mode, it's possible in ADIOS2 to create datasets that contain the same iteration twice in different steps. 

In ADIOS2, such a dataset can only be successfully read if the `snapshot` attribute is defined:

* If the snapshot attribute is defined: openPMD knows which iteration is contained in the current step and can give it to the user
* If the snapshot attribute is not defined:
    * Backends other than ADIOS2 allow random-accessing their data and we just iterate through the iterations in rising order
    * In ADIOS2, the dataset is read step by step, but openPMD does not know which step has which iteration. As a fallback, we assume that the iterations are in rising order, creating a mismatch in such datasets.

Opening such a Series anyway might look somewhat like this (using the error-recovery abilities of #1237):
```
openpmd-ls ../samples/append_groupbased.bp                                                                                                                
openPMD series: append_groupbased                                                                                                                                      
openPMD standard: 1.1.0                                                                                                                                                
openPMD extensions: 0                                                                                                                                                  
                                                                                                                                                                       
data author: unknown                                                                                                                                                   
data created: 2022-05-17 12:20:15 +0000                                                                                                                                
data backend: ADIOS2                                                                                                                                                   
generating machine: unknown                                                                                                                                            
generating software: openPMD-api (version: 0.15.0-dev)                                                                                                                 
generating software dependencies: unknown                                                                                                                              
                                                                                                                                                                       
number of iterations: 8 (groupBased)                                                                                                                                   
  all iterations: 0 1 2 3 4 [ADIOS2] Warning: Attribute with name /data/7/meshes/E/x/value has no type in backend.                                                     
Cannot read record component 'x' and will skip it due to read error:                                                                                                   
Read Error in backend ADIOS2                                                                                                                                           
Object type:    Attribute                                                                                                                                              
Error type:     NotFound                                                                                                                                               
Further description:    /data/7/meshes/E/x/value
[ADIOS2] Warning: Attribute with name /data/7/meshes/E/x/value has no type in backend.
Cannot read record component 'x' and will skip it due to read error:
Read Error in backend ADIOS2
Object type:    Attribute
Error type:     NotFound
Further description:    /data/7/meshes/E/x/value
7 10 [ADIOS2] Warning: Attribute with name /data/11/meshes/E/x/value has no type in backend.
Cannot read record component 'x' and will skip it due to read error:
Read Error in backend ADIOS2
Object type:    Attribute
Error type:     NotFound
Further description:    /data/11/meshes/E/x/value
[ADIOS2] Warning: Attribute with name /data/11/meshes/E/x/value has no type in backend.
Cannot read record component 'x' and will skip it due to read error:
Read Error in backend ADIOS2
Object type:    Attribute
Error type:     NotFound
Further description:    /data/11/meshes/E/x/value
11 

number of meshes: 1
  all meshes:
    E

number of particle species: 0
```

Solution to this:

* **In the current release cycle:** To workaround this, we should resolve #1274 in the current release cycle and add a random-access read mode to ADIOS2.
* **In the next release cycle:** We will revise the ADIOS2 schemas and hopefully make the capabilities of `snapshot` available to users on a broader scale. Currently, those features are only available in the experimental new ADIOS2 schema.

franzpoeschel:
Next note: Once this PR and #1218 are merged, the `append_mode` test should be extended to test the `AppendAfterSteps` option of BP5. Both PRs are necessary to do that properly, but before the upcoming release we should make sure that truncation works as expected.

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -109,6 +109,12 @@ namespace internal
      * To be used for reading
      */
     constexpr FlushParams defaultFlushParams{};
+
+    enum class SeriesStatus : unsigned char
```
ax3l, line 136: Please add Doxygen, esp. for each parameter

include/openPMD/Iteration.hpp
```diff
@@ -282,21 +284,64 @@ class Iteration : public Attributable
...
     void read_impl(std::string const &groupPath);
 
+    /**
+     * Status after beginning an IO step. Currently includes:
+     * * The advance status (OK, OVER, RANDOMACCESS)
+     * * The opened iterations, in case the snapshot attribute is found
+     */
+    struct BeginStepStatus
+    {
+        using AvailableIterations_t = std::optional<std::deque<uint64_t> >;
```
ax3l, line 295: Needs [`<cstdint>`](https://en.cppreference.com/w/cpp/header/cstdint)

include/openPMD/Series.hpp
```diff
@@ -82,6 +84,12 @@ namespace internal
          * the same instance.
          */
         std::optional<WriteIterations> m_writeIterations;
+        /**
+         * For writing: Remember which iterations have been written in the
+         * currently active output step. Use this later when writing the
+         * snapshot attribute.
+         */
+        std::set<uint64_t> m_currentlyActiveIterations;
```
ax3l, line 93: Needs [`<cstdint>`](https://en.cppreference.com/w/cpp/header/cstdint)

include/openPMD/Series.hpp
```diff
@@ -37,8 +37,10 @@
 #include <mpi.h>
 #endif
 
+#include <deque>
```
ax3l, line 41:
```suggestion
#include <cstdint>
#include <deque>
```

include/openPMD/Iteration.hpp
```diff
@@ -28,7 +28,9 @@
 #include "openPMD/backend/Attributable.hpp"
 #include "openPMD/backend/Container.hpp"
 
+#include <deque>
```
ax3l, line 32:
```suggestion
#include <cstdint>
#include <deque>
```

include/openPMD/Streaming.hpp
```diff
@@ -20,7 +20,8 @@ namespace openPMD
 enum class AdvanceStatus : unsigned char
 {
     OK, /* stream goes on */
```
ax3l: Please use `///<` per element, so Doxygen can pick the doc string up :)

src/Series.cpp
```diff
@@ -1064,7 +1088,7 @@ void Series::readOneIterationFileBased(std::string const &filePath)
     series.iterations.readAttributes(ReadMode::OverrideExisting);
 }
 
-void Series::readGorVBased(bool do_init)
+std::optional<std::deque<uint64_t>> Series::readGorVBased(bool do_init)
```
ax3l, line 1142: Double check includes here, including `cstdint`

include/openPMD/Series.hpp
franzpoeschel: Should we add a type alias `IterationIndex` or sth like it for 0.15? It's currently hardcoded everywhere as uint64_t.

franzpoeschel:
Somehow the Icpc issue is back and I can even reproduce it locally. Does somehow not fully deactivate the warning, will have a look..

franzpoeschel:
With `-Werror`, I get the same build failure even on dev. Given that this warning is a compiler bug, should we maybe just deactivate the `-Werror` flag for now?
EDIT: Latest commit uses `-Werror -wd1011`, this fixes it

include/openPMD/IO/IOTask.hpp
```diff
@@ -521,6 +522,13 @@ struct OPENPMDAPI_EXPORT Parameter<Operation::WRITE_ATT>
 
     std::string name = "";
     Datatype dtype = Datatype::UNDEFINED;
+    /*
+     * If true, this attribute changes across IO steps.
+     * It should only be written in backends that support IO steps,
+     * otherwise writing should be skipped.
+     * The frontend is responsible for handling both situations.
+     */
```
franzpoeschel, line 557: Note to self: snapshot attribute should indiscriminately be written in v-based iteration encoding

src/Series.cpp
```diff
@@ -1407,9 +1425,15 @@ AdvanceStatus Series::advance(
          * opening an iteration's file by beginning a step on it.
          * So, return now.
          */
+        iteration.get().m_closed = internal::CloseStatus::ClosedInBackend;
         return AdvanceStatus::OK;
     }
 
+    if (mode == AdvanceMode::ENDSTEP)
+    {
+        flushStep(/* doFlush = */ false);
```
franzpoeschel, line 1529: Note to self: There is some PR further down the road that defines a second version of `Series::advance()`, check if that also has this change.

franzpoeschel: The other instance doesnt need it, its purpose is to be independent of any concrete Iteration.

franzpoeschel:
```
append_mode
/home/runner/work/openPMD-api/openPMD-api/test/SerialIOTest.cpp:5886
/home/runner/work/openPMD-api/openPMD-api/test/SerialIOTest.cpp:5886: FAILED:
  ERROR: attribute /date has been defined and its value cannot be changed, in
  call to DefineAttribute
  ```

This sporadically happens with this PR in the ASAN UBSAN CI run
Apparently, putting verbose output in the relevant test avoids the issue, making this hard to debug in the CI
I'll have to try recreating the environment locally, I wasnt able to reproduce this so far

**Comment by franzpoeschel** [comment]
I've recreated the environment and reproduced the issue locally now, it seems like an ADIOS2 glitch:

```
~│                                                                                                                                           │#0  0x00007f21bac88672 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6                                                   [272/304]
 ~│                                                                                                                                           │#1  0x00007f21b9d0064e in adios2::core::Attribute<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >& adios2::c
 ~│                                                                                                                                           │ore::IO::DefineAttribute<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(std::__cxx11::basic_string<char, st
 ~│                                                                                                                                           │d::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, s
 ~│                                                                                                                                           │td::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>
 ~│                                                                                                                                           │, std::allocator<char> >) () from /usr/local/lib/../lib/libadios2_core.so.2
 ~│                                                                                                                                           │#2  0x00007f21ba0436a5 in void adios2::format::BP4Deserializer::DefineAttributeInEngineIO<std::__cxx11::basic_string<char, std::char_traits<ch
 1│ Dump of assembler code for function __cxa_throw:                                                                                          │ar>, std::allocator<char> > >(adios2::format::BPBase::ElementIndexHeader const&, adios2::core::Engine&, std::vector<char, std::allocator<char>
 2│    0x00007f21bac88660 <+0>: endbr64                                                                                                       │ > const&, unsigned long) const () from /usr/local/lib/../lib/libadios2_core.so.2
 3│    0x00007f21bac88664 <+4>: push   %r13                                                                                                   │#3  0x00007f21b9fa6da7 in adios2::format::BP4Deserializer::ParseAttributesIndexPerStep(adios2::format::BufferSTL const&, adios2::core::Engine&
 4│    0x00007f21bac88666 <+6>: mov    %rdx,%r13                                                                                              │, unsigned long, unsigned long) () from /usr/local/lib/../lib/libadios2_core.so.2
 5│    0x00007f21bac88669 <+9>: push   %r12                                                                                                   │#4  0x00007f21b9fa5f48 in adios2::format::BP4Deserializer::ParseMetadata(adios2::format::BufferSTL const&, adios2::core::Engine&, bool) () fro
 6│    0x00007f21bac8866b <+11>:    mov    %rsi,%r12                                                                                          │m /usr/local/lib/../lib/libadios2_core.so.2
 7│    0x00007f21bac8866e <+14>:    push   %rbp                                                                                               │#5  0x00007f21b9de34b1 in adios2::core::engine::BP4Reader::InitBuffer(std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::dur
 8│    0x00007f21bac8866f <+15>:    mov    %rdi,%rbp                                                                                          │ation<double, std::ratio<1l, 1000000000l> > > const&, std::chrono::duration<double, std::ratio<1l, 1l> > const&, std::chrono::duration<double,
 9│    0x00007f21bac88672 <+18>:    nop                                                                                                       │ std::ratio<1l, 1l> > const&) () from /usr/local/lib/../lib/libadios2_core.so.2
10│    0x00007f21bac88673 <+19>:    callq  0x7f21bac7aa70 <__cxa_get_globals@plt>                                                             │#6  0x00007f21b9de1d82 in adios2::core::engine::BP4Reader::Init() () from /usr/local/lib/../lib/libadios2_core.so.2
11│    0x00007f21bac88678 <+24>:    mov    %r13,%rdx                                                                                          │#7  0x00007f21b9ddfb65 in adios2::core::engine::BP4Reader::BP4Reader(adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char
12│    0x00007f21bac8867b <+27>:    mov    %r12,%rsi                                                                                          │>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm) () from /usr/local/lib/../lib/libadios2_core.so.2
13│    0x00007f21bac8867e <+30>:    mov    %rbp,%rdi                                                                                          │#8  0x00007f21b9d11229 in std::shared_ptr<adios2::core::Engine> adios2::core::IO::MakeEngine<adios2::core::engine::BP4Reader>(adios2::core::IO
14│    0x00007f21bac88681 <+33>:    addl   $0x1,0x8(%rax)                                                                                     │&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm) () from /usr/lo
15│    0x00007f21bac88685 <+37>:    callq  0x7f21bac7aae0 <__cxa_init_primary_exception@plt>                                                  │cal/lib/../lib/libadios2_core.so.2
16│    0x00007f21bac8868a <+42>:    movl   $0x1,(%rax)                                                                                        │#9  0x00007f21ba36abb9 in std::_Function_handler<std::shared_ptr<adios2::core::Engine> (adios2::core::IO&, std::__cxx11::basic_string<char, st
17│    0x00007f21bac88690 <+48>:    lea    0x60(%rax),%rbp                                                                                    │d::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm), std::shared_ptr<adios2::core::Engine> (*)(adios2::co
18│    0x00007f21bac88694 <+52>:    mov    %rbp,%rdi                                                                                          │re::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mode, adios2::helper::Comm)>::_M_invo
19│    0x00007f21bac88697 <+55>:    callq  0x7f21bac79bb0 <_Unwind_RaiseException@plt>                                                        │ke(std::_Any_data const&, adios2::core::IO&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::M
20│    0x00007f21bac8869c <+60>:    mov    %rbp,%rdi                                                                                          │ode&&, adios2::helper::Comm&&) () from /usr/local/lib/../lib/libadios2_core_mpi.so.2
21│    0x00007f21bac8869f <+63>:    callq  0x7f21bac78690 <__cxa_begin_catch@plt>                                                             │#10 0x00007f21b9cf7174 in adios2::core::IO::Open(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios
22│    0x00007f21bac886a4 <+68>:    callq  0x7f21bac78180 <_ZSt9terminatev@plt>                                                               │2::Mode, adios2::helper::Comm) () from /usr/local/lib/../lib/libadios2_core.so.2
23│ End of assembler dump.                                                                                                                    │#11 0x00007f21b9cf8300 in adios2::core::IO::Open(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios
 ~│                                                                                                                                           │2::Mode) () from /usr/local/lib/../lib/libadios2_core.so.2
 ~│                                                                                                                                           │#12 0x00007f21bae606a1 in adios2::IO::Open(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, adios2::Mod
 ~│                                                                                                                                           │e) () from /usr/local/lib/libadios2_cxx11.so.2
 ~│                                                                                                                                           │#13 0x00007f21c49060ce in openPMD::detail::BufferedActions::getEngine (this=0x616000042c80) at /home/franz/git-repos/openPMD-api/src/IO/ADIOS/
 ~│                                                                                                                                           │ADIOS2IOHandler.cpp:2472
 ~│                                                                                                                                           │#14 0x00007f21c4918945 in openPMD::detail::BufferedActions::configure_IO (this=0x616000042c80, impl=...) at /home/franz/git-repos/openPMD-api/
 ~│                                                                                                                                           │src/IO/ADIOS/ADIOS2IOHandler.cpp:2450
** Dump of assembler code for function __cxa_throw: (7f21bac88660 - 7f21bac886a4) **                                                          │#15 0x00007f21c491090c in openPMD::detail::BufferedActions::BufferedActions (this=0x616000042c80, impl=..., file=...) at /home/franz/git-repos
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

/home/franz/git-repos/openPMD-api/test/SerialIOTest.cpp:5886: FAILED:
  ERROR: attribute /date has been defined and its value cannot be changed, in
  call to DefineAttribute

test cases: 1 | 1 failed
assertions: 2 | 1 passed | 1 failed
```

I'll try to find out under which circumstances it happens (ADIOS2 version, ADIOS2 build mode)

**Comment by franzpoeschel** [comment]
Alright, I think I have it figured out:

* The ASAN environment makes the test run slower
* Slow enough that when appending, the `/date` attribute sometimes changes to the next second
* When reading, ADIOS2 v2.7 checks that duplicate attributes are identical (due to using the same internal code as for defining attributes in writing) and throws an error 
* This behavior seems to be fixed in ADIOS2 v2.8
* It's easy to trigger this behavior by running other CI configurations with ADIOS2 v2.7 under valgrind

The best fix for this would be by avoiding to write duplicate attributes, need to check if that is possible some way

**Comment by franzpoeschel** [comment]
The issue is already present in dev, I'll try going for a fix in #1218 because it has some features that should help in fixing this.

File: src/ReadIterations.cpp [hunk 12]
```diff
@@ -292,6 +292,8 @@ std::optional<SeriesIterator *> SeriesIterator::loopBody()
         {
             return std::nullopt;
         }
+        std::cout << "Trying to access iteration "
+                  << currentIterationIndex.value() << std::endl;
```
franzpoeschel: todo: remove this again

franzpoeschel:
I've addressed all review comments and cleaned up the commit history today. Commit descriptions are mostly very detailed. Tests ran green, so ready for review :) @ax3l

src/Iteration.cpp
```diff
@@ -282,6 +282,13 @@ void Iteration::flushVariableBased(
         Parameter<Operation::OPEN_PATH> pOpen;
         pOpen.path = "";
         IOHandler()->enqueue(IOTask(this, pOpen));
+        /*
+         * In v-based encoding, the snapshot attribute must always be written,
```
ax3l, line 286:
```suggestion
         * In variableBased encoding, the snapshot attribute must always be written,
```

ax3l (approved):

PR #952, opened 2021-03-27
Dask: Array

ax3l:
Add bindings for all record components to dask arrays (ND mesh record components and 1D particle record components).

## Follow-Up To Do

Multi-node / multi-worker setup (aka distributed) and already seen with `compute="processes"`:
```
distributed.protocol.pickle - INFO - Failed to serialize <openpmd_api.DaskArray.DaskRecordComponent object at 0x7fb2e8216670>. Exception: cannot pickle 'openpmd_api.openpmd_api_cxx.Mesh_Record_Component' object
```
and same for data frames:
```
openpmd_api.openpmd_api_cxx.ParticleSpecies
```
Similar to `h5py`, things might get tricky here: https://stackoverflow.com/questions/63236728/dask-to-hdf5-write-fails-h5py-cannot-be-pickled

Maybe we find a way to pickle a `Record_Component` by transporting the info needed to initialize a non-MPI series and re-creating the record component from it: file name, deferred open, iteration, record & record component name.

## References

- https://github.com/openPMD/openPMD-api/pull/935#issuecomment-800844453
- https://github.com/dask/dask/issues/7475

### DIFF
--- a/src/binding/python/openpmd_api/DaskArray.py
+++ b/src/binding/python/openpmd_api/DaskArray.py
@@ -1,0 +1,140 @@
+"""
+This file is part of the openPMD-api.
+
+Copyright 2021 openPMD contributors
+Authors: Axel Huebl
+License: LGPLv3+
+"""
+import math
+import numpy as np
+try:
+    from dask.array import from_array
+    found_dask = True
+except ImportError:
+    found_dask = False
+
+
+class DaskRecordComponent:
+    # shape, .ndim, .dtype and support numpy-style slicing
+    def __init__(self, record_component):
+        self.rc = record_component
+
+    @property
+    def shape(self):
+        # fixme: https://github.com/openPMD/openPMD-api/issues/808
+        return tuple(self.rc.shape)
+
+    @property
+    def ndim(self):
+        return self.rc.ndim
+
+    @property
+    def dtype(self):
+        return self.rc.dtype
+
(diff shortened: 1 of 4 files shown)


src/binding/python/openpmd_api/DaskArray.py
```diff
@@ -0,0 +1,133 @@
...
+        return self.rc.dtype
+
+    def __getitem__(self, slices):
+        """here we support what Record_Component implements: a tuple of slices,
+        a slice or an index; we do not support fancy indexing
+        """
+        data = self.rc[slices]
+        self.rc.series_flush()
+        data *= self.rc.unit_SI
+        # print("data.shape=", data.shape)
```
ax3l, line 48: This is called surprisingly often

src/binding/python/openpmd_api/DaskDataFrame.py
```diff
@@ -74,7 +74,7 @@ def particles_to_daskdataframe(particle_species):
     if chunks is None:
         for k_r, r in particle_species.items():
             for k_rc, rc in r.items():
-                chunks = rc.available_chunks()
+                chunks = rc.shape()
```
ax3l: double-check if still needed

ax3l: looks like its not needed

PR #960, opened 2021-04-08
HDF5 Dask DataFrame issue: File ID

ax3l:
When removing those lines - and only with HDF5 - I get the following error on `electrons.to_dask()`:
```
openpmd_api/DaskDataFrame.py in particles_to_daskdataframe(particle_species)
     65         for k_rc, rc in r.items():
     66             if not rc.constant:
---> 67                 chunks = rc.available_chunks()
     68                 break
     69         if chunks:

RuntimeError: [HDF5] File ID not found with file name
```

Seen with the 3D openPMD HDF5 example data set, at iteration 400.

Printing the records and components in the loop shows:
- charge Scalar (skipped because constant)
- mass Scalar (skipped because constant)
- momentum x (calls available_chunks)

The patch from #959 is applied.

### DIFF
--- a/examples/11_particle_dataframe.py
+++ b/examples/11_particle_dataframe.py
@@ -27,13 +27,13 @@ s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
 electrons = s.iterations[400].particles["electrons"]
 
 # all particles
-df = electrons.to_df()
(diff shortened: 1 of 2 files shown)


ax3l:
demonstrator moved to #961

PR #962, opened 2021-04-08
HDF5: Fix Segfault with libSplash Files

### LINKED ISSUE #939: Segfault when reading libSplash file
**Describe the bug, To Reproduce**
Steps:
* Use the HDF5 plugin in PIConGPU to write data (that plugin is now removed, last commit containing it is 9180c81ba). Field data is sufficient to demonstrate.
* Use Pip to download the current openpmd_api
* Run `openpmd-ls dump%T.h5` or similar
* Observe a segfault

**Expected behavior**
Should not segfault. A self-built instance of the openPMD API reads the file just fine.

**Software Environment**
 - openpmd_api 0.13.2 from pip
 - operating system: observed on HZDR Hemera cluster (CentOS) and on a local x64 Ubuntu 20.10 machine
 - name and version of Python implementation: Python 3.8.6 from Ubuntu packages
 - version of *: whatever it is that the Pip package uses

**Additional context**
Valgrind seems to suggest there's something going wrong during attribute reading:
```
==164527== Invalid free() / delete / delete[] / realloc()
==164527==    at 0x483DA3F: free (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==164527==    by 0x67B9850: openPMD::HDF5IOHandlerImpl::readAttribute(openPMD::Writable*, openPMD::Parameter<(openPMD::Operation)18>&) (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x67C445A: openPMD::AbstractIOHandlerImpl::flush() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x67B574D: openPMD::HDF5IOHandler::flush() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x6797021: openPMD::Attributable::readAttributes() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x6771E32: openPMD::RecordComponent::readBase() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x67A4FF4: openPMD::MeshRecordComponent::read() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x6765D57: openPMD::Mesh::read() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x67617BD: openPMD::Iteration::read() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x6778D2A: openPMD::Series::read() (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x677D550: openPMD::Series::readGroupBased(bool) (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x67804E6: openPMD::Series::init(std::shared_ptr<openPMD::AbstractIOHandler>, std::unique_ptr<openPMD::Series::ParsedInput, std::default_delete<openPMD::Series::ParsedInput> >) (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
```
The final crash comes from an HDF5 sanity check:
```
==164527== Process terminating with default action of signal 6 (SIGABRT)
==164527==    at 0x4A718CB: raise (raise.c:50)
==164527==    by 0x4A56863: abort (abort.c:79)
==164527==    by 0x4A56748: __assert_fail_base.cold (assert.c:92)
==164527==    by 0x4A68A95: __assert_fail (assert.c:101)
==164527==    by 0x6F9FE16: H5MM_final_sanity_check (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
==164527==    by 0x6E107F7: H5_term_library (in /home/franz/.local/lib/python3.8/site-packages/openpmd_api/openpmd_api_cxx.cpython-38-x86_64-linux-gnu.so)
```

Is there a reason why Valgrind gives me 10km of Python errors? Thought that `openpmd-ls` was C++ only? My locally built version of `openpmd-ls` runs without error under Valgrind.

What is the best way to share a reproducing file?

ax3l:
Fix #939: a segfault (triggered reproducibly with pip builds) with legacy libSplash HDF5 files.
Introduced in #641 (`0.10.3-alpha` and newer).

From the docs:
```
Name: H5Tget_member_name
  ...
  The HDF5 Library allocates a buffer to receive the name of the
  field. The caller must subsequently free the buffer with
  H5free_memory.
```

I did not need this in my test to fix the segfault, but revisiting the logic I also noticed that passing `nullptr`s to `str(n)cmp` is undefined behavior and fixed that as well.

Refs.:
- https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-GetMemberName
- https://support.hdfgroup.org/HDF5/doc/RM/RM_H5.html#Library-FreeMemory (`IN: Buffer to be freed. Can be NULL.`)
- https://github.com/HDFGroup/hdf5/blob/hdf5-1_12_0/src/H5.c#L971-L984
- https://github.com/HDFGroup/hdf5/blob/hdf5-1_12_0/src/H5MM.c#L509-L527
- https://github.com/ComputationalRadiationPhysics/libSplash/pull/285

### DIFF
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -1466,7 +1466,7 @@ HDF5IOHandlerImpl::readAttribute(Writable* writable,
                 char* m1 = H5Tget_member_name(attr_type, 1);
                 char* m2 = H5Tget_member_name(attr_type, 2);
                 if( m0 == nullptr || m1 == nullptr || m2 == nullptr )
-                    isLegacyLibSplashAttr = false;
+                    isLegacyLibSplashAttr = false;  // NOLINT(bugprone-branch-clone)
                 else if(strcmp("x", m0) != 0 || strcmp("y", m1) != 0 || strcmp("z", m2) != 0)
                     isLegacyLibSplashAttr = false;
                 H5free_memory(m2);
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -1424,8 +1424,9 @@ HDF5IOHandlerImpl::readAttribute(Writable* writable,
             {
                 char* m0 = H5Tget_member_name(attr_type, 0);
                 char* m1 = H5Tget_member_name(attr_type, 1);
-                if( (strncmp("TRUE" , m0, 4) == 0) && (strncmp("FALSE", m1, 5) == 0) )
-                    attrIsBool = true;
+                if( m0 != nullptr && m1 != nullptr )
+                    if( (strncmp("TRUE" , m0, 4) == 0) && (strncmp("FALSE", m1, 5) == 0) )
+                        attrIsBool = true;
                 H5free_memory(m1);
                 H5free_memory(m0);
             }
@@ -1446,8 +1447,9 @@ HDF5IOHandlerImpl::readAttribute(Writable* writable,
             {
                 char* m0 = H5Tget_member_name(attr_type, 0);
                 char* m1 = H5Tget_member_name(attr_type, 1);
-                if( (strncmp("r" , m0, 4) == 0) && (strncmp("i", m1, 5) == 0) )
-                    isComplexType = true;
+                if( m0 != nullptr && m1 != nullptr )
+                    if( (strncmp("r" , m0, 1) == 0) && (strncmp("i", m1, 1) == 0) )
+                        isComplexType = true;
                 H5free_memory(m1);
                 H5free_memory(m0);
             }
@@ -1463,7 +1465,9 @@ HDF5IOHandlerImpl::readAttribute(Writable* writable,
                 char* m0 = H5Tget_member_name(attr_type, 0);
                 char* m1 = H5Tget_member_name(attr_type, 1);
                 char* m2 = H5Tget_member_name(attr_type, 2);
-                if(strcmp("x", m0) != 0 || strcmp("y", m1) != 0 || strcmp("z", m2) != 0)
+                if( m0 == nullptr || m1 == nullptr || m2 == nullptr )
+                    isLegacyLibSplashAttr = false;
+                else if(strcmp("x", m0) != 0 || strcmp("y", m1) != 0 || strcmp("z", m2) != 0)
                     isLegacyLibSplashAttr = false;
                 H5free_memory(m2);
                 H5free_memory(m1);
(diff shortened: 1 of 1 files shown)


src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -1465,9 +1465,9 @@ HDF5IOHandlerImpl::readAttribute(Writable* writable,
                 char* m2 = H5Tget_member_name(attr_type, 2);
                 if(strcmp("x", m0) != 0 || strcmp("y", m1) != 0 || strcmp("z", m2) != 0)
```
ax3l: I tested this and this fixed the issue we saw.
But looking at the difference between `free` and `H5free_memory` I wonder if, as a follow-up, we should consider that `H5Tget_member_name` might return a `nullptr`:
https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-GetMemberName
```
Returns:
    Returns a valid pointer to a string allocated with malloc() if successful; otherwise returns NULL.
```

The behavior of using a `nullptr` in `strcmp` seems to be implementation-defined (we might be lucky atm.):
https://stackoverflow.com/questions/21865041/in-c-what-exactly-happens-when-you-pass-a-null-pointer-to-strcmp
According to other answers therein, this might be undefined behavior...

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -1463,11 +1465,13 @@ HDF5IOHandlerImpl::readAttribute(Writable* writable,
                 char* m0 = H5Tget_member_name(attr_type, 0);
                 char* m1 = H5Tget_member_name(attr_type, 1);
                 char* m2 = H5Tget_member_name(attr_type, 2);
-                if(strcmp("x", m0) != 0 || strcmp("y", m1) != 0 || strcmp("z", m2) != 0)
+                if( m0 == nullptr || m1 == nullptr || m2 == nullptr )
                     isLegacyLibSplashAttr = false;
```
```suggestion
                    isLegacyLibSplashAttr = false;  // NOLINT(bugprone-branch-clone)
```

PR #460, opened 2019-02-05
Differentiate accessTypes on use

anokfireball:
For frontend, use a seperate accessType variable that may change during the first read of a fileBased Series (required for setting values in case of READ_ONLY).
For the backend, continue using the same immutable accessType as before (desired by the user).

Fix #394.

PR #963, opened 2021-04-09
ParticleSpecies & RecordComponent Serialize

ax3l:
Preparation to derive information necessary to construct a new series that can access the an equivalent `ParticleSpecies` or  `RecordComponent` (see #952).

- [x] rebase after #966 was merged
- [x] cleanup

### DIFF
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -34,68 +34,8 @@ int main()
         "../samples/git-sample/data%T.h5",
         Access::READ_ONLY
     );
-    cout << "Read a Series with openPMD standard version "
-         << series.openPMD() << '\n';
-
-    cout << "The Series contains " << series.iterations.size() << " iterations:";
-    for( auto const& i : series.iterations )
-        cout << "\n\t" << i.first;
-    cout << '\n';
-
-    Iteration i = series.iterations[100];
-    cout << "Iteration 100 contains " << i.meshes.size() << " meshes:";
-    for( auto const& m : i.meshes )
-        cout << "\n\t" << m.first;
-    cout << '\n';
-    cout << "Iteration 100 contains " << i.particles.size() << " particle species:";
-    for( auto const& ps : i.particles ) {
-        cout << "\n\t" << ps.first;
-        for( auto const& r : ps.second ) {
-            cout << "\n\t" << r.first;
-            cout << '\n';
-        }
-    }
-
-    openPMD::ParticleSpecies electrons = i.particles["electrons"];
-    std::shared_ptr<double> charge = electrons["charge"][openPMD::RecordComponent::SCALAR].loadChunk<double>();
-    series.flush();
-    cout << "And the first electron particle has a charge = " << charge.get()[0];
-    cout << '\n';
-
-    MeshRecordComponent E_x = i.meshes["E"]["x"];
-    Extent extent = E_x.getExtent();
-    cout << "Field E/x has shape (";
-    for( auto const& dim : extent )
-        cout << dim << ',';
-    cout << ") and has datatype " << E_x.getDatatype() << '\n';
-
-    Offset chunk_offset = {1, 1, 1};
-    Extent chunk_extent = {2, 2, 1};
-    auto chunk_data = E_x.loadChunk<double>(chunk_offset, chunk_extent);
-    cout << "Queued the loading of a single chunk from disk, "
-            "ready to execute\n";
-    series.flush();
-    cout << "Chunk has been read from disk\n"
-         << "Read chunk contains:\n";
-    for( size_t row = 0; row < chunk_extent[0]; ++row )
-    {
-        for( size_t col = 0; col < chunk_extent[1]; ++col )
-            cout << "\t"
-                 << '(' << row + chunk_offset[0] << '|' << col + chunk_offset[1] << '|' << 1 << ")\t"
-                 << chunk_data.get()[row*chunk_extent[1]+col];
-        cout << '\n';
-    }
-
-    auto all_data = E_x.loadChunk<double>();
-    series.flush();
-    cout << "Full E/x starts with:\n\t{";
-    for( size_t col = 0; col < extent[1] && col < 5; ++col )
-        cout << all_data.get()[col] << ", ";
-    cout << "...}\n";
-
-    /* The files in 'series' are still open until the object is destroyed, on
-     * which it cleanly flushes and closes all open file handles.
-     * When running out of scope on return, the 'Series' destructor is called.
-     */
+    auto it = series.iterations[400];
+    auto px = it.particles["electrons"]["momentum"]["x"];
+    px.serialize();
     return 0;
 }
--- a/examples/2_read_serial.cpp
+++ b/examples/2_read_serial.cpp
@@ -34,8 +34,68 @@ int main()
         "../samples/git-sample/data%T.h5",
         Access::READ_ONLY
     );
-    auto it = series.iterations[400];
-    auto px = it.particles["electrons"]["momentum"]["x"];
-    px.serialize();
+    cout << "Read a Series with openPMD standard version "
+         << series.openPMD() << '\n';
+
+    cout << "The Series contains " << series.iterations.size() << " iterations:";
+    for( auto const& i : series.iterations )
+        cout << "\n\t" << i.first;
+    cout << '\n';
+
+    Iteration i = series.iterations[100];
+    cout << "Iteration 100 contains " << i.meshes.size() << " meshes:";
+    for( auto const& m : i.meshes )
+        cout << "\n\t" << m.first;
+    cout << '\n';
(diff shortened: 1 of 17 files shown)


src/RecordComponent.cpp
```diff
@@ -176,6 +178,39 @@ RecordComponent::empty() const
...
+        throw std::runtime_error("serialize does not work with MPI");
+
+    // information to construct a new serial read-only series
+    auto const str_path = IOHandler()->directory; // samples/git-samples/
+    auto const str_name = series.name();          // data%T
+    auto const str_ext = suffix(series.m_format); // .bp, .h5, .json, ...
+    auto const filepath = str_path + str_name + str_ext;
+
+    // information to find iteration
+    //auto const int it = ...; // 400
```
ax3l:
```suggestion
    std::vector< std::string > const my_path = this->myPath();
    int const  it = std::stoi(my_path.at(1));  // 400
    
    std::string str_r, str_rc, str_part;
    if( my_path.at(2) == "meshes" )
    {
        str_r = my_path.at(3);  // "E"
        str_rc = my_path.at(4);  // "x"
    } else
    {
        str_part = my_path.at(3);  // "x"
        str_r = my_path.at(4);  // "E"
        str_rc = my_path.at(5);  // "x"
    }
```

Probably a good strategy: serialize the whole `vector< string >` and de-clutter during re-initalization/de-serialization.

src/RecordComponent.cpp
```diff
@@ -176,6 +178,39 @@ RecordComponent::empty() const
...
+    auto & series = this->retrieveSeries();
+    auto const str_backend = series.backend(); // HDF5, ADIOS2, ...
+    if( str_backend.find( "MPI" ) != std::string::npos )
+        throw std::runtime_error("serialize does not work with MPI");
+
+    // information to construct a new serial read-only series
+    auto const str_path = IOHandler()->directory; // samples/git-samples/
+    auto const str_name = series.name();          // data%T
+    auto const str_ext = suffix(series.m_format); // .bp, .h5, .json, ...
+    auto const filepath = str_path + str_name + str_ext;
```
ax3l: @franzpoeschel I might add this logic as the first argument in `.myPath()` logic with this PR, just to keep it all together. What do you think?

examples/2_read_serial.cpp
```diff
@@ -34,68 +34,8 @@ int main()
...
-    series.flush();
-    cout << "Full E/x starts with:\n\t{";
-    for( size_t col = 0; col < extent[1] && col < 5; ++col )
-        cout << all_data.get()[col] << ", ";
-    cout << "...}\n";
-
-    /* The files in 'series' are still open until the object is destroyed, on
-     * which it cleanly flushes and closes all open file handles.
-     * When running out of scope on return, the 'Series' destructor is called.
-     */
```
franzpoeschel: Be aware that this file is also included as a go-to example in our `serial.rst`, so it should not be replaced in its entirety.

src/RecordComponent.cpp
```diff
@@ -176,6 +178,39 @@ RecordComponent::empty() const
...
 }
 
+void
+RecordComponent::serialize()
+{
+    // verify: read-only
+    if( IOHandler()->m_frontendAccess != Access::READ_ONLY )
+        throw std::runtime_error("serialize only works on read-only access");
+    //if( IOHandler()->m_backendAccess != Access::READ_ONLY )
+    // as well??
```
franzpoeschel: Ref.: https://github.com/openPMD/openPMD-api/pull/460
Tldr: It should not matter at this point

src/RecordComponent.cpp
franzpoeschel: So you want to add that information to the output of `.myPath()`?
If yes, I would suggest to make this a little struct
```cpp
struct MyPath
{
    std::string directory;
    std::string seriesName;
    std::string seriesExtension;
    std::string filePath() const;
    std::vector< std::string > openPMDGroup;
}
```
It's already error-prone enough to handle the `std::vector<std::string>` correctly, but at least somewhat justifiable since it's an openPMD path reconstructed by the `Attributable` backbone in openPMD. Also, this would keep the old functionality of `.myPath()` separate.

franzpoeschel:
I have implemented some of the suggestions above on a branch building on this one https://github.com/franzpoeschel/openPMD-api/tree/topic-serialize-suggestion

ax3l:
Thank you for the suggestion, I like the `openPMDGroup` touch, that makes it way more general :)
I didn't realize we have a `myPath` member on Attributable already, hah!

examples/2_read_serial.cpp
ax3l: yes, only temporarily manipulated for quick tests :)

src/binding/python/MeshRecordComponent.cpp
```diff
@@ -52,5 +53,40 @@ void init_MeshRecordComponent(py::module &m) {
...
+                if (t.size() != 2)
+                    throw std::runtime_error("Invalid state!");
+
+                std::string const filename = t[0].cast< std::string >();
+                std::vector< std::string > const openPMDGroup =
+                        t[1].cast< std::vector< std::string > >();
+
+                // Create a new openPMD Series and keep it alive forevaaa
+                // TODO: how much of a hack is that, heh? :D omg.
+                static auto series = openPMD::Series(
```
ax3l: Not sure how much of a hack this `static` member is to fix up lifetimes, but it seems to work :)

include/openPMD/binding/python/Pickle.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+                    std::string const filename = t[0].cast< std::string >();
+                    std::vector< std::string > const group =
+                        t[1].cast< std::vector< std::string > >();
+
+                    // Create a new openPMD Series and keep it alive.
+                    // This is a big hack for now, but it works for our use
+                    // case, which is spinning up remote serial read series
+                    // for DASK.
+                    static auto series = openPMD::Series(
```
ax3l, line 83: Not sure how much of a hack this `static` member is to fix up lifetimes, but it seems to work pretty well :)
We only have little control when it gets destroyed, which is likely as late as process end/module unload.

ax3l:
And lift-off :rocket: 

https://user-images.githubusercontent.com/1353258/124344217-d5db8980-db85-11eb-8582-f6421f78256c.mp4

[DistributedDaskOpenPMD.zip](https://github.com/openPMD/openPMD-api/files/6757880/DistributedDaskOpenPMD.zip)

include/openPMD/binding/python/Pickle.hpp
franzpoeschel: What happens when a user wants to unpickle two different Series within the same application? The `static auto series` will not be constructed a second time and then be stuck with the old file path.
Do you do this for avoiding the Series from going out of scope?
Would something like `return seriesAccessor( std::move( series ), group)` help? This would ensure that the `seriesAccessor` takes no reference.

franzpoeschel: Ah, I think I see the problem.
```cpp
    add_pickle(
        cl,
        [](openPMD::Series & series, std::vector< std::string > const & group ) {
            uint64_t const n_it = std::stoull(group.at(1));
            return series.iterations[n_it].meshes[group.at(3)][group.at(4)];
        }
    )
```
This lets the `Series` go out of scope and returns only the record component..

I think that this is a fundamental design issue with our Python bindings: Python has garbage collection, but we don't use it. Realistically, calling `del series` should *not* be sufficient in Python to destroy the Series, since the `series` should stay active as long as there is an active dependent member like a contained `RecordComponent`.
So, moving away from a C++-centered memory model for our Python API should probably do the trick.

Three quick and dirty fixes, each quicker and dirtier than the last:

1. Is it somehow possible in PyBind to quickly tell the garbage collection that in the above code the `series` and the returned `MeshRecordComponent` depend on each other? This should at least implement the above solution locally.
2. Make a subclass `PickledMeshRecordComponent` that privately stores the `series`.
3. Just do:
    ```cpp
    static openPMD::Series series;
    series = openPMD::Series(...); // this is executed every time
    ```
    (This solution will only allow one deserialized object to be used at a time though)

ax3l: > fundamental design issue with our Python bindings: Python has garbage collection, but we don't use it. 

Offline discussed: What we mean is we don't use the possibility to tell the garbage collector dependencies between our object hierarchy yet (in both directions) yet.

> Is it somehow possible in PyBind to quickly tell the garbage collection that in the above code the series and the returned MeshRecordComponent depend on each other? This should at least implement the above solution locally.

Gotta investigate that, would be a nice and clean solution :+1: 

Currently not tested and likely to fail in this situation:
- unpickle of multiple objects of the same class (I test different classes to be pickled/unpickled)

> 3.

good follow-up to at least have the last object (of the same kind) always be in usable state.

PR #966, opened 2021-04-13
Retrieve paths of objects in the openPMD hierarchy

franzpoeschel:
Needed for "serialization" in DASK contexts.

TODO:
- [x] testing

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -204,6 +204,9 @@ OPENPMD_protected:
     internal::SeriesInternal & retrieveSeries();
 
     void seriesFlush( FlushLevel );
+
+    std::vector< std::string > myPath() const;
+
     void flushAttributes();
     void readAttributes();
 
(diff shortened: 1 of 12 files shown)


include/openPMD/backend/Attributable.hpp
```diff
@@ -203,6 +203,8 @@ class AttributableImpl
     internal::SeriesInternal const & retrieveSeries() const;
     internal::SeriesInternal & retrieveSeries();
 
+    std::vector< std::string > myPath() const;
```
ax3l, line 215: Can you please add a doxygen string for this one? :) 
This would be a good place to explain the concept of what's in that vector of strings.

ax3l (approved):
Awesome, thanks a lot (and great coverage!)

include/openPMD/backend/Attributable.hpp
franzpoeschel: I think there are also some other places where a little comment will help, so yep will do

franzpoeschel: Done

ax3l (approved):
Thank you, looks great! :tada:

PR #971, opened 2021-05-07
Set a default QueueLimit of 2 in the ADIOS2/SST engine

franzpoeschel:
For reference, see [here](https://github.com/ornladios/ADIOS2/issues/2601#issuecomment-834175861).

For an explanation, I'll copy the in-line comment:
```cpp
            /*
             * By default, the SST engine of ADIOS2 does not set a limit on its
             * internal queue length.
             * If the reading end is slower than the writing end, this will
             * lead to a congestion in the queue and hence an increasing
             * memory usage while the writing code goes forward.
             * We could set a default queue limit of 1, thus forcing the
             * two codes to proceed entirely in lock-step.
             * We prefer a default queue limit of 2, which is still lower than
             * the default infinity, but allows writer and reader to process
             * data asynchronously as long as neither code fails to keep up the
             * rhythm. The writer can produce the next iteration while the
             * reader still deals with the old one.
             * Thus, a limit of 2 is a good balance between 1 and infinity,
             * keeping pipeline parallelism a default without running the risk
             * of using unbound memory.
             */
```

TODO:
- [x] Merge #981 first

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -2399,6 +2399,27 @@ namespace detail
              */
             m_IO.SetParameter( "StatsLevel", "0" );
         }
+        if( m_engineType == "sst" && notYetConfigured( "QueueLimit" ) )
+        {
+            /*
+             * By default, the SST engine of ADIOS2 does not set a limit on its
+             * internal queue length.
+             * If the reading end is slower than the writing end, this will
+             * lead to a congestion in the queue and hence an increasing
+             * memory usage while the writing code goes forward.
+             * We could set a default queue limit of 1, thus forcing the
+             * two codes to proceed entirely in lock-step.
+             * We prefer a default queue limit of 2, which is still lower than
+             * the default infinity, but allows writer and reader to process
+             * data asynchronously as long as neither code fails to keep up the
+             * rhythm. The writer can produce the next iteration while the
+             * reader still deals with the old one.
+             * Thus, a limit of 2 is a good balance between 1 and infinity,
+             * keeping pipeline parallelism a default without running the risk
+             * of using unbound memory.
+             */
+            m_IO.SetParameter( "QueueLimit", "2" );
+        }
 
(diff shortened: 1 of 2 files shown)


src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2399,6 +2399,27 @@ namespace detail
              */
             m_IO.SetParameter( "StatsLevel", "0" );
         }
+        if( m_engineType == "sst" && notYetConfigured( "QueueLimit" ) )
```
ax3l, line 2402: That sounds great!
Can you please describe the new option keyword in the user-facing `.rst` docs in the ADIOS2 section?

ax3l:
CI reports this on macOS:
```
adios2_streaming

/Users/runner/work/openPMD-api/openPMD-api/test/ParallelIOTest.cpp:946: FAILED:
  ERROR: SstReader did not find active Writer contact info in file "../samples/
  adios2_stream.sst".  Timeout or non-current SST contact file?
```

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: Done

franzpoeschel:
> CI reports this on macOS:
> 
> ```
> adios2_streaming
> ...............................................................................
> 
> /Users/runner/work/openPMD-api/openPMD-api/test/ParallelIOTest.cpp:946: FAILED:
>   {Unknown expression after the reported line}
> due to unexpected exception with message:
>   ERROR: SstReader did not find active Writer contact info in file "../samples/
>   adios2_stream.sst".  Timeout or non-current SST contact file?
> ```

I've rebased this on #981 which I hope should solve the issue. It doesn't seem to be an issue solely of this PR.

ax3l:
Thanks, I merged #981 now, so we can rebase it out again to keep the GitHub diff pretty :D

ax3l (approved):
:rocket: :sparkles:

PR #974, opened 2021-05-12
[WIP] GCC on PPC64le: Visibility Warning (Py)

ax3l:
There is an odd visibility warning when installing via with the PPC64le
GCC compiler (v 6.4.0) on Summit.

```
warning: 'store_chunk(openPMD::RecordComponent&, pybind11::array&, const Offset&, const Extent&, const std::vector<bool>&)::<lambda(auto:8)> [with auto:8 = bool]::<lambda(CXXType*)>' declared with greater visibility than the type of its field 'store_chunk(openPMD::RecordComponent&, pybind11::array&, const Offset&, const Extent&, const std::vector<bool>&)::<lambda(auto:8)> [with auto:8 = bool]::<lambda(CXXType*)>::<a capture>' [-Wattributes]
```
```
warning: 'store_chunk(openPMD::RecordComponent&, pybind11::array&, const Offset&, const Extent&, const std::vector<bool>&)::<lambda(auto:8)> [with auto:8 = bool]::<lambda(CXXType*)>' declared with greater visibility than the type of its field 'store_chunk(openPMD::RecordComponent&, pybind11::array&, const Offset&, const Extent&, const std::vector<bool>&)::<lambda(auto:8)> [with auto:8 = bool]::<lambda(CXXType*)>::<a capture>' [-Wattributes]
```

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -300,6 +300,7 @@ store_chunk(RecordComponent & r, py::array & a, Offset const & offset, Extent co
         using CXXType = decltype(cxxtype);
         a.inc_ref();
         void* data = a.mutable_data();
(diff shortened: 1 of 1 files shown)


franzpoeschel:
I've seen this one also on x64 architectures. Most results that you find about this warnings seem to point to a [GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80947), so I did not bother with it further.

Rel. [similar question on SO](https://stackoverflow.com/questions/44390898/gcc-6-x-warning-about-lambda-visibility)

ax3l:
Ok, looks like with GCC 8+ this is fixed and we are good.

PR #979, opened 2021-05-14
HDF5Auxiliary: Check String Sizes

ax3l:
Investigate origin of HDF5 1.10.7 check in [nightly installation tests](https://dev.azure.com/axelhuebl/openPMD-api/_build?definitionId=1&_a=summary) with Spack, seen for `3_write_serial.cpp` test:
```
HDF5-DIAG: Error detected in HDF5 (1.10.7) thread 0:
  #000: H5T.c line 2159 in H5Tset_size(): size must be positive
    major: Invalid arguments to routine
    minor: Bad value
```

Ref.: https://support.hdfgroup.org/HDF5/doc/RM/RM_H5T.html#Datatype-SetSize
Code: https://github.com/HDFGroup/hdf5/blob/1.12/master/src/H5T.c#L1839-L1841

### DIFF
--- a/src/IO/HDF5/HDF5Auxiliary.cpp
+++ b/src/IO/HDF5/HDF5Auxiliary.cpp
@@ -31,8 +31,15 @@
 #   include <array>
 #   include <complex>
 #   include <stack>
+#   include <stdexcept>
 #   include <string>
 #   include <typeinfo>
+
+#   if openPMD_USE_VERIFY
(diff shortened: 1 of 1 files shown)


ax3l:
Hm, cannot reproduce the `size` argument issue...

ax3l:
The problem in nightly CI might be that Spack built the dependencies with GCC 11.1.0 but we then built the post-install test with GCC 9.3.0: 58bdb6f1ea424aea8b24669688acb6d4827e11ba

ax3l:
Save to merge these checks anyway, to check API return values.

PR #981, opened 2021-05-17
SST tests: start reader a second after the writer

franzpoeschel:
If a previous test run left a stale .sst file, this avoids that the reader sees that file.
Theoretically, the timing of reader relative to the writer should not matter, but that's a thing for the ADIOS2 team to test.


ax3l (approved):
Thanks, let's see if that makes it more robust :)

PR #1001, opened 2021-06-03
Sphinx: rsvg converter for LaTeX

ax3l:
The imagemagick on RTD (Ubuntu 18.04) is too old and creates artifacts. We now use rsvg to generate a PDF for latex and keep the SVG for HTML & EPUB.

Follow-up to #997

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,6 +1,14 @@
-requirements_file: docs/requirements.txt
+version: 2
+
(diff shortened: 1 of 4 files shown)


ax3l:
Haha, looks great in PDF with my local rsvg version (2.50.3) but the RTD rsvg version (2.40.20) chops up the PDF figure for LaTeX.

PR #1002, opened 2021-06-04
Wheels: 0.13.4.post0

ax3l:
ADIOS: 2.6.0 -> 2.7.1 (Windows: we patch in https://github.com/ornladios/ADIOS2/pull/2746)
    
Add support for improved blosc support: https://github.com/ornladios/ADIOS2/pull/2592

### DIFF
--- a/library_builders.sh
+++ b/library_builders.sh
@@ -94,7 +94,7 @@ function build_adios2 {
         # ADIOS 2.7.1 & Blosc 1.20.1/1.21.0
         #   /usr/local/lib/libblosc.a(blosc.c.o): In function `blosc_init.part.9':
         # blosc.c:(.text+0x43e): undefined reference to `pthread_atfork'
-        export LDFLAGS="-pthread ${LDFLAGS}"
+        export LDFLAGS="-pthread"
     else
         # ZPL in EVPATH disabled because it does not build with older macOS
         #       https://github.com/GTkorvo/evpath/issues/47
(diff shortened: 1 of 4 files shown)


library_builders.sh
```diff
@@ -91,21 +92,26 @@ function build_adios2 {
     if [ "$(uname -s)" = "Linux" ]
     then
         EVPATH_ZPL="ON"
+        # ADIOS 2.7.1 & Blosc 1.20.1/1.21.0
+        #   /usr/local/lib/libblosc.a(blosc.c.o): In function `blosc_init.part.9':
+        # blosc.c:(.text+0x43e): undefined reference to `pthread_atfork'
```
ax3l, line 97: same problem appears again in my env on summit at import time...

ax3l: Maybe we can try (here and for blosc below):
```
export LDFLAGS="-lpthread -lpthread_nonshared"
```
https://gitlab.kitware.com/cmake/cmake/-/issues/20709#note_760439

ax3l: I think this line in blosc is outdated:
https://github.com/Blosc/c-blosc/blob/7435f28dd08606bd51ab42b49b0e654547becac4/blosc/CMakeLists.txt#L68
Is: `CMAKE_THREAD_PREFER_PTHREAD` ([docs](https://cmake.org/cmake/help/v3.2/module/FindThreads.html))
Should be: `THREADS_PREFER_PTHREAD_FLAG` ([docs](https://cmake.org/cmake/help/latest/module/FindThreads.html))

ax3l: started a draft in #1013

Issue #996, opened 2021-06-01
Appending iterations ADIOS2 


pordyna:
Hi, I have a quick question on how to use the api with the ADIOS B4 backend.  What is the correct way to append new iterations to the same file with ADIOS2? I know that I can switch to the group based iteration encoding by removing `%T` from the series path but when I try to add a new iteration with `Access::READ_WRITE` I get
```
 ADIOS2 does currently not yet implement ReadWrite (Append) mode. Replacing with Read mode.
Unhandled exception of type 'St16invalid_argument' with message 'ERROR: Engine Open Mode not valid for function
```
Should I use `Access::CREATE` instead or could that remove previous iterations?

**Software Environment:**
 - version of openPMD-api: 0.13.2
 - version of ADIOS: 2.7.1.106
 - installed openPMD-api via: module system

ax3l:
Thanks for the question!

Contrary to what we document [here](https://github.com/openPMD/openPMD-api/blob/0.13.4/include/openPMD/IO/Access.hpp#L26-L33), `openPMD::Access::CREATE` does indeed work to add new iterations to both file-based (`..._%T.bp`) and groupBased (no `%T` in the file-name) encoding. I close and re-open a Series to add new iterations and this works.

Practically speaking, this should be the behavior of `READ_WRITE` (of a _Series_) - maybe that's something to fix @franzpoeschel?

pordyna:
Okay, thanks for the clarification!

ax3l:
Little correction: at the moment, you cannot re-open an ADIOS2 non-filebased Series for write without overwriting written iterations. That's something on our todo list using ADIOS2's append mode.

So either keep the Series object alive for all writing time, then all iteration encodings will work, or if you really must close it between iterations, then use fileBased.

pordyna:
thanks, good to know. I think I can close this issue now.

ax3l:
Awesome. I was wondering to leave it open, so we can keep track of it as we add `READ_WRITE` for ADIOS2. What do you think @franzpoeschel ?

PR #1007, opened 2021-06-10
ADIOS2 Append Mode

franzpoeschel:
Our `READ_WRITE` mode is not quite adequate for use in ADIOS2, since ADIOS2 only allows either reading or writing, but our `READ_WRITE` mode's workflow needs that both are working (#996). However, Append mode exposes an interface equivalent to that of the Write mode, but steps already on disk are not deleted, new ones are just appended. Hence, add a new `Access::APPEND` that unlike `READ_WRITE` does not allow any reading, but does not overwrite anything either.

- [x] Simple proof-of-concept implementation
- [x] Test all possible edge cases:
    * Since there is no reading involved, it's on the user to use the exact same configuration of openPMD in append mode as had been used for the appended-to file.
    * This is now testing new and old ADIOS schema, group- file- and variablebased encoding
    * Overwriting old iterations is still a work-in-progress, non-linear read patterns and truncation of files upon opening are to come in a later PR
- [x] Keep possible a workflow where in file-based iteration encoding, already existing iterations can be read and new ones can be written
  The `READ_WRITE` mode currently does this. I would suggest to use new Append mode to support this more properly. Will need some tweaks in the frontend though where APPEND mode is currently treated as equivalent to WRITE mode.
  Idea: In filebased iteration encoding, coerce the backendAccess to CREATE. Any written iteration will then be overwritten in doubt.
- [x] Bring this to the other backends. This should become the new default `READ_WRITE` mode light, unless you really need to read data.
- [ ] Discussion point: Since reading does not work *at all* in this mode, I don't see any way to guarantee that an update to a dataset is written with the same overall settings (schema, usage of steps, ..). Users will need to make sure to create consistent datasets...
- [ ] Discussion point: Checkpoint-restarting. Probably one of the main uses of this. Dump every 100 time steps, checkpoint every 1000, crash at 1500, restart from 1000. Consequence: Steps {1000,1100,…,1500} are written twice. See #949 for an exploration on how to deal with this, but it's not really ideal.
    Update: See https://github.com/ornladios/ADIOS2/issues/2775
    Will still not be trivial to implement, since a preceding read access might be necessary to figure out how many steps should be dropped. Would also require to specify from which iteration to restart writing.
- [x] Documentation page on access modi
- [x] File-based appending in ADIOS2: Throw error if trying to append to already written iteration?
    Update: Given the proposed workflow with https://github.com/ornladios/ADIOS2/issues/2775, it would be most consistent to just overwrite iterations. Maybe use `Access::CREATE` again for the backendAccess.
- [x] ADIOS1 UPDATE: OperationUnsupportedInBackend
- [x] Merge #1080 first
- [x] Document altered meaning of `createFile` task

Also, this introduces a new `Error.hpp`, so I don't need to again throw `std::runtime_errors` around.

### DIFF
--- a/docs/source/usage/workflow.rst
+++ b/docs/source/usage/workflow.rst
@@ -12,15 +12,16 @@ The openPMD-api distinguishes between a number of different access modes:
 * **Read/Write mode**: Creates a new Series if not existing, otherwise opens an existing Series for reading and writing.
   New datasets and iterations will be inserted as needed.
   Not fully supported by all backends:
+
   * ADIOS1: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
   * ADIOS2: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
-    Since this happens on a per-file level, this mode allows to read from existing iterations and write to new iterations at the same time in file-based iteration encoding.
+  Since this happens on a per-file level, this mode allows to read from existing iterations and write to new iterations at the same time in file-based iteration encoding.
 * **Append mode**: Restricted mode for appending new iterations to an existing Series that is supported by all backends at least in file-based iteration encoding, and by all but ADIOS1 in other encodings.
   The API is equivalent to that of the *Create* mode, meaning that no reading is supported whatsoever.
   If the Series does not exist yet, this behaves equivalently to the *Create* mode.
   Existing iterations will not be deleted, newly-written iterations will be inserted.
 
-  **Warning:** If writing an iteration that already exists, the behavior is implementation-defined and depends on the chosen backend and iteration encoding:
+  **Warning:** When writing an iteration that already exists, the behavior is implementation-defined and depends on the chosen backend and iteration encoding:
 
   * The new iteration might fully replace the old one.
   * The new iteration might be merged into the old one.
--- a/docs/source/usage/workflow.rst
+++ b/docs/source/usage/workflow.rst
@@ -1,5 +1,38 @@
 .. _workflow:
 
+Access modes
+============
+
+The openPMD-api distinguishes between a number of different access modes:
+
+* **Create mode**: Used for creating a new Series from scratch.
+  Any file possibly existing in the specified location will be overwritten.
+* **Read-only mode**: Used for reading from an existing Series.
+  No modifications will be made.
+* **Read/Write mode**: Creates a new Series if not existing, otherwise opens an existing Series for reading and writing.
+  New datasets and iterations will be inserted as needed.
+  Not fully supported by all backends:
+  * ADIOS1: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
+  * ADIOS2: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
+    Since this happens on a per-file level, this mode allows to read from existing iterations and write to new iterations at the same time in file-based iteration encoding.
+* **Append mode**: Restricted mode for appending new iterations to an existing Series that is supported by all backends at least in file-based iteration encoding, and by all but ADIOS1 in other encodings.
+  The API is equivalent to that of the *Create* mode, meaning that no reading is supported whatsoever.
+  If the Series does not exist yet, this behaves equivalently to the *Create* mode.
+  Existing iterations will not be deleted, newly-written iterations will be inserted.
+
+  **Warning:** If writing an iteration that already exists, the behavior is implementation-defined and depends on the chosen backend and iteration encoding:
+
+  * The new iteration might fully replace the old one.
+  * The new iteration might be merged into the old one.
+  * (To be removed in a future update) The old and new iteration might coexist in the resulting dataset.
+
+  We suggest to fully define iterations when using Append mode (i.e. as if using Create mode) to avoid implementation-specific behavior.
+  Appending to an openPMD Series is only supported on a per-iteration level.
+
+  **Warning:** There is no reading involved in using Append mode.
+  It is a user's responsibility to ensure that the appended dataset and the appended-to dataset are compatible with each other.
+  The results of using incompatible backend configurations are undefined.
+
 Workflow
 ========
 
--- a/include/openPMD/Error.hpp
+++ b/include/openPMD/Error.hpp
@@ -75,7 +75,6 @@ namespace error
      *
      * Example: A nullpointer is observed somewhere.
      */
-
     class Internal : public Error
     {
     public:
--- a/include/openPMD/Error.hpp
+++ b/include/openPMD/Error.hpp
@@ -69,5 +69,17 @@ namespace error
 
         BackendConfigSchema(std::vector<std::string>, std::string what);
     };
+
+    /**
+     * @brief Internal errors that should not happen. Please report.
+     *
+     * Example: A nullpointer is observed somewhere.
+     */
+
+    class Internal : public Error
+    {
+    public:
+        Internal(std::string const &what);
+    };
 } // namespace error
 } // namespace openPMD
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -121,6 +121,23 @@ namespace internal
  */
 class AbstractIOHandler
 {
+    friend class Series;
+
+private:
+    void setIterationEncoding(IterationEncoding encoding)
+    {
+        /*
+         * In file-based iteration encoding, the APPEND mode is handled entirely
+         * by the frontend, the backend should just treat it as CREATE mode
+         */
+        if (encoding == IterationEncoding::fileBased &&
+            m_backendAccess == Access::APPEND)
+        {
+            // do we really want to have those as const members..?
+            *const_cast<Access *>(&m_backendAccess) = Access::CREATE;
+        }
+    }
+
 public:
 #if openPMD_HAVE_MPI
     AbstractIOHandler(std::string path, Access at, MPI_Comm)
@@ -153,6 +170,7 @@ public:
     virtual std::string backendName() const = 0;
 
     std::string const directory;
+    // why do these need to be separate?
     Access const m_backendAccess;
     Access const m_frontendAccess;
     std::queue<IOTask> m_work;
--- a/src/Error.cpp
+++ b/src/Error.cpp
@@ -45,5 +45,9 @@ namespace error
               concatVector(errorLocation_in) + "': " + std::move(what))
         , errorLocation(std::move(errorLocation_in))
     {}
+
+    Internal::Internal(std::string const &what)
+        : Error("Internal error: " + what + "\nThis is a bug. Please report.")
+    {}
 } // namespace error
 } // namespace openPMD
--- a/src/Error.cpp
+++ b/src/Error.cpp
@@ -47,7 +47,10 @@ namespace error
     {}
 
     Internal::Internal(std::string const &what)
-        : Error("Internal error: " + what + "\nThis is a bug. Please report.")
+        : Error(
+              "Internal error: " + what +
+              "\nThis is a bug. Please report at ' "
+              "https://github.com/openPMD/openPMD-api/issues'.")
     {}
 } // namespace error
 } // namespace openPMD
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -315,9 +315,6 @@ void ADIOS2IOHandlerImpl::createFile(
         m_iterationEncoding = parameters.encoding;
         associateWithFile(writable, shared_name);
         this->m_dirty.emplace(shared_name);
-        getFileData(shared_name, IfFileNotOpen::OpenImplicitly).m_mode =
-            adios2::Mode::Write; // WORKAROUND
-        // ADIOS2 does not yet implement ReadWrite Mode
 
         writable->written = true;
         writable->abstractFilePosition = std::make_shared<ADIOS2FilePosition>();
@@ -1074,21 +1071,16 @@ adios2::Mode ADIOS2IOHandlerImpl::adios2AccessMode(std::string const &fullPath)
         if (auxiliary::directory_exists(fullPath) ||
             auxiliary::file_exists(fullPath))
         {
-            std::cerr << "ADIOS2 does currently not yet implement ReadWrite "
-                         "(Append) mode. "
-                      << "Replacing with Read mode." << std::endl;
             return adios2::Mode::Read;
         }
         else
         {
-            std::cerr << "ADIOS2 does currently not yet implement ReadWrite "
-                         "(Append) mode. "
-                      << "Replacing with Write mode." << std::endl;
             return adios2::Mode::Write;
         }
-    default:
-        return adios2::Mode::Undefined;
+    case Access::APPEND:
+        return adios2::Mode::Append;
     }
+    throw std::runtime_error("Unreachable!");
 }
 
 json::TracingJSON ADIOS2IOHandlerImpl::nullvalue = {
@@ -2235,6 +2227,7 @@ namespace detail
                         delayOpeningTheFirstStep = true;
                         break;
                     case adios2::Mode::Write:
+                    case adios2::Mode::Append:
                         /*
                          * File engines, write mode:
                          * Default for old layout is no steps.
@@ -2442,6 +2435,7 @@ namespace detail
         {
             switch (m_mode)
             {
+            case adios2::Mode::Append:
             case adios2::Mode::Write: {
                 // usesSteps attribute only written upon ::advance()
                 // this makes sure that the attribute is only put in case
@@ -2686,17 +2680,14 @@ namespace detail
                 switch (ba.m_mode)
                 {
                 case adios2::Mode::Write:
+                case adios2::Mode::Append:
                     eng.PerformPuts();
                     break;
                 case adios2::Mode::Read:
                     eng.PerformGets();
                     break;
-                case adios2::Mode::Append:
-                    // TODO order?
-                    eng.PerformGets();
-                    eng.PerformPuts();
-                    break;
                 default:
+                    throw error::Internal("[ADIOS2] Unexpected access mode.");
                     break;
                 }
             },
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -237,13 +237,40 @@ void HDF5IOHandlerImpl::createFile(
         std::string name = m_handler->directory + parameters.name;
         if (!auxiliary::ends_with(name, ".h5"))
             name += ".h5";
-        unsigned flags;
-        if (m_handler->m_backendAccess == Access::CREATE)
+        unsigned flags{};
+        switch (m_handler->m_backendAccess)
+        {
+        case Access::CREATE:
             flags = H5F_ACC_TRUNC;
-        else
+            break;
+        case Access::APPEND:
+            if (auxiliary::file_exists(name))
+            {
+                flags = H5F_ACC_RDWR;
+            }
+            else
+            {
+                flags = H5F_ACC_TRUNC;
+            }
+            break;
+        case Access::READ_WRITE:
             flags = H5F_ACC_EXCL;
-        hid_t id =
-            H5Fcreate(name.c_str(), flags, H5P_DEFAULT, m_fileAccessProperty);
+            break;
+        case Access::READ_ONLY:
+            // condition has been checked above
+            throw std::runtime_error("Control flow error");
+        }
+
+        hid_t id{};
+        if (flags == H5F_ACC_RDWR)
+        {
+            id = H5Fopen(name.c_str(), flags, m_fileAccessProperty);
+        }
+        else
+        {
+            id = H5Fcreate(
+                name.c_str(), flags, H5P_DEFAULT, m_fileAccessProperty);
+        }
         VERIFY(id >= 0, "[HDF5] Internal error: Failed to create HDF5 file");
 
         writable->written = true;
@@ -702,7 +729,14 @@ void HDF5IOHandlerImpl::openFile(
     Access at = m_handler->m_backendAccess;
     if (at == Access::READ_ONLY)
         flags = H5F_ACC_RDONLY;
-    else if (at == Access::READ_WRITE || at == Access::CREATE)
+    /*
+     * Within the HDF5 backend, APPEND and READ_WRITE mode are
+     * equivalent, but the openPMD frontend exposes no reading
+     * functionality in APPEND mode.
+     */
+    else if (
+        at == Access::READ_WRITE || at == Access::CREATE ||
+        at == Access::APPEND)
         flags = H5F_ACC_RDWR;
     else
         throw std::runtime_error("[HDF5] Unknown file Access");
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -258,7 +258,8 @@ void HDF5IOHandlerImpl::createFile(
             break;
         case Access::READ_ONLY:
             // condition has been checked above
-            throw std::runtime_error("Control flow error");
+            throw std::runtime_error(
+                "[HDF5] Control flow error in createFile backend access mode.");
         }
 
         hid_t id{};
@@ -470,8 +471,8 @@ void HDF5IOHandlerImpl::createDataset(
                 status = H5Ldelete(node_id, name.c_str(), H5P_DEFAULT);
                 VERIFY(
                     status == 0,
-                    "[HDF5] Internal error: Failed to delete old dataset from "
-                    "group for overwriting.");
+                    "[HDF5] Internal error: Failed to delete old dataset '" +
+                        name + "' from group for overwriting.");
                 break;
             }
         }
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -436,6 +436,46 @@ void HDF5IOHandlerImpl::createDataset(
             "[HDF5] Internal error: Failed to open HDF5 group during dataset "
             "creation");
 
+        if (m_handler->m_backendAccess == Access::APPEND)
+        {
+            // The dataset might already exist in the file from a previous run
+            // We delete it, otherwise we could not create it again with
+            // possibly different parameters.
+            // This is inefficient, as only the link to the dataset will be
+            // removed, but not the actual dataset
+            // But such is life if overwriting an iteration in Append mode
+            H5G_info_t group_info;
+            herr_t status = H5Gget_info(node_id, &group_info);
+            VERIFY(
+                status == 0,
+                "[HDF5] Internal error: Failed to get HDF5 group info for " +
+                    concrete_h5_file_position(writable) +
+                    " during dataset creation");
+            for (hsize_t i = 0; i < group_info.nlinks; ++i)
+            {
+                if (H5G_DATASET != H5Gget_objtype_by_idx(node_id, i))
+                {
+                    continue;
+                }
+                ssize_t name_length =
+                    H5Gget_objname_by_idx(node_id, i, nullptr, 0);
+                std::vector<char> charbuffer(name_length + 1);
+                H5Gget_objname_by_idx(
+                    node_id, i, charbuffer.data(), name_length + 1);
+                if (std::strncmp(
+                        name.c_str(), charbuffer.data(), name_length + 1) != 0)
+                {
+                    continue;
+                }
+                status = H5Ldelete(node_id, name.c_str(), H5P_DEFAULT);
+                VERIFY(
+                    status == 0,
+                    "[HDF5] Internal error: Failed to delete old dataset from "
+                    "group for overwriting.");
+                break;
+            }
+        }
+
         Datatype d = parameters.dtype;
         if (d == Datatype::UNDEFINED)
         {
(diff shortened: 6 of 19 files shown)


ax3l:
> Checkpoint-restarting. Probably one of the main uses of this. Dump every 100 time steps, checkpoint every 1000, crash at 1500, restart from 1000. Consequence: Steps {1000,1100,…,1500} are written twice. See #949 for an exploration on how to deal with this, but it's not really ideal.

I think useful would be a mode that overwrites a step exactly when it is re-opened again, but leaves older (and even newer) steps as they are. That's what people most expect and know from file-based output: files stay around until you overwrite one by one.

Since output from checkpoint-restart should give again the same result as the prior run, this is also ok (no "invalid" older files, so to say).

Alternatively, one could at a certain point (e.g. once during re-open/append) delete all steps _after_ the restart step.

franzpoeschel:
> I think useful would be a mode that overwrites a step exactly when it is re-opened again, but leaves older (and even newer) steps as they are. That's what people most expect and know from file-based output: files stay around until you overwrite one by one.

That is possible to implement with file-based iteration encoding and APPEND mode, but it won't be possible with either group- or variable-based encoding. Adding steps to an existing file in ADIOS2 allows no reading or modifying of old steps, you're just appending to a BP file without caring what had previously been written to it. 

The problem is not file-based encoding, but variable-/step-based encodings with steps enabled *for output data* after loading a checkpoint.

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -97,6 +97,25 @@ enum class FlushLevel : unsigned char
...
+         * In file-based iteration encoding, the APPEND mode is handled entirely
+         * by the frontend, the backend should just treat it as CREATE mode
+         * @todo maybe replace the encoding parameter of the OPEN_FILE task
+         *     with this?
+         */
+        if( encoding == IterationEncoding::fileBased &&
+            m_backendAccess == Access::APPEND )
+        {
+            // do we really want to have those as const members..?
+            *const_cast< Access * >( &m_backendAccess ) = Access::CREATE;
```
franzpoeschel: I have now added some handling in the backend so that this isn't necessary any more. Could remove it.

ax3l:
> The problem is not file-based encoding, but variable-/step-based encodings with steps enabled for output data after loading a checkpoint.

Uff, right. I put this on the ADIOS2 agenda. I think we could try to gracefully warn and skip such outputs in user-code if they exist, but definitely have to think about a workflow.

franzpoeschel:
~~I quickly deleted the member and looked where compilation failed. It seems that there currently is no other bug of the sort.~~ wrong PR

ax3l:
Discussion update:
- RFE with ADIOS2: Append mode, with parameter to truncate after the Nth (ADIOS) step: https://github.com/ornladios/ADIOS2/issues/2775
- high-level solution: using the openPMD “step” variable to list the same step again and our logic picks up the “latest” one in reads (will work, but accumulates data if steps are written multiple times)

franzpoeschel:
I think this is ready for review @ax3l 
Note that there is further info found in the added documentation, otherwise I have tried to split this into a sane number of commits.

include/openPMD/IO/AbstractIOHandler.hpp
franzpoeschel: Update: Eh, let' keep it. Clean overwriting is better and in file-based encoding it's easy to achieve.

franzpoeschel:
I've done further cleanup now. If it runs green, ready for review @ax3l

docs/source/usage/workflow.rst
```diff
@@ -1,5 +1,38 @@
 .. _workflow:
 
+Access modes
+============
+
+The openPMD-api distinguishes between a number of different access modes:
+
+* **Create mode**: Used for creating a new Series from scratch.
+  Any file possibly existing in the specified location will be overwritten.
```
ax3l, line 9: For file-based I/O, I think there might be a potentially confusing corner case:
if one writes the same series again, but with different interval, then some files might be overwritten and others untouched.

I am thinking about this from time to time and see pro and cons for, e.g., actively cleaning/moving out all potential files (even for iterations that are not created again) when _opening_ in truncating create mode.

More of a thing we could discuss next time.

docs/source/usage/workflow.rst
```diff
@@ -1,5 +1,38 @@
...
+* **Read-only mode**: Used for reading from an existing Series.
+  No modifications will be made.
+* **Read/Write mode**: Creates a new Series if not existing, otherwise opens an existing Series for reading and writing.
+  New datasets and iterations will be inserted as needed.
+  Not fully supported by all backends:
+  * ADIOS1: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
```
ax3l, line 16: rst lists are weird: newline needed before starting them:
```suggestion

  * ADIOS1: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
```

docs/source/usage/workflow.rst
```diff
@@ -1,5 +1,38 @@
...
+  * ADIOS2: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
+    Since this happens on a per-file level, this mode allows to read from existing iterations and write to new iterations at the same time in file-based iteration encoding.
+* **Append mode**: Restricted mode for appending new iterations to an existing Series that is supported by all backends at least in file-based iteration encoding, and by all but ADIOS1 in other encodings.
+  The API is equivalent to that of the *Create* mode, meaning that no reading is supported whatsoever.
+  If the Series does not exist yet, this behaves equivalently to the *Create* mode.
+  Existing iterations will not be deleted, newly-written iterations will be inserted.
+
+  **Warning:** If writing an iteration that already exists, the behavior is implementation-defined and depends on the chosen backend and iteration encoding:
```
ax3l:
```suggestion
  **Warning:** When writing an iteration that already exists, the behavior is implementation-defined and depends on the chosen backend and iteration encoding:
```

docs/source/usage/workflow.rst
```diff
@@ -1,5 +1,38 @@
...
+* **Create mode**: Used for creating a new Series from scratch.
+  Any file possibly existing in the specified location will be overwritten.
+* **Read-only mode**: Used for reading from an existing Series.
+  No modifications will be made.
+* **Read/Write mode**: Creates a new Series if not existing, otherwise opens an existing Series for reading and writing.
+  New datasets and iterations will be inserted as needed.
+  Not fully supported by all backends:
+  * ADIOS1: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
+  * ADIOS2: Automatically coerced to *Create* mode if the file does not exist yet and to *Read-only* mode if it exists.
+    Since this happens on a per-file level, this mode allows to read from existing iterations and write to new iterations at the same time in file-based iteration encoding.
```
ax3l:
```suggestion
  Since this happens on a per-file level, this mode allows to read from existing iterations and write to new iterations at the same time in file-based iteration encoding.
```

docs/source/usage/workflow.rst
```diff
@@ -1,5 +1,38 @@
...
+  Existing iterations will not be deleted, newly-written iterations will be inserted.
+
+  **Warning:** If writing an iteration that already exists, the behavior is implementation-defined and depends on the chosen backend and iteration encoding:
+
+  * The new iteration might fully replace the old one.
+  * The new iteration might be merged into the old one.
+  * (To be removed in a future update) The old and new iteration might coexist in the resulting dataset.
+
+  We suggest to fully define iterations when using Append mode (i.e. as if using Create mode) to avoid implementation-specific behavior.
+  Appending to an openPMD Series is only supported on a per-iteration level.
```
ax3l, line 31: Maybe a bit too strict here to understand what is meant or might be repetitive with what is said above?
Because one could add here: `Extending the shape of existing data sets is possible, see the examples.`

include/openPMD/Error.hpp
```diff
@@ -62,5 +62,17 @@ namespace error
     public:
         WrongAPIUsage( std::string what );
     };
+
+    /**
+     * @brief Internal errors that should not happen. Please report.
+     *
+     * Example: A nullpointer is observed somewhere.
+     */
+
```
ax3l:
```suggestion
```

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -131,6 +148,7 @@ class AbstractIOHandler
     virtual std::string backendName() const = 0;
 
     std::string const directory;
+    // why do these need to be separate?
```
ax3l, line 173: Do we want to answer this question? :)

I am not sure we documented this well when this was introduced (also visible in the missing doxygen strings for those two members).

src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -1077,6 +1078,14 @@ namespace openPMD
         {
             case Access::CREATE:
             case Access::READ_WRITE:
+            case Access::APPEND:
+                /*
+                 * Always truncate when writing, we alway write entire JSON
+                 * datasets, never partial ones.
```
ax3l: Writing with `storeChunk` can still be _partial_, even in serial writes. Thus the text here is a bit unclear in the argument, I think.

(I am ok with the logic to truncate.)

test/SerialIOTest.cpp
```diff
@@ -1629,8 +1629,24 @@ void fileBased_write_test(const std::string & backend)
         REQUIRE(o.iterations[5].time< double >() == 5.0);
     }
 
-    // extend existing series with new step and auto-detection of iteration padding
+    if( backend == "bp" )
     {
+        // Append + filebased iteration encoding works for all backends
```
ax3l: The `if "bp"` above is confusing (outdated?) if we want to run this for all backends now :)

test/SerialIOTest.cpp
```diff
@@ -1673,6 +1689,7 @@ void fileBased_write_test(const std::string & backend)
     }
 
     // write with auto-detection and in-consistent padding
+    if( backend != "bp" )
```
ax3l: Double-checking: still skip for `bp`?

include/openPMD/IO/AbstractIOHandler.hpp
guj: I ran into that a few times. I believe it would be consistent to treat all files in the file mode as they were one file. When its time to overwrite, remove all existing files related to the  series.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1063,21 +1061,16 @@ ADIOS2IOHandlerImpl::adios2AccessMode( std::string const & fullPath )
         if( auxiliary::directory_exists( fullPath ) ||
             auxiliary::file_exists( fullPath ) )
         {
-            std::cerr << "ADIOS2 does currently not yet implement ReadWrite "
-                         "(Append) mode. "
-                      << "Replacing with Read mode." << std::endl;
             return adios2::Mode::Read;
```
guj, line 1074: Now that the APPEND mode  is added.  Should READ_WRITE mode be rejected for ADIOS?  Might be semantically cleaner.

docs/source/usage/workflow.rst
franzpoeschel: In that case, let's maybe restrict the wording to make clear that this refers to the Append mode only:

> The Append mode is intended for adding new iterations to a Series, not for the modification of existing iterations.

include/openPMD/IO/AbstractIOHandler.hpp
franzpoeschel: > Do we want to answer this question? :)

I'd suggest to do it in a separate PR, need to figure out what the answer is at first though
I wasn't always quite sure which one of these to use when writing this PR

> I ran into that a few times. I believe it would be consistent to treat all files in the file mode as they were one file. When its time to overwrite, remove all existing files related to the series.

We could think about this for CREATE mode, since its semantics are "overwrite anything that exists". The purpose READ_WRITE and APPEND modes is explicitly *not* to delete existing data. For APPEND mode, the most sensible solution seems to be truncation (i.e. "delete anything past iteration 500") which would be a follow-up PR.

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: READ_WRITE is still useful for file-based iteration encoding since it allows you to read old iterations while at the same time to write new ones. This feature is not found in any other iteration encodings. For the other iteration encodings, there is no use in having READ_WRITE mode for ADIOS2, yeah.
We should however not deactivate it immediately, but allow users to transition to APPEND mode since it's a breaking change.

src/IO/JSON/JSONIOHandlerImpl.cpp
franzpoeschel: What I mean in here is that this backend always re-writes the entire JSON file and does not just update parts of it. This makes it ok to just truncate the file. So maybe:
```suggestion
                 * Always truncate when writing, we alway write entire JSON
                 * files, never partial ones.
```

test/SerialIOTest.cpp
franzpoeschel: Outdated, yeah
While working on the PR, read_write mode was for a while completely inaccessible for ADIOS2 because I had planned to replace it with Append mode entirely
Then I noticed that there are still reasons to use read_write with ADIOS2 in file-based iteration encoding, so those specialized test paths can now be deleted again

ax3l:
@franzpoeschel Just for after the holidays: This one needs a rebase please :)

The first commit (`error::Internal`) is probably already merged via #1037, and a few of the other commits have tiny merge conflicts, I think.

franzpoeschel:
> @franzpoeschel Just for after the holidays: This one needs a rebase please :)
> 
> The first commit (`error::Internal`) is probably already merged via #1037, and a few of the other commits have tiny merge conflicts, I think.

I've rebased this now. The introduced error types are different ones.

test/SerialIOTest.cpp
franzpoeschel: I've removed the skip now

docs/source/usage/workflow.rst
franzpoeschel: Fair point. I guess the most sensible solution would be within the bounds of a potential truncate option.

ax3l:
@franzpoeschel please rebase due to clang-format changes

You might find this helpful: https://github.com/pre-commit/pre-commit/issues/1696#issuecomment-974621448 :)

franzpoeschel:
> @franzpoeschel please rebase due to clang-format changes

Done :)
> 
> You might find this helpful: [pre-commit/pre-commit#1696 (comment)](https://github.com/pre-commit/pre-commit/issues/1696#issuecomment-974621448) :)

I've used [this script](https://github.com/franzpoeschel/reformat-rebase) which entirely avoids pre-commit getting active at all

franzpoeschel:
clang-6 is being a bit weird with checking for exceptions:
```
 /home/runner/work/openPMD-api/openPMD-api/test/SerialIOTest.cpp:5573: FAILED:
  REQUIRE_THROWS_AS( write.flush(), error::OperationUnsupportedInBackend )
  Operation unsupported in ADIOS1: Appending to existing file on disk (use
  Access::CREATE to overwrite)
```

And the Windows / MSVC errors I don't understand at all tbf
But I'll try to have a look

UPDATE:

The MSVC errors seem to stem from pybind: https://github.com/microsoft/onnxruntime/issues/9735#issuecomment-969577865
>everybody using pybind11 with VS2022 and STL should hit the issue.

share/openPMD/thirdParty/pybind11/include/pybind11/pybind11.h
```diff
@@ -10,6 +10,24 @@
 
 #pragma once
 
+/*
```
franzpoeschel: This one seems to fix the CI issues that I saw. Might want to put it in a separate PR and merge it with the MSVC block below.

franzpoeschel:
Many CI runs currently show this behavior:
```
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
read missing: successfully caught! json
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
read missing: successfully caught! h5
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: ../../../src/H5F.c line 509 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: ../../../src/H5Fint.c line 1400 in H5F__open(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: ../../../src/H5Fint.c line 1546 in H5F_open(): unable to open file: time = Wed Apr  6 14:19:34 2022
, name = '../samples/git-sample/data00000100.h5', tent_flags = 1
    major: File accessibilty
    minor: Unable to open file
  #003: ../../../src/H5FD.c line 734 in H5FD_open(): open failed
    major: Virtual File Layer
  #004: ../../../src/H5FDsec2.c line 346 in H5FD_sec2_open(): unable to open file: name = '../samples/git-sample/data00000100.h5', errno = 13, error message = 'Permission denied', flags = 1, o_flags = 2
    major: File accessibilty
    minor: Unable to open file
openPMD series: empty_alternate_fbpic_%T
openPMD standard: 1.0.0
openPMD extensions: 1

number of iterations: 1

number of meshes: 4

number of particle species: 1
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
[AbstractIOHandlerImpl] IO Task CLOSE_FILE failed with exception. Removing task from IO queue and passing on the exception.

SerialIOTests is a Catch v2.13.4 host application.
Run with -? for options

git_hdf5_sample_fileBased_read_test
/home/runner/work/openPMD-api/openPMD-api/test/SerialIOTest.cpp:2975

/home/runner/work/openPMD-api/openPMD-api/test/SerialIOTest.cpp:2975: FAILED:
  [HDF5] Trying to close a file that is not present in the backend
```
No idea what's going on yet, will have a look

franzpoeschel:
Ok, I get nearly exactly the same error message locally when I remove my reading rights via `chmod a-r`:
```
Filters: git_hdf5_sample_fileBased_read_test
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: ../../../src/H5F.c line 509 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: ../../../src/H5Fint.c line 1400 in H5F__open(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: ../../../src/H5Fint.c line 1546 in H5F_open(): unable to open file: time = Thu Apr  7 10:52:21 2022
, name = '../samples/git-sample/data00000100.h5', tent_flags = 0
    major: File accessibilty
    minor: Unable to open file
  #003: ../../../src/H5FD.c line 734 in H5FD_open(): open failed
    major: Virtual File Layer
  #004: ../../../src/H5FDsec2.c line 346 in H5FD_sec2_open(): unable to open file: name = '../samples/git-sample/data00000100.h5', errno = 13, error message = 'Permission denied', flags = 0, o_flags = 0
    major: File accessibilty
    minor: Unable to open file
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
[AbstractIOHandlerImpl] IO Task CLOSE_FILE failed with exception. Removing task from IO queue and passing on the exception.

SerialIOTests is a Catch v2.13.4 host application.
Run with -? for options

git_hdf5_sample_fileBased_read_test
/home/franzpoeschel/git-repos/openPMD-api/test/SerialIOTest.cpp:2975

/home/franzpoeschel/git-repos/openPMD-api/test/SerialIOTest.cpp:2975: FAILED:
  [HDF5] Trying to close a file that is not present in the backend
```

Seems like a further CI issue to me..

ax3l:
I think the problem here is usually if we try to open a file in "write" mode, although it should be read-only.
We set HDF5 files to "read-only" with `chmod` to catch that bug: #394 / #460 - we want to make sure that we do not accidentally trigger "write" mode if we only read (`+r-w` permissions _must_ be sufficient).

franzpoeschel:
> I think the problem here is usually if we try to open a file in "write" mode, although it should be read-only. We set HDF5 files to "read-only" with `chmod` to catch that bug: #394 / #460 - we want to make sure that we do not accidentally trigger "write" mode if we only read (`+r-w` permissions _must_ be sufficient).

Ah, so there's a potential bug to hunt down. Will check it.

franzpoeschel:
Ok, by taking my writing permissions I can reproduce the same issue offline:

```
Filters: git_hdf5_sample_fileBased_read_test
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: ../../../src/H5F.c line 509 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: ../../../src/H5Fint.c line 1400 in H5F__open(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: ../../../src/H5Fint.c line 1546 in H5F_open(): unable to open file: time = Fri Apr 29 14:11:16 2022
, name = '../samples/git-sample/data00000100.h5', tent_flags = 1
    major: File accessibilty
    minor: Unable to open file
  #003: ../../../src/H5FD.c line 734 in H5FD_open(): open failed
    major: Virtual File Layer
  #004: ../../../src/H5FDsec2.c line 346 in H5FD_sec2_open(): unable to open file: name = '../samples/git-sample/data00000100.h5', errno = 13, error message = 'Permission denied', flags = 1, o_flags = 2
    major: File accessibilty
    minor: Unable to open file
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
[AbstractIOHandlerImpl] IO Task CLOSE_FILE failed with exception. Removing task from IO queue and passing on the exception.
```

However, the failing line is this one:
```cpp
            Series o =
                Series("../samples/git-sample/data%T.h5", Access::READ_WRITE);
```
So, read-write access mode is entirely justified?

franzpoeschel:
The `dev` branch gives me the same error, but lets the test pass:

```
Filters: git_hdf5_sample_fileBased_read_test
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 0:
  #000: ../../../src/H5F.c line 509 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: ../../../src/H5Fint.c line 1400 in H5F__open(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: ../../../src/H5Fint.c line 1546 in H5F_open(): unable to open file: time = Fri Apr 29 14:23:12 2022
, name = '../samples/git-sample/data00000100.h5', tent_flags = 1
    major: File accessibilty
    minor: Unable to open file
  #003: ../../../src/H5FD.c line 734 in H5FD_open(): open failed
    major: Virtual File Layer
  #004: ../../../src/H5FDsec2.c line 346 in H5FD_sec2_open(): unable to open file: name = '../samples/git-sample/data00000100.h5', errno = 13, error message = 'Permission denied', flags = 1, o_flags = 2
    major: File accessibilty
    minor: Unable to open file
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
git sample not accessible. ([HDF5] Failed to open HDF5 file ../samples/git-sample/data00000100.h5)
All tests passed (13 assertions in 1 test case)
```

franzpoeschel:
Ah, ok, something changed in the error handling and the error is replaced with a new one:
```
[AbstractIOHandlerImpl] IO Task OPEN_FILE failed with exception. Removing task from IO queue and passing on the exception.
[AbstractIOHandlerImpl] IO Task CLOSE_FILE failed with exception. Removing task from IO queue and passing on the exception.

SerialIOTests is a Catch v2.13.4 host application.
Run with -? for options

git_hdf5_sample_fileBased_read_test
/home/franzpoeschel/git-repos/openPMD-api/test/SerialIOTest.cpp:2989

/home/franzpoeschel/git-repos/openPMD-api/test/SerialIOTest.cpp:2989: FAILED:
  [HDF5] Trying to close a file that is not present in the backend
```

I'll check if I can fix that some way.

franzpoeschel:
Ready for review again :) @ax3l

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -409,6 +436,46 @@ void HDF5IOHandlerImpl::createDataset(
             "[HDF5] Internal error: Failed to open HDF5 group during dataset "
             "creation");
 
+        if (m_handler->m_backendAccess == Access::APPEND)
+        {
+            // The dataset might already exist in the file from a previous run
+            // We delete it, otherwise we could not create it again with
+            // possibly different parameters.
+            // This is inefficient, as only the link to the dataset will be
+            // removed, but not the actual dataset
```
franzpoeschel: Is it inefficient? I don't really know about HDF5's garbage collection mechanics. Do they need to be triggered manually? Do they apply in this situation?

ax3l (approved):
This looks great, thank you :tada:

ax3l:
Should we clarify the two discussion points first in the PR description?
first one: I guess yes.
second one: moved to https://github.com/ornladios/ADIOS2/issues/2775

src/IO/HDF5/HDF5IOHandler.cpp
ax3l: I think they are unlinked and not removed unless one does a HDF5 repack. I think that's also the right thing to do.

maybe @jeanbez knows is the below code looks solid to remove a HDF5 object? :)

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -409,6 +436,46 @@ void HDF5IOHandlerImpl::createDataset(
...
+                    node_id, i, charbuffer.data(), name_length + 1);
+                if (std::strncmp(
+                        name.c_str(), charbuffer.data(), name_length + 1) != 0)
+                {
+                    continue;
+                }
+                status = H5Ldelete(node_id, name.c_str(), H5P_DEFAULT);
+                VERIFY(
+                    status == 0,
+                    "[HDF5] Internal error: Failed to delete old dataset from "
```
ax3l: If you include `name` in the error message, the error gets more useful for users.

src/Error.cpp
```diff
@@ -45,5 +45,9 @@ namespace error
               concatVector(errorLocation_in) + "': " + std::move(what))
         , errorLocation(std::move(errorLocation_in))
     {}
+
+    Internal::Internal(std::string const &what)
+        : Error("Internal error: " + what + "\nThis is a bug. Please report.")
```
ax3l: add the URL where to report here?
`https://github.com/openPMD/openPMD-api/issues`

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -237,13 +237,40 @@ void HDF5IOHandlerImpl::createFile(
...
+            }
+            break;
+        case Access::READ_WRITE:
             flags = H5F_ACC_EXCL;
-        hid_t id =
-            H5Fcreate(name.c_str(), flags, H5P_DEFAULT, m_fileAccessProperty);
+            break;
+        case Access::READ_ONLY:
+            // condition has been checked above
+            throw std::runtime_error("Control flow error");
```
```suggestion
            throw std::runtime_error("[HDF5] Control flow error in createFile backend access mode.");
```

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: > I think that's also the right thing to do.

So you mean we should repack here?

src/Error.cpp
```diff
@@ -45,5 +45,12 @@ namespace error
               concatVector(errorLocation_in) + "': " + std::move(what))
         , errorLocation(std::move(errorLocation_in))
     {}
+
+    Internal::Internal(std::string const &what)
+        : Error(
+              "Internal error: " + what +
+              "\nThis is a bug. Please report at ' "
+              "https://github.com/openPMD/openPMD-api/issues'.")
```
ax3l, line 53: Hard to copy, did read `' url'.`
```suggestion
              "\nThis is a bug. Please report at "
              "https://github.com/openPMD/openPMD-api/issues")
```

src/IO/HDF5/HDF5IOHandler.cpp
ax3l: No, repack is an expensive operation one does in post. It copies everything.
Just calling delete here is the right thing to do. It might actually free disk space, but in doubt it is ok if it does not (and only hides the entry).

ax3l (approved):

ax3l:
@franzpoeschel please needs a rebase. Collided with #1264

src/IO/HDF5/HDF5IOHandler.cpp
jeanbez: @ax3l I was not sure, so I checked with Tang/Suren and they mentioned that when removing an object with H5Ldelete, the HDF5 library should be able to detect and recycle the file space when no other reference to the deleted object exists. Regarding the code below it's more efficient to do H5Lexists to test whether a group/dset/attr exists, than iterating over existing items in a group like the one in this code snippet.

ax3l: Thanks a lot for the details, @jeanbez.

Can we also simply unlink a group and HDF5 garbage collects all underlying inodes of that group? (Similar for data sets?) The challenge we have here is that a user might have added custom attributes that we do not know of.

franzpoeschel: The latest commit [7afb99d](https://github.com/openPMD/openPMD-api/pull/1007/commits/7afb99dbb71bde944a42969db99122732524222f) now uses H5Lexists, thanks for the tip!

jeanbez: @ax3l Yes, according to the HDF5 team the refcount will be decremented on the group when it's unlinked from its parent, and the group will be deleted if its refcount drops to zero.  If a group is deleted, it will unlink all of its children, possibly causing them to be deleted as well.

jeanbez: Update: empty space in the file can only be reused until the file is closed, unless persistent free space tracking is on (it is off by default)

PR #1009, opened 2021-06-11
Document ADIOS2 memory usage

franzpoeschel:
The mode of operation in ADIOS2 greatly influences the memory usage. This documents the common options and shows some graphs measured with PIConGPU and KDE Heaptrack.


docs/source/backends/adios2.rst
```diff
@@ -105,6 +105,47 @@ The new layout may be activated **for experimental purposes** in two ways:
...
+.. image:: ./memory_filebased.png
+  :alt: Memory usage of file-based iteration encoding
+
+* **variable-based iteration encoding and group-based iteration encoding with steps**:
+  One buffer is created and reused across all iterations.
+  It is drained to disk when closing a step.
+  If carefully selecting the correct ``InitialBufferSize``, this is merely one single allocation held across all iterations.
+  If selecting the ``InitialBufferSize`` too small, reallocations will occur.
+  As usual with ``std::vector`` (which ADIOS2 uses internally), a reallocation will occupy both the old and new memory for a short time, leading to small memory spikes.
+  These memory spikes can easily lead to OOM situations, motivating that the ``InitialBufferSize`` should not be chosen too small.
```
```suggestion
  These memory spikes can easily lead to out-of-memory (OOM) situations, motivating that the ``InitialBufferSize`` should not be chosen too small.
```

ax3l (approved):
Thank you! :tada: 

Since we use openPMD-api in super-builds and download it often on the fly, we need to make sure the repo size stays small, esp. with binaries. I thus I reduced the PNGs with a Linux tool (Ubuntu package: `pngquant`) that I found (lossy compression):
```bash
pngquant --speed 1 file.png -o file-out.png
```

I also added two tiny fixes to the text.

PR #1010, opened 2021-06-14
8 Parallel Benchmark: 3D

ax3l:
The 4D size of the parallel benchmark is not very representative of the layouts that we do in our primary applications. Thus, reshape to 3D (with same number of elements).

Instead of using this example, we can test arbitrary ND fields in tests.


franzpoeschel:
I actually stumbled upon this around a week ago when looking for a good example to show a user and had a similar thought that this should better be 3D. So, thanks for this PR :)

franzpoeschel (approved):

PR #1011, opened 2021-06-15
Support for custom geometries

franzpoeschel:
We have some use cases that need to define custom geometries as they don't really fit with any of the predefined ones. Example: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3560 (this needs support for binning and logarithmic scales).

Currently, setting a custom geometry is fine, but trying to read one fails. This PR adds a failing test and will add support for easier writing to and reading from datasets with custom geometries.

### DIFF
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -58,7 +58,7 @@ public:
         thetaMode,
         cylindrical,
         spherical,
-        custom
+        other
     };  //Geometry
 
     /** @brief Enumerated datatype for the memory layout of N-dimensional data.
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -88,7 +88,7 @@ public:
      * @param   geometry    geometry of the mesh of the mesh record, as string
      * @return  Reference to modified mesh.
      */
-    Mesh& setGeometry(std::string const& geometry);
+    Mesh& setGeometry(std::string geometry);
 
     /**
      * @throw   no_such_attribute_error If Mesh::geometry is not Mesh::Geometry::thetaMode.
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -57,7 +57,8 @@ public:
         cartesian,
         thetaMode,
         cylindrical,
-        spherical
+        spherical,
+        custom
     };  //Geometry
 
     /** @brief Enumerated datatype for the memory layout of N-dimensional data.
@@ -78,6 +79,12 @@ public:
      * @return  Reference to modified mesh.
      */
     Mesh& setGeometry(Geometry g);
+    /** Set the geometry of the mesh of the mesh record.
+     *
+     * @param   g    geometry of the mesh of the mesh record, as string
+     * @return  Reference to modified mesh.
+     */
+    Mesh& setGeometry(std::string const& geometry);
 
     /**
      * @throw   no_such_attribute_error If Mesh::geometry is not Mesh::Geometry::thetaMode.
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -84,6 +84,7 @@ public:
      */
     Mesh& setGeometry(Geometry g);
     /** Set the geometry of the mesh of the mesh record.
+     *
      * If the geometry is unknown to the openPMD-api, the string is prefixed
      * with "other:" automatically unless the prefix is already present.
      *
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -70,9 +70,13 @@ public:
     };  //DataOrder
 
     /**
-     * @return String representing the geometry of the mesh of the mesh record.
+     * @return Enum representing the geometry of the mesh of the mesh record.
      */
     Geometry geometry() const;
+    /**
+     * @return String representing the geometry of the mesh of the mesh record.
+     */
+    std::string geometryString() const;
     /** Set the geometry of the mesh of the mesh record.
      *
      * @param   g    geometry of the mesh of the mesh record.
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -84,6 +84,8 @@ public:
      */
     Mesh& setGeometry(Geometry g);
     /** Set the geometry of the mesh of the mesh record.
+     * If the geometry is unknown to the openPMD-api, the string is prefixed
+     * with "other:" automatically unless the prefix is already present.
      *
      * @param   geometry    geometry of the mesh of the mesh record, as string
      * @return  Reference to modified mesh.
--- a/include/openPMD/Mesh.hpp
+++ b/include/openPMD/Mesh.hpp
@@ -85,7 +85,7 @@ public:
     Mesh& setGeometry(Geometry g);
     /** Set the geometry of the mesh of the mesh record.
      *
-     * @param   g    geometry of the mesh of the mesh record, as string
+     * @param   geometry    geometry of the mesh of the mesh record, as string
      * @return  Reference to modified mesh.
      */
     Mesh& setGeometry(std::string const& geometry);
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -45,12 +45,17 @@ Mesh::Mesh()
 Mesh::Geometry
 Mesh::geometry() const
 {
-    std::string ret = getAttribute("geometry").get< std::string >();
+    std::string ret = geometryString();
     if( "cartesian" == ret ) { return Geometry::cartesian; }
     else if( "thetaMode" == ret ) { return Geometry::thetaMode; }
     else if( "cylindrical" == ret ) { return Geometry::cylindrical; }
     else if( "spherical" == ret ) { return Geometry::spherical; }
     else { return Geometry::custom; }
+}
+
+std::string Mesh::geometryString() const
+{
+    return getAttribute( "geometry" ).get< std::string >();
 }
 
 Mesh&
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -83,9 +83,9 @@ Mesh::setGeometry(Mesh::Geometry g)
     return *this;
 }
 
-Mesh & Mesh::setGeometry( std::string const & geometry )
+Mesh & Mesh::setGeometry( std::string geometry )
 {
-    setAttribute( "geometry", geometry );
+    setAttribute( "geometry", std::move( geometry ) );
     return *this;
 }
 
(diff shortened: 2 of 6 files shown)


include/openPMD/Mesh.hpp
```diff
@@ -69,15 +70,25 @@ class Mesh : public BaseRecord< MeshRecordComponent >
...
      * @param   g    geometry of the mesh of the mesh record.
      * @return  Reference to modified mesh.
      */
     Mesh& setGeometry(Geometry g);
+    /** Set the geometry of the mesh of the mesh record.
+     *
+     * @param   geometry    geometry of the mesh of the mesh record, as string
+     * @return  Reference to modified mesh.
+     */
+    Mesh& setGeometry(std::string const& geometry);
```
ax3l: Since we can move out strings now, the proper interface for this is likely by value?
```suggestion
    Mesh& setGeometry(std::string geometry);
```

src/Mesh.cpp
```diff
@@ -70,10 +75,20 @@ Mesh::setGeometry(Mesh::Geometry g)
...
             break;
+        case Geometry::custom:
+            // use the std::string overload to be more specific
+            setAttribute("geometry", std::string("custom"));
+            break;
     }
     return *this;
 }
 
+Mesh & Mesh::setGeometry( std::string const & geometry )
```
ax3l:
```suggestion
Mesh & Mesh::setGeometry( std::string geometry )
```

src/Mesh.cpp
```diff
@@ -70,10 +75,20 @@ Mesh::setGeometry(Mesh::Geometry g)
...
+{
+    setAttribute( "geometry", geometry );
```
ax3l:
```suggestion
    setAttribute( "geometry", std::move(geometry) );
```

include/openPMD/Mesh.hpp
```diff
@@ -57,7 +57,8 @@ class Mesh : public BaseRecord< MeshRecordComponent >
         cartesian,
         thetaMode,
         cylindrical,
-        spherical
+        spherical,
+        custom
```
ax3l: the naming we use in the openPMD standard is `other` for such kinds :)

src/Mesh.cpp
```diff
@@ -392,6 +407,9 @@ openPMD::operator<<(std::ostream& os, openPMD::Mesh::Geometry const& go)
         case openPMD::Mesh::Geometry::spherical:
             os<<"spherical";
             break;
+        case openPMD::Mesh::Geometry::custom:
+            os<<"custom";
```
ax3l:
```suggestion
        case openPMD::Mesh::Geometry::other:
            os<<"other";
```

src/binding/python/Mesh.cpp
```diff
@@ -47,7 +47,9 @@ void init_Mesh(py::module &m) {
             &Mesh::setUnitDimension,
             python::doc_unit_dimension)
 
-        .def_property("geometry", &Mesh::geometry, &Mesh::setGeometry)
+        .def_property("geometry", &Mesh::geometry, py::overload_cast<Mesh::Geometry>(&Mesh::setGeometry))
+        .def_property(
+            "geometry_string", &Mesh::geometryString, py::overload_cast<std::string const &>(&Mesh::setGeometry))
```
ax3l:
```suggestion
            "geometry_string", &Mesh::geometryString, py::overload_cast<std::string>(&Mesh::setGeometry))
```

test/CoreTest.cpp
```diff
@@ -841,3 +841,53 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+            "customGeometry" );
+        REQUIRE( B.geometry() == Mesh::Geometry::custom );
+        REQUIRE( B.geometryString() == "customGeometry" );
+
+        auto e_chargeDensity = read.iterations[ 0 ].meshes[ "e_chargeDensity" ];
+        REQUIRE(
+            e_chargeDensity.getAttribute( "geometry" ).get< std::string >() ==
+            "custom" );
+        REQUIRE( e_chargeDensity.geometry() == Mesh::Geometry::custom );
+        REQUIRE( e_chargeDensity.geometryString() == "custom" );
```
ax3l:
```suggestion
        REQUIRE( e_chargeDensity.geometryString() == "other" );
```

include/openPMD/Mesh.hpp
ax3l: https://github.com/openPMD/openPMD-standard/blob/upcoming-2.0.0/STANDARD.md#required-attributes-for-each-mesh-record

We can store this as follows: `other:<userString>`

ax3l:
Thanks for the addition! I added small suggestions to harmonize this a bit more with the standard and simplify the passed argument in the API

include/openPMD/Mesh.hpp
franzpoeschel: I used a const reference since that's what `Attributable::setAttribute` uses: https://github.com/openPMD/openPMD-api/blob/a5b026e861eb2906bf203915d20be30a359a9e0f/include/openPMD/backend/Attributable.hpp#L360
If we go with move semantics, we should update `setAttribute` accordingly

franzpoeschel:
The last commit uses move semantics. Note that I also changed the signature of `Attributable::setAttribute()`, otherwise it would have had no purpose.

src/binding/python/Mesh.cpp
```diff
@@ -81,5 +83,6 @@ void init_Mesh(py::module &m) {
         .value("thetaMode", Mesh::Geometry::thetaMode)
         .value("cylindrical", Mesh::Geometry::cylindrical)
         .value("spherical", Mesh::Geometry::spherical)
+        .value("other", Mesh::Geometry::other)
```
ax3l, line 87: can you please add a tiny python test using this interface in `test/python/unittest/API/APITest.py`?

test/CoreTest.cpp
```diff
@@ -841,3 +841,53 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+    {
+        Series write( "../samples/custom_geometry.json", Access::CREATE );
+        auto E = write.iterations[ 0 ].meshes[ "E" ];
+        E.setAttribute( "geometry", "customGeometry" );
+        auto E_x = E[ "x" ];
+        E_x.resetDataset( { Datatype::INT, { 10 } } );
+        E_x.storeChunk( sampleData, { 0 }, { 10 } );
+
+        auto B = write.iterations[ 0 ].meshes[ "B" ];
+        B.setGeometry( "customGeometry" );
```
ax3l, line 857:
```suggestion
        B.setGeometry( "other:customGeometry" );
```

test/CoreTest.cpp
```diff
@@ -841,3 +841,53 @@ TEST_CASE( "no_file_ending", "[core]" )
...
                         Catch::Equals("Unknown file format! Did you specify a file ending?"));
 }
+
+TEST_CASE( "custom_geometries", "[core]" )
+{
+    std::vector< int > sampleData( 10, 0 );
+    {
+        Series write( "../samples/custom_geometry.json", Access::CREATE );
+        auto E = write.iterations[ 0 ].meshes[ "E" ];
+        E.setAttribute( "geometry", "customGeometry" );
```
ax3l: I would use this in PIConGPU like this:
```suggestion
        E.setAttribute( "geometry", "other:customGeometry" );
```

test/CoreTest.cpp
```diff
@@ -841,3 +841,53 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+    }
+
+    {
+        Series read( "../samples/custom_geometry.json", Access::READ_ONLY );
+        auto E = read.iterations[ 0 ].meshes[ "E" ];
+        REQUIRE(
+            E.getAttribute( "geometry" ).get< std::string >() ==
+            "customGeometry" );
+        REQUIRE( E.geometry() == Mesh::Geometry::other );
+        REQUIRE( E.geometryString() == "customGeometry" );
```
ax3l:
```suggestion
        REQUIRE( E.geometryString() == "other:customGeometry" );
```

test/CoreTest.cpp
```diff
@@ -841,3 +841,53 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+
+        auto B = read.iterations[ 0 ].meshes[ "B" ];
+        REQUIRE(
+            B.getAttribute( "geometry" ).get< std::string >() ==
+            "customGeometry" );
+        REQUIRE( B.geometry() == Mesh::Geometry::other );
+        REQUIRE( B.geometryString() == "customGeometry" );
```
ax3l:
```suggestion
        REQUIRE( B.geometryString() == "other:customGeometry" );
```

test/CoreTest.cpp
franzpoeschel: Should we maybe prepend the prefix unconditionally? So that the call stays `B.setGeometry("customGeometry")`, but `"other:customGeometry"` is written?

ax3l: Yes, I thought that would be even better :)

Do we want users to be able to set also the defined geometries by string? Could be handy as well, then we would need to check against an allow-list before prepending the `other:`.

src/binding/python/Mesh.cpp
franzpoeschel: done

test/CoreTest.cpp
franzpoeschel: Yeah, I did exactly that now

test/CoreTest.cpp
franzpoeschel: I've left the test this way on purpose to demonstrate that `setAttribute` performs no automatic coercion and `setGeometry` should be used for that purpose.

franzpoeschel:
Why api:breaking though? This just adds new functionality, but does not alter any old ones. Or is this because of `setAttribute()`?

ax3l:
Addeds this because of the changed signature in `AttributableImpl::setAttribute`. It's not a biggy though, and we will land it in 0.14.0 anyway.

include/openPMD/Mesh.hpp
```diff
@@ -69,15 +70,27 @@ class Mesh : public BaseRecord< MeshRecordComponent >
...
+     */
+    std::string geometryString() const;
     /** Set the geometry of the mesh of the mesh record.
      *
      * @param   g    geometry of the mesh of the mesh record.
      * @return  Reference to modified mesh.
      */
     Mesh& setGeometry(Geometry g);
+    /** Set the geometry of the mesh of the mesh record.
+     * If the geometry is unknown to the openPMD-api, the string is prefixed
```
ax3l, line 88: Doxygen title/body is separated by a newline:
```suggestion
     *
     * If the geometry is unknown to the openPMD-api, the string is prefixed
```

ax3l (approved):
Thanks a lot! :tada:

PR #1013, opened 2021-06-16
Wheels: 0.13.4.post1

ax3l:
Try to mitigate import error of
```
undefined reference to `pthread_atfork'
```
on PPC64le.

I think this line in blosc is outdated:
https://github.com/Blosc/c-blosc/blob/7435f28dd08606bd51ab42b49b0e654547becac4/blosc/CMakeLists.txt#L68
Is: `CMAKE_THREAD_PREFER_PTHREAD` ([docs](https://cmake.org/cmake/help/v3.2/module/FindThreads.html))
Should be: `THREADS_PREFER_PTHREAD_FLAG` ([docs](https://cmake.org/cmake/help/latest/module/FindThreads.html))

See:
https://github.com/openPMD/openPMD-api/pull/1002#discussion_r652975641

Upstream: https://github.com/Blosc/c-blosc/pull/318

In the end, we fixed it due to an error in `FindBlosc.cmake` in ADIOS2: https://github.com/ornladios/ADIOS2/pull/2768

## To Do

- [ ] bump version (via patch)


library_builders.sh
```diff
@@ -137,7 +133,14 @@ function build_blosc {
     cd build-c-blosc
     PY_BIN=$(which python)
     CMAKE_BIN="$(${PY_BIN} -m pip show cmake 2>/dev/null | grep Location | cut -d' ' -f2)/cmake/data/bin/"
-    PATH=${CMAKE_BIN}:${PATH} cmake -DDEACTIVATE_SNAPPY=ON -DBUILD_SHARED=OFF -DBUILD_TESTS=OFF -DBUILD_BENCHMARKS=OFF -DCMAKE_INSTALL_PREFIX=${BUILD_PREFIX} ../c-blosc-*
+    PATH=${CMAKE_BIN}:${PATH} cmake          \
+      -DDEACTIVATE_SNAPPY=ON                 \
+      -DBUILD_SHARED=OFF                     \
+      -DBUILD_TESTS=OFF                      \
+      -DBUILD_BENCHMARKS=OFF                 \
+      -DTHREADS_PREFER_PTHREAD_FLAG=ON       \
```
ax3l:
```suggestion
      -DTHREADS_PREFER_PTHREAD_FLAG=ON       \
      -DCMAKE_VERBOSE_MAKEFILES=ON           \
```

library_builders.sh
```diff
@@ -137,7 +133,15 @@ function build_blosc {
...
+      -DCMAKE_VERBOSE_MAKEFILES=ON           \
```
ax3l:
```suggestion
      -DCMAKE_VERBOSE_MAKEFILE=ON            \
```

PR #1017, opened 2021-06-23
Always check the return status of IO::Open() and Engine::BeginStep() in ADIOS2

franzpoeschel:
Issue noticed [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/3631). We sometimes don't check the return status of `Engine::BeginStep()`, leading to confusing late errors.
Also, fix a similar bug to check whether an Adios Engine is successfully opened. I accidentally only checked whether I filled the containing `auxiliary::Option`.

Tbh., having a definition for `auxiliary::Option::operator bool()` calls for bugs of this kind, we could delete the operator and just use the existing `auxiliary::Option::has_value()` instead.

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -2476,7 +2476,13 @@ namespace detail
                         }
                         else
                         {
-                            m_engine.get().BeginStep();
+                            if( m_engine.get().BeginStep() !=
+                                adios2::StepStatus::OK )
+                            {
+                                throw std::runtime_error(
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you for the fix!

ax3l:
> Tbh., having a definition for auxiliary::Option::operator bool() calls for bugs of this kind, we could delete the operator and just use the existing auxiliary::Option::has_value() instead.

I saw the lines, it's a bit tricky. But on the other hand also part of `std::optional`...

franzpoeschel:
I quickly deleted the member and looked where compilation failed. It seems that there currently is no other bug of the sort.

PR #1018, opened 2021-06-23
Don't flush Series a second time after throwing an error

### LINKED ISSUE #1006: Series flushed twice if error occurs
When some error happens during `Series::flush()`, C++ will destroy all objects currently found on the stack, including the instance `Series`. `Series::~Series()` calls `Series::flush()`. Consequence: An error happening during `Series::flush()` will lead to `Series::flush()` being attempted a second time, on a broken object, often leading to completely different and mostly unrelated error messages.

This is annoying during development, I've often wasted time now looking into the wrong error that wasn't actually the problem. I imagine, having unrelated, seemingly random error messages is often not the best user experience either.

Suggestion: In `Series::flush`, catch any error, set a flag `unhandledError` or something. If that flag is found, the destructor does not call `Series::flush()`.

franzpoeschel:
Fix #1006. Remember in a flag whether the last flush was successful. If it was not, don't flush again upon `Series::~Series`.

Example: `close_iteration_test` was very noisy until now since it deliberately left the `Series` in an invalid state:
```
Filters: close_iteration_test
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
openPMD series: close_iterations_%T
openPMD standard: 1.1.0
openPMD extensions: 0

number of iterations: 2

number of meshes: 1

number of particle species: 0
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
openPMD series: close_iterations_%T
openPMD standard: 1.1.0
openPMD extensions: 0

number of iterations: 2

number of meshes: 1

number of particle species: 0
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
openPMD series: close_iterations_%T
openPMD standard: 1.1.0
openPMD extensions: 0

number of iterations: 2

number of meshes: 1

number of particle species: 0
[~Series] An error occurred: [Series] Detected illegal access to iteration that has been closed previously.
```

After this PR:
```
Filters: close_iteration_test
openPMD series: close_iterations_%T
openPMD standard: 1.1.0
openPMD extensions: 0

number of iterations: 2

number of meshes: 1

number of particle species: 0
openPMD series: close_iterations_%T
openPMD standard: 1.1.0
openPMD extensions: 0

number of iterations: 2

number of meshes: 1

number of particle species: 0
openPMD series: close_iterations_%T
openPMD standard: 1.1.0
openPMD extensions: 0

number of iterations: 2

number of meshes: 1

number of particle species: 0
```

### DIFF
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -349,7 +349,8 @@ OPENPMD_private:
     std::future< void > flush_impl(
         iterations_iterator begin,
         iterations_iterator end,
-        FlushLevel );
+        FlushLevel,
+        bool flushIOHandler );
     void flushFileBased( iterations_iterator begin, iterations_iterator end );
     /*
      * Group-based and variable-based iteration layouts share a lot of logic
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -347,10 +347,21 @@ OPENPMD_private:
     std::unique_ptr< ParsedInput > parseInput(std::string);
     void init(std::shared_ptr< AbstractIOHandler >, std::unique_ptr< ParsedInput >);
     void initDefaults( IterationEncoding );
+    /**
+     * @brief Internal call for flushing a Series.
+     *
+     * Any flushing of the Series will pass through this call.
+     * 
+     * @param begin Start of the range of iterations to flush.
+     * @param end End of the range of iterations to flush.
+     * @param level Flush level, as documented in AbstractIOHandler.hpp.
+     * @param flushIOHandler Tasks will always be enqueued to the backend.
+     *     If this flag is true, tasks will be flushed to the backend.
+     */
     std::future< void > flush_impl(
         iterations_iterator begin,
         iterations_iterator end,
-        FlushLevel,
+        FlushLevel level,
         bool flushIOHandler );
     void flushFileBased( iterations_iterator begin, iterations_iterator end );
     /*
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -96,6 +96,7 @@ public:
      */
     StepStatus m_stepStatus = StepStatus::NoStep;
     bool m_parseLazily = false;
+    bool m_lastFlushSuccessful = true;
 }; // SeriesData
 
 class SeriesInternal;
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -1424,6 +1424,14 @@ SeriesInternal::~SeriesInternal()
     // we must not throw in a destructor
     try
     {
+        /*
+         * Scenario: A user calls `Series::flush()` but does not check for thrown
+         * exceptions. The exception will propagate further up, usually thereby
+         * popping the stack frame that holds the `Series` object.
+         * `Series::~Series()` will run. This check avoids that the `Series` is
+         * needlessly flushed a second time. Otherwise, error messages can get
+         * very confusing.
+         */
         if( get().m_lastFlushSuccessful )
         {
             flush();
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -502,6 +502,8 @@ SeriesImpl::flush_impl(
     bool flushIOHandler )
 {
     IOHandler()->m_flushLevel = level;
+    auto & series = get();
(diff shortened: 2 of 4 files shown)


include/openPMD/Series.hpp
```diff
@@ -349,7 +350,8 @@ class SeriesImpl : public AttributableImpl
     std::future< void > flush_impl(
```
ax3l, line 361: Can you please add a doxygen string here, describing the parameters?

Do we really need a new parameter or can we access the member variable directly? It looks like we pass `true` most of the time?

src/Series.cpp
```diff
@@ -1223,28 +1236,8 @@ SeriesImpl::advance(
...
-        }
-        IOHandler()->m_flushLevel = oldFlushLevel;
-    }
-    catch( ... )
-    {
-        IOHandler()->m_flushLevel = oldFlushLevel;
-        throw;
-    }
+    flush_impl(
+        begin, end, FlushLevel::UserFlush, /* flushIOHandler = */ false );
```
ax3l, line 1242: Why don't we flush the IOHandler at this point?
Needs an inline comment, this gets quite involved.

include/openPMD/Series.hpp
franzpoeschel: > Do we really need a new parameter or can we access the member variable directly? It looks like we pass true most of the time?

This is not the new flag that I added. I needed to add this parameter so I could call `flush_impl()` from within `advance()`. This way, all flushing goes through `flush_impl()`, allowing me to add this logic here only once.

src/Series.cpp
franzpoeschel: I've added/clarified the comment now. Note that the logic of `advance` hasn't changed, it was just so far a slightly modified manual version of `flush_impl()`, but I have now added this workflow to `flush_impl()`.

include/openPMD/Series.hpp
```diff
@@ -346,10 +347,22 @@ class SeriesImpl : public AttributableImpl
...
+     * @param level Flush level, as documented in AbstractIOHandler.hpp.
+     * @param flushIOHandler Tasks will always be enqueued to the backend.
+     *     If this flag is true, tasks will be flushed to the backend.
+     */
     std::future< void > flush_impl(
         iterations_iterator begin,
         iterations_iterator end,
-        FlushLevel );
+        FlushLevel level,
+        bool flushIOHandler );
```
ax3l, line 365: Do we want to add `= true` as default parameter, since this is usually what we need and expect?
That was the old behavior and seems logical that `flush_impl()` calls `flush()`?

That way we could also keep all but one call to it unchanged.

What do you think?

include/openPMD/Series.hpp
ax3l: I don't understand :)
Isn't `bool flushIOHandler` the new flag?

ax3l:
The current doc update already helps - let's discuss the leftover question in person in our next meeting :)

Thanks again for the fix!

ax3l (approved):

include/openPMD/Series.hpp
franzpoeschel: `bool flushIOHandler` and `bool SeriesData::m_lastFlushSuccessful` are two different things. `flushIOHandler` is not manipulated, only read.

include/openPMD/Series.hpp
franzpoeschel: On the one hand, I like to keep those things explicit in internal usage (example: I already had a merge conflict where I had added different flags to the same function, but since those flags had no defaults, git showed me the conflict for each invocation, making things easier to fix). 
However, there's a very clear call for a default in this instance, so I can do a follow-up PR and add it if you like?

include/openPMD/Series.hpp
ax3l: discussed offline :+1:

include/openPMD/Series.hpp
ax3l: Sounds good!

PR #1019, opened 2021-06-24
Test for ADIOS2 Git sample

franzpoeschel:
Add testing for the ADIOS2 example data files: https://github.com/openPMD/openPMD-example-datasets/pull/14

- [x] Test attributes
- [x] Test content

This does not test attributes as extensively as in the HDF5 sample, since we don't need to test the entire openPMD hierarchy a second time. Hence, some but not all meshes and particles are tested.

Also, I noticed that our HDF5 sample has some wrong unitDimensions.


ax3l (approved):
Thanks a lot!

I added windows support and some cleanups :)

ax3l:
Dang, I just found a bug in that data set: https://github.com/openPMD/openPMD-example-datasets/pull/14#issuecomment-873337294

Try to:
- determine the extend of a constant record component (particle mass/charge), it should be the same size as the non-constant records
- try to read the constant record component (mass or charge) like a non-constant one

PR #1022, opened 2021-06-25
Fix detection of loadChunk() calls with wrong type

franzpoeschel:
First commit adds a failing test, second commit will add a fix


sbastrakov:
Wow, i've faced this just a couple hours ago. It can be rather nasty, as currently `loadChunk` does not throw when types are not matched, but returns a pointer which corrupts the heap when destroyed.

ax3l (approved):
Oh, tricky.
Thanks for adding a test as well!

PR #1030, opened 2021-07-07
Bug fix: Use Series::writeIterations() without explicit flushing

franzpoeschel:
First commit adds a failing test
Second commit will add a fix

Probably related: https://github.com/openPMD/openPMD-api/issues/994

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -4297,3 +4297,54 @@ TEST_CASE( "deferred_parsing", "[serial]" )
         deferred_parsing( t );
     }
 }
+
+// @todo merge this back with the chaotic_stream test of PR #949
+// (bug noticed while working on that branch)
+void no_explicit_flush( std::string filename )
+{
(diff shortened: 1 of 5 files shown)


ax3l (approved):
Thank you! :tada:

test/SerialIOTest.cpp
```diff
@@ -4297,3 +4297,54 @@ TEST_CASE( "deferred_parsing", "[serial]" )
...
+        REQUIRE( index == 10 );
+    }
+}
+
+TEST_CASE( "no_explicit_flush", "[serial]" )
+{
+    for( auto const & t : testedFileExtensions() )
+    {
+        no_explicit_flush( "../samples/chaotic_stream_filebased_%T." + t );
+        no_explicit_flush( "../samples/chaotic_stream." + t );
```
ax3l: Do you want to rename those to
```suggestion
        no_explicit_flush( "../samples/no_explicit_flush_filebased_%T." + t );
        no_explicit_flush( "../samples/no_explicit_flush." + t );
```
?

I just wonder if it's a leftover from extracting it from the other branch.

PR #1035, opened 2021-07-08
availableChunks open logic (HDF5: Early Chunk Read)

### LINKED ISSUE #961: Early Query: availableChunks & HDF5
**Describe the bug**
It seems that querying `availableChunks` early in a HDF5 file during read causes an issue with opening logic.

**To Reproduce**

```C++

#include <iostream>


using std::cout;
using namespace openPMD;

int main()
{
    Series s = Series(
        "../samples/git-sample/data%T.h5",
        Access::READ_ONLY
    );

    auto electrons = s.iterations[400].particles["electrons"];

    for( auto & r : electrons )
    {
        cout << r.first << ": ";
        for( auto & r_c : r.second )
        {
            cout << r_c.first << "\n";
            if( !r_c.second.constant() )
                auto chunks = r_c.second.availableChunks();
        }
    }

    return 0;
}
```

```
charge: 
        Scalar
mass: 
      Scalar
momentum: x
  what():  [HDF5] File ID not found with file name
Aborted (core dumped)
```

https://github.com/openPMD/openPMD-api/blob/6fd39365ff4bf6e350faec66a5ebac2f3804e8dd/src/IO/HDF5/HDF5IOHandler.cpp#L399-L401

The value of `fname->second` is `../samples/git-sample/data00000400.h5`.
The elements in `m_fileNamesWithID.size()` are `0`.

**Expected behavior**
This works with ADIOS2, so I think it's a backend bug for HDF5.

**Software Environment**
 - version of openPMD-api: latest `dev`
 - installed openPMD-api via: source
 - operating system: Ubuntu 20.04
 - machine: local
 - name and version of Python implementation: N/A
 - version of HDF5: 1.10.7
 - version of ADIOS1: 1.13.1
 - version of ADIOS2: 2.7.1
 - name and version of MPI: serial run, but build with OpenMPI

**Additional context**

ax3l:
This adds a test for early reading of chunks (via `availableChunks`) with HDF5 that appears quite often with post-processing routines based on chunks.

I have not yet found the reason for this fail, but it seems to be HDF5 specific (details documented in #961). Either the writable is a bit off or the file not yet open.

- [x] add test
- [x] fix #961
- [x] merge https://github.com/ax3l/openPMD-api/pull/1
- [x] update [MPI - Collective Behavior](https://openpmd-api.readthedocs.io/en/latest/details/mpi.html):
  - [x] document the "first call to loadChunk/storeChunk/availableChunks is collective" unless we first called `::open()`;
        expand that `Iteration::open()` is used only for (MPI || deferred iteration parsing), elaborate on open() relation to flushing logic
  - [x] add `availableChunks` to table (behavior independent unless it's the first call to a freshly opened series)

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -47,9 +47,10 @@ namespace traits
 } // traits
 class AbstractFilePosition;
 class AttributableImpl;
+class Iteration;
 namespace internal
 {
-class SeriesInternal;
+    class SeriesInternal;
 }
 
 class no_such_attribute_error : public std::runtime_error
@@ -234,6 +235,9 @@ OPENPMD_protected:
 
     internal::SeriesInternal const & retrieveSeries() const;
     internal::SeriesInternal & retrieveSeries();
+
+    Iteration const & containingIteration() const;
+    Iteration & containingIteration();
 
     void seriesFlush( FlushLevel );
 
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -236,8 +236,16 @@ OPENPMD_protected:
     internal::SeriesInternal const & retrieveSeries() const;
     internal::SeriesInternal & retrieveSeries();
 
+    /** Returns the corresponding Iteration
+     *
+     * Return the openPMD::iteration that this Attributable is contained in.
+     * This walks up the linked parents until it finds the Iteration object.
+     * Throws an error otherwise, e.g., for Series objects.
+     * @{
+     */
     Iteration const & containingIteration() const;
     Iteration & containingIteration();
+    /** @} */
 
     void seriesFlush( FlushLevel );
 
--- a/src/backend/Attributable.cpp
+++ b/src/backend/Attributable.cpp
@@ -178,7 +178,8 @@ Iteration const & AttributableImpl::containingIteration() const
 Iteration & AttributableImpl::containingIteration()
 {
     return const_cast< Iteration & >(
-        static_cast< Iteration const * >( this )->containingIteration() );
+        static_cast< AttributableImpl const * >( this )
+            ->containingIteration() );
 }
 
 std::string Attributable::MyPath::filePath() const
--- a/src/backend/Attributable.cpp
+++ b/src/backend/Attributable.cpp
@@ -19,6 +19,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 #include "openPMD/backend/Attributable.hpp"
+#include "openPMD/Iteration.hpp"
 #include "openPMD/Series.hpp"
 #include "openPMD/auxiliary/DerefDynamicCast.hpp"
 #include "openPMD/auxiliary/StringManip.hpp"
@@ -132,6 +133,52 @@ internal::SeriesInternal & AttributableImpl::retrieveSeries()
 {
     return const_cast< internal::SeriesInternal & >(
         static_cast< AttributableImpl const * >( this )->retrieveSeries() );
+}
+
+Iteration const & AttributableImpl::containingIteration() const
+{
+    std::vector< Writable const * > searchQueue;
+    searchQueue.reserve( 7 );
+    Writable const * findSeries = &writable();
+    while( findSeries )
+    {
+        searchQueue.push_back( findSeries );
+        // we don't need to push the last Writable since it's the Series anyway
+        findSeries = findSeries->parent;
+    }
+    // End of the queue:
+    // Iteration -> Series.iterations -> Series
+    if( searchQueue.size() < 3 )
+    {
+        throw std::runtime_error(
+            "containingIteration(): Must be called for an object contained in "
+            "an iteration." );
+    }
+    auto end = searchQueue.rbegin();
+    internal::AttributableData const * attr = ( *( end + 2 ) )->attributable;
+    /*
+     * We now know the unique instance of Attributable that corresponds with
+     * the iteration.
+     * Since the class Iteration itself still follows the old class design,
+     * we will have to take a detour via Series.
+     */
+    auto & series = auxiliary::deref_dynamic_cast< internal::SeriesInternal >(
+        ( *searchQueue.rbegin() )->attributable );
+    for( auto const & pair : series.iterations )
+    {
+        if( &pair.second.get() == attr )
+        {
+            return pair.second;
+        }
+    }
+    throw std::runtime_error(
+        "Containing iteration not found in containing Series." );
+}
+
+Iteration & AttributableImpl::containingIteration()
+{
+    return const_cast< Iteration & >(
+        static_cast< Iteration const * >( this )->containingIteration() );
 }
 
 std::string Attributable::MyPath::filePath() const
--- a/src/backend/Attributable.cpp
+++ b/src/backend/Attributable.cpp
@@ -156,6 +156,8 @@ Iteration const & AttributableImpl::containingIteration() const
     }
     auto end = searchQueue.rbegin();
     internal::AttributableData const * attr = ( *( end + 2 ) )->attributable;
+    if( attr == nullptr )
+        throw std::runtime_error( "containingIteration(): attributable must not be a nullptr." );
     /*
      * We now know the unique instance of Attributable that corresponds with
      * the iteration.
--- a/src/backend/BaseRecordComponent.cpp
+++ b/src/backend/BaseRecordComponent.cpp
@@ -64,17 +64,10 @@ BaseRecordComponent::availableChunks()
         Offset offset( m_dataset->extent.size(), 0 );
         return ChunkTable{ { std::move( offset ), m_dataset->extent } };
     }
-
-    // set dirty, so Series::flush will open the file if needed
-    this->dirty() = true;
-    this->seriesFlush();
-    this->dirty() = false;
-
     Parameter< Operation::AVAILABLE_CHUNKS > param;
     IOTask task( this, param );
     IOHandler()->enqueue( task );
     IOHandler()->flush();
-
     return std::move( *param.chunks );
 }
 } // namespace openPMD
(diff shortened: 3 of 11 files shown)


test/SerialIOTest.cpp
```diff
@@ -2402,6 +2402,35 @@ TEST_CASE( "git_hdf5_sample_fileBased_read_test", "[serial][hdf5]" )
...
+
+        auto electrons = s.iterations[400].particles["electrons"];
+
+        for( auto & r : electrons )
+        {
+            std::cout << r.first << ": ";
+            for( auto & r_c : r.second )
+            {
+                std::cout << r_c.first << "\n";
+                if( !r_c.second.constant() )
```
ax3l, line 2422: Memo to myself and unrelated to the issue: this `if` can probably be removed now that #942 is implemented :)

Also: let's do something with the `chunks` variable (e.g. print it).

franzpoeschel:
Ref.: https://github.com/openPMD/openPMD-api/pull/862/files
tldr: No file handles are open after parsing in file-based iteration encoding. Fixed by using `auto electrons = s.iterations[400].open().particles["electrons"];`.

This test should fail for other backends, too. I claim "no bug".

ax3l:
Thanks for the ideas!

Why I tend to think that this is a usability bug and we should auto-open iterations in `availableChunks()` calls is because we do the same for `loadChunks()`.

I mean also from the API contract it's logical that if a _serial_
```c++
Series s = Series(...);

auto electrons = s.iterations[400].particles["electrons"];
auto chunks = electrons["position"]["x"].loadChunk();
```
works than a
```c++
Series s = Series(...);

auto electrons = s.iterations[400].particles["electrons"];
auto chunks = electrons["position"]["x"].availableChunks();
```
should work the same way. Especially since the latter would be called before the former now with #802-guided processing.

> Ref.: https://github.com/openPMD/openPMD-api/pull/862/files

Note that #862 is a MPI-parallel context where we want to trigger a MPI-collective open call with a collective guarantee because we use `loadChunks()` and `flush()` in a non-collective way as the following first data calls (see test in that PR). The problem here is serial.

If course, if someone called in an MPI-parallel context `availableChunks()` as the first thing and only from one rank, then they would be again required to first perform a collective `::open()` on the iteration, the same way as they need to for `loadChunk()`.

> This test should fail for other backends, too.

The problem became apparent to me in serial reads with `src/binding/python/openpmd_api/DaskDataFrame.py`, where we have no access to the iteration. In my tests, ADIOS2 reads work and HDF5 reads fail (both are serial, file-based encoding).

src/backend/BaseRecordComponent.cpp
```diff
@@ -64,10 +64,17 @@ BaseRecordComponent::availableChunks()
         Offset offset( m_dataset->extent.size(), 0 );
         return ChunkTable{ { std::move( offset ), m_dataset->extent } };
     }
+
+    // set dirty, so Series::flush will open the file if needed
+    this->dirty() = true;
+    this->seriesFlush();
+    this->dirty() = false;
+
```
franzpoeschel: I'm a bit doubtful about this..
1. It makes the `availableChunks` call collective, we should at least document that
2. `availableChunks` will now silently flush the whole Series. This will break user code that peruses our new flushing logic and its guarantees of no writes occurring until flushing. If we really want to do this, this function must not do anything more than merely opening those files.

franzpoeschel:
> Thanks for the ideas!
> 
> Why I tend to think that this is a usability bug and we should auto-open iterations in `availableChunks()` calls is because we do the same for `loadChunks()`.
> 
> I mean also from the API contract it's logical that if a _serial_
> 
> ```c++
> Series s = Series(...);
> 
> auto electrons = s.iterations[400].particles["electrons"];
> auto chunks = electrons["position"]["x"].loadChunk();
> ```
> 

This code does not open the iteration, `s.flush()` does. Interacting with an iteration fundamentally requires opening it in openPMD, but `Series::flush()` as well as the Streaming API will do it automatically. The fundamental difference between `availableChunks()` and `loadChunk()` at this point is that the latter is a deferred operation and hence receives all the properties of a deferred operation, including that it is performed by flushing.
> works than a
> 
> ```c++
> Series s = Series(...);
> 
> auto electrons = s.iterations[400].particles["electrons"];
> auto chunks = electrons["position"]["x"].availableChunks();
> ```
> 
> should work the same way. Especially since the latter would be called before the former now with #802-guided processing.
> 
> > Ref.: https://github.com/openPMD/openPMD-api/pull/862/files

This would make `availableChunks()` collective too, while `loadChunk()` can remain non-collective (collective business is dealt with in `Series::flush()`). Just as imbalanced as before, though even a bit trickier to debug.
> 
> Note that #862 is a MPI-parallel context where we want to trigger a MPI-collective open call with a collective guarantee because we use `loadChunks()` and `flush()` in a non-collective way as the following first data calls (see test in that PR). The problem here is serial.

We're making the problem parallel by implicit opening of files (and with the current implementation: implicit flushing).
> 
> If course, if someone called in an MPI-parallel context `availableChunks()` as the first thing and only from one rank, then they would be again required to first perform a collective `::open()` on the iteration, the same way as they need to for `loadChunk()`.
> 
> > This test should fail for other backends, too.
> 
> The problem became apparent to me in serial reads with `src/binding/python/openpmd_api/DaskDataFrame.py`, where we have no access to the iteration. In my tests, ADIOS2 reads work and HDF5 reads fail (both are serial, file-based encoding).

It's weird that this would even work in ADIOS2, I'll need to have a look at what's going on there.

Do you want `Iteration::open()` fundamentally to be necessary in parallel contexts only? I'm a bit hesitant to add exceptions to using `Iteration::open()` here and there, but if we set that as a fundamental guideline, and have some clean general concepts, that's another story then.

franzpoeschel:
After following the behavior of ADIOS2 in the debugger, [this line](https://github.com/openPMD/openPMD-api/blob/d276b05ffdb3fa0fdb3b808c56f237a98c8f0a32/src/IO/ADIOS/ADIOS2IOHandler.cpp#L1036) implicitly opens an ADIOS2 engine if not yet happened. I would declare this a bug, such things should only happen if explicitly requested by the frontend.

ax3l:
We will discuss this further on Wednesday.

I think the cleanest way would be to:
- implicitly open the file if still closed; if this is the first call than this makes it collective (otherwise this is an independent call)
  - serial reads are also significantly simpler that way, e.g. the demonstrator test in this PR
- make sure `Iteration::open()` works for this context, currently it's not the case (tried it)

I would thus also discuss if we want to apply #1045 really or if we instead rather implement a similar solution here.

We currently don't need `Iteration::open()` besides for the described independent-first-access context above. If we want to change that we need to improve error reporting on non-opened backends I think and we probably need to advertise that `Iteration::open()` always needs to be called for consistency.

franzpoeschel:
Memo to self: Try why `Iteration::open()` won't work in this context

test/SerialIOTest.cpp
```diff
@@ -2402,6 +2402,35 @@ TEST_CASE( "git_hdf5_sample_fileBased_read_test", "[serial][hdf5]" )
...
+TEST_CASE( "git_hdf5_early_chunk_query", "[serial][hdf5]" )
+{
+    try
+    {
+        Series s = Series(
+            "../samples/git-sample/data%T.h5",
+            Access::READ_ONLY
+        );
+
+        auto electrons = s.iterations[400].particles["electrons"];
```
ax3l, line 2414: tried also this:
```suggestion
        auto it = s.iterations[400];
        auto electrons = it.particles["electrons"];
        it.open(); // or here...
```
was surprised this did not work, IIRC

test/SerialIOTest.cpp
```diff
@@ -2402,6 +2402,35 @@ TEST_CASE( "git_hdf5_sample_fileBased_read_test", "[serial][hdf5]" )
     }
 }
 
+TEST_CASE( "git_hdf5_early_chunk_query", "[serial][hdf5]" )
+{
+    try
+    {
+        Series s = Series(
+            "../samples/git-sample/data%T.h5",
```
ax3l, line 2410: Discussed: add (file based) adios2 test, too due to #1045

test/SerialIOTest.cpp
franzpoeschel: works for me

ax3l: Ok, cool! Thanks, then I must have done something wrong or mixed it up in my mind :tada:

franzpoeschel:
> elaborate on open() relation to flushing logic

The changes in https://github.com/ax3l/openPMD-api/pull/1 make `open()` independent from flushing logic.

include/openPMD/backend/Attributable.hpp
```diff
@@ -235,6 +236,9 @@ class AttributableImpl
     internal::SeriesInternal const & retrieveSeries() const;
     internal::SeriesInternal & retrieveSeries();
 
+    Iteration const & containingIteration() const;
+    Iteration & containingIteration();
```
ax3l, line 247:
```suggestion
    /** Returns the corresponding Iteration
     *
     * Return the openPMD::iteration that this Attributable is contained in.
     * This walks up the linked parents until it finds the Iteration object.
     * Throws an error otherwise, e.g., for Series objects.
     * @{
     */
    Iteration const & containingIteration() const;
    Iteration & containingIteration();
    /** @} */
```

src/backend/Attributable.cpp
```diff
@@ -134,6 +135,53 @@ internal::SeriesInternal & AttributableImpl::retrieveSeries()
...
+    // End of the queue:
+    // Iteration -> Series.iterations -> Series
+    if( searchQueue.size() < 3 )
+    {
+        throw std::runtime_error(
+            "containingIteration(): Must be called for an object contained in "
+            "an iteration." );
+    }
+    auto end = searchQueue.rbegin();
+    internal::AttributableData const * attr = ( *( end + 2 ) )->attributable;
```
ax3l, line 158:
```suggestion
    internal::AttributableData const * attr = ( *( end + 2 ) )->attributable;
    if( attr == nullptr )
        throw std::runtime_error( "containingIteration(): attributable must not be a nullptr." );
```

test/SerialIOTest.cpp
```diff
@@ -4288,6 +4326,48 @@ void deferred_parsing( std::string const & extension )
                 std::numeric_limits< float >::epsilon() );
         }
     }
+    {
+        Series series(
```
ax3l, line 4351: Note: also fixes & tests read_write lazy parsing in this diff.

PR #1043, opened 2021-07-09
JSON update

### LINKED ISSUE #1037: JSON Update
This issue is supposed to collect things that we want to add to the JSON configuration of openPMD backends:

- [ ] Implement the extended layout for specifying dataset-specific configurations as experimented with already [in PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3513#issuecomment-788820242).
- [x] more global options: explicit selection of backend and iteration encoding, etcc (?)
- [x] Priority of configuration: explicit API calls > JSON > environment variables > automatically detected stuff (e.g. file extension) > defaults
    Many places already do this right, but document this and check if it's done correctly everywhere
- [ ] Replace legacy compression configurations with JSON - add helper functions to build backend-agnostic JSON strings?
- [ ] JSON configuration for ADIOS1, JSON (heh) and HDF5
- [x] capitalization. I favor capitalization-insensitivity, keeping consistent with how ADIOS2 does it.
- [x] More lenient datatypes in ADIOS2: To users, it seems weird that they should specify `adios2.engine.usesteps = true` (as a boolean) but `adios2.engine.parameters.OpenTimeoutSecs = "30"` (as a string).
- [ ] potentially install and expose a helper include if people want to use our shipped nlohmann-json
- [x] Alternatively read TOML files. Easy implementation could be done by implementing a converting reader for TOML files into instances of `nlohmann::json`. Clean implementation would be by building our own simple interface for config files that could have either JSON or TOML as a backend.

franzpoeschel:
Fix #1037, see there for a description.
This will probably not address all of the points noted therein, currently we have:
- [x] capitalization. I favor capitalization-insensitivity, keeping consistent with how ADIOS2 does it.
- [x] More lenient datatypes in ADIOS2: To users, it seems weird that they should specify `adios2.engine.usesteps = true` (as a boolean) but `adios2.engine.parameters.OpenTimeoutSecs = "30"` (as a string).
- [x] Use `TracingJSON` everywhere: partially implemented, warnings not yet shown for global options
- [x] Merge #1101 first
- [x] This PR removes some keys from `Dataset`: `chunks`, `compression` and `transform`. So far, only `transform` was even implemented (I've replaced that implementation with a JSON-based one now), and only in ADIOS1 which is legacy anyway. For keeping compatibility with user code, we should maybe keep the fields for now, but deprecate them and remove all implementation surrounding them. Discussed: remove them fully in the API if we change the API contract (implementation) to break only once.
- [x] Warn for global options: https://github.com/openPMD/openPMD-api/issues/1095
- [x] Warning if file ending collides with JSON option
- [x] ~~Are [these options](https://adios2.readthedocs.io/en/latest/engines/engines.html) in ADIOS2 capitalization insensitive? If not, I'll need to add some handling for that.~~ Eh, let's not rely on this in ADIOS2. I've added exceptions for the lower case transformation now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -581,6 +581,11 @@ if(openPMD_HAVE_ADIOS1)
         target_compile_definitions(openPMD.ADIOS1.Parallel PRIVATE openPMD_HAVE_MPI=0)
     endif()
 
+    target_include_directories(openPMD.ADIOS1.Serial SYSTEM PRIVATE
+        $<TARGET_PROPERTY:openPMD::thirdparty::nlohmann_json,INTERFACE_INCLUDE_DIRECTORIES>)
+    target_include_directories(openPMD.ADIOS1.Parallel SYSTEM PRIVATE
+        $<TARGET_PROPERTY:openPMD::thirdparty::nlohmann_json,INTERFACE_INCLUDE_DIRECTORIES>)
+
     set_target_properties(openPMD.ADIOS1.Serial PROPERTIES
         POSITION_INDEPENDENT_CODE ON
         CXX_VISIBILITY_PRESET hidden
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -777,6 +777,7 @@ set(openPMD_TEST_NAMES
     Auxiliary
     SerialIO
     ParallelIO
+    JSON
 )
 # command line tools
 set(openPMD_CLI_TOOL_NAMES
@@ -863,8 +864,10 @@ if(openPMD_BUILD_TESTING)
             target_link_libraries(${testname}Tests PRIVATE CatchMain)
         endif()
 
-        target_include_directories(${testname}Tests SYSTEM PRIVATE
-            $<TARGET_PROPERTY:openPMD::thirdparty::nlohmann_json,INTERFACE_INCLUDE_DIRECTORIES>)
+        if(${testname} STREQUAL JSON)
+            target_include_directories(${testname}Tests SYSTEM PRIVATE
+                $<TARGET_PROPERTY:openPMD::thirdparty::nlohmann_json,INTERFACE_INCLUDE_DIRECTORIES>)
+        endif()
     endforeach()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -857,6 +857,9 @@ if(openPMD_BUILD_TESTING)
         else()
             target_link_libraries(${testname}Tests PRIVATE CatchMain)
         endif()
+
+        target_include_directories(${testname}Tests SYSTEM PRIVATE
+            $<TARGET_PROPERTY:openPMD::thirdparty::nlohmann_json,INTERFACE_INCLUDE_DIRECTORIES>)
     endforeach()
 endif()
 
--- a/examples/7_extended_write_serial.py
+++ b/examples/7_extended_write_serial.py
@@ -102,8 +102,27 @@ if __name__ == "__main__":
     # component this describes the datatype and shape of data as it should be
     # written to disk
     d = Dataset(partial_mesh.dtype, extent=[2, 5])
-    d.set_compression("zlib", 9)
-    d.set_custom_transform("blosc:compressor=zlib,shuffle=bit,lvl=1;nometa")
+    dataset_config = """
+{
+  "adios1": {
+    "dataset": {
+      "transform": "blosc:compressor=zlib,shuffle=bit,lvl=1;nometa"
+    }
+  },
+  "adios2": {
+    "dataset": {
+      "operators": [
+        {
+          "type": "zlib",
+          "parameters": {
+            "clevel": 9
+          }
+        }
+      ]
+    }
+  }
+}"""
+    d.options = dataset_config
     mesh["x"].reset_dataset(d)
 
     electrons = cur_it.particles["electrons"]
--- a/examples/7_extended_write_serial.py
+++ b/examples/7_extended_write_serial.py
@@ -8,6 +8,7 @@ License: LGPLv3+
 """
 from openpmd_api import Series, Access, Dataset, Mesh_Record_Component, \
     Unit_Dimension
+import json
 import numpy as np
 
 
@@ -102,27 +103,24 @@ if __name__ == "__main__":
     # component this describes the datatype and shape of data as it should be
     # written to disk
     d = Dataset(partial_mesh.dtype, extent=[2, 5])
-    dataset_config = """
-{
-  "adios1": {
-    "dataset": {
-      "transform": "blosc:compressor=zlib,shuffle=bit,lvl=1;nometa"
-    }
-  },
-  "adios2": {
-    "dataset": {
-      "operators": [
-        {
-          "type": "zlib",
-          "parameters": {
-            "clevel": 9
-          }
+    dataset_config = {
+        "adios1": {
+            "dataset": {
+                "transform": "blosc:compressor=zlib,shuffle=bit,lvl=1;nometa"
+            }
+        },
+        "adios2": {
+            "dataset": {
+                "operators": [{
+                    "type": "zlib",
+                    "parameters": {
+                        "clevel": 9
+                    }
+                }]
+            }
         }
-      ]
     }
-  }
-}"""
-    d.options = dataset_config
+    d.options = json.dumps(dataset_config)
     mesh["x"].reset_dataset(d)
 
     electrons = cur_it.particles["electrons"]
--- a/examples/8_benchmark_parallel.cpp
+++ b/examples/8_benchmark_parallel.cpp
@@ -148,10 +148,10 @@ int main(
     // * The number of iterations. Effectively, the benchmark will be repeated for this many
     //   times.
 #if openPMD_HAVE_ADIOS1 || openPMD_HAVE_ADIOS2
-    benchmark.addConfiguration("", 0, "bp", dt, 10);
+    benchmark.addConfiguration("bp", dt, 10);
 #endif
 #if openPMD_HAVE_HDF5
-    benchmark.addConfiguration("", 0, "h5", dt, 10);
+    benchmark.addConfiguration("h5", dt, 10);
 #endif
 
     // Execute all previously configured benchmarks. Will return a MPIBenchmarkReport object
--- a/examples/8_benchmark_parallel.cpp
+++ b/examples/8_benchmark_parallel.cpp
@@ -148,10 +148,14 @@ int main(
     // * The number of iterations. Effectively, the benchmark will be repeated for this many
     //   times.
 #if openPMD_HAVE_ADIOS1 || openPMD_HAVE_ADIOS2
-    benchmark.addConfiguration("bp", dt, 10);
+    benchmark.addConfiguration(
+        R"({"adios2": {"dataset":{"operators":[{"type": "blosc"}]}}})",
+        "bp",
+        dt,
+        10 );
 #endif
 #if openPMD_HAVE_HDF5
-    benchmark.addConfiguration("h5", dt, 10);
+    benchmark.addConfiguration( "{}", "h5", dt, 10 );
 #endif
 
     // Execute all previously configured benchmarks. Will return a MPIBenchmarkReport object
--- a/include/openPMD/Dataset.hpp
+++ b/include/openPMD/Dataset.hpp
@@ -49,16 +49,10 @@ public:
     Dataset( Extent );
 
     Dataset& extend(Extent newExtent);
-    Dataset& setChunkSize(Extent const&);
-    Dataset& setCompression(std::string const&, uint8_t const);
-    Dataset& setCustomTransform(std::string const&);
 
     Extent extent;
     Datatype dtype;
     uint8_t rank;
-    Extent chunkSize;
-    std::string compression;
-    std::string transform;
     std::string options = "{}"; //!< backend-dependent JSON configuration
 };
 } // namespace openPMD
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -281,6 +281,16 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CREATE_DATASET > : public Abstrac
     Extent extent = {};
     Datatype dtype = Datatype::UNDEFINED;
     std::string options = "{}";
+
+    // template parameter so we don't have to include the JSON lib here
+    // this function is useful for the createDataset() methods in,
+    // IOHandlerImpl's, so putting that here is the simplest way to make it
+    // available for them
+    template< typename TracingJSON >
+    static void warnUnusedParameters(
+        TracingJSON &,
+        std::string const & currentBackendName,
+        std::string const & warningMessage );
 };
 
 template<>
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -268,8 +268,7 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CREATE_DATASET > : public Abstrac
     Parameter() = default;
     Parameter(Parameter const & p) : AbstractParameter(),
         name(p.name), extent(p.extent), dtype(p.dtype),
-        chunkSize(p.chunkSize), compression(p.compression),
-        transform(p.transform), options(p.options) {}
+        options(p.options) {}
 
     std::unique_ptr< AbstractParameter >
     clone() const override
@@ -281,9 +280,6 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CREATE_DATASET > : public Abstrac
     std::string name = "";
     Extent extent = {};
     Datatype dtype = Datatype::UNDEFINED;
-    Extent chunkSize = {};
-    std::string compression = "";
-    std::string transform = "";
     std::string options = "{}";
 };
 
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -282,10 +282,12 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CREATE_DATASET > : public Abstrac
     Datatype dtype = Datatype::UNDEFINED;
     std::string options = "{}";
 
-    // template parameter so we don't have to include the JSON lib here
-    // this function is useful for the createDataset() methods in,
-    // IOHandlerImpl's, so putting that here is the simplest way to make it
-    // available for them
+    /** Warn about unused JSON paramters
+     *
+     * Template parameter so we don't have to include the JSON lib here.
+     * This function is useful for the createDataset() methods in,
+     * IOHandlerImpl's, so putting that here is the simplest way to make it
+     * available for them. */
     template< typename TracingJSON >
     static void warnUnusedParameters(
         TracingJSON &,
--- a/include/openPMD/Series.hpp
+++ b/include/openPMD/Series.hpp
@@ -408,6 +408,9 @@ OPENPMD_private:
         }    }
 
     std::unique_ptr< ParsedInput > parseInput(std::string);
+    // template parameter so we don't have to include the JSON lib here
+    template< typename TracingJSON >
+    void parseJsonOptions( TracingJSON & options, ParsedInput & );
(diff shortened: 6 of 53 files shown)


franzpoeschel:
Warning for unused global options should work now, needs a bit more testing though. I configured an ADIOS2 dataset with the following string:
```json
{
  "resizable": true,
  "asdf": "asdf",
  "adios2": {
    "unused": "dataset parameter",
    "dataset": {
      "unused": "too",
      "operators": [
        {
          "type": "blosc",
          "parameters": {
            "clevel": 3,
            "doshuffle": "BLOSC_BITSHUFFLE"
          }
        }
      ]
    },
    "hdf5": {
      "this": "should not warn"
    }
  }
}
```
Resulting warning:
```
Warning: parts of the JSON configuration for ADIOS2 dataset '/data/0/meshes/E/y' remain unused:
{"adios2":{"dataset":{"unused":"too"},"unused":"dataset parameter"},"asdf":"asdf"}
```
There is no warning for the HDF5 dataset as the HDF5 part is not parsed.

franzpoeschel:
Only thing really still missing for finishing this PR is to introduce JSON configuration for stuff in ADIOS1 that has so far been configured via environment variables. Since this PR is already large enough, I'll suggest to postpone it to a follow-up.
Otherwise, I've split this thing into logical commits now, so reviewing by commit should be doable. I'll also go through things once again myself for a little cleanup, otherwise this is quite ready for review @ax3l :)

~~PS: What's also missing is to add JSON config to the MPI benchmarks. This PR removes the compression from them and doesn't replace them for now.~~ Added that now

franzpoeschel:
I've done some cleanup now and the CI is now also running fine

ax3l:
I merged #1101 now :)

franzpoeschel:
> I merged #1101 now :)

I've rebased

franzpoeschel:
Added today:
1. Detection of contradicting filename extension and backend parameter:
    ```ipython
    In [2]: io.Series("rabimmel.bp", io.Access.create, """{"backend": "json"}""")                   
    [Warning] Supplied filename extension ''.bp' contradicts the backend specified via the 'backend' key. Will go on with backend json.
    Out[2]: <openPMD.Attributable with '8' attributes>
    ```
2. Used the JSON `backend` key in some tests to select ADIOS1 or ADIOS2 without having to use `Series::backend()`
3. For the implementation of point 2, add a function `std::string json::merge( std::string const & defaultValue, std::string const & overwrite )`. I have included this function into public headers, for the reason, see its doxy string:
    ```cpp
    namespace json
    {
        /**
         * @brief Merge two JSON datasets into one.
         *
         * Merging rules:
         * 1. If both `defaultValue` and `overwrite` are JSON objects, then the
         *    resulting JSON object will contain the union of both objects' keys.
         *    If a key is specified in both objects, the values corresponding to the
         *    key are merged recursively.
         *    Keys that point to a null value after this procedure will be pruned.
         * 2. In any other case, the JSON dataset `defaultValue` is replaced in its
         *    entirety with the JSON dataset `overwrite`.
         *
         * Note that item 2 means that datasets of different type will replace each
         * other without error.
         * It also means that array types will replace each other without any notion
         * of appending or merging.
         *
         * Possible use case:
         * An application uses openPMD-api and wants to do the following:
         * 1. Set some default backend options as JSON parameters.
         * 2. Let its users specify custom backend options additionally.
         *
         * By using the json::merge() function, this application can then allow
         * users to overwrite default options, while keeping any other ones.
         *
         * @param defaultValue
         * @param overwrite
         * @return std::string
         */
        std::string merge(
            std::string const & defaultValue,
            std::string const & overwrite );
    }
    ```

CMakeLists.txt
```diff
@@ -841,6 +841,9 @@ if(openPMD_BUILD_TESTING)
         else()
             target_link_libraries(${testname}Tests PRIVATE CatchMain)
         endif()
+
+        target_include_directories(${testname}Tests SYSTEM PRIVATE
```
ax3l: I would like to not pull the nlohmann-json header into all tests.
Is that needed or could we limit it to selected test if needed?

test/AuxiliaryTest.cpp
```diff
@@ -42,6 +43,116 @@ struct TestHelper : public LegacyAttributable
...
+        }
+      ]
+    }
+  }
+}
+)END";
+    nlohmann::json jsonUpper = nlohmann::json::parse( upper );
+    nlohmann::json jsonLower = nlohmann::json::parse( lower );
+    REQUIRE( jsonUpper.dump() != jsonLower.dump() );
+    json::lowerCase( jsonUpper );
```
ax3l: I think this test can be a separate test binary, say JSONtests, that way we avoid pulling in a third party header publicly in all tests.

include/openPMD/auxiliary/JSON_internal.hpp
```diff
@@ -165,17 +169,31 @@ namespace auxiliary
...
+    nlohmann::json & lowerCase( nlohmann::json & );
+
+    auxiliary::Option< std::string > asStringDynamic( nlohmann::json const & );
+
+    auxiliary::Option< std::string >
+    asLowerCaseStringDynamic( nlohmann::json const & );
+
+    extern std::vector< std::string > backendKeys;
+
+    void warnGlobalUnusedOptions( TracingJSON const & config );
```
ax3l: Can you add doxygen string to these new JSON APIs pls? :)

src/Series.cpp
```diff
@@ -1463,21 +1463,32 @@ void SeriesInterface::openIteration( uint64_t index, Iteration iteration )
...
+    void getJsonOptionLowerCase(
+        json::TracingJSON & config, std::string const & key, Dest & dest )
+    {
+        if( config.json().contains( key ) )
+        {
+            dest = json::asLowerCaseStringDynamic( config[ key ].json() );
+        }
+    }
+
+    void parseJsonOptions(
```
ax3l: Cannot quickly understand the difference between parse and get variant here. Maybe add short doxygen strings?

examples/7_extended_write_serial.py
```diff
@@ -102,8 +102,27 @@
     # component this describes the datatype and shape of data as it should be
     # written to disk
     d = Dataset(partial_mesh.dtype, extent=[2, 5])
-    d.set_compression("zlib", 9)
-    d.set_custom_transform("blosc:compressor=zlib,shuffle=bit,lvl=1;nometa")
+    dataset_config = """
```
ax3l: Let's use a Python dictionary here as in the other PR/example we did recently.

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -381,7 +381,7 @@ HDF5IOHandlerImpl::createDataset(Writable* writable,
             VERIFY(status == 0, "[HDF5] Internal error: Failed to set chunk size during dataset creation");
         }
 
-        std::string const& compression = parameters.compression;
+        std::string const& compression = ""; // @todo read from JSON
```
ax3l, line 399: marking to make sure todo is resolved in later commits.

include/openPMD/Dataset.hpp
```diff
@@ -49,16 +49,10 @@ class Dataset
     Dataset( Extent );
 
     Dataset& extend(Extent newExtent);
-    Dataset& setChunkSize(Extent const&);
```
ax3l, line 52: Write in our *Upgrade Guide* (`NEWS.rst`) that these APIs were removed from `Dataset`.

examples/8_benchmark_parallel.cpp
```diff
@@ -148,10 +148,14 @@ int main(
     // * The number of iterations. Effectively, the benchmark will be repeated for this many
     //   times.
 #if openPMD_HAVE_ADIOS1 || openPMD_HAVE_ADIOS2
-    benchmark.addConfiguration("bp", dt, 10);
+    benchmark.addConfiguration(
+        R"({"adios2": {"dataset":{"operators":[{"type": "blosc"}]}}})",
```
ax3l, line 152: Maybe we need a test or option for this, since blosc is not always there.

On the other hand: ADIOS2 will just warn if not, and keeps going without it.

My go-to compressor is parallel `zstd` in blosc though :)

include/openPMD/IO/IOTask.hpp
```diff
@@ -281,6 +281,16 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CREATE_DATASET > : public Abstrac
     Extent extent = {};
     Datatype dtype = Datatype::UNDEFINED;
     std::string options = "{}";
+
+    // template parameter so we don't have to include the JSON lib here
```
ax3l: Let's make this a proper Doxygen string, so the parser picks it up :)

test/SerialIOTest.cpp
```diff
@@ -3394,6 +3394,8 @@ TEST_CASE( "serial_adios2_json_config", "[serial][adios2]" )
 )END";
     std::string datasetConfig = R"END(
```
ax3l, line 3456: let's add a comment on resiable, asdf and hdf5 here, to avoid confusion for us later on when reading this.

include/openPMD/Series.hpp
```diff
@@ -346,6 +346,9 @@ class SeriesInterface : public AttributableInterface
         }    }
 
     std::unique_ptr< ParsedInput > parseInput(std::string);
+    // template parameter so we don't have to include the JSON lib here
```
ax3l: Let's add proper Doxygen for new member functions :)

src/Series.cpp
```diff
@@ -1479,15 +1479,75 @@ namespace
     {
         if( config.json().contains( key ) )
         {
-            dest = json::asLowerCaseStringDynamic( config[ key ].json() );
+            auto maybeString =
+                json::asLowerCaseStringDynamic( config[ key ].json() );
+            if( maybeString.has_value() )
+            {
+                dest = std::move( maybeString.get() );
```
ax3l, line 1491: Oh, we don't have the `<utility>` header in this `.cpp` file yet, it seems.

src/Series.cpp
```diff
@@ -1479,15 +1479,75 @@ namespace
...
+            }
+            else
+            {
+                throw error::BackendConfigSchema(
```
ax3l, line 1495: Does this one need a new internal include?

test/CoreTest.cpp
```diff
@@ -862,6 +868,92 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+        Series series(
+            "../samples/optionsViaJsonOverwritesAutomaticDetection.sst",
+            Access::CREATE,
+            "{\"adios2\": {\"engine\": {\"type\": \"bp4\"}}}" );
+    }
+    REQUIRE( auxiliary::directory_exists(
+        "../samples/optionsViaJsonOverwritesAutomaticDetection.bp" ) );
+
+#if openPMD_HAVE_ADIOS1
+    setenv( "OPENPMD_BP_BACKEND", "ADIOS1", 1 );
```
ax3l, line 925: We currently handle this by having a CI test that sets the variable (and runs all BP tests with ADIOS1 then that support it).
Maybe that's simpler to keep so?

test/SerialIOTest.cpp
```diff
@@ -3316,6 +3316,10 @@ TEST_CASE( "serial_adios2_json_config", "[serial][adios2]" )
     }
     std::string writeConfigBP3 = R"END(
 {
+  "unused": "global parameter",
+  "hdf5": {
+    "unused": "hdf5 parameter please dont warn"
```
ax3l: dont -> do not

src/auxiliary/JSON.cpp
```diff
@@ -240,8 +257,25 @@ namespace json
...
-        return lowerCase( json, currentPath );
+        currentPath.reserve( 7 );
+        return lowerCase(
+            json, currentPath, []( std::vector< std::string > const & path ) {
+                std::vector< std::string > const ignoredPaths[] = {
+                    { "adios2", "engine", "parameters" },
+                    { "adios2",
+                      "dataset",
+                      "operators",
+                      "\vnum",
```
ax3l, line 272: Whats this one?

Potentially also add an inline comment for it?

test/SerialIOTest.cpp
```diff
@@ -4139,7 +4179,9 @@ void variableBasedSeries( std::string const & file )
 
 TEST_CASE( "variableBasedSeries", "[serial][adios2]" )
 {
+#if openPMD_HAVE_ADIOS2
```
ax3l: You could also mask out the whole test by moving this around the `TEST_CASE` a whole - then it does not show up in the list of available tests if ADIOS2 is not compiled in.

ax3l (commented):
Thank you for the great ordering by commits and the clean docs in them.

I went through all of them and only have a few comments, let me know what you think. Great work! :tada:

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: Compression is currently unimplemented in the HDF5 backend, so I've left this part unimplemented too. 
If a user specifies a compression, a warning will be printed for the unused configuration parameter.

examples/8_benchmark_parallel.cpp
franzpoeschel: Just to be sure, you mean zstd *instead* of blosc? Or is there a way to pick zstd inside blosc? This current configuration will just select whatever defaults that ADIOS2 defines.

src/Series.cpp
franzpoeschel: `#include "openPMD/Error.hpp"` is already included

test/CoreTest.cpp
franzpoeschel: The purpose of this test is to make it explicit that the JSON option `"backend": "adios2"` should have priority over the environment variable. Sure if we removed the `setenv()` this would still be covered by CI, but having the line in the test makes it explicit what this is about.

src/auxiliary/JSON.cpp
ax3l: added :)

include/openPMD/auxiliary/JSON_internal.hpp
```diff
@@ -0,0 +1,235 @@
...
+        std::string const & options, MPI_Comm comm, bool considerFiles );
+
+#endif
+
+    /**
+     * Recursively transform all keys in a JSON dataset to lower case.
+     * String values are unaffected.
+     * JSON objects at the following openPMD-defined locations are not affected:
+     * * adios2.engine.parameters
+     * * adios2.dataset.operators.<number>.parameters
```
ax3l: Avoid warning of the kind:
```
include/openPMD/auxiliary/JSON_internal.hpp:193: error: Unsupported xml/html tag <number> found (warning treated as error, aborting now)
```
```suggestion
     * * `adios2.engine.parameters`
     * * `adios2.dataset.operators.<number>.parameters`
```

test/AuxiliaryTest.cpp
ax3l: fixed :)

CMakeLists.txt
ax3l: fixed :)

include/openPMD/Dataset.hpp
ax3l: @franzpoeschel probably overlooked: please add this to `NEWS.rst` as we announce breaking changes there :)

examples/8_benchmark_parallel.cpp
ax3l: > Just to be sure, you mean zstd instead of blosc? Or is there a way to pick zstd inside blosc?

The latter! blosc is a meta-compressor and implements threaded compression routines for zlib, zstd, lz4, blosclz, etc.:
https://www.blosc.org/pages/blosc-in-depth

here is an example how I use blosc with threaded zstd compressor for openPMD output in WarpX:
- https://warpx.readthedocs.io/en/21.12/usage/parameters.html#full-diagnostics

And this is how I did it in ADIOS1 in PIConGPU:
- https://github.com/ComputationalRadiationPhysics/picongpu/blob/0.5.0/docs/TBG_macros.cfg#L188-L195

See also Fig. 4 in https://arxiv.org/abs/1706.00522 - the threaded compressors are all via blosc.

src/Series.cpp
ax3l: added :)

include/openPMD/IO/IOTask.hpp
ax3l: @franzpoeschel probably overlooked: still need to be transformed to a doxygen string :)

include/openPMD/Series.hpp
ax3l: fixed :)

test/SerialIOTest.cpp
ax3l: added, thanks a lot :)

test/CoreTest.cpp
ax3l: > The purpose of this test is to make it explicit that the JSON option `"backend": "adios2"` should have priority over the environment variable.

Got it, makes sense to me. Thanks!

ax3l: (Just because I did not directly understand it: Maybe add this as a comment for future me? :) )

test/SerialIOTest.cpp
ax3l: fixed :)

test/SerialIOTest.cpp
ax3l: updated :)

include/openPMD/Dataset.hpp
franzpoeschel: Ah yes, I rebased this branch to get the latest version of NEWS.rst and then forgot going forward with it

test/CoreTest.cpp
ax3l: added :)

include/openPMD/Dataset.hpp
ax3l: Added :)

include/openPMD/IO/IOTask.hpp
ax3l: Still missing, I'll do it quickly :)

include/openPMD/IO/IOTask.hpp
```diff
@@ -281,10 +280,17 @@ struct OPENPMDAPI_EXPORT Parameter< Operation::CREATE_DATASET > : public Abstrac
...
     Datatype dtype = Datatype::UNDEFINED;
-    Extent chunkSize = {};
-    std::string compression = "";
-    std::string transform = "";
     std::string options = "{}";
+
+    // template parameter so we don't have to include the JSON lib here
+    // this function is useful for the createDataset() methods in,
+    // IOHandlerImpl's, so putting that here is the simplest way to make it
+    // available for them
```
```suggestion
    /** Warn about unused JSON paramters
     *
     * Template parameter so we don't have to include the JSON lib here.
     * This function is useful for the createDataset() methods in,
     * IOHandlerImpl's, so putting that here is the simplest way to make it
     * available for them. */
```

test/CoreTest.cpp
```diff
@@ -866,6 +871,118 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+            "../samples/optionsViaJson%06T",
+            Access::CREATE,
+            encodingFileBased );
+        series.iterations[1456];
+    }
+    std::string encodingGroupBased =
+        R"({"backend": "json", "iteration_encoding": "group_based"})";
+    {
+        /*
+         * … and if a pattern is detected, but the JSON config says to use
```
ax3l: Plain ASCII:
```suggestion
         * ... and if a pattern is detected, but the JSON config says to use
```

test/CoreTest.cpp
```diff
@@ -866,6 +871,118 @@ TEST_CASE( "no_file_ending", "[core]" )
...
+                Series series(
+                    "../samples/optionsViaJson",
+                    Access::CREATE,
+                    encodingFileBased );
+            }(),
+            error::WrongAPIUsage );
+    }
+    {
+        /*
+         * … but specifying both the pattern and the option in JSON should work.
```
ax3l: Plain ASCII:
```suggestion
         * ... but specifying both the pattern and the option in JSON should work.
```

ax3l (approved):
Thanks a lot :tada: Great work! :rocket: :sparkles:

PR #1045, opened 2021-07-12
ADIOS2: Don't implicitly open files

franzpoeschel:
Bug found [here](https://github.com/openPMD/openPMD-api/pull/1035#issuecomment-878199483). The ADIOS2 backend will implicitly open a file if it has not been opened yet. Files should only be opened if the frontend explicitly requested it, otherwise that's an error.

(Why should we be so strict? Implicit opening can make unsuspected calls accidentally collective. Better to have a clear error stating that something went wrong rather than a hanging parallel program.)

- [x] depends on improved frontend open logic in #1035 to be finalized/merged first
- [x] needs test for bp as in #1035

### DIFF
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -384,7 +384,14 @@ private:
      */
     ADIOS2FilePosition::GD groupOrDataset( Writable * );
 
-    detail::BufferedActions & getFileData( InvalidatableFile file );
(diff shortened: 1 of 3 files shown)


ax3l:
We will review on Wednesday if we want to follow the route in this PR or instead add this functionality to HDF5, too:
https://github.com/openPMD/openPMD-api/pull/1035#issuecomment-878831677

ax3l (approved):
Thanks a lot!

I added the `.bp` test to the PR now to compete it :)

PR #1047, opened 2021-07-15
Docs: Benchmark 8 Update

ax3l:
Update documentation page.

Follow-up to  #1010

### DIFF
--- a/docs/source/usage/benchmarks.rst
+++ b/docs/source/usage/benchmarks.rst
@@ -8,15 +8,14 @@ Parallel benchmark 8
 
 Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`, this benchmark executes a simple parallel read-write test.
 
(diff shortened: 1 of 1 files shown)


docs/source/usage/benchmarks.rst
```diff
@@ -8,15 +8,14 @@ Parallel benchmark 8
 
 Build based on the helper functions in the :ref:`benchmark utilities <utilities-benchmark>`, this benchmark executes a simple parallel read-write test.
 
-In particular, this test case writes and reads a 4D array of type ``uint64_t``, sliced 1D along the first dimension.
+In particular, this test case writes and reads a 3D array of type ``uint64_t``, sliced 1D along the first dimension.
```
ax3l, line 11: @franzpoeschel looking at this again, writing `double`s would probably be more realistic :sweat_smile: 
(Also with respect to compressors, if one might add them.)

PR #1051, opened 2021-07-19
Update Adios2 git sample tests

franzpoeschel:
Follow-up to https://github.com/openPMD/openPMD-example-datasets/pull/15
This tests against the numbers in the new dataset

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -3721,6 +3721,8 @@ TEST_CASE( "git_adios2_sample_test", "[serial][adios2]" )
     REQUIRE( charge_scalar.getDatatype() == Datatype::DOUBLE );
     REQUIRE( charge_scalar.getDimensionality() == 1 );
     REQUIRE( charge_scalar.getExtent() == Extent{ 96781 } );
+    double const charge_value = -3.0620612669736147e-3;
+    REQUIRE( charge_scalar.getAttribute("value").get< double >() == charge_value );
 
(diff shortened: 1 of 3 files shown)


test/SerialIOTest.cpp
```diff
@@ -3719,21 +3730,21 @@ TEST_CASE( "git_adios2_sample_test", "[serial][adios2]" )
     REQUIRE( mass.count( RecordComponent::SCALAR ) == 1 );
 
     RecordComponent & mass_scalar = mass[ RecordComponent::SCALAR ];
-    REQUIRE( areEqual( mass_scalar.unitSI(), 6.3356338938136719e-27 ) );
+    REQUIRE( areEqual( mass_scalar.unitSI(), 2.9749182215581054e-28 ) );
     REQUIRE( mass_scalar.getDatatype() == Datatype::DOUBLE );
     REQUIRE( mass_scalar.getDimensionality() == 1 );
-    REQUIRE( mass_scalar.getExtent() == Extent{ 0 } );
+    REQUIRE( mass_scalar.getExtent() == Extent{ 96781 } );
```
ax3l, line 3738: You could also read "mass" and "charge" with `loadChunk` to make sure constant records are read well :)

ax3l: (A bit more as a hack, you can also compare the `value` attribute to it.)

ax3l: added a commit :)

ax3l (approved):

PR #1053, opened 2021-07-20
Warnings: Unreachable Code & Unused Param

ax3l:
- fix unreachable code: `break` after `return` for `Datatype`
- fix unused parameter on non-Linux platforms for example 8b


ax3l:
sanitizer problem already present in `dev` and unrelated...

PR #1056, opened 2021-07-21
SerialIOTest: OPENPMD_TEST_NFILES_MAX

ax3l:
Limit the number of open files for individual tests.
Caused hangs with the sanitizer runs.

Follow-up to #1054

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -173,7 +173,10 @@ TEST_CASE( "adios2_char_portability", "[serial][adios2]" )
 
 void
 write_and_read_many_iterations( std::string const & ext ) {
-    constexpr unsigned int nIterations = 200; // original: 1000
+    // the idea here is to trigger the maximum allowed number of file handles,
+    // e.g., the upper limit in "ulimit -n" (default: often 1024). Once this
+    // is reached, files should be closed automatically for open iterations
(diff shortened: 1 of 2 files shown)


test/SerialIOTest.cpp
```diff
@@ -173,7 +173,10 @@ TEST_CASE( "adios2_char_portability", "[serial][adios2]" )
 
 void
 write_and_read_many_iterations( std::string const & ext ) {
-    constexpr unsigned int nIterations = 200; // original: 1000
+    // the idea here is to trigger the maximum allowed number of file handles,
+    // e.g., the upper limit in "ulimit -n" (default: often 1024). Once this
+    // is reached, files should be closed automatically for open iterations
+    unsigned int nIterations = auxiliary::getEnvNum( "OPENPMD_TEST_NFILES_MAX", 1000 );
```
franzpoeschel: Just to make sure I understand this correctly: We're now opting *out* of the 1000-file test, the default is 1000?

ax3l: Yes, that's correct. We could also opt-in.

franzpoeschel: Also since the default `ulimit -n` is often 1024, maybe take the chance to bump the default to 1030 or so?

PR #1058, opened 2021-07-22
Release: 0.14.0

ax3l:
### Resize, Dask, openpmd-pipe and new ADIOS2 Iteration Encoding

This release adds support for resizable data sets. For data-processing, support for Dask (parallel) and Pandas (serial) are added and lazy reader parsing of iterations is now supported. ADIOS2 adds an experimental variable-based iteration encoding. An openPMD Series can now be flushed from non-``Series`` objects and write buffers can be requested upfront to avoid unnecessary data copies in some situations.

Thanks to @ax3l, @franzpoeschel, @eschnett, @guj, @jakirkham and @dmitry-ganyushin for contributing to this release!

### Python

```python
import openpmd_api
```

### C++

```C++
#include <iostream>

int main() {
    // query compile-time macros:

    return 0;
}
```

### CLI

```bash
openpmd-ls --version
```


.rodare.json
```diff
@@ -102,6 +102,12 @@
             "affiliation": "NVIDIA",
             "name": "Kirkham, John",
             "type": "Other"
+        },
+        {
+            "affiliation": "Perimeter Institute for Theoretical Physics",
```
ax3l, line 107: @eschnett thank you for your contributions already!

We will push out this long-accumulated major release and then move your contributions to co-author once we merge your two open PRs in for the next release window, if that is ok with you, too :)

CHANGELOG.rst
```diff
@@ -5,32 +5,124 @@ Changelog
 
 0.14.0
 ------
-**Date:** TBA
+**Date:** 2021-06-XX
```
ax3l: Memo to myself: update me with final date

CHANGELOG.rst
```diff
@@ -5,32 +5,124 @@ Changelog
...
 
-...
+Resize, Dask, openpmd-pipe and new ADIOS2 Iteration Encoding
 
-...
+This release adds support for resizable data sets.
```
ax3l, line 12: User-facing dcumentation to do's for @ax3l and @franzpoeschel:
- [x] `openpmd-pipe.py` tool should be described in CLI section: https://openpmd-api.readthedocs.io/en/latest/utilities/cli.html: #1062 #1069
- [x] the keyword `variableBased` does not occur in the docs (we write variable-based in text in the ADIOS2 section, which makes this hard to search for): #1064
- [x] we need a user-facing example demonstrating the `Dataset` resize feature: #1060

CHANGELOG.rst
ax3l: ## To Do

- [x] docs: https://github.com/openPMD/openPMD-api/pull/1058#discussion_r674565930
- [x] CI splitting & NVHPC: #1052
- [x] Series throw file not found type: #1059
- [x] update version in `CITATION.cff` #1070

PR #1060, opened 2021-07-22
Doc: Resize Examples (Particles)

ax3l:
Add user-facing examples on how to write  resizeable data sets.

### DIFF
--- a/.github/workflows/tooling.yml
+++ b/.github/workflows/tooling.yml
@@ -68,4 +68,6 @@ jobs:
         export ASAN_OPTIONS=detect_stack_use_after_return=1:detect_leaks=1:check_initialization_order=true:strict_init_order=true:detect_stack_use_after_scope=1:fast_unwind_on_malloc=0
         export LSAN_OPTIONS=suppressions="$SOURCEPATH/.github/ci/sanitizer/clang/Leak.supp"
         export LD_PRELOAD=/usr/lib/clang/10/lib/linux/libclang_rt.asan-x86_64.so
-        ctest --test-dir build --output-on-failure
+        ctest --test-dir build -E 3b --output-on-failure
+        export OPENPMD_HDF5_CHUNKS="auto"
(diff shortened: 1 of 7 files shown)


.github/workflows/tooling.yml
```diff
@@ -68,4 +68,6 @@ jobs:
         export ASAN_OPTIONS=detect_stack_use_after_return=1:detect_leaks=1:check_initialization_order=true:strict_init_order=true:detect_stack_use_after_scope=1:fast_unwind_on_malloc=0
         export LSAN_OPTIONS=suppressions="$SOURCEPATH/.github/ci/sanitizer/clang/Leak.supp"
         export LD_PRELOAD=/usr/lib/clang/10/lib/linux/libclang_rt.asan-x86_64.so
-        ctest --test-dir build --output-on-failure
+        ctest --test-dir build -E 3b --output-on-failure
+        export OPENPMD_HDF5_CHUNKS="auto"
+        ctest --test-dir build -R 3b --output-on-failure
```
franzpoeschel, line 73: Why does this test require changing things in the tooling? Shouldn't our Cmake cover this?

franzpoeschel:
Nice, thank you for this! I like that you also added examples for constant record components.
The only thing that might be missing (but it's not too important anyway), is that you can use the `Dataset` constructor without specifying a datatype for resizing:
```cpp
    /**
     * @brief Constructor that sets the datatype to undefined.
     *
     * Helpful for resizing datasets, since datatypes need not be given twice.
     *
     */
    Dataset( Extent );
```
(And I think, I forgot adding that to the Python bindings.. Will open a quick PR for that)

franzpoeschel (approved):

ax3l:
Sounds great, thanks! Yes, let's add the doxygen string with #1063 as we rebase it :tada:

ax3l:
Ah dang, I misread your comment. :D

PR #1061, opened 2021-07-23
Docs: Move IOTasks Quoted Lines

ax3l:
Moved a little in the source file.


franzpoeschel:
This is likely to get moved in a future commit somewhere down the road. Maybe add a little comment in `IOTask.hpp` "If you move this line, please also edit design.rst"?

ax3l:
Good idea, added! :)

PR #1062, opened 2021-07-23
Document openpmd-pipe script

franzpoeschel:
This adds documentation for `openpmd-pipe`.

Todo:
- [x] ~~Add the image/PDF below to demonstrate its usage in a streaming setup, or is that too much for this place?
[streaming_workflow.pdf](https://github.com/openPMD/openPMD-api/files/6868027/streaming_workflow.pdf)~~

Also, it doesn't really matter, but in the documentation for `openpmd-ls` in the same file, the line before `   python -m openpmd_api.ls --help` should probably say `bash`, not `python3`.

This also fixes a small bug that made it impossible to pipe between different iteration encodings. See the second commit for that.


ax3l:
Thank you! :tada:

>  Add the image/PDF below to demonstrate its usage in a streaming setup, or is that too much for this place?

Probably not needed conceptionally with the text you have and would be confusion (showing a concrete workflow that we have no example for).

But you could add a usage example that similar to the test we have checked in converts two files, for instance.
You can also show that usage example in the help text of `openpmd-pipe --help`.
I am thinking about a block like this:
https://github.com/openPMD/openPMD-api/blob/c6b32d65810e755704448d51510bb48e21ae679b/include/openPMD/cli/ls.hpp#L47-L51

> Also, it doesn't really matter, but in the documentation for openpmd-ls in the same file, the line before python -m openpmd_api.ls --help should probably say bash, not python3.

Oh yes, copy-paste typo. Yes, please fix with this PR :)

> This also fixes a small bug that made it impossible to pipe between different iteration encodings. See the second commit for that.

Awesome, good catch! :)

---

Somewhat unrelated, I just realized most of the logic of pipe is implemented in `__init__.py`. That's a bit unusual, I would implement int either in `__main__.py` directly or a third file alongside it that then gets imported. `__init__.py` is usually free of implementation logic and only takes care about imports and module declarations.

franzpoeschel:
> Somewhat unrelated, I just realized most of the logic of pipe is implemented in `__init__.py`. That's a bit unusual, I would implement int either in `__main__.py` directly or a third file alongside it that then gets imported. `__init__.py` is usually free of implementation logic and only takes care about imports and module declarations.

To be fair, I did it this way since I don't really know about Python packaging that well and after a while of playing around, this one worked. I've moved things to `__main__.py` now, but I don't think it's quite right yet, see the diff of `src/cli/pipe.py`: I now need to import `openpmd_api.pipe.__main__`..

I added the examples to the documentation

ax3l:
> To be fair, I did it this way since I don't really know about Python packaging that well and after a while of playing around, this one worked.

Oh no worries, just saw it on the side. Will take a look and fix if needed, thanks a ton! :)

> --help update

Thanks, that's superb! :+1:

ax3l:
I also added it to the `setup.py` entry points for console scripts, so we can easily get both the `openpmd-ls` and `openpmd-pipe` tools via [pipx](https://github.com/pypa/pipx) installed :)

ax3l (approved):

PR #1063, opened 2021-07-23
Add Dataset(Extent) constructor to Python bindings

franzpoeschel:
There's no Python-based test for resizing since resizing did not add any new API, but was instead a backend-only (~middleend rather) thing. If we want to test this, I suggest rebasing it onto https://github.com/openPMD/openPMD-api/pull/1060?
Otherwise, this just adds a small missing constructor to the Python bindings.

### DIFF
--- a/examples/3b_write_resizable_particles.py
+++ b/examples/3b_write_resizable_particles.py
@@ -43,11 +43,14 @@ if __name__ == "__main__":
     rc_xo.make_constant(0.0)
     rc_yo.make_constant(0.0)
 
+    # after this call, the provided data buffers can be used again or deleted
+    series.flush()
+
     # extend and append more particles
     x = np.array([5., 6., 7.], dtype=np.double)
     y = np.array([-7., -8., -9.], dtype=np.double)
     offset += dataset.extent[0]
-    dataset.extend([dataset.extent[0] + x.shape[0]])
+    dataset = io.Dataset([dataset.extent[0] + x.shape[0]])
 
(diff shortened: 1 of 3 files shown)


ax3l:
Thank you! When we rebase this, let's add the doc string here as well:
https://github.com/openPMD/openPMD-api/pull/1060#issuecomment-885555127 :tada:

ax3l (approved):

examples/3b_write_resizable_particles.py
```diff
@@ -47,7 +47,7 @@
     x = np.array([5., 6., 7.], dtype=np.double)
     y = np.array([-7., -8., -9.], dtype=np.double)
     offset += dataset.extent[0]
-    dataset.extend([dataset.extent[0] + x.shape[0]])
+    dataset = io.Dataset([dataset.extent[0] + x.shape[0]])
```
ax3l, line 53: In line 55 on macOS:
```
  File "/Users/runner/work/openPMD-api/openPMD-api/build/bin/3b_write_resizable_particles.py", line 55, in <module>
    rc_x[offset:] = x
RuntimeError: Datatypes of chunk data (DOUBLE) and record component (UNDEFINED) do not match.
[HDF5] Datatype::UNDEFINED caught during dataset creation (serial HDF5)
```

ax3l: Ah, I forgot to flush after the first write :)

PR #1065, opened 2021-07-23
Rename SeriesImpl → SeriesInterface, AttributableImpl → AttributableInterface

franzpoeschel:
Up for discussion.
The old name is not wrong, but misleading. Those classes *do* implement the logic of `Series` and `Attributable`, but from our users' perspective, the more important aspect is that they define the interface of those classes. The current names may look like we accidentally exposed an internal class.

TODO:
- [x] Discuss this
- [x] I've renamed things automatically. Read through the diff carefully to check that things are fine. (Especially the docs)

### DIFF
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -42,7 +42,7 @@ Especially when a Series has many iterations, this can be a costly operation and
 
 When parsing non-eagerly, each iteration needs to be explicitly opened with ``Iteration::open()`` before accessing.
 (Notice that ``Iteration::open()`` is generally recommended to be used in parallel contexts to avoid parallel file accessing hazards).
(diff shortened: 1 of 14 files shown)


eschnett (approved):
As a name change this makes a lot of sense. +3.

In Julia, these classes would typically be called `AbstractSeries` and `Series`. The term "interface" carries the same meaning; where things are actually implemented doesn't matter at all. (You could in the future refactor to have a private base class `SeriesImpl`.)

ax3l:
Sounds good to me, too! Let's get this still in 0.14.0, because we introduced this in the current release cycle.
@franzpoeschel some of the CI is still failing, probably needs just another careful sweep through the renamed parts.

franzpoeschel:
Yeah, I forgot that there was an instance of `AttributableImpl` in the tests.

ax3l (approved):
Thanks a lot! :tada:

PR #1066, opened 2021-07-23
Doc: Benchmark 8b Pack Parameter

guj:


docs/source/usage/benchmarks.rst
```diff
@@ -75,6 +75,11 @@ To run:
 then the file generated are:  ../samples/8a_parallel_3Db_*
 
 
+Optional input parameter: pack
+
+Often a processor will write out a few miniBlocks. Using the example above, if writer side uses 1024 processors, each processor will handle 16 blocks from the 32x32x16 grid. The selection of this 16 blocks can be from 1x1x16, or 2x2x4, etc. To specify which, the associated parameter used will be "pack=1 1 16" or "pack=2 2 4". Without specifing this parameter, an default will be applied.  This parameter does not expected to impact the performance of writing, it will likely make a difference for certain reading patterns if the underlying storage is using subfiles.
```
ax3l: The term `miniBlock` is not defined in the text, so I would rewrite it.

```suggestion
Often a processor will write out a few small blocks.
Using the example above, if the writer side uses 1024 processors, each processor will handle 16 blocks from the 32x32x16 grid.
The selection of these 16 blocks can be from 1x1x16, to 2x2x4, etc..
To specify which, the ``pack`` parameter can be specified, e.g., ``"pack=1 1 16"`` or ``"pack=2 2 4"``.
Without specifying this parameter, a default will be applied.
This parameter does not expected to impact the performance of writing, it will likely make a difference for certain reading patterns if the underlying storage is using subfiles.
```

The central sentences:
> The selection of these 16 blocks can be from 1x1x16, to 2x2x4, etc..
> To specify which, the ``pack`` parameter can be specified, e.g., ``"pack=1 1 16"`` or ``"pack=2 2 4"``.

I do not yet understand. Can you please elaborate?

ax3l:
@guj thank you for the update! :)

I need a little more detail to clarify some parts in the description, otherwise it's great! :+1:

docs/source/usage/benchmarks.rst
guj: Using a smaller example, if a grid is 2x2, and you have 2 processors,  to divide work load, we can either let each processor work on a row (1x2), or a col (2x1). 

Does it make sense?

ax3l: Ah I see, let me add a little rephrasing suggestion so I get this right.

docs/source/usage/benchmarks.rst
```diff
@@ -75,6 +75,16 @@ To run:
 then the file generated are:  ../samples/8a_parallel_3Db_*
 
 
+Optional input parameter: pack
+
+Often a processor will write out a few small blocks.
+Using the example above, if the writer side uses 1024 processors, each processor will handle 16 blocks from the 32x32x16 grid.
+The selection of these 16 blocks can be from 1x1x16, to 2x2x4, etc..
+To specify which, the ``pack`` parameter can be specified, e.g., ``"pack=1 1 16"`` or ``"pack=2 2 4"``.
```
```suggestion
These 16 blocks per processor can be packed from a selection of 1x1x16 blocks, or 2x2x4, etc.
The ``pack`` parameter specifies this selection, e.g., ``"pack=1 1 16"`` or ``"pack=2 2 4"``.
```

ax3l: As a comment: just to mention, this is not exactly how AMReX does this.
AMReX uses either a space-filling Z curve or a simpler "knapsack" distribution when assigning boxes to MPI ranks. But I think the setup here is a reasonable idealization of the pattern for benchmarking, as long as we stay aware of this.

ax3l (approved):
Thanks a lot, @guj! :tada:

PR #1068, opened 2021-07-26
Document regexes for showing only attributes or datasets in new ADIOS2 schema

franzpoeschel:
This functionality was previously available via the `-A` or `-a` switches of `bpls`. Since we only use variables now, we have to become a bit more explicit and use regexes.
Document this so I don't have to look up how regex lookaheads work every time I want to do this.

### DIFF
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -105,6 +105,14 @@ The new layout may be activated **for experimental purposes** in two ways:
 
 The ADIOS2 backend will automatically recognize the layout that has been used by a writer when reading a dataset.
 
+.. tip::
+   This schema does not use ADIOS2 attributes any more, making the ``bpls -a`` and ``bpls -A`` switches useless for openPMD datasets.
+   Their functionality can be emulated via regexes:
+
+   * Print datasets and attributes: Default behavior
+   * Print datasets only: ``bpls -e '.*/__data__$'``
+   * Print attributes only: ``bpls -e '^(.(?!/__data__$))*$'``
+
 Memory usage
 ------------
 
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -106,7 +106,8 @@ The new layout may be activated **for experimental purposes** in two ways:
 The ADIOS2 backend will automatically recognize the layout that has been used by a writer when reading a dataset.
 
 .. tip::
-   This schema does not use ADIOS2 attributes any more, making the ``bpls -a`` and ``bpls -A`` switches useless for openPMD datasets.
+
+   This schema does not use ADIOS2 attributes anymore, thus ``bpls -a`` and ``bpls -A`` attribute switches do not show openPMD attributes.
    Their functionality can be emulated via regexes:
 
    * Print datasets and attributes: Default behavior


docs/source/backends/adios2.rst
```diff
@@ -105,6 +105,14 @@ The new layout may be activated **for experimental purposes** in two ways:
 
 The ADIOS2 backend will automatically recognize the layout that has been used by a writer when reading a dataset.
 
+.. tip::
+   This schema does not use ADIOS2 attributes any more, making the ``bpls -a`` and ``bpls -A`` switches useless for openPMD datasets.
```
ax3l: I think this needs an empty newline to render properly.

Maybe also a suggestion in the meaning:
```suggestion

   This schema does not use ADIOS2 attributes anymore, thus ``bpls -a`` and ``bpls -A`` attribute switches do not show openPMD attributes.
```

docs/source/backends/adios2.rst
```diff
@@ -105,6 +105,14 @@ The new layout may be activated **for experimental purposes** in two ways:
...
+   Their functionality can be emulated via regexes:
+
+   * Print datasets and attributes: Default behavior
```
ax3l, line 113: Is this what we mean here?
```suggestion
   * Print datasets and attributes: Default behavior (``bpls -a -l``)
```

docs/source/backends/adios2.rst
```diff
@@ -105,6 +105,14 @@ The new layout may be activated **for experimental purposes** in two ways:
...
+   * Print datasets only: ``bpls -e '.*/__data__$'``
+   * Print attributes only: ``bpls -e '^(.(?!/__data__$))*$'``
```
ax3l, line 115: Maybe suggest the user `-l` for long here?
```suggestion
   * Print datasets only: ``bpls -l -e '.*/__data__$'``
   * Print attributes only: ``bpls -l -e '^(.(?!/__data__$))*$'``
```

ax3l (commented):
Great idea to document this. I added small suggestions inline :)

docs/source/backends/adios2.rst
franzpoeschel: The only ADIOS2 attributes that we still use are internal ones such as `__openPMD_internal/openPMD2_adios2_schema`. So, the `-a` flag has (nearly) no influence. By default behavior, I mean "what happens if I call `bpls` without further parameters).

docs/source/backends/adios2.rst
franzpoeschel: This listing is supposed to show how to select attributes or datasets from a BP file. The `-l` switch doesn't really have anything to do with that.
We can add a little sentence below that the `-l` flag is useful to show the values of scalars?

ax3l: Oh I just add it by muscle memory because I always find it useful, that's why I was wondering if we just doc it that way.

docs/source/backends/adios2.rst
ax3l: Ah, so essentially
```suggestion
   * Print datasets and attributes: No extra flags needed
```

ax3l (approved):
Works for me like this :)

docs/source/backends/adios2.rst
```diff
@@ -105,6 +105,15 @@ The new layout may be activated **for experimental purposes** in two ways:
 
 The ADIOS2 backend will automatically recognize the layout that has been used by a writer when reading a dataset.
 
+.. tip::
+
+   This schema does not use ADIOS2 attributes anymore, thus ``bpls -a`` and ``bpls -A`` attribute switches do not show openPMD attributes.
+   Their functionality can be emulated via regexes:
```
ax3l, line 111:
```suggestion
   Their functionality can be emulated in ``bpls`` via regexes:
```

PR #1069, opened 2021-07-27
Fix: openpmd-pipe --help

franzpoeschel:
Follow-up to #1062, I forgot a little switch in the examples

### DIFF
--- a/src/binding/python/openpmd_api/pipe/__main__.py
+++ b/src/binding/python/openpmd_api/pipe/__main__.py
@@ -47,7 +47,8 @@ with the largest extent.
 
 Examples:
     {0} --infile simData.h5 simData_%T.bp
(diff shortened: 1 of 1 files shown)


src/binding/python/openpmd_api/pipe/__main__.py
```diff
@@ -47,7 +47,8 @@ def parse_args(program_name):
 
 Examples:
     {0} --infile simData.h5 simData_%T.bp
```
ax3l: Is this missing here as well?
```suggestion
    {0} --infile simData.h5 --outfile simData_%T.bp
```

franzpoeschel: You're absolutely right. I guess this was a copy-paste error then

ax3l (approved):
Awesome, good catch! :)

PR #1073, opened 2021-07-30
Test: Interleaved Write & Close

ax3l:
Add a test that performs writes to interleaved iterations and then closes iterations as they are done writing.

Issues:
- [x] as soon as `Iteration::close()` is used in this example, the writes segfault
```
#0  0x0000555555700b20 in ?? ()
#1  0x0000555555608a68 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() ()
#2  0x00007ffff7f45bc8 in void openPMD::detail::OldAttributeWriter::operator()<unsigned char>(openPMD::ADIOS2IOHandlerImpl*, openPMD::Writable*, openPMD::Parameter<(openPMD::Operation)18> const&) ()
   from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#3  0x00007ffff7f58cd6 in decltype (({parm#2}.(operator()<char>))((forward<openPMD::ADIOS2IOHandlerImpl*>)({parm#3}), (forward<openPMD::Writable*&>)({parm#3}), (forward<openPMD::Parameter<(openPMD::Operation)18> const&>)({parm#3}))) openPMD::switchType<openPMD::detail::OldAttributeWriter, openPMD::ADIOS2IOHandlerImpl*, openPMD::Writable*&, openPMD::Parameter<(openPMD::Operation)18> const&>(openPMD::Datatype, openPMD::detail::OldAttributeWriter, openPMD::ADIOS2IOHandlerImpl*&&, openPMD::Writable*&, openPMD::Parameter<(openPMD::Operation)18> const&) () from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#4  0x00007ffff7f008eb in openPMD::ADIOS2IOHandlerImpl::writeAttribute(openPMD::Writable*, openPMD::Parameter<(openPMD::Operation)18> const&) () from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#5  0x00007ffff7eb9c98 in openPMD::AbstractIOHandlerImpl::flush() () from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#6  0x00007ffff7ef51ef in openPMD::ADIOS2IOHandlerImpl::flush() () from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#7  0x00007ffff7ef5319 in openPMD::ADIOS2IOHandler::flush() () from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#8  0x00007ffff7ea873e in std::_Sp_counted_ptr_inplace<openPMD::ADIOS2IOHandler, std::allocator<openPMD::ADIOS2IOHandler>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() ()
   from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#9  0x0000555555608a68 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() ()
#10 0x00007ffff7e57afa in openPMD::internal::SeriesData::~SeriesData() () from /home/axel/src/openPMD/openPMD-api/build/lib/libopenPMD.so
#11 0x00005555555ffe08 in openPMD::Series::~Series() ()
...
```
- [x] in WarpX, a similar pattern throws https://github.com/ECP-WarpX/WarpX/pull/2150#issuecomment-890193164
```
  what():  [ADIOS2] Requested file has not been opened yet: openpmd_00003.bp
```

### DIFF
--- a/include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
+++ b/include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
@@ -78,11 +78,15 @@ protected:
     /**
      * Get the writable's containing file.
      * @param writable The writable whose containing file to figure out.
+     * @param preferParentFile If true, the file is set to the parent's file if
+     *     present. Otherwise, the parent file is only considered if no own file
+     *     is defined.
      * @return The containing file of the writable. If its parent is associated
      * with another file, update the writable to match its parent and return
      * the refreshed file.
      */
-    InvalidatableFile refreshFileFromParent( Writable * writable );
+    InvalidatableFile
+    refreshFileFromParent( Writable * writable, bool preferParentFile );
 
     /**
      * Figure out the file position of the writable.
@@ -195,26 +199,40 @@ std::string AbstractIOHandlerImplCommon< FilePositionType >::fullPath(
     }
 }
 
-
-template < typename FilePositionType >
+template< typename FilePositionType >
 InvalidatableFile
 AbstractIOHandlerImplCommon< FilePositionType >::refreshFileFromParent(
-    Writable * writable )
+    Writable * writable, bool preferParentFile )
 {
-    if ( writable->parent )
-    {
+    auto getFileFromParent = [ writable, this ]() {
         auto file = m_files.find( writable->parent )->second;
         associateWithFile( writable, file );
         return file;
+    };
+    if( preferParentFile && writable->parent )
+    {
+        return getFileFromParent();
     }
     else
     {
-        return m_files.find( writable )->second;
+        auto it = m_files.find( writable );
+        if( it != m_files.end() )
+        {
+            return m_files.find( writable )->second;
+        }
+        else if( writable->parent )
+        {
+            return getFileFromParent();
+        }
+        else
+        {
+            throw std::runtime_error(
+                "Internal error: Root object must be opened explicitly." );
+        }
     }
 }
 
-
-template < typename FilePositionType >
+template< typename FilePositionType >
 std::shared_ptr< FilePositionType >
 AbstractIOHandlerImplCommon< FilePositionType >::setAndGetFilePosition(
     Writable * writable, bool write )
--- a/include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
+++ b/include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
@@ -80,7 +80,7 @@ protected:
      * @param writable The writable whose containing file to figure out.
      * @param preferParentFile If true, the file is set to the parent's file if
      *     present. Otherwise, the parent file is only considered if no own file
-     *     is defined.
+     *     is defined. This is usually needed when switching between iterations when opening paths.
      * @return The containing file of the writable. If its parent is associated
      * with another file, update the writable to match its parent and return
      * the refreshed file.
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -487,6 +487,69 @@ TEST_CASE( "close_iteration_test", "[serial]" )
 }
 
 void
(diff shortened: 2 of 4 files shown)


franzpoeschel:
This seems to uncover (at least) two independent bugs. I'll push a fix for the first bug (explicit opening of files in ADIOS2 at createFile), the second one I'm still trying to get behind (wrong association of objects to files)

include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
```diff
@@ -78,11 +78,15 @@ class AbstractIOHandlerImplCommon : public AbstractIOHandlerImpl
     /**
      * Get the writable's containing file.
      * @param writable The writable whose containing file to figure out.
+     * @param preferParentFile If true, the file is set to the parent's file if
+     *     present. Otherwise, the parent file is only considered if no own file
+     *     is defined.
```
ax3l: @franzpoeschel shall we add a typical use case here?
```suggestion
     *     is defined. This is usually needed when switching between iterations when opening paths.
```

test/SerialIOTest.cpp
```diff
@@ -543,9 +553,24 @@ close_iteration_interleaved_test( std::string file_ending )
...
 TEST_CASE( "close_iteration_interleaved_test", "[serial]" )
 {
+    bool const bp_prefer_adios1 =
+        ( auxiliary::getEnvString( "OPENPMD_BP_BACKEND", "NOT_SET" ) == "ADIOS1" );
+
     for( auto const & t : testedFileExtensions() )
     {
-        close_iteration_interleaved_test( t );
+        //! @FIXME ADIOS1 bugs with Iteration::close()
+        if( bp_prefer_adios1 )
```
ax3l, line 562: We can investigate this in a separate PR.

test/SerialIOTest.cpp
```diff
@@ -543,9 +553,24 @@ close_iteration_interleaved_test( std::string file_ending )
...
+
+        close_iteration_interleaved_test( t, IterationEncoding::fileBased );
+        close_iteration_interleaved_test( t, IterationEncoding::groupBased );
+
+        // run this test for ADIOS2 & JSON only
+        if( t == "h5" )
+            continue;
+        if( t == "bp" && bp_prefer_adios1 )
+            continue;
+        close_iteration_interleaved_test( t, IterationEncoding::variableBased );
```
ax3l, line 573: Here we ideally should throw front-end errors that we can easily catch when `write.setIterationEncoding` is called for non-ADIOS2 backends.

Curiously, JSON just works (but is probably not doing something else than `groupBased`, from what I saw in the data diff?).

include/openPMD/IO/AbstractIOHandlerImplCommon.hpp
ax3l: I'll add this already in :)

ax3l:
Performed the WarpX test (crashes seen in https://github.com/ECP-WarpX/WarpX/pull/2150) this works now, too :tada:

test/SerialIOTest.cpp
ax3l: Quoting @franzpoeschel on slack: The basic idea was that this works for other backends as long as they only write one iteration:
https://github.com/openPMD/openPMD-api/blob/56c5353dd1354a100d941fc9f11a7f9c13c0c9ed/test/SerialIOTest.cpp#L3592

Here we write multiple.
Maybe we can guard somehow about this, not sure if/how we documented that `variableBased` works for only one iteration for other backends :)

PR #1078, opened 2021-08-05
ADIOS1: Test Interleaved It + Close

ax3l:
Same as #1073 but in ADIOS1.

@franzpoeschel @guj any idea? :)


franzpoeschel:
I'll have a look tomorrow.

franzpoeschel:
Umm.. where's the error? :D

ax3l:
Ha, does it work now? :laughing: 

Let's restart CI for once and if it works, well then we merge it in.

Issue #983, opened 2021-05-17
Reading: Improved Error Handling


ax3l:
When writing a "long time" into an openPMD iteration, which is especially the case for WarpX back-transformed diagnostics from the boosted frame, we already see the files on disk (valid ADIOS2) but they are not yet completely written with data or openPMD meta-data. An example meta-data that we write at the end (rightfully) is the base attribute `/openPMD`.

The typical workflow we need to handle is that users might already read part of a written series. Currently, we read all files of a file-based iteration encoding into the `Container` of Iterations, even if the `/openPMD` attribute does not yet exist for some of the iterations. This leads to read issues when accessing the Series. For example:
```
RuntimeError: [ADIOS2] Requested attribute (/openPMD) not found in backend.
```
and the Series cannot be constructed, although some valid (closed) iterations already exist.

Instead, we should gracefully throw errors and/or silently ignore the iterations in a Series that are not (yet) valid.

This is part of a larger effort to provide more graceful error handling, especially on the reading side.

PR #1080, opened 2021-08-05
Add header for openPMD-specific errors

franzpoeschel:
This is just the scaffolding for the more fine-grained error-handling that we want to implement.
This adds:
* A class `openPMD::Error` which should serve as base class for all of openPMD's error types. It should not be instatiated directly, but via child classes.
* A namespace `openPMD::error` which bundles all those error types.
* An example error type `openPMD::error::OperationUnsupportedInBackend`, without using it.

Related: #983 (this is not a fix, but the fundament for a fix)

TODO:
- [x] Python?


include/openPMD/Error.hpp
```diff
@@ -0,0 +1,48 @@
...
+private:
+    std::string m_what;
+
+protected:
+    Error( std::string what ) : m_what( what )
+    {
+    }
+
+public:
+    virtual const char * what() const noexcept final;
```
ax3l: Intentionally `final` though?

include/openPMD/Error.hpp
```diff
@@ -0,0 +1,48 @@
...
+ * in the openPMD::error namespace.
+ *
+ */
+class Error : public std::exception
+{
+private:
+    std::string m_what;
+
+protected:
+    Error( std::string what ) : m_what( what )
```
ax3l, line 22: We might want to default the other constructors as well here:
https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming)#Rule_of_Five

src/binding/python/Error.cpp
```diff
@@ -0,0 +1,19 @@
...
+    auto & baseError = py::register_exception< Error >( m, "Error" );
+    py::register_exception< error::OperationUnsupportedInBackend >(
+        m, "ErrorOperationUnsupportedInBackend", baseError );
+
+#ifndef NDEBUG
+    m.def( "test_throw", []( std::string description ) {
+        throw error::OperationUnsupportedInBackend( "json", description );
+    } );
+#endif
+}
```
ax3l:
```suggestion
}

```

include/openPMD/Error.hpp
franzpoeschel: I probably had a reason for it, but I can't remember it and I can't come up with one any more, so I'll take it out again.

ax3l (approved):
Thanks a lot! :tada:

PR #1081, opened 2021-08-06
File-based iteration encoding: Re-parse padding when necessary

### LINKED ISSUE #1048: infinite loop when filename does not contain '%T' but IterationEncoding is fileBased
**Describe the bug**
When setting a series' file name without the `%T` and switching the `IterationEncoding` to `fileBased` the code runs into an infinite loop. 

**To Reproduce**
C++
```C++

void error_openPMD()
{
    ::openPMD::Series series = ::openPMD::Series("error.h5", ::openPMD::Access::CREATE);
    series.iterations[10];
    series.setIterationEncoding(::openPMD::IterationEncoding::fileBased);
    series.flush(); 
};

int main()
{
    error_openPMD();
};

```

**Expected behavior**
Such a wrong configuration should be caught and an error should be thrown. 

**Software Environment**
 - version of openPMD-api: 0.13.2
 - installed openPMD-api via: conda-forge,
 - operating system: MacOS 10.15.7
 - version of HDF5: 1.10.6

**Additional context**
This infinite loop can only be triggered if the user initially set a wrong file name.

franzpoeschel:
Close #1048

- [x] Merge #1080 first
- [x] rebase after merge of #1080

Until now, the methods `SeriesInterface::setName` and `SeriesInterface::setIterationEncoding` do not modify the parsed filename prefix, padding and postfix.
Reparse them when necessary and detect wrong specifications.

Meta: We might want to introduce a label middle-end? This is not a user-facing change (except for fixing a bug), so I don't want to add the label Frontend-C++14. It's a change in the implementation of our frontend, i.e. the middle-end.

### DIFF
--- a/test/CoreTest.cpp
+++ b/test/CoreTest.cpp
@@ -688,14 +688,14 @@ TEST_CASE( "structure_test", "[core]" )
 
 TEST_CASE( "wrapper_test", "[core]" )
 {
-    Series o = Series("./new_openpmd_output.json", Access::CREATE);
+    Series o = Series("./new_openpmd_output_%T.json", Access::CREATE);
 
     o.setOpenPMDextension(42);
     o.setIterationEncoding(IterationEncoding::fileBased);
     Series copy = o;
     REQUIRE(copy.openPMDextension() == 42);
     REQUIRE(copy.iterationEncoding() == IterationEncoding::fileBased);
-    REQUIRE(copy.name() == "new_openpmd_output");
+    REQUIRE(copy.name() == "new_openpmd_output_%T");
     copy.setOpenPMD("1.2.0");
     copy.setIterationEncoding(IterationEncoding::groupBased);
     copy.setName("other_name");
--- a/include/openPMD/Error.hpp
+++ b/include/openPMD/Error.hpp
@@ -50,5 +50,17 @@ namespace error
         OperationUnsupportedInBackend(
             std::string backend_in, std::string what );
     };
+
+    /**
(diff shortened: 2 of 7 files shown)


franzpoeschel:
I don't understand [this line](https://github.com/openPMD/openPMD-api/blob/dd0abbc7506ef665ba4ae9bbfa618644ec8aa3be/src/Series.cpp#L333) in the old implementation of `setName()`. It checks if the *old* name contains `%T` right before overwriting it? Why should that make a difference? Is it supposed to be possible to set a name without expansion pattern after initializing a Series with an expansion pattern?

I implemented a prototype of this in commit 9f2c38d, should we go that way?

test/CoreTest.cpp
```diff
@@ -688,14 +688,14 @@ TEST_CASE( "structure_test", "[core]" )
 
 TEST_CASE( "wrapper_test", "[core]" )
 {
-    Series o = Series("./new_openpmd_output.json", Access::CREATE);
+    Series o = Series("./new_openpmd_output_%T.json", Access::CREATE);
 
     o.setOpenPMDextension(42);
     o.setIterationEncoding(IterationEncoding::fileBased);
```
franzpoeschel, line 694: Fun fact: This is exactly the kind of wrong usage that #1048 reports. Since this PR adds sanity-checks to catch such behavior, I'm changing this Series to be file-based now.

ax3l:
Merged #1080, ready for rebase :)

ax3l (approved):
Looks great, thanks a lot!

PR #1083, opened 2021-08-09
Bug fix: Don't forget closing files

franzpoeschel:
Noticed while working on #949, our refactored flush logic will sometimes forget to close a file

TODO
- [x] Same fix probably needed for write mode too
- [x] Find a test that shows this behavior without needing #949
- [x] ~~Probably `IterationOpened` needs a third entry `CloseMe` or something~~ no it does not

### DIFF
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -547,18 +547,18 @@ SeriesInterface::flushFileBased( iterations_iterator begin, iterations_iterator 
     if( IOHandler()->m_frontendAccess == Access::READ_ONLY )
         for( auto it = begin; it != end; ++it )
         {
+            // Phase 1
             switch( openIterationIfDirty( it->first, it->second ) )
             {
                 using IO = IterationOpened;
-            case IO::RemainsClosed:
-                continue;
             case IO::HasBeenOpened:
-                // continue below
+                it->second.flush();
+                break;
+            case IO::RemainsClosed:
                 break;
             }
 
-            it->second.flush();
-
+            // Phase 2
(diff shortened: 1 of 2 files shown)


franzpoeschel:
I've added a failing test now and will push the fix again after the CI has run

franzpoeschel:
This fails less tests than I had hoped for, but it does fail some. (Also, it fails for me locally)
I'll push the fix to see that things go green.

src/Series.cpp
```diff
@@ -547,18 +547,18 @@ SeriesInterface::flushFileBased( iterations_iterator begin, iterations_iterator
     if( IOHandler()->m_frontendAccess == Access::READ_ONLY )
         for( auto it = begin; it != end; ++it )
         {
+            // Phase 1
```
ax3l, line 550: What does Phase mean in this context? :)

ax3l (approved):
Thank you, great catch! :+1:

src/Series.cpp
franzpoeschel: It's just the 3 discrete steps that each iteration of the loop consists of: Flushing the frontend, closing the file, flushing the backend. I just added those comments as an optical guideline to find things faster.

ax3l: Cool, thanks! Do we maybe want to add this to the comment, e.g. "Phase 1: flush frontend" etc.? Good to keep the context in such long logic chains :)

ax3l:
I already applied this to 0.14.2 :)

PR #1084, opened 2021-08-10
HDF5: Fix String Vlen Attribute Reads

ax3l:
We unofficially try to also support HDF5 variable lengths strings in reading, just to increase compatibility.
The standard mandates fixed-lengths strings, so we only add this in reads.

This was never working it seems.


ax3l:
I dunno how to add an automated test for this besides writing something with `h5py` manually...

ax3l:
Since I tested this locally, I will merge it in to fix the existing bug.

PR #1085, opened 2021-08-10
Fix reading of vector attributes with only one contained value

franzpoeschel:
Noticed by @PrometheusPi 
Some of our backends can't distinguish attributes of scalar types and attributes of vector types containing only a single value
We already check for that situation in [some places](https://github.com/openPMD/openPMD-api/blob/dd0abbc7506ef665ba4ae9bbfa618644ec8aa3be/src/Mesh.cpp#L321), in [others](https://github.com/openPMD/openPMD-api/blob/dd0abbc7506ef665ba4ae9bbfa618644ec8aa3be/src/Mesh.cpp#L135) we forget it.

We will keep forgetting it, so this PR modifies the `Attribute::get()` function template to do some more automatic conversions:
* Convert from scalar values to vectors with a single value
* Convert between `std::vector` and `std::array` types (some backends cannot distinguish those either, so I'll take the chance to cover that)

First commit has a failing test, I'll push the fix afterwards. I will also take the freedom to do some simplifications in `Attribute.hpp`.

TODO:
- [x] Remove those manual checks that are scattered across our code base?

### DIFF
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -148,7 +148,7 @@ struct DoConvert<T, std::vector< U >, false>
     {
         std::vector< U > u;
         u.reserve( 1 );
-        u.push_back( *pv );
+        u.push_back( static_cast< U >( *pv ) );
         return u;
     }
 
@@ -209,7 +209,7 @@ struct DoConvert<std::vector< T >, std::array< U, n >, false>
         }
         for( size_t i = 0; i < n; ++i )
         {
-            u[ i ] = ( *pv )[ i ];
+            u[ i ] = static_cast< U >( ( *pv )[ i ] );
         }
         return u;
     }
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -235,7 +235,7 @@ getCast( Attribute const & a )
 {
     auto v = a.getResource();
 
-    // icpc does not like variantSrc::visit
+    // icpc 2021.3.0 does not like variantSrc::visit (with mpark-variant)
     // we use variantSrc::visit for the other compilers to avoid having an
     // endless list of if-then-else
     // also, once we switch to C++17, we might throw this out in
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -240,7 +240,7 @@ getCast( Attribute const & a )
     // endless list of if-then-else
     // also, once we switch to C++17, we might throw this out in
     // favor of a hopefully working std::visit
-#ifdef __INTEL_COMPILER
+#if defined(__ICC) || defined(__INTEL_COMPILER)
     if(auto pvalue_c = variantSrc::get_if< char >( &v ) )
         return DoConvert<char, U>{}(pvalue_c);
     else if(auto pvalue_uc = variantSrc::get_if< unsigned char >( &v ) )
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -138,6 +138,93 @@ struct DoConvert<std::vector< T >, std::vector< U >, false>
     }
 };
 
+// conversion cast: turn a single value into a 1-element vector
+template< typename T, typename U >
+struct DoConvert<T, std::vector< U >, false>
+{
+    static constexpr bool convertible = std::is_convertible<T, U>::value;
+
+    template< typename PV, typename UU = U >
+    auto operator()( PV pv )
+    -> typename std::enable_if< convertible, std::vector< UU > >::type
+    {
+        std::vector< U > u;
+        u.reserve( 1 );
+        u.push_back( *pv );
+        return u;
+    }
+
+    template< typename PV, typename UU = U >
+    auto operator()( PV )
+    -> typename std::enable_if< !convertible, std::vector< UU > >::type
+    {
+        throw std::runtime_error(
+            "getCast: no scalar to vector conversion possible.");
+    }
+};
+
+// conversion cast: array to vector
+// if a backend reports a std::array<> for something where the frontend expects
+// a vector
+template< typename T, typename U, size_t n >
+struct DoConvert<std::array< T, n > , std::vector< U >, false>
+{
+    static constexpr bool convertible = std::is_convertible<T, U>::value;
+
+    template< typename PV, typename UU = U >
+    auto operator()( PV pv )
+    -> typename std::enable_if< convertible, std::vector< UU > >::type
+    {
+        std::vector< U > u;
+        u.reserve( n );
+        std::copy( pv->begin(), pv->end(), std::back_inserter(u) );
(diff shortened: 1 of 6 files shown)


test/CoreTest.cpp
```diff
@@ -950,3 +950,73 @@ TEST_CASE( "load_chunk_wrong_datatype", "[core]" )
...
+            attr.get< std::vector< char > >() == std::vector< char >{ 'x' } );
+        REQUIRE(
+            attr.get< std::vector< unsigned char > >() ==
+            std::vector< unsigned char >{ 'x' } );
+        REQUIRE(
+            attr.get< std::vector< signed char > >() ==
+            std::vector< signed char >{ 'x' } );
+    }
+    {
+        std::array< double, 7 > array{ 0, 1, 2, 3, 4, 5, 6 };
```
ax3l:
```suggestion
        std::array< double, 7 > array{{ 0, 1, 2, 3, 4, 5, 6 }};
```

test/CoreTest.cpp
```diff
@@ -950,3 +950,73 @@ TEST_CASE( "load_chunk_wrong_datatype", "[core]" )
...
+            attr.get< std::vector< double > >() ==
+            std::vector< double >{ 0, 1, 2, 3, 4, 5, 6 } );
+        REQUIRE(
+            attr.get< std::vector< int > >() ==
+            std::vector< int >{ 0, 1, 2, 3, 4, 5, 6 } );
+    }
+    {
+        std::vector< double > vector{ 0, 1, 2, 3, 4, 5, 6 };
+        std::array< double, 7 > arraydouble{ 0, 1, 2, 3, 4, 5, 6 };
+        std::array< int, 7 > arrayint{ 0, 1, 2, 3, 4, 5, 6 };
```
ax3l:
```suggestion
        std::array< double, 7 > arraydouble{{ 0, 1, 2, 3, 4, 5, 6 }};
        std::array< int, 7 > arrayint{{ 0, 1, 2, 3, 4, 5, 6 }};
```

ax3l:
Thanks, good idea!

> Remove those manual checks that are scattered across our code base?

:+1: 

CI: note that we see a double free in one of the tests currently.

franzpoeschel:
> CI: note that we see a double free in one of the tests currently.

It looks like `icpc` doesn't really understand `variantSrc::visit`. I've gone back to the old way for Intel compilers again.

franzpoeschel:
> Remove those manual checks that are scattered across our code base?

Done that now, but I'm leaving [this one](https://github.com/openPMD/openPMD-api/blob/101723a95db6050d137d011b3c3ed652b2182c88/src/backend/Attributable.cpp#L306) in to stay on the safe side and always read `unitDimension` to its dedicated type.

include/openPMD/backend/Attribute.hpp
```diff
@@ -150,6 +235,12 @@ getCast( Attribute const & a )
 {
     auto v = a.getResource();
 
+    // icpc does not like variantSrc::visit
+    // we use variantSrc::visit for the other compilers to avoid having an
+    // endless list of if-then-else
+    // also, once we switch to C++17, we might throw this out in
+    // favor of a hopefully working std::visit
+#ifdef __INTEL_COMPILER
```
ax3l: The problem is only with the legacy `icpc` compiler from intel but not the new LLVM-based `icpx` compilers.
Thus, I would guard this like this:
```suggestion
#if defined(__ICC) || defined(__INTEL_COMPILER)
```
(That's what CMake uses as identifier: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Compiler/Intel-DetermineCompiler.cmake#L2)

The IntelLLVM (`icpx`) identifier on the other hand is:
```
(defined(__clang__) && defined(__INTEL_CLANG_COMPILER)) || defined(__INTEL_LLVM_COMPILER)
```
https://gitlab.kitware.com/cmake/cmake/-/blob/v3.21.1/Modules/Compiler/IntelLLVM-DetermineCompiler.cmake#L2

ax3l (commented):
I think that's a good approach and the PR looks great, thanks a lot!
I'll just slightly modify the ICC guard and then it can go in.

ax3l (approved):

include/openPMD/backend/Attribute.hpp
```diff
@@ -150,6 +235,12 @@ getCast( Attribute const & a )
 {
     auto v = a.getResource();
 
+    // icpc does not like variantSrc::visit
```
ax3l: Documenting version where we saw this, in case this gets fixed at some point
```suggestion
    // icpc 2021.3.0 does not like variantSrc::visit (with mpark-variant)
```

ax3l (approved):

ax3l:
@franzpoeschel given the original issue that @PrometheusPi saw: what backend and attribute exactly was that?
Was that with PIConGPU?

PrometheusPi:
@ax3l This is triggered by 0.6.0-dev PIConGPU data created with openPMD-api using the adios2 backend. The error appeared when using the latest openPMD-viewer (see also discussion on SLACK):

```
In [3]: ts=OpenPMDTimeSeries('/__workdir__/jet_preexpansion_502/simOutput/openPMD/',backend='openpmd-api')
RuntimeError                              Traceback (most recent call last)

<ipython-input-3-b2fd3d9a7bec> in <module>
----> 1 ts=OpenPMDTimeSeries('/__workdir__/jet_preexpansion_502/simOutput/openPMD/',backend='openpmd-api')

~/.conda/envs/analysis/lib/python3.9/site-packages/openpmd_viewer/openpmd_timeseries/main.py in init(self, path_to_dir, check_all_files, backend)
     85 
     86         # - Extract parameters from the first file
---> 87         t, params0 = self.data_reader.read_openPMD_params(self.iterations[0])
     88         self.t[0] = t
     89         self.extensions = params0['extensions']

~/.conda/envs/analysis/lib/python3.9/site-packages/openpmd_viewer/openpmd_timeseries/data_reader/data_reader.py in read_openPMD_params(self, iteration, extract_parameters)
    135 
    136         elif self.backend == 'openpmd-api':
--> 137             return io_reader.read_openPMD_params(
    138                     self.series, iteration, extract_parameters)
    139 

~/.conda/envs/analysis/lib/python3.9/site-packages/openpmd_viewer/openpmd_timeseries/data_reader/io_reader/params_reader.py in read_openPMD_params(series, iteration, extract_parameters)
     75             metadata = {}
     76             metadata['geometry'] = field.get_attribute('geometry')
---> 77             metadata['axis_labels'] = field.axis_labels
     78 
     79             # Swap the order of the labels if the code that wrote the HDF5 file

RuntimeError: getCast: no cast possible.
```

Most likely caused by reading `/data/0/fields/picongpu_idProvider/axisLabels` but @franzpoeschel knows this better.

ax3l:
Clarified offline:
- seen with `idProvider` (1d mesh)
- `axisLabels` is only entry then
- ADIOS1/2 attributes are always 1D arrays (of size 1-N)

PR #1087, opened 2021-08-11
setAttribute: Reject Empty Strings

ax3l:
Some backends, especially HDF5, do not allow us to define zero-sized strings. We thus need to catch this in the frontend and forward the restriction to the user.

Part of the debugging effort for the 0.14.* release in WarpX: https://github.com/ECP-WarpX/WarpX/pull/2150

Follow-up to #979

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -31,6 +31,7 @@
 #include <vector>
 #include <string>
 #include <cstddef>
+#include <type_traits>
 
 // expose private and protected members for invasive testing
 #ifndef OPENPMD_protected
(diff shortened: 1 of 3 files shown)


include/openPMD/backend/Attributable.hpp
```diff
@@ -420,6 +432,7 @@ AttributableInterface::setAttribute( std::string const & key, T value )
         return false;
     }
 }
+
 inline bool
 AttributableInterface::setAttribute( std::string const & key, char const value[] )
```
franzpoeschel, line 449: Should we add it to this overload then, too?

ax3l: I first started this but then skipped it because I would need to move the body out into another private member function.
But then it's also C++14 compatible, so can do :)

ax3l: Ah, but since that's in a header file it would change the API/ABI and I want to patch this into a patch release... hm...
Maybe not really worth it...?

ax3l: Actually, I can just make it an optional `if constexpr` and else an `if`
update: ah no, cannot because `.empty()` does not work for arbitrary types `T`

ax3l: Ah, found a simple update for C++14 :)

PR #1089, opened 2021-08-17
Don't read iterations if they have already been parsed

franzpoeschel:
Until now the condition to not reread an iteration in the streaming API was if the `Iteration.closed` attribute was present.
In datasets where such an attribute was never written, this leads to reparsing of the entire series in every iteration (i.e. quadratic parsing cost). Iterations are not changed after they are first seen, so rereading is not necessary.


ax3l (approved):
Thank you for the fix! :tada:

PR #1093, opened 2021-08-20
switchType: Do not take action object as argument any more

franzpoeschel:
Most instances did not use this anyway, it is unnecessary since it can be made part of the variadic arguments if needed. In other places, this just requires instantiating unneeded empty objects.

TODO:
- [x] Either remove the template operator macro if MSVC plays nice, or adapt it to this PR
- [x] Fix documentation

### DIFF
--- a/include/openPMD/benchmark/mpi/MPIBenchmark.hpp
+++ b/include/openPMD/benchmark/mpi/MPIBenchmark.hpp
@@ -236,17 +236,13 @@ namespace openPMD
             );
 
             template< typename T >
-            void operator()(
+            static void call(
+                BenchmarkExecution< Clock > &,
                 MPIBenchmarkReport< typename Clock::duration > & report,
                 int rootThread = 0
             );
 
-
-            template< int n >
-            void operator()(
-                MPIBenchmarkReport< typename Clock::duration > &,
-                int
-            );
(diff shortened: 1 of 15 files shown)


ax3l (commented):
Discussed on Zoom: stateful information can be passed via the variadic parameters of the signature, so no need to pass an instantiated object for the functor.

ax3l (approved):
Thanks, lgtm! :)

include/openPMD/benchmark/mpi/MPIBenchmark.hpp
```diff
@@ -274,7 +270,7 @@ namespace openPMD
         }
         for( Datatype dt: datatypes )
         {
-            switchType(
+            switchType< BenchmarkExecution< Clock > >(
                 dt,
                 exec,
```
ax3l, line 275: Bit confused here, do we still need the `exec` parameter passed then?

ax3l: Ah, looks like it below.

franzpoeschel: Yup, this one is for proving the comment above ;) "stateful information can be passed via the variadic parameters of the signature"

PR #1099, opened 2021-09-03
RecordComponent::resetDataset(): Reject undefined datatypes

franzpoeschel:
Currently, `RecordComponent::resetDataset()` does not check if the passed dataset has an undefined datatype. In case this happened erroneously, this will result in a backend error which is probably not the best UX. Instead, throw a clear error as soon as possible.

Slightly API breaking since it's theoretically possible to call `resetDataset()` twice, so the first call doesn't count. But in reality, that's useless.

### DIFF
--- a/src/RecordComponent.cpp
+++ b/src/RecordComponent.cpp
@@ -202,6 +202,15 @@ RecordComponent::flush(std::string const& name)
         }
     } else
     {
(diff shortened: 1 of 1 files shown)


ax3l (approved):
LGTM :+1:

PR #1100, opened 2021-09-03
Remove Datatype::DATATYPE

franzpoeschel:
Is this even used anywhere? 
I think this was initially used in some backend somewhere? If yes, it should not leak into the API datatypes and something like `auxiliary::Option` should be preferred.

Alternatively, deprecate for now and leave this PR open for a later point.

### DIFF
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -37,14 +37,11 @@
 namespace openPMD
 {
 
-constexpr int LOWEST_DATATYPE = 0;
-constexpr int HIGHEST_DATATYPE = 1000;
-
 /** Concrete datatype of an object available at runtime.
  */
 enum class Datatype : int
 {
-    CHAR = LOWEST_DATATYPE, UCHAR, // SCHAR,
+    CHAR, UCHAR, // SCHAR,
     SHORT, INT, LONG, LONGLONG,
     USHORT, UINT, ULONG, ULONGLONG,
     FLOAT, DOUBLE, LONG_DOUBLE,
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -71,8 +71,6 @@ enum class Datatype : int
 
     BOOL,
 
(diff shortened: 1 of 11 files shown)


include/openPMD/Datatype.hpp
```diff
@@ -71,8 +71,6 @@ enum class Datatype : int
 
     BOOL,
 
-    DATATYPE = HIGHEST_DATATYPE,
```
ax3l, line 74: Let's check if `HIGHEST_DATATYPE` is used somewhere

include/openPMD/Datatype.hpp
```diff
@@ -71,8 +71,6 @@ enum class Datatype : int
...
-
     UNDEFINED
```
ax3l, line 71: Potentially for iterating: last data type as `NUM_DATATYPES` and then it will automatically be the number of elements in the enum (because we start with `= 0`). (Often used for pre-variant implementations.)

franzpoeschel: But that would only really work if we did *not* assign it to `HIGHEST_DATATYPE`? Because that is defined as 1000, so you would enumerate from 0 to 1000, but we only have 30 datatypes or so.
See [here](https://stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class):
```cpp
enum class Example { A, B, C, D, E, Count };
```
I'd suggest, if we want to do such a thing, we just keep `Datatype::UNDEFINED` at the end of the enum and use that for enumerating.

Latest commit removes HIGHEST_DATATYPE and LOWEST_DATATYPE, they were only really used internally in `switchType` where it didn't make sense.

include/openPMD/Datatype.hpp
franzpoeschel: It's not

ax3l (approved):

ax3l:
@franzpoeschel please rebase since #1093 was merged :)

PR #1101, opened 2021-09-06
Two suggestions to fix CommonADIOS1IOHandlerImpl

franzpoeschel:
The current macro hack we have works, but is annoying:

1. It's impossible to include other openPMD headers into CommonADIOS1IOHandlerImpl.cpp, since that file is itself included into the openPMD namespace. The first commit fixes the namespaces.
2. IDEs usually don't understand what is going on and these macro things are not really compatible with the upcoming C++20 modules. The second commit suggests replacing things with a CRT pattern. Also has the advantage of having a separate class declaration for all things that are actually shared between the parallel and serial implementations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -426,11 +426,13 @@ set(IO_ADIOS1_SEQUENTIAL_SOURCE
         src/Error.cpp
         src/auxiliary/Filesystem.cpp
         src/ChunkInfo.cpp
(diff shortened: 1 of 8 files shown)


ax3l (approved):
Thank you, that looks good!
I like replacing macros with CRT anytime :D

ax3l:
Just saw that there is a regression in this PR when building an all-serial ADIOS1 build (via #1167).

PR #1104, opened 2021-09-15
Doc: More Locations -DPython_EXECUTABLE

ax3l:
Mention the `-DPython_EXECUTABLE` twice more in build examples.


franzpoeschel:
Thank you for this. That is the place where I searched for it and didn't find it :D
Since `-DopenPMD_USE_PYTHON` without specifying the executable can lead to weird errors on systems with multiple Python installations, I think it's good to use both together in examples

franzpoeschel (approved):

PR #1105, opened 2021-09-16
CMake: Set RPATHs on Installed Targets

ax3l:
Set RPATHS on installed targets.
Adding `INSTALL_RPATH_USE_LINK_PATH` simplifies the usage of openPMD-api when installed against shared dependencies or when building Python bindings while also building the core library as shared (non-default, but possible). Furthermore, we add `$ORIGIN` (or `@loader_path` on macOS) to express the internal dependency on our ADIOS1 wrapper libs.

Also: cleans up CMake target C++ standard Control via a function.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ openpmd_option(ADIOS2         "ADIOS2 backend (.bp files)"                AUTO)
 openpmd_option(PYTHON         "Enable Python bindings"                    AUTO)
 
 option(openPMD_INSTALL               "Add installation targets"             ON)
+option(openPMD_INSTALL_RPATH         "Add RPATHs to installed binaries"     ON)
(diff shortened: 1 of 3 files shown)


franzpoeschel:
Should we add this to the output of "openPMD build configuration:"?

~~Also, am I doing something wrong? When I build with `openPMD_INSTALL_RPATH=ON`, I don't see the rpath, neither with `readelf -d /install/location/libopenPMD.so` nor with `objdump -x /install/location/libopenPMD.so | grep RPATH`.~~ I wasn't looking properly

ax3l:
Yes, it's already printed in the installation summary :)

franzpoeschel:
Ah, next time I'll look closer

franzpoeschel (approved):

PR #1111, opened 2021-09-21
Lazy parsing: Make findable in docs and use in openpmd-ls

### LINKED ISSUE #1109: Doc: Lazy/Deferred Parsing in Reads
I think we need to add more docs on deferred/lazy reading in the manual.

Putting both keywords in the search
- https://openpmd-api.readthedocs.io/en/latest/search.html?q=deferred
- https://openpmd-api.readthedocs.io/en/latest/search.html?q=lazy

We need to place those keywords strategically in this section:
https://openpmd-api.readthedocs.io/en/latest/details/backendconfig.html#backend-independent-json-configuration

Currently gives no results that document:
```C++
Series series(
    basename + "%06T." + extension,
    Access::READ_ONLY,
    "{\"defer_iteration_parsing\": true}"
);
```

### LINKED ISSUE #1110: openpmd-ls: Use defer_iteration_parsing ?
Should we use `defer_iteration_parsing` (#1109) in `openpmd-ls`?

Constructor: https://github.com/openPMD/openPMD-api/blob/0.14.2/include/openPMD/cli/ls.hpp#L99-L102

Executed logic: https://github.com/openPMD/openPMD-api/blob/0.14.2/src/helper/list_series.cpp#L39

franzpoeschel:
Close #1109
Close #1110

### DIFF
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -25,7 +25,7 @@ In order to activate steps, it is imperative to use the :ref:`Streaming API <usa
 
 ADIOS2 release 2.6.0 contained a bug (fixed in ADIOS 2.7.0, see `PR #2348 <https://github.com/ornladios/ADIOS2/pull/2348>`_) that disallows random-accessing steps in file-based engines.
 With this ADIOS2 release, files written with steps may only be read using the streaming API.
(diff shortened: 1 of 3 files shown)


ax3l:
Thanks a lot!

Can we also add the doc hints for `usesteps`? It's part of https://github.com/openPMD/openPMD-api/issues/1109#issuecomment-923144264 and it might get lost if we auto-close it with the PR :)

franzpoeschel:
Done :)

ax3l (approved):
Discussed offline:
Lazy parsing is not (yet) default because one has to either:
- `Iteration::open()` the iteration before access explicitly or
- use the streaming API (`readIterations()`)

PR #1116, opened 2021-09-27
GH Actions: CI Savings

ax3l:
This uses GH Action CI resources more carefully by:

- Abort running pushes in CI if an update was pushed to a PR.
- Skip all but short style checks if a PR is opened as a draft.

### Convert an open PR to draft:
convert

### Open your next PR as a draft PR:
draft_pr

## Notes on Avoiding to Run GH Actions on Older Forks of this Repo

if you have an older fork of this repo on your account, all tests are run twice. Once on your repo (and not shown anywhere), once on our PR. (This default changed to opt-in if you created a recent fork.)

So, if you own an older fork of this repo on GitHub, you can also safe some CO2 if you like:
Go to your fork, click on the tab `Actions` -> Select workflow -> Click on `...` -> Select "Disable Workflow"

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -3,7 +3,7 @@ name: 🐧 Intel
 on: [push, pull_request]
 
 concurrency:
-  group: ${{ github.head_ref }}-intel
+  group: ${{ github.ref }}-${{ github.head_ref }}-intel
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -2,6 +2,10 @@ name: 🐧 Intel
 
 on: [push, pull_request]
 
+concurrency:
+  group: ${{ github.head_ref }}-intel
+  cancel-in-progress: true
+
 jobs:
   icc_cxxonly:
     name: ICC C++ only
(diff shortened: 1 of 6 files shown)


franzpoeschel:
Great, thank you for this!
You didn't change anything for `.appveyor.yml`, right? If I look into your CI run, MSVC did compile some files, but canceled then (e.g. [here](https://ci.appveyor.com/project/ax3l/openpmd-api/builds/40970685/job/9l80e68l5fcq4202)). Do you know what's up with that?
Also, when a PR is converted from draft to ready, this doesn't trigger the CI to run again, right? But I guess there's no way around that, so we'll just have to live with that.

ax3l:
Yes, I think AppVeyor just had a hick-up. It's not affected

> Also, when a PR is converted from draft to ready, this doesn't trigger the CI to run again, right? But I guess there's no way around that, so we'll just have to live with that.

I think it does actually. Same was as close/reopen triggers CI, too.

franzpoeschel:
> I think it does actually. Same was as close/reopen triggers CI, too.

Maybe I just didn't wait long enough when trying it out then. Looks good to me then :)

franzpoeschel (approved):

PR #1118, opened 2021-10-07
Fix: Read Inconsistent Zero Pads

ax3l:
Some codes mess up the zero-padding in `fileBased` encoding, e.g., when specifying padding to 5 digits but creating >100'000 output steps.

Files like those cannot yet be parsed using an `openpmd_%T.h5` file open string and fell back to no padding, which fails to open the file:
```
openpmd_00000.h5
openpmd_02000.h5
openpmd_101000.h5
openpmd_01000.h5
openpmd_100000.h5
openpmd_104000.h5
```

Error:
```
RuntimeError: [HDF5] Failed to open HDF5 file diags/diag1/openpmd_0.h5
```

## To Do

- [x] clean up
- [x] pass existing tests
- [x] add new test with above pattern

### DIFF
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -1618,15 +1618,19 @@ namespace
     }
 
     std::function<Match(std::string const &)>
-    buildMatcher(std::string const &regexPattern) {
+    buildMatcher(std::string const &regexPattern, int padding) {
         std::regex pattern(regexPattern);
 
-        return [pattern](std::string const &filename) -> Match {
+        return [pattern, padding](std::string const &filename) -> Match {
             std::smatch regexMatches;
             bool match = std::regex_match(filename, regexMatches, pattern);
-            int padding = match ? regexMatches[1].length() : 0;
-            return {match, padding, match ? std::stoull(regexMatches[1]) : 0};
-        };
+            int processedPadding = padding != 0
+                ? padding
+                : ( match ? regexMatches[ 1 ].length() : 0 );
+            return {
+                match,
+                processedPadding,
+                match ? std::stoull( regexMatches[ 1 ] ) : 0 };        };
     }
 
     std::function<Match(std::string const &)> matcher(
@@ -1673,7 +1677,7 @@ Given file pattern: ')END"
             nameReg += "+)";
         }
         nameReg += postfix + filenameSuffix + "$";
-        return buildMatcher(nameReg);
+        return buildMatcher(nameReg, padding);
     }
 } // namespace [anonymous]
 } // namespace openPMD
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -30,6 +30,7 @@
 #include "openPMD/Series.hpp"
 #include "openPMD/version.hpp"
 
+#include <cctype>
 #include <exception>
 #include <iomanip>
 #include <iostream>
@@ -76,8 +77,12 @@ namespace
      *          bool is True if file could be of type f and matches the iterationEncoding. False otherwise.
      *          int is the amount of padding present in the iteration number %T. Is 0 if bool is False.
      */
-    std::function<Match(std::string const &)>
-    matcher(std::string const &prefix, int padding, std::string const &postfix, Format f);
+    std::function<Match(std::string const &)> matcher(
+        std::string const &prefix,
+        int padding,
+        std::string const &postfix,
+        Format f,
+        std::string const &fullPattern);
 } // namespace [anonymous]
 
 struct SeriesInterface::ParsedInput
@@ -796,7 +801,7 @@ SeriesInterface::readFileBased( )
 
     auto isPartOfSeries = matcher(
         series.m_filenamePrefix, series.m_filenamePadding,
-        series.m_filenamePostfix, series.m_format);
+        series.m_filenamePostfix, series.m_format, series.m_name);
     bool isContained;
     int padding;
     uint64_t iterationIndex;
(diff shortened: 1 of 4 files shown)


src/Series.cpp
```diff
@@ -1171,7 +1173,8 @@ SeriesInterface::iterationFilename( uint64_t i )
         return series.m_overrideFilebasedFilename.get();
     }
     std::stringstream iteration( "" );
-    iteration << std::setw( series.m_filenamePadding )
+    int filenamePadding = series.m_filenamePadding.at(i);
```
ax3l: I have the feeling the `SeriesInterface::iterationFilename` function might be better placed in `Iteration`?
That could potentially safe us the `m_filenamePadding` member.

franzpoeschel:
I've reverted your changes and pushed a suggested implementation that should fix this
A quick test seems to confirm it:
```
In [1]: import openpmd_api as io                

In [2]: io.Series("data%T.json", io.Access.create)                                              
Out[2]: <openPMD.Attributable with '8' attributes>

In [3]: series = io.Series("data%T.json", io.Access.create)                                     
In [4]: series.iterations[3000]                 
Out[4]: <openPMD.Iteration of at t = '0.000000 s'>

In [5]: del series      

In [6]: series = io.Series("data%02T.json", io.Access.read_only)                                

In [7]: [key for key in series.iterations]      
Out[7]: [3000]
```

The trick was to modify the regex a bit that parses a directory's filenames
e.g. the regex for a padding of 5: `^<prefix>(([1-9][[:digit:]]*)?([[:digit:]]{5}))<postfix>$`:
* The part after the question mark: Exactly five digits, these need to be given at the minimum
* The part before the question mark: Optionally, further digits, but only if the first digit is not 0

In consequence, this regex now accepts `00001`, `50000`, `550000`, `999999999999999999`, but **not** `001`, `1` or `000050000`

ax3l:
Cool, I think that works well!

I think a general thing we could also consider is that prefixes that are weirdly named, e.g., `diag1` that then read `diag10004.bp` can cause problems. I generally implement this in user-code with a separating underscore to avoid the ambiguity.

Should we maybe add a warning if the user-provided name does end in a number of the prefix before the `%T`?

ax3l:
I pushed test bug it looks like that one passes already before the patch - as long as we use `%T`.

Maybe that's all that is needed in openPMD-viewer then.

ax3l:
Ah, I think the problem was not in `Series::Series` but when we actually read the data and do a `series.flush()` on a specific iteration. Let's refine the test.

franzpoeschel:
> Should we maybe add a warning if the user-provided name does end in a number of the prefix before the `%T`?

Theoretically we could start soft-enforcing the underscore by giving a warning if anything but an underscore is used?

A prefix `diag1` should be technically unproblematic since the `1` is a hardcoded part of the regex and will not be considered for the iteration number. But user-side confusion is very likely in such cases, yeah



> I pushed test bug it looks like that one passes already before the patch - as long as we use `%T`.

Yep, if using `%T`, the regex will just match for `[[digit]]+`, so any number of digits will be accepted. So, `%T` is a good catch-all for inconsistent paddings. This could well fix your use case, but I still think that going forward with this PR is necessary. `%06T` should be able to read all files that it produces.

franzpoeschel:
> Theoretically we could start soft-enforcing the underscore by giving a warning if anything but an underscore is used?

I've pushed a commit that does this. In case we want to only warn if the prefix ends in a digit, that should be easy enough to fix.

Apparently, there is an instance of `data%T` in our test suite:
```
[Warning] In file-based iteration encoding, it is strongly recommended
to prepend the expansion pattern of the filename with an underscore '_'.
Example: 'data_%T.json' or 'simOutput_%06T.h5'
Given file pattern: 'data%T'
```

ax3l:
Not sure if we need to warn on not ending in an underscore, ending in a digit is maybe worth a warning.

test/SerialIOTest.cpp
```diff
@@ -1660,32 +1682,65 @@ void fileBased_write_test(const std::string & backend)
...
+
+    // read back with auto-detection (allow relaxed/overflow padding)
     {
-        Series s = Series("../samples/subdir/serial_fileBased_write%08T." + backend, Access::READ_ONLY);
-        REQUIRE(s.iterations.size() == 6);
+        Series s = Series("../samples/subdir/serial_fileBased_write%T." + backend, Access::READ_ONLY);
+        REQUIRE(s.iterations.size() == 8);
+        REQUIRE(s.iterations.contains(4));
+        REQUIRE(s.iterations.contains(10));
+        REQUIRE(s.iterations.contains(123456));
```
franzpoeschel, line 1732: You initially had this line as `REQUIRE(!s.iterations.contains(123456))`. 
The whole point of this PR is that it *should* be read.

test/SerialIOTest.cpp
```diff
@@ -1642,15 +1642,37 @@ void fileBased_write_test(const std::string & backend)
...
         Series o = Series("../samples/subdir/serial_fileBased_write%01T." + backend, Access::READ_WRITE);
 
-        REQUIRE(o.iterations.empty());
+        REQUIRE(o.iterations.size() == 1);
+        /*
+         * 123456 has no padding, it's just a very long number.
+         * So even when opening the series with a padding of 1,
+         * that iteration will be opened.
+         */
+        REQUIRE(o.iterations.count(123456) == 1);
```
ax3l, line 1675: I was initially thinking here: should we read `123456` if our padding was specifically asked to be `00`?

But if we would skip it, then this would be a bit inconsistent with reading unpadded numbers.
Thus, I think the logic change here in the test that you added is the proper way to handle this :+1:

test/SerialIOTest.cpp
ax3l: Oh, must have been a copy paste mistake. Thanks for highlighting!

test/SerialIOTest.cpp
```diff
@@ -1660,32 +1682,65 @@ void fileBased_write_test(const std::string & backend)
...
+        REQUIRE(s.iterations.contains(4));
+        REQUIRE(s.iterations.contains(10));
+        REQUIRE(s.iterations.contains(123456));
+
+        REQUIRE(s.iterations[3].time< double >() == 3.0);
+        REQUIRE(s.iterations[4].time< double >() == 4.0);
+        REQUIRE(s.iterations[5].time< double >() == 5.0);
+        REQUIRE(s.iterations[10].time< double >() == 10.0);
+        REQUIRE(s.iterations[123456].time< double >() ==
+            static_cast<double>(123456));
```
ax3l: Style consistency in the block:
```suggestion
            static_cast< double >(123456));
```

ax3l:
@franzpoeschel, thanks a lot for your help and updates :sparkling_heart:

ax3l:
CI: `conda install` seems to be hanging today, nothing to worry

test/SerialIOTest.cpp
franzpoeschel: > should we read 123456 if our padding was specifically asked to be 00?

Yep, it should, because that number is not padded ;)

ax3l:
@camille12225 reports that even with this port in 0.14.3, she still sees issues with inconsistent padding in files.

I added a report to #1173

PR #1119, opened 2021-10-07
added BP5 options

guj:


src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2392,6 +2392,24 @@ namespace detail
                 m_IO.SetParameter(
                     "SubStreams", std::to_string( num_substreams ) );
             }
+
+            // BP5 parameters
+            auto numAgg = auxiliary::getEnvNum( "OPENPMD_ADIOS2_BP5_NumAgg", 0 );
```
ax3l, line 2397: Please don't forget documenting those options in
  https://openpmd-api.readthedocs.io/en/0.14.2/backends/adios2.html

We should also expose them as JSON options:
  https://openpmd-api.readthedocs.io/en/0.14.2/details/backendconfig.html#adios2
that allows us to set them programmatically

guj: OK, updated the backends/adios2.html

franzpoeschel: The ADIOS2 backend just forwards any key-value pair under `adios2.engine.parameters` to ADIOS2, so exposing these as JSON parameters happens automatically, i.e. they're available also without this PR.

Two problems with this PR:
1. If the parameter is set via JSON *and* via environment variable, the JSON configuration should take precedence over the environment variable. Currently, this PR prioritizes environment variables over the JSON configuration. Can you please use the same logic as for the existing parameters? These implement the correct precedence.
2. ~~Capitalization, PR #1043 will make JSON parameters capitalization-insensitive. Since I actually missed this part in the PR, I will fix it there, so we can ignore it for this PR.~~ This can be fixed without affecting this PR at all

guj: Thanks Franz. code is updated accordingly.

franzpoeschel:
I've pushed two small changes to enforce file endings for BP5 as is done for the other ADIOS2 engines as well, and to fix a small typo. Otherwise looks good to me

franzpoeschel (approved):

guj:
> I've pushed two small changes to enforce file endings for BP5 as is done for the other ADIOS2 engines as well, and to fix a small typo. Otherwise looks good to me

Thanks Franz.

franzpoeschel:
@ax3l What I said yesterday about `OPENPMD_BP_BACKEND` was wrong: that variable does not decide whether to use BP3, BP4 or BP5, it decides whether to use ADIOS1 or ADIOS2. It has nothing to do with this PR, so I'll keep my approval.

Failing runs are what I mentioned yesterday about our Windows CI currently having some troubles.

docs/source/backends/adios2.rst
```diff
@@ -54,6 +54,11 @@ environment variable                  default    description
 ``OPENPMD2_ADIOS2_SCHEMA``            ``0``      ADIOS2 schema version (see below)
 ``OPENPMD_ADIOS2_STATS_LEVEL``        ``0``      whether to generate statistics for variables in ADIOS2. (``1``: yes, ``0``: no).
 ``OPENPMD_BP_BACKEND``                ``ADIOS2`` Chose preferred ``.bp`` file backend if ``ADIOS1`` and ``ADIOS2`` are available.
+``OPENPMD_ADIOS2_BP5_BufferChunkMB``  ``0``      ADIOS2 BP5 engine: applies when using either EveryoneWrites or EveryoneWritesSerial aggregation
+``OPENPMD_ADIOS2_BP5_MaxShmMB``       ``0``      ADIOS2 BP5 engine: applies when using TwoLevelShm aggregation
+``OPENPMD_ADIOS2_BP5_NumSubFiles``    ``0``      ADIOS2 BP5 engine: num of subfiles
+``OPENPMD_ADIOS2_BP5_NumAgg``         ``0``      ADIOS2 BP5 engine: num of aggregators
+``OPENPMD_ADIOS2_BP5_TypeAgg``        ````       ADIOS2 BP5 engine: aggregation type. (EveryoneWrites, EveryoneWritesSerial, TwoLevelShm)
```
ax3l: Render issue:
```suggestion
``OPENPMD_ADIOS2_BP5_TypeAgg``        *empty*    ADIOS2 BP5 engine: aggregation type. (EveryoneWrites, EveryoneWritesSerial, TwoLevelShm)
```

ax3l:
Discussed today: let's document the new JSON BP5 options that we get from the new ADIOS engine also in https://github.com/openPMD/openPMD-api/blob/dev/docs/source/details/backendconfig.rst#adios2

ax3l (commented):
Added a small inline comment on rst

guj:
> Discussed today: let's document the new JSON BP5 options that we get from the new ADIOS engine also in https://github.com/openPMD/openPMD-api/blob/dev/docs/source/details/backendconfig.rst#adios2

So Franz will update this?

franzpoeschel:
We never documented the concrete options that can be forwarded to ADIOS2, but I added a link to the ADIOS2 documentation now where they can be looked up @ax3l

ax3l (approved):
Thank you both! :tada:

franzpoeschel:
Note for the milestone: BP5 crashes when not using steps. We should enable steps by default in BP5 before putting out 0.14.5. @ax3l

ax3l:
@franzpoeschel ok, does this need a separate PR?

franzpoeschel:
> @franzpoeschel ok, does this need a separate PR?

I think that we might do it as part of https://github.com/openPMD/openPMD-api/pull/1215 and just collect BP5-specific adaptations therein

ax3l:
As discussed: we do not add BP5 options to the 0.14.* release line but in an upcoming 0.15.0 release.

PR #1126, opened 2021-10-18
[WIP] ADIOS2: Fix Parallel Resize

ax3l:
We found in WarpX (https://github.com/ECP-WarpX/WarpX/pull/1898) that MPI-parallel resizes (at least in file-based encoding) do not yet work with the ADIOS2 backend (parallel HDF5 is fine).
The current behavior is that following resizes are not performed or do not show up in read as a changed extent/shape at least. (Tests do not crash in write because ADIOS let's data be written independent of extent, which is meta-data.)

- [x] fix test: was uncovered before
- [ ] fix implementation

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -390,7 +390,7 @@ extendDataset( std::string const & ext )
     std::iota( data2.begin(), data2.end(), 25 );
     {
         Series write( filename, Access::CREATE, MPI_COMM_WORLD );
-        if( ext == "bp" && write.backend() != "ADIOS2" )
+        if( ext == "bp" && write.backend() != "MPI_ADIOS2" )
         {
             // dataset resizing unsupported in ADIOS1
             return;


test/SerialIOTest.cpp
```diff
@@ -4154,8 +4154,9 @@ iterate_nonstreaming_series(
         Series writeSeries( file, Access::CREATE );
         if( variableBasedLayout )
         {
-            if( writeSeries.backend() != "ADIOS2" )
+            if( writeSeries.backend() == "ADIOS1" )
```
franzpoeschel: Can you revert this change? This `if` condition is supposed to filter all backends that support variable-based iteration encoding, which is only ADIOS2. See the HDF5 error message also:
```
  #000: H5D.c line 152 in H5Dcreate2(): unable to create dataset
    major: Dataset
  #001: H5Dint.c line 338 in H5D__create_named(): unable to create and link to dataset
    major: Dataset
  #002: H5L.c line 1605 in H5L_link_object(): unable to create new link to object
    major: Links
  #003: H5L.c line 1846 in H5L__create_real(): can't insert link
    major: Links
    minor: Unable to insert object
  #004: H5Gtraverse.c line 848 in H5G_traverse(): internal path traversal failed
    major: Symbol table
    minor: Object not found
  #005: H5Gtraverse.c line 624 in H5G__traverse_real(): traversal operator failed
    major: Symbol table
    minor: Callback failed
  #006: H5L.c line 1641 in H5L__link_cb(): name already exists
    major: Links
    minor: Object already exists
```

test/ParallelIOTest.cpp
```diff
@@ -390,7 +390,7 @@ extendDataset( std::string const & ext )
     std::iota( data2.begin(), data2.end(), 25 );
     {
         Series write( filename, Access::CREATE, MPI_COMM_WORLD );
-        if( ext == "bp" && write.backend() != "ADIOS2" )
+        if( ext == "bp" && write.backend() == "MPI_ADIOS1" )
```
franzpoeschel: This does not yet uncover the bug in the tests for me? The parallel tests are running clean also with this change.

franzpoeschel:
> The current behavior is that following resizes are not performed or do not show up in read as a changed extent/shape at least. 

This was the exact same behavior in ADIOS2 before resizing was implemented there, so I guess it was forgotten somewhere. 
See [this issue](https://github.com/ornladios/ADIOS2/issues/2530) for it.

test/SerialIOTest.cpp
ax3l: ups, thought it only runs over `.bp`. Will simplify

test/ParallelIOTest.cpp
```diff
@@ -390,7 +390,7 @@ extendDataset( std::string const & ext )
     std::iota( data2.begin(), data2.end(), 25 );
     {
         Series write( filename, Access::CREATE, MPI_COMM_WORLD );
```
ax3l, line 392: Should we move the `write` below the `return` for ADIOS1?
Otherwise we create that empty ghost series.

franzpoeschel: It's a bit unfortunate, but we have to do it this way unfortunately, see `write.backend()` in the if condition. However, I can add a commit in #1043 that uses `{"backend": "adios2"}` to explicitly select backends up front.

franzpoeschel: I added that commit

ax3l:
Marked as WIP as I am waiting for further tests to generally pass in https://github.com/ECP-WarpX/WarpX/pull/1898 before trying the next reproducer.

PR #1127, opened 2021-10-18
Fix AppVeyor: Python Executable

ax3l:
Avoid mismatching system Python and Conda Python in Windows CI.

Ref.: https://www.appveyor.com/docs/windows-images-software/

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,32 +1,18 @@
 environment:
   matrix:
-    - TARGET_ARCH: x86
-      CONDA_PY: 36
-      CONDA_INSTALL_LOCN: C:\\Miniconda36
-      APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-      platform: x86
-      SHARED: ON
-
     - TARGET_ARCH: x64
-      CONDA_PY: 37
+      CONDA_PY: 3.7
       CONDA_INSTALL_LOCN: C:\\Miniconda37-x64
       APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
       platform: x64
-      SHARED: ON
+      SHARED: OFF
 
     - TARGET_ARCH: x86
-      CONDA_PY: 37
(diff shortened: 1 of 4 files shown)


ax3l:
@franzpoeschel fyi, I disable the openpmd-pipe test for now in Windows, because we will need to write it a bit more generically to get the Python executable path set up properly. We should also check if `sh` is present before relying on it.

.appveyor.yml
```diff
@@ -121,12 +106,12 @@ before_build:
...
-  - cmd: cmake -G "%OPENPMD_CMAKE_GENERATOR%" -DCMAKE_BUILD_TYPE=%CONFIGURATION% -DBUILD_SHARED_LIBS=%SHARED% -DBUILD_TESTING=ON -DCMAKE_INSTALL_PREFIX="%CONDA_INSTALL_LOCN%" -DCMAKE_INSTALL_BINDIR="Library\bin" ".."
+  - cmd: cmake -G "%OPENPMD_CMAKE_GENERATOR%" -DCMAKE_BUILD_TYPE=%CONFIGURATION% -DBUILD_SHARED_LIBS=%SHARED% -DBUILD_TESTING=ON -DopenPMD_USE_PYTHON=ON -DPython_EXECUTABLE="%CONDA_INSTALL_LOCN%\python.exe" -DCMAKE_INSTALL_PREFIX="%CONDA_INSTALL_LOCN%" -DCMAKE_INSTALL_BINDIR="Library\bin" ".."
 
 build_script:
-  - cmd: cmake --build . --config %CONFIGURATION%
+  - cmd: cmake --build . --config %CONFIGURATION% -j 2
   - cmd: cmake --build . --config %CONFIGURATION% --target install
 
 test_script:
   - cmd: ctest -V -C %CONFIGURATION%
```
ax3l, line 116: This probably only works because we install first :sweat_smile: 
CTest Python path is still a bit flaky for multi-config generators. Ok for a follow-up PR.

PR #1128, opened 2021-10-19
CXX Std: Remember <variant> Impl.

ax3l:
We use `<variant>` or `<mpark/variant.hpp>` in our public API interface for datatypes, depending on the C++ standard.

This pull request makes sure that the same implementation is used in downstream code, even if the C++ standard is switched. This avoids ABI issues when, e.g., using a C++14 built openPMD-api in a C++17 downstream code.

Fix #1121

X-ref:
- https://github.com/ECP-WarpX/WarpX/pull/2440
- https://github.com/ECP-WarpX/WarpX/issues/1776#issuecomment-942689086
- https://github.com/conda-forge/warpx-feedstock/pull/30

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,26 +479,36 @@ target_include_directories(openPMD PUBLIC
     $<INSTALL_INTERFACE:include>
 )
 
+# is this a C++17 compile? If yes, skip MPark.Variant
+try_compile(openPMD_HAS_CXX17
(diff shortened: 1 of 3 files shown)


franzpoeschel (approved):

franzpoeschel:
Very nice, thank you for the fix :)

ax3l:
Dang, forgot to add `cmake/` to `MANIFEST.in`.

Fix in: #1140

PR #1130, opened 2021-10-22
8b minor updates

guj:
- printing out the iterations visited
- printing out iteration encoding of the file
- a bug fix


ax3l:
Thank you for the PR!

@guj can you please rebase your PR against `dev` so it only contains the fix for the test?

> a bug fix

Can you please describe this fix in more detail?

guj:
> Thank you for the PR!
> 
> @guj can you please rebase your PR against `dev` so it only contains the fix for the test?
> 
> > a bug fix
> 
> Can you please describe this fix in more detail?

I will close this one.

PR #1133, opened 2021-10-29
Include check for paged allocation

jeanbez:
Include a check for paged allocation for the minimum required version of HDF5 1.10.1


src/IO/HDF5/ParallelHDF5IOHandler.cpp
```diff
@@ -70,8 +70,10 @@ ParallelHDF5IOHandlerImpl::ParallelHDF5IOHandlerImpl(
         hsize_t page_size;
         tstream >> page_size;
 
+#if H5_VERSION_GE(1,10,1)
```
ax3l: Do you want to move this even up to guard the whole block? (line 66-77)

jeanbez: Sure, makes sense!

ax3l (approved):
Thank you! :tada:

PR #1135, opened 2021-11-03
Release 0.14.3 backports

ax3l:
https://github.com/openPMD/openPMD-api/milestone/2

### Read Bugs, C++17 Mixing and HDF5 Performance

This release makes reads more robust by fixing small API, file-based parsing and test bugs. Building the library in C++14 and using it in C++17 will not result in incompatible ABIs anymore. HDF5 1.10.1+ performance was improved significantly.

Thanks to @ax3l, @franzpoeschel and @jeanbez for contributing to this release!

### Python

```python
import openpmd_api
```

### C++

```C++
#include <iostream>

int main() {
    // query compile-time macros:

    return 0;
}
```

### CLI

```bash
openpmd-ls --version
```


CHANGELOG.rst
```diff
@@ -3,6 +3,49 @@
 Changelog
 =========
 
+0.14.2
```
ax3l, line 6: Upsi :see_no_evil:

Should have read `0.14.3`. Classic.

PR #1137, opened 2021-11-04
Fix parsing for grid spacing with type long double

franzpoeschel:
`Mesh.cpp` has an explicit instantiation for long double grid spacings:
```cpp
template
Mesh&
Mesh::setGridSpacing(std::vector< long double > const & gs);
```

This is important for our Python API as floats default to `long double` there:
```python
temperature.grid_spacing = [1.0, 1.0]
```
Result of that:
```
  long double  /data/9/meshes/temperature/gridSpacing       attr   = {1, 1}
```

But upon trying to open that Series again:
```
  Unexpected Attribute datatype for 'gridSpacing'
```
Because our reading routines only check for float and double.

First commit: Failing test
Second commit: Fix

We should maybe scan our codebase for similar errors.

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -331,6 +331,8 @@ Mesh::read()
         setGridSpacing(a.get< std::vector< float > >());
     else if( *aRead.dtype == DT::VEC_DOUBLE || *aRead.dtype == DT::DOUBLE )
         setGridSpacing(a.get< std::vector< double > >());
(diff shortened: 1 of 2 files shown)


franzpoeschel:
Python seems to be really committed to using `long double`, I'm finding no way to force it to another datatype:
```python
temperature.grid_spacing = np.array([1, 1], dtype=np.double)
```
Even this one creates long doubles. This could mean that it is currently impossible in our API to specify a dataset's grid spacing in Python.

franzpoeschel:
It looks like the CI is again in a mood

ax3l (approved):
Thanks a lot!

PR #1138, opened 2021-11-04
Wheels 0.14.3 ppc64le

ax3l:
Test patch in https://github.com/pypa/cibuildwheel/pull/906

Follow-up to #1136


ax3l:
Replaced by #1145

ax3l:
Re-opened, because emulation takes too long to fit in GH Action constrains (360min per build).

PR #1141, opened 2021-11-08
Add new Python-based linear/streaming API tests

franzpoeschel:
TODO:
- [x] Add existing Python-based streaming tests to CMake test suite, we forgot that so far
- [x] Extend that test to include demonstrate some more openPMD-api features from a recent workshop

### DIFF
--- a/examples/10_streaming_write.py
+++ b/examples/10_streaming_write.py
@@ -34,8 +34,8 @@ if __name__ == "__main__":
     # create a series and specify some global metadata
     # change the file extension to .json, .h5 or .bp for regular file writing
     series = io.Series("simData.sst", io.Access_Type.create, config)
-    series.set_author("Franz")
-    series.set_software("python snippet")
+    series.set_author("Franz Poeschel <f.poeschel@hzdr.de>")
+    series.set_software("openPMD-api-python-examples")
 
     # now, write a number of iterations (or: snapshots, time steps)
     for i in range(10):
@@ -65,7 +65,7 @@ if __name__ == "__main__":
                                dtype=np.dtype("double"))
         for dim in ["x", "y", "z"]:
             pos = electronPositions[dim]
-            pos.reset_dataset(io.Dataset(np.dtype("double"), [length]))
+            pos.reset_dataset(io.Dataset(local_data.dtype, [length]))
             pos[()] = local_data
 
         # optionally: flush now to clear buffers
--- a/examples/10_streaming_write.py
+++ b/examples/10_streaming_write.py
@@ -3,31 +3,92 @@ import openpmd_api as io
 import numpy as np
 import sys
 
+# pass-through for ADIOS2 engine parameters
+# https://adios2.readthedocs.io/en/latest/engines/engines.html
+config = """
+{
+  "adios2": {
+    "engine": {
+      "parameters": {
+        "Threads": "4"
+      }
+    },
+    "dataset": {
+      "operators": [
+        {
+          "type": "bzip2"
+        }
+      ]
+    }
+  }
+}
+"""
+
 if __name__ == "__main__":
-    if 'adios2' not in io.variants or not io.variants['adios2']:
-        print('This example requires ADIOS2')
-        sys.exit(0)
-
-    series = io.Series("stream.sst", io.Access_Type.create)
-
+    # this block is for our CI, SST engine is not present on all systems
     backends = io.file_extensions
     if "sst" not in backends:
         print("SST engine not available in ADIOS2.")
         sys.exit(0)
 
-    datatype = np.dtype("double")
-    length = 10
-    global_extent = [10]
-    dataset = io.Dataset(datatype, global_extent)
+    # create a series and specify some global metadata
+    # change the file extension to .json, .h5 or .bp for regular file writing
+    series = io.Series("simData.sst", io.Access_Type.create, config)
+    series.set_author("Franz")
+    series.set_software("python snippet")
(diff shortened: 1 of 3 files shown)


ax3l:
trigger CI :D

examples/10_streaming_write.py
```diff
@@ -3,31 +3,92 @@
...
+
+    # now, write a number of iterations (or: snapshots, time steps)
+    for i in range(10):
+        # Use `series.write_iterations()` instead of `series.iterations`
+        # for streaming support (while still retaining file-writing support).
+        # Direct access to `series.iterations` is only necessary for random-access
+        # of iterations. By using `series.write_iterations()`, the openPMD-api
+        # will adhere to streaming semantics while writing. In particular, this
+        # means that only one iteration can be written at a time and
+        # an iteration can no longer be modified after closing it.
```
ax3l: Fix line len:
```suggestion
        # Direct access to `series.iterations` is only necessary for
        # random-access of iterations. By using `series.write_iterations()`,
        # the openPMD-api will adhere to streaming semantics while writing.
        # In particular, this means that only one iteration can be written at a
        # time and an iteration can no longer be modified after closing it.
```

examples/10_streaming_write.py
```diff
@@ -3,31 +3,92 @@
...
         sys.exit(0)
 
-    datatype = np.dtype("double")
-    length = 10
-    global_extent = [10]
-    dataset = io.Dataset(datatype, global_extent)
+    # create a series and specify some global metadata
+    # change the file extension to .json, .h5 or .bp for regular file writing
+    series = io.Series("simData.sst", io.Access_Type.create, config)
+    series.set_author("Franz")
```
ax3l: That's the syntax we recommend in the standard, so let's examplify it :)
```suggestion
    series.set_author("Franz Poeschel <f.poeschel@hzdr.de>")
```
People tend to copy-paste from examples, so let's do best practices :)

examples/10_streaming_write.py
```diff
@@ -3,31 +3,92 @@
...
+    series.set_software("python snippet")
```
ax3l: Maybe more concrete:
```suggestion
    series.set_software("openPMD-api-python-examples")
```

examples/10_streaming_write.py
```diff
@@ -3,31 +3,92 @@
...
+        # custom attribute
+        electronPositions.set_attribute("comment", "I'm a comment")
+
+        length = 10
+        local_data = np.arange(i * length, (i + 1) * length,
+                               dtype=np.dtype("double"))
         for dim in ["x", "y", "z"]:
             pos = electronPositions[dim]
-            pos.reset_dataset(dataset)
+            pos.reset_dataset(io.Dataset(np.dtype("double"), [length]))
```
ax3l: Maybe simpler:
```suggestion
            pos.reset_dataset(io.Dataset(local_data.dtype, [length]))
```

examples/10_streaming_write.py
```diff
@@ -3,31 +3,92 @@
...
 
-if __name__ == "__main__":
-    if 'adios2' not in io.variants or not io.variants['adios2']:
-        print('This example requires ADIOS2')
-        sys.exit(0)
-
-    series = io.Series("stream.sst", io.Access_Type.create)
+# pass-through for ADIOS2 engine parameters
+# https://adios2.readthedocs.io/en/latest/engines/engines.html
+config = """
```
ax3l: Let's replace this with the [import json](https://docs.python.org/3/library/json.html) objects because it's more readable and a standard lib in Python :)

franzpoeschel: I think, in that case we should make it possible to write something like:
```python
series = io.Series("asdf.bp", io.Access.create, {"adios2": {"engine": {"type": "bp4"}}}
```
That's currently not possible because the `Series` constructor accepts strings only, so a string conversion needs to happen inside Python.

ax3l: That says the same - we populate a dictionary and then pass the json string representation of it to `io.Series`:
```py
import json

config = {'adios2': {'engine': {}, 'dataset': {}}}
config['adios2']['engine'] = {'parameters': {'Threads': '4'}}
config['adios2']['dataset'] = {'operators': [{'type': 'bzip2'}]}

series = io.Series("stream.sst", io.Access_Type.create, json.dumps(config))
```
or
```py
import json

config = {
  'adios2': {
    'engine': {
      'parameters': { 'Threads': '4' }
    },
    'dataset': {
      'operators': [
        { 'type': 'bzip2' }
      ]
    }
  }
}

series = io.Series("stream.sst", io.Access_Type.create, json.dumps(config))
```

That way, we can benefit from Python dictionaries and their controlled parsing in the language itself :)

franzpoeschel: What I was trying to say is that it would just be a bit neater to avoid the `json.dumps()` and make it look like we understand the JSON value natively. But it might be too much hassle.

franzpoeschel: I've used your first suggestion now

ax3l: Cool, thx! I think that's good for the current PR :)

> What I was trying to say is that it would just be a bit neater to avoid the json.dumps() and make it look like we understand the JSON value natively. But it might be too much hassle.

We can also add a new overload in Python that understands dictionaries and do `json.dumps(config_dict)` internally :)

ax3l (approved):
Thank you! Splendid :+1:

PR #1144, opened 2021-11-08
Include prefix (directory) in write and checks for read parameter

jeanbez:
Some updates on the 8a/8b benchmarks:
- 8a (write) add an additional option to the configuration input file to specify a custom path to write the files (by default keep using `../samples/`
- 8b (read) check for valid file (show message if we do not find a file) and valid pattern (from those available in the documentation)

### DIFF
--- a/examples/8a_benchmark_write_parallel.cpp
+++ b/examples/8a_benchmark_write_parallel.cpp
@@ -389,7 +389,7 @@ void parse(TestInput& input, std::string line)
     return;
   }
 
-  if ( vec[0].compare("prefix") == 0 ) {
+  if ( vec[0].compare("fileLocation") == 0 ) {
     input.m_Prefix = vec[1];
     return;
   }
--- a/examples/8a_benchmark_write_parallel.cpp
+++ b/examples/8a_benchmark_write_parallel.cpp
@@ -339,6 +339,8 @@ public:
   unsigned long  m_XFactor = 0; // if not overwritten, use m_MPISize
   unsigned long  m_YFactor = 8;
   unsigned long  m_ZFactor = 8;
+
+  std::string m_Prefix = "../samples";
 }; // class TestInput
 
 
@@ -387,6 +389,11 @@ void parse(TestInput& input, std::string line)
     return;
   }
 
+  if ( vec[0].compare("prefix") == 0 ) {
+    input.m_Prefix = vec[1];
+    return;
+  }
+
   // now vec[1] is N-dim integers
   std::vector<unsigned long> numbers;
   std::istringstream tmp(vec[1]);
@@ -395,7 +402,7 @@ void parse(TestInput& input, std::string line)
 
   if ( (numbers.size() == 0) || ((numbers.size() - input.m_Dim) != 0) ) {
     if ( input.m_MPIRank == 0 )
-      std::cout<<vec[1]<<" Expecting "<<input.m_Dim<<" dimensions. But given input is"<<numbers.size()<<std::endl;
+      std::cout<<vec[1]<<" Expecting "<<input.m_Dim<<" dimensions. But given input is "<<numbers.size()<<std::endl;
     return;
   }
 
@@ -568,7 +575,7 @@ void AbstractPattern::run()
 
     { // file based
       std::ostringstream s;
-      s << "../samples/8a_parallel_"<<m_GlobalMesh.size()<<"D"<<balance<<"_%07T"<<m_Input.m_Backend;
+      s << m_Input.m_Prefix << "/8a_parallel_"<<m_GlobalMesh.size()<<"D"<<balance<<"_%07T"<<m_Input.m_Backend;
 
       std::string filename = s.str();
 
@@ -589,7 +596,7 @@ void AbstractPattern::run()
 #ifdef NEVER // runs into error for ADIOS. so temporarily disabled
     { // group based
       std::ostringstream s;
-      s << "../samples/8a_parallel_"<<m_GlobalMesh.size()<<"D"<<balance<<m_Input.m_Backend;
+      s << m_Input.m_Prefix << "/8a_parallel_"<<m_GlobalMesh.size()<<"D"<<balance<<m_Input.m_Backend;
       std::string filename = s.str();
 
       {
--- a/examples/8a_benchmark_write_parallel.cpp
+++ b/examples/8a_benchmark_write_parallel.cpp
@@ -340,6 +340,7 @@ public:
   unsigned long  m_YFactor = 8;
   unsigned long  m_ZFactor = 8;
 
+  //! prefix for the output directory
   std::string m_Prefix = "../samples";
 }; // class TestInput
 
--- a/examples/8b_benchmark_read_parallel.cpp
+++ b/examples/8b_benchmark_read_parallel.cpp
@@ -253,7 +253,6 @@ public:
 
       if ( 0 == m_MPIRank )
       {
-         std::cout << "  "<<series.iterationEncoding() << std::endl;
          std::cout << "  Num Iterations in " << filename << " : " << numIterations << std::endl << std::endl;
       }
 
@@ -744,9 +743,9 @@ main( int argc, char *argv[] )
     MPI_Comm_size( MPI_COMM_WORLD, &input.m_MPISize );
     MPI_Comm_rank( MPI_COMM_WORLD, &input.m_MPIRank );
 
-    if (argc < 2) {
+    if (argc < 3) {
       if (input.m_MPIRank == 0)
-          std::cout<<"Usage: "<<argv[0]<<" input_file_prefix"<<std::endl;
+          std::cout<<"Usage: "<<argv[0]<<" input_file_prefix pattern"<<std::endl;
       MPI_Finalize();
       return 0;
     }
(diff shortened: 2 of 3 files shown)


examples/8b_benchmark_read_parallel.cpp
```diff
@@ -764,19 +763,23 @@ main( int argc, char *argv[] )
...
-        if ( types[1] == 'y')
+        else if ( types[1] == 'y')
           input.m_Pattern = 65;
-        if ( types[1] == 'z')
+        else if ( types[1] == 'z')
           input.m_Pattern = 75;
+        else
+          std::cout << "For pattern 'f' you should provide the dimension as well: fx, fy, or fz" << std::endl;
       } else {
-          input.m_Pattern = atoi(argv[2]);
```
guj, line 779: Please put this line back in, and do not follow with an error message. 

A pattern that reaches here, as an integer,  is still valid.  Although it is not a common pattern that is advertised to public.

examples/8b_benchmark_read_parallel.cpp
```diff
@@ -764,19 +763,23 @@ main( int argc, char *argv[] )
...
+        std::cout << "Unknown pattern. Available options are: m, sx, sy, sz, fx, fy, fz" << std::endl;
```
guj: Please remove this line, see above.

src/Series.cpp
```diff
@@ -840,8 +840,10 @@ SeriesInterface::readFileBased( )
     {
         /* Frontend access type might change during SeriesInterface::read() to allow parameter modification.
          * Backend access type stays unchanged for the lifetime of a Series. */
-        if(IOHandler()->m_backendAccess == Access::READ_ONLY  )
+        if(IOHandler()->m_backendAccess == Access::READ_ONLY  ) {
+            std::cerr << "Unable to find the given file." << std::endl;
```
guj: There is exception already, why an error message? If error message is needed, should also provide file path?

examples/8b_benchmark_read_parallel.cpp
```diff
@@ -253,7 +253,6 @@ class TestInput
 
       if ( 0 == m_MPIRank )
       {
-         std::cout << "  "<<series.iterationEncoding() << std::endl;
```
guj: Why removing this line?

examples/8a_benchmark_write_parallel.cpp
```diff
@@ -387,6 +389,11 @@ void parse(TestInput& input, std::string line)
     return;
   }
 
+  if ( vec[0].compare("prefix") == 0 ) {
```
guj: First, would be good to call it something like "DefaultFileLocation", to spell out the purpose.

Second, if a user provided such a path,  shall we first check whether it is a valid one with right permissions, If not, ask the user to create it first.

examples/8b_benchmark_read_parallel.cpp
ax3l: Hm, I think let's document all feature sets though :D

examples/8a_benchmark_write_parallel.cpp
```diff
@@ -339,6 +339,8 @@ class TestInput
   unsigned long  m_XFactor = 0; // if not overwritten, use m_MPISize
   unsigned long  m_YFactor = 8;
   unsigned long  m_ZFactor = 8;
+
+  std::string m_Prefix = "../samples";
```
ax3l, line 344: Let's add a doxygen string here:
```suggestion
  //! prefix for the output directory
  std::string m_Prefix = "../samples";
```

examples/8a_benchmark_write_parallel.cpp
ax3l: For the second part: `openPMD::Series()` will create similar to `mkdir -p` directories that are passed and not yet existing.

Permission issues would be forwarded through it (can maybe be improved generally, if they should not propagate well).

examples/8b_benchmark_read_parallel.cpp
jeanbez: Oh, I did not know about those additional patterns. I thought that was something left there from deprecated versions. I've already included it back again! Do you have a list of all available patterns to include in the documentation? I believe that would be helpful. Also are there any checks for this, or do all integers mean something?

examples/8b_benchmark_read_parallel.cpp
jeanbez: Done!

src/Series.cpp
jeanbez: I've included the error message because the exception is not displayed for the user. Should it be displaying that message?

examples/8a_benchmark_write_parallel.cpp
jeanbez: I've renamed the variable to `fileLocation` to make it clearer. Thanks for the feedback! I leave it up to you to decide the expected behavior: (1) we check and ask the user to create if it is not there or (2) we transparently create the directory and continue the execution. What would be the best?

examples/8b_benchmark_read_parallel.cpp
jeanbez: My mistake! Thanks for reporting it

examples/8b_benchmark_read_parallel.cpp
guj: Thanks. 
There are a few other patterns that I am playing with,  will document as advanced options in a seperate pull request.

examples/8b_benchmark_read_parallel.cpp
guj: Thanks

src/Series.cpp
guj: Interesting. I tried 8b with an invalid file, it gives me the exception msg.

examples/8a_benchmark_write_parallel.cpp
guj: Yes, I know Series() creates recursively. 

I would like to give extra precaution when doing so for user provided path.   

Guess I will leave it to Axel to decide on.  @ax3l

src/Series.cpp
jeanbez: @guj if you do not provide any file path, does it still show the exception msg?

guj: Without file path, 8b can not run. It will ask for input.

examples/8a_benchmark_write_parallel.cpp
jeanbez: @ax3l what is your take on this?

jeanbez:
@ax3l I noticed a failed test in the documentation for ADIOS, which I did not update. Is that because of changes in the testing framework? Do I need to fix this here?

examples/8a_benchmark_write_parallel.cpp
ax3l: Let's leave the default behavior: user provides a path to write to, we will create it if it does not exist.
That's what we do everywhere in the usage, so it will be consistent.

ax3l:
> I noticed a failed test in the documentation for ADIOS, which I did not update. Is that because of changes in the testing framework? Do I need to fix this here?

Ah yes, probably from an update of Doxygen in CI.
Franz fixed this already in another PR, thus it will be gone now when you push again to the PR for final updates :)

examples/8b_benchmark_read_parallel.cpp
```diff
@@ -744,9 +744,9 @@ main( int argc, char *argv[] )
     MPI_Comm_size( MPI_COMM_WORLD, &input.m_MPISize );
     MPI_Comm_rank( MPI_COMM_WORLD, &input.m_MPIRank );
 
-    if (argc < 2) {
+    if (argc < 3) {
```
ax3l, line 751: Following the LGTM warning: should the line 759 below read `>= 4` now?
@jeanbez @guj

examples/8b_benchmark_read_parallel.cpp
```diff
@@ -744,9 +744,9 @@ main( int argc, char *argv[] )
...
       if (input.m_MPIRank == 0)
-          std::cout<<"Usage: "<<argv[0]<<" input_file_prefix"<<std::endl;
+          std::cout<<"Usage: "<<argv[0]<<" input_file_prefix pattern"<<std::endl;
```
ax3l, line 753: Please document the new syntax in the manual here: `docs/source/usage/benchmarks.rst` :)

jeanbez: It is already documented there, only the code did not print any message if no pattern was provided.

examples/8b_benchmark_read_parallel.cpp
jeanbez: Yep, actually, we do not need that if in line 759

examples/8b_benchmark_read_parallel.cpp
ax3l: I think I am looking for a line similar to
```
$ cat --help
Usage: cat [OPTION]... [FILE]...
Concatenate FILE(s) to standard output.
```
For instance:
```
Usage: 8a_benchmark_write_parallel <input-file> [pattern]
```
in case that `<input-file>` is mandatory and `[pattern]` is optional

ax3l: I see, this is the content, just written in a free format instead of using the unix usual conventions:
https://github.com/openPMD/openPMD-api/blob/dev/docs/source/usage/benchmarks.rst#reading

Already there, does not need to be improved in the current PR then (but we should do that eventually @guj )

ax3l (approved):
Thanks a lot! :)

examples/8b_benchmark_read_parallel.cpp
guj: The input-fie is mandatory and pattern is optional. When pattern is not specified,  the old default behavior is to run all patterns.   It is later changed  to not doing anything. 

It is ok if you both like  forcing the pattern to be specified.

guj (approved):

examples/8b_benchmark_read_parallel.cpp
ax3l: Oh we want to keep it optional - I was just saying that we need to improve the doc page to show a standard UNIX-like, easy to understand usage in it.

PR #1145, opened 2021-11-09
Wheels 0.14.3 ppc64le emu

ax3l:
PPC64le builds are currently emulated on GH Actions because Travis' ppc64le docker has network hangs.

Follow-up to #1136


ax3l:
Argh, emulation even breaks the 360min timeout length of GitHub actions.
Have to re-visit #1138

PR #1146, opened 2021-11-11
Read backend configuration from .toml files

franzpoeschel:
Approach for implementation: Read from .toml file, but convert to nlohmann::json internally

Config files are much more readable in TOML:
```toml
unused = "global parameter"

[hdf5]
unused = "hdf5 parameter please dont warn"

[adios2]
unused = "parameter"

[adios2.engine]
type = "bp3"
unused = "as well"

[adios2.engine.parameters]
BufferGrowthFactor = 2
Profile = "On"

# double brackets, because the operators are a list
[[adios2.dataset.operators]]
type = "blosc"

# parameters for the current list entry
[adios2.dataset.operators.parameters]
parameters.clevel = "1"
parameters.doshuffle = "BLOSC_BITSHUFFLE"
```

TODO:
- [x] parallel testing
- [x] merge #1043 and #1148 first
- [x] error messages don't yet distinguish between JSON and TOML
- [x] make TOML usable for `Dataset` config too (how to do that?)
- [x] ~~Make TOML an optional dependency?~~ discussed: compiler coverage here is good.
- [x] Separate toml11 thirdparty lib addition into separate PR #1148
- [x] Documentation thirdparty lib: as with nlohmann::json (news file, readme, ...) #1148
- [x] Update `CommonADIOS1IOHandlerImpl< ChildClass >::initJson()` after merging #1162 (there will be no merge conflict)
- Add a CI config that tests this again upstream toml11 -> covered in nightly install tests

PR #1148, opened 2021-11-18
Thirdparty library: toml11 by Toru Niina

franzpoeschel:
Needed for #1146 (toml).
If needed, I can also remove all commits but the first and keep those inside #1146 instead

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,10 @@ Changes to "0.14.0"
 Features
 """"""""
 
+- CMake
+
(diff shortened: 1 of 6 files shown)


ax3l:
Thanks a lot!

>  If needed, I can also remove all commits but the first and keep those inside #1146 instead

Applied this now in 06386e630c904e0620497134390c55196dc3f320 and rebased this PR :)

I fixed tiny formatting problems in the README, Changelog and CMake on the way.

ax3l (approved):

PR #1150, opened 2021-11-29
Parsing: Remove invalid records from data structures entirely

franzpoeschel:
Our parsing logic in `ParticleSpecies.cpp` catches runtime errors when parsing a record, and if such an error happens, ignores the record. (Needed for old ADIOS1 output from PIConGPU that contained some nonstandard records).
Until now, the half-parsed records were not deleted.
Our checks in `Series::~Series()` caught up on that and threw errors.
This PR makes sure to delete records whose parsing has been canceled.

### DIFF
--- a/include/openPMD/backend/Container.hpp
+++ b/include/openPMD/backend/Container.hpp
@@ -366,6 +366,17 @@ OPENPMD_protected:
             return m_originalContainer.at( std::forward< K >( k ) );
         }
 
+        /**
+         * Remove key from the list of accessed keys.
+         * If the key is not accessed after this again, it will be deleted along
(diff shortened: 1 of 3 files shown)


include/openPMD/backend/Container.hpp
```diff
@@ -366,6 +366,12 @@ class Container : public LegacyAttributable
             return m_originalContainer.at( std::forward< K >( k ) );
         }
 
+        template< typename K >
```
franzpoeschel, line 374: Memo to self: Add description

ax3l (approved):
Thanks a lot! :tada:

PR #1152, opened 2021-12-02
Don't apply compression operators multipe times in variable-based iteration encoding

franzpoeschel:
I just noticed this while reading the source file that we apply compression operators in every step anew


ax3l:
Thanks!

Shall I mark this for backport to 0.14.4, too?

ax3l (approved):

franzpoeschel:
> Shall I mark this for backport to 0.14.4, too?

I don't see a reason why not to, yeah

PR #1155, opened 2021-12-07
Automatically deactive span based Put API in ADIOS2 when operators are present

franzpoeschel:
Ref. [this issue](https://github.com/ornladios/ADIOS2/issues/2965)
Tldr: Span-based Put API can help save memory, but is incompatible with compression operators

This PR automatically deactivates the span-based Put API on a per-variable basis if the variable has operators defined.
Alternatively, a JSON option can be used to activate / deactivate it completely.

TODO:
- [x] Documentation
- [x] Testing
- [x] Merge #1043 first

### DIFF
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -94,6 +94,11 @@ Explanation of the single keys:
 
   * ``type`` supported ADIOS operator type, e.g. zfp, sz
   * ``parameters`` is an associative map of string parameters for the operator (e.g. compression levels)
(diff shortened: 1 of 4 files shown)


ax3l:
Oh, this cannot be backported to the `0.14.*` line if it depends on the JSON PR.

No worries though, we can follow up with `0.15.0` early next year.

williamfgc:
FYI, I'm putting [ADIOS2#2981](https://github.com/ornladios/ADIOS2/pull/2981), so an exception is thrown at the back end level.

ax3l:
Rebased since #1043 was merged :)

ax3l (approved):

PR #1162, opened 2022-01-07
Fix reading of dataset-specific configuration in ADIOS1

franzpoeschel:
Follow-up to #1043, I forgot one layer in the JSON schema inside `createDataset`. This fixes that and adds a small test.

### DIFF
--- a/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
+++ b/src/IO/ADIOS/CommonADIOS1IOHandler.cpp
@@ -552,19 +552,23 @@ CommonADIOS1IOHandlerImpl< ChildClass >::createDataset(Writable* writable,
         {
             json::TracingJSON options = json::parseOptions(
                 parameters.options, /* considerFiles = */ false );
-            auto maybeTransform = datasetTransform( options );
-            if( maybeTransform.has_value() )
+            if( options.json().contains( "adios1" ) )
             {
-                transform = maybeTransform.get();
-            }
+                options = options["adios1"];
+                auto maybeTransform = datasetTransform( options );
+                if( maybeTransform.has_value() )
+                {
+                    transform = maybeTransform.get();
+                }
 
-            auto shadow = options.invertShadow();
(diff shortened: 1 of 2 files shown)


src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -552,19 +552,23 @@ CommonADIOS1IOHandlerImpl< ChildClass >::createDataset(Writable* writable,
...
-            {
-                std::cerr << "Warning: parts of the JSON configuration for "
-                             "ADIOS1 dataset '"
-                          << name << "' remain unused:\n"
-                          << shadow << std::endl;
+                auto shadow = options.invertShadow();
+                if( shadow.size() > 0 )
+                {
+                    std::cerr << "Warning: parts of the JSON configuration for "
+                                "ADIOS1 dataset '"
```
ax3l, line 568:
```suggestion
                    std::cerr << "Warning: parts of the JSON configuration for "
                                 "ADIOS1 dataset '"
```

src/IO/ADIOS/CommonADIOS1IOHandler.cpp
```diff
@@ -1762,6 +1766,13 @@ void CommonADIOS1IOHandlerImpl< ChildClass >::initJson(
     {
         m_defaultTransform = std::move( maybeTransform.get() );
     }
+    auto shadow = config.invertShadow();
+    if( shadow.size() > 0 )
+    {
+        std::cerr << "Warning: parts of the JSON configuration for ADIOS1 "
+                        "remain unused:\n"
```
ax3l, line 1773:
```suggestion
        std::cerr << "Warning: parts of the JSON configuration for ADIOS1 "
                     "remain unused:\n"
```

ax3l (approved):
Thanks, LGTM :+1: 

Two suggestions for alignment/spaces, but not a big deal when we add `clang-format` soon.

PR #1164, opened 2022-01-11
C++17

ax3l:
Transition to require a C++17 capable compiler.

This allows us to drop the public dependency on `MPark.Variant` (kudos and thanks for the great time!).

### To Do

- [x] update CMake
- [x] update docs
- [x] remove internally shipped `MPark.Variant` version
- [x] refactor `Variant` and `Option` logic to only use `std::variant` as source
- [x] replace `Option` with `std::optional`: https://github.com/ax3l/openPMD-api/pull/4
- [x] remove `__cplusplus` switches that guard pre-C++17 code
- [x] replace SFINAE constructs with `constexpr if`: https://github.com/ax3l/openPMD-api/pull/3
- [x] replace nested namespaces with `namespace openPMD::detail::foo::bar` (this is better for clang-format)
  - done as part of the C++17 modernizations in `clang-tidy`:
    - `modernize-concat-nested-namespaces`
    - `modernize-use-nodiscard`


ax3l:
Clang 6 issues with `<variant>` from GCC's libstdc++:
- https://stackoverflow.com/a/46507150/2719194
- https://bugs.llvm.org/show_bug.cgi?id=31852
- https://bugs.llvm.org/show_bug.cgi?id=33222

.github/workflows/linux.yml
```diff
@@ -7,23 +7,23 @@ concurrency:
...
     - uses: actions/checkout@v2
     - name: Spack Cache
       uses: actions/cache@v2
-      with: {path: /opt/spack, key: clang5_nopy_nompi_h5 }
+      with: {path: /opt/spack, key: clang6_nopy_nompi_h5_libcpp }
     - name: Install
       run: |
         sudo apt-get update
-        sudo apt-get install clang-5.0
+        sudo apt-get install clang-6.0 libc++-7-dev
```
ax3l:
```suggestion
        sudo apt-get install clang-6.0 libc++-7-dev libc++abi-7-dev
```

.github/workflows/linux.yml
```diff
@@ -32,27 +32,28 @@ jobs:
...
     - uses: actions/checkout@v2
     - name: Spack Cache
       uses: actions/cache@v2
-      with: {path: /opt/spack, key: clang5_nopy_ompi_h5_ad1_ad2_bp3 }
+      with: {path: /opt/spack, key: clang6_nopy_ompi_h5_ad1_ad2_bp3_libcpp }
     - name: Install
       run: |
         sudo apt-get update
-        sudo apt-get install clang-5.0 gfortran libopenmpi-dev python3
+        sudo apt-get install clang-6.0 libc++-7-dev gfortran libopenmpi-dev python3
```
ax3l:
```suggestion
        sudo apt-get install clang-6.0 libc++-7-dev libc++abi-7-dev gfortran libopenmpi-dev python3
```

src/auxiliary/Filesystem.cpp
```diff
@@ -200,11 +197,11 @@ namespace
      * Only some of these are actually instanciated,
      * so suppress warnings for the others.
      */
-    OPENPMDAPI_UNUSED
+    [[maybe_unused]]
```
ax3l, line 200: Causes false-positive diagnostic of `-Wignored-attributes` in clang 6:
```
'maybe_unused' attribute only applies to variables, functions, methods, types, enumerations
```

Clang 7 and later (and all other compilers) do not warn here.

README.md
```diff
@@ -100,10 +100,9 @@ Our manual shows full [read & write examples](https://openpmd-api.readthedocs.io
 
 Required:
 * CMake 3.15.0+
-* C++14 capable compiler, e.g. g++ 5.0+, clang 5.0+, VS 2017+
+* C++17 capable compiler, e.g., g++ 7+, clang 6+, MSVC 19.15+, icpc 19+, icpx
```
ax3l: I wonder if we should advertise/support only clang 7+, primarily due to the incompatibilities of `<variant>` in earlier versions with GCC's libstdc++ (and to a minor degree: the small false warning in `Filesystem.cpp`)
```suggestion
* C++17 capable compiler, e.g., g++ 7+, clang 7+, MSVC 19.15+, icpc 19+, icpx
```

Ubuntu 18.04 (bionic/oldstable) ships clang 6 by default, but Ubuntu 20.04 (focal/stable) is already at clang 10.

We could also leave the `try_variant` test in, so that we can fail old/buggy compilers already in the configure step.

docs/source/dev/dependencies.rst
```diff
@@ -10,14 +10,13 @@ Required
 --------
 
 * CMake 3.15.0+
-* C++14 capable compiler, e.g. g++ 5.0+, clang 5.0+, VS 2017+
+* C++17 capable compiler, e.g., g++ 7+, clang 6+, MSVC 19.15+, icpc 19+, icpx
```
ax3l: Potentially useful to limit: see arguments above
```suggestion
* C++17 capable compiler, e.g., g++ 7+, clang 7+, MSVC 19.15+, icpc 19+, icpx
```

src/auxiliary/Filesystem.cpp
ax3l: Also needs `-Wno-unused-const-variable` now :D
```
src/auxiliary/Filesystem.cpp:201:47: error: unused variable 'value' [-Werror,-Wunused-const-variable]
    MPI_Datatype const MPI_Types< unsigned >::value = MPI_UNSIGNED;
                                              ^
```

PR #1166, opened 2022-01-12
ADIOS2 fixes: Incompatibilites with current ADIOS2 master branch

### LINKED ISSUE #1163: Gracefully handle unsupported compression in current ADIOS2 master
**Describe the bug**
ADIOS2 generally throws an exception when adding an unsupported compression (e.g. adding Blosc compression when Blosc has not been compiled).
So far, we catch this exception and turn it into a warning. 

Something changed in the way ADIOS2 throws exceptions for these, so the openPMD-api currently does not catch them any more.

**Expected behavior**

Two options (we should discuss this before fixing anything):

1. Stick with the old behavior, i.e. fix the catching of the exception. This makes some CI workflows easier (which I think was the original reason I overrode the behavior of ADIOS2)
2. Change our behavior and adopt ADIOS2's behavior: Just forward the exception. This would mean that we have to fix some tests.

Opinion? @ax3l

franzpoeschel:
Fix #1163:

1. ADIOS2 now sometimes throws different exception types, so we must catch different types
2. ADIOS2 now throws an error if using the span-based Put() API in combination with operators, so we must adapt our tests

Notes on the second point:
* Destructing should always work, even if errors are thrown. We cannot do much against segfaults in our flushing routines during destruction, but we should check if this can be solved without leaking the memory.
* Should we add an API call `Series::invalidate()` to indicate that this `Series` object should perform no further operations (especially: only free memory during destruction, do nothing else, no flushing)?

### DIFF
--- a/include/openPMD/IO/AbstractIOHandlerImpl.hpp
+++ b/include/openPMD/IO/AbstractIOHandlerImpl.hpp
@@ -25,6 +25,7 @@
 #include "openPMD/auxiliary/DerefDynamicCast.hpp"
 
 #include <future>
+#include <iostream>
 
 
 namespace openPMD
@@ -123,8 +124,14 @@ public:
                         availableChunks(i.writable, deref_dynamic_cast< Parameter< O::AVAILABLE_CHUNKS > >(i.parameter.get()));
                         break;
                 }
-            } catch (unsupported_data_error&)
+            } catch (...)
             {
+                std::cerr
+                    << "[AbstractIOHandlerImpl] IO Task "
+                    << internal::operationAsString( i.operation )
+                    << " failed with exception. Removing task"
+                    << " from IO queue and passing on the exception."
+                    << std::endl;
                 (*m_handler).m_work.pop();
                 throw;
             }
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -76,6 +76,15 @@ OPENPMDAPI_EXPORT_ENUM_CLASS(Operation)
     AVAILABLE_CHUNKS //!< Query chunks that can be loaded in a dataset
 }; // note: if you change the enum members here, please update docs/source/dev/design.rst
 
+namespace internal
+{
+    /*
+     * The returned strings are compile-time constants, so no worries about
+     * pointer validity.
+     */
+    std::string operationAsString( Operation );
+}
+
 struct OPENPMDAPI_EXPORT AbstractParameter
 {
     virtual ~AbstractParameter() = default;
--- a/src/IO/IOTask.cpp
+++ b/src/IO/IOTask.cpp
@@ -73,4 +73,87 @@ void Parameter< Operation::CREATE_DATASET >::warnUnusedParameters<
         }
     }
 }
+
+namespace internal
+{
+    std::string
+    operationAsString( Operation op )
+    {
+        switch( op )
+        {
+        case Operation::CREATE_FILE:
+            return "CREATE_FILE";
+            break;
+        case Operation::OPEN_FILE:
+            return "OPEN_FILE";
+            break;
+        case Operation::CLOSE_FILE:
+            return "CLOSE_FILE";
+            break;
+        case Operation::DELETE_FILE:
+            return "DELETE_FILE";
+            break;
+        case Operation::CREATE_PATH:
+            return "CREATE_PATH";
+            break;
+        case Operation::CLOSE_PATH:
+            return "CLOSE_PATH";
+            break;
+        case Operation::OPEN_PATH:
+            return "OPEN_PATH";
+            break;
+        case Operation::DELETE_PATH:
+            return "DELETE_PATH";
+            break;
+        case Operation::LIST_PATHS:
+            return "LIST_PATHS";
+            break;
+        case Operation::CREATE_DATASET:
+            return "CREATE_DATASET";
+            break;
+        case Operation::EXTEND_DATASET:
+            return "EXTEND_DATASET";
+            break;
+        case Operation::OPEN_DATASET:
+            return "OPEN_DATASET";
+            break;
+        case Operation::DELETE_DATASET:
+            return "DELETE_DATASET";
+            break;
+        case Operation::WRITE_DATASET:
+            return "WRITE_DATASET";
+            break;
+        case Operation::READ_DATASET:
+            return "READ_DATASET";
+            break;
+        case Operation::LIST_DATASETS:
+            return "LIST_DATASETS";
+            break;
+        case Operation::GET_BUFFER_VIEW:
+            return "GET_BUFFER_VIEW";
+            break;
+        case Operation::DELETE_ATT:
+            return "DELETE_ATT";
+            break;
+        case Operation::WRITE_ATT:
+            return "WRITE_ATT";
+            break;
+        case Operation::READ_ATT:
+            return "READ_ATT";
+            break;
+        case Operation::LIST_ATTS:
+            return "LIST_ATTS";
+            break;
+        case Operation::ADVANCE:
+            return "ADVANCE";
+            break;
+        case Operation::AVAILABLE_CHUNKS:
+            return "AVAILABLE_CHUNKS";
+            break;
+        default:
+            return "unknown";
+            break;
+        }
+    }
+}
 } // openPMD
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -4409,7 +4409,6 @@ TEST_CASE( "automatically_deactivate_span", "[serial][adios2]" )
     }
 
     // enable span-based API indiscriminately
-    try
     {
         std::string enable = R"END(
 {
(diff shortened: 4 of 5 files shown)


test/SerialIOTest.cpp
```diff
@@ -4427,17 +4427,23 @@ TEST_CASE( "automatically_deactivate_span", "[serial][adios2]" )
...
   }
 })END";
-        Series write( "../samples/span_based.bp", Access::CREATE, enable );
-        auto E_uncompressed = write.iterations[ 0 ].meshes[ "E" ][ "x" ];
-        auto E_compressed = write.iterations[ 0 ].meshes[ "E" ][ "y" ];
+        /*
+         * Use a pointer so we can leak the memory later.
+         * We will leave write in an invalid state and the destructor results
+         * in a segfault.
+         * This is probably a bug in itself and should be fixed.
```
ax3l: Is this a bug in our lib or in ADIOS2?

test/SerialIOTest.cpp
```diff
@@ -4427,17 +4427,23 @@ TEST_CASE( "automatically_deactivate_span", "[serial][adios2]" )
...
 {
   "adios2": {
     "use_span_based_put": true
   }
 })END";
-        Series write( "../samples/span_based.bp", Access::CREATE, enable );
-        auto E_uncompressed = write.iterations[ 0 ].meshes[ "E" ][ "x" ];
-        auto E_compressed = write.iterations[ 0 ].meshes[ "E" ][ "y" ];
+        /*
+         * Use a pointer so we can leak the memory later.
```
ax3l: From the change I cannot see this immediately: Why is this change now necessary and wasn't before?
Due to the second point in the PR description
> ADIOS2 now throws an error if using the span-based Put() API in combination with operators, so we must adapt our tests

cc @williamfgc?
```suggestion
        /* Use a pointer so we can leak the memory later.
```

> Destructing should always work, even if errors are thrown.

Agreed.

> We cannot do much against segfaults in our flushing routines during destruction, but we should check if this can be solved without leaking the memory.

Can we potentially check which parts segfault in flush and guard them with checks? This should be related to the new operator checks, no?

> Should we add an API call `Series::invalidate()` to indicate that this `Series` object should perform no further operations (especially: only free memory during destruction, do nothing else, no flushing)?

Possible for the state of a `Series`, but that means we need to check on every member function access if the class is in valid state (and so for every other object down the dependency tree)...

test/SerialIOTest.cpp
franzpoeschel: Probably on our side, I'm currently trying to figure it out.

test/SerialIOTest.cpp
franzpoeschel: > Possible for the state of a Series, but that means we need to check on every member function access if the class is in valid state (and so for every other object down the dependency tree)...

Not necessarily? I would propose the following semantics for `Series::invalidate()`:
* Any access to an invalidated Series is undefined behavior
* Destruction must work

To be fair, these (repeatedly occuring) issues make me increasingly wary of putting too much logic into destructors. Sometimes, you just want to cleanly get rid of allocated memory and `invalidate()` would essentially indicate this.

franzpoeschel: Also, in case that you don't want to introduce this sort of undefined behavior, we can put a check on `Series::get()` which nearly every method in the Series API calls at some time.

But in my opinion, if calling something like `Series::invalidate()`, you are explicitly asking for that kind of situation, so I don't see a big problem with undefined behavior here

test/SerialIOTest.cpp
franzpoeschel: Fixed it now, see the commit message for details

test/SerialIOTest.cpp
franzpoeschel: > From the change I cannot see this immediately: Why is this change now necessary and wasn't before?
> Due to the second point in the PR description

Ah, I missed this part. See [this PR](https://github.com/ornladios/ADIOS2/pull/2981) and the [corresponding issue](https://github.com/ornladios/ADIOS2/issues/2965).
Previously, doing this just silently created invalid datasets and it took a lot of time to find out why they were invalid, so there is now a clear error message for that situation in ADIOS2.

williamfgc: @ax3l @franzpoeschel let us know if there is more of these. As much as adios2 throws exceptions, they is always a missing case. Thanks!

test/SerialIOTest.cpp
ax3l: Ah perfect, thanks! Now we should remove this comment block again :)

include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -123,7 +123,7 @@ class AbstractIOHandlerImpl
                         availableChunks(i.writable, deref_dynamic_cast< Parameter< O::AVAILABLE_CHUNKS > >(i.parameter.get()));
                         break;
                 }
-            } catch (unsupported_data_error&)
+            } catch (...)
             {
                 (*m_handler).m_work.pop();
```
ax3l, line 135: Should we write a warning here? I think this catch should not happen in a well-defined program and could potentially complicate debugging later on if we do this silently.

test/SerialIOTest.cpp
```diff
@@ -4469,24 +4474,28 @@ TEST_CASE( "automatically_deactivate_span", "[serial][adios2]" )
...
-    {
-        /*
-         * Using the span-based API in combination with compression is
-         * unsupported in ADIOS2.
-         * Currently, this is silently ignored, but in future there might be
-         * an error of some sort.
-         */
-        std::cerr << "Ignoring expected error: " << e.what() << std::endl;
+
+        // Leak write on purpose
```
```suggestion
```

test/SerialIOTest.cpp
ax3l: Pushed :)

include/openPMD/IO/AbstractIOHandlerImpl.hpp
franzpoeschel: Yep, probably a good idea and useful as a starting point for debugging. 
(Also, the current implementation has led to some quite painful debugging experiences for me, because the failing task is tried *again* upon destruction, leading to very weird error behavior. Making this change is the right call imo)

franzpoeschel: Another note: This code is not used by the ADIOS1 backend. Should we add the same logic there?

franzpoeschel: Pushed an update, current output of the test:
```
Filters: automatically_deactivate_span
[AbstractIOHandlerImpl] IO Task GET_BUFFER_VIEW failed with exception. Removing task from IO queue and passing on the exception.
Ignoring expected error: ERROR: span currently not supported with operations, remove calls to Variable<T>::AddOperation or use Variable<T>::RemoveOperations, in call to Variable<T>::Span Engine::Put
```

Note that this only removes the failing task from the IO queue but no others. We don't really do any error recovery at the moment, we already talked about improving that aspect in the future. When we do this, it will probably show which implementation makes more sense (removing only the failing task or clearing the queue).

include/openPMD/IO/IOTask.hpp
```diff
@@ -76,6 +76,15 @@ OPENPMDAPI_EXPORT_ENUM_CLASS(Operation)
     AVAILABLE_CHUNKS //!< Query chunks that can be loaded in a dataset
 }; // note: if you change the enum members here, please update docs/source/dev/design.rst
 
+namespace internal
+{
+    /*
+     * The returned strings are compile-time constants, so no worries about
+     * pointer validity.
+     */
+    char const * operationAsString( Operation );
```
ax3l: Let's use C++ types :)
```suggestion
    std::string operationAsString( Operation );
```

(`#include <string>` already part of this file)

src/IO/IOTask.cpp
```diff
@@ -62,4 +62,84 @@ void Parameter< Operation::CREATE_DATASET >::warnUnusedParameters<
         std::cerr << warningMessage << shadow.dump() << std::endl;
     }
 }
+
+namespace internal
+{
+    char const * operationAsString( Operation op )
```
ax3l:
```suggestion
    std::string
    operationAsString( Operation op )
```

src/IO/IOTask.cpp
```diff
@@ -62,4 +62,84 @@ void Parameter< Operation::CREATE_DATASET >::warnUnusedParameters<
...
+            return "LIST_ATTS";
+            break;
+        case Operation::ADVANCE:
+            return "ADVANCE";
+            break;
+        case Operation::AVAILABLE_CHUNKS:
+            return "AVAILABLE_CHUNKS";
+            break;
+        }
+        throw std::runtime_error( "Unreachable!" );
```
ax3l: Avoid compiler warnings:
```suggestion
        case default:
            return "unknown";
            break;
        }
```

include/openPMD/IO/AbstractIOHandlerImpl.hpp
ax3l: > Another note: This code is not used by the ADIOS1 backend. Should we add the same logic there?

Yes, good idea. Do you like to push that in a separate PR? :)

ax3l (approved):
Thanks a lot! :slightly_smiling_face:

ax3l:
Squashed for easier back-porting.

ax3l (approved):

include/openPMD/IO/AbstractIOHandlerImpl.hpp
franzpoeschel: Done :) https://github.com/openPMD/openPMD-api/pull/1179

PR #1174, opened 2022-01-15
[WIP] Python: Apply "Black" Styling

ax3l:
Auto-format Python code according to [black](https://black.readthedocs.io).

Checks docs, too.
This might not be the best idea, since some files like "first read"/"first write" use carefully crafted, short lines in code examples, so they can go side-by-side with C++ code on many screens.

- [ ] Rebase after #1175
- [ ] Add flake8 configs: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length


NEWS.rst
```diff
@@ -163,7 +163,13 @@ The new order allows to make use of defaults in many cases in order reduce compl
    electrons["position"]["x"].reset_dataset(d)
 
    # old code
-   electrons["position"]["x"].store_chunk([0, ], particlePos_x.shape, particlePos_x)
+   electrons["position"]["x"].store_chunk(
+       [
+           0,
```
ax3l, line 168:
```suggestion
           0
```

franzpoeschel:
> This might not be the best idea, since some files like "first read"/"first write" use carefully crafted, short lines in code examples, so they can go side-by-side with C++ code on many screens.

Same issue then with the C++ reformatting? #1032
We could exempt the examples from reformatting in both PRs? Or even just specific examples.

ax3l:
At the moment, I am just concerned about:
> Checks docs, too.

and I think we will serve these two files with an exception :) (have to check how to config that ^^)

I think the `clang-format` one does not reformat inline code in docs that is not included from example files elsewhere.

PR #1179, opened 2022-01-20
ADIOS1: Remove task from IO queue if it fails with exception

franzpoeschel:
Follow-up to #1166, see [this comment](https://github.com/openPMD/openPMD-api/pull/1166#discussion_r788045146)

We could theoretically put it into 0.14.4, it's the same change we already did for the other backends. It doesn't really fix any bugs, but it makes errors a lot easier to debug if they occur.


ax3l (approved):
Thank you, splendid! :)

ax3l:
@franzpoeschel do we need to introduce the same for the JSON and HDF5 backend?

franzpoeschel:
@ax3l No, these use `AbstractIOHandlerImpl::flush()` which is already updated by #1166. ADIOS1 overrides that function which I had forgotten to take care of in #1166.

ax3l:
OK perfect, and that patch is already in 0.14.4 :tada:

PR #1183, opened 2022-01-25
Avoid SFINAE in Attribute conversion

franzpoeschel:
`Attribute.hpp` is included near everywhere, including user code. 
For attribute datatype conversion we can now use `if constexpr` and avoid SFINAE constructs.


ax3l (approved):
Yay, better compile time & less code w/o SFINAE! :tada:

PR #1191, opened 2022-02-01
Allow setting operations in ADIOS2 read API

franzpoeschel:
We don't currently have an API for setting *dataset-specific* JSON configurations in the read API, so this addition is only for the global JSON/TOML config passed via the `Series` constructor.

Usage example:
```cpp
std::string readConfig = R"END(
{
  "adios2": {
    "engine": {
      "type": "bp4"
    },
    "dataset": {
      "operators": [
        {
          "type": "blosc",
          "parameters": {
            "nthreads": 8
          }
        }
      ]
    }
  }
}
)END";
Series read( "mydataset.bp", Access::READ_ONLY, readConfig );
```

TODO:
- [x] Before merging this, can you check that this does what you need? @guj

### DIFF
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -35,6 +35,7 @@ Parameters that are directly passed through to an external library and not inter
 
 The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
 This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
+Dataset-specific configurations are (currently) only available during dataset creation, but not when reading datasets.
 
 A JSON/TOML configuration may either be specified as an inline string that can be parsed as a JSON/TOML object, or  alternatively as a path to a JSON/TOML-formatted text file (only in the constructor of ``openPMD::Series``):
 
@@ -128,6 +129,29 @@ Explanation of the single keys:
   The openPMD-api will automatically use a fallback implementation for the span-based Put() API if any operator is added to a dataset.
   This workaround is enabled on a per-dataset level.
   The workaround can be completely deactivated by specifying ``{"adios2": {"use_span_based_put": true}}`` or it can alternatively be activated indiscriminately for all datasets by specifying ``{"adios2": {"use_span_based_put": false}}``.
+
+Operations specified inside ``adios2.dataset.operators`` will be applied to ADIOS2 datasets in writing as well as in reading.
+Beginning with ADIOS2 2.8.0, this can be used to specify decompressor settings:
+
+.. code-block:: json
+
+  {
+    "adios2": {
+      "dataset": {
+        "operators": [
+          {
+            "type": "blosc",
+            "parameters": {
+              "nthreads": 2
+            }
+          }
+        ]
+      }
+    }
+  }
+
+In older ADIOS2 versions, this specification will be without effect in read mode.
+Dataset-specific configurations are (currently) only possible when creating datasets, not when reading.
 
 Any setting specified under ``adios2.dataset`` is applicable globally as well as on a per-dataset level.
 Any setting under ``adios2.engine`` is applicable globally only.
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -3592,6 +3592,16 @@ this = "should not warn"
     "engine": {
       "type": "bp3",
       "unused": "parameter"
+    },
+    "dataset": {
+      "operators": [
+        {
+          "type": "blosc",
+          "parameters": {
+            "nthreads": 8
+          }
+        }
+      ]
     }
   }
 }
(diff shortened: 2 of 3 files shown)


franzpoeschel:
If dataset-specific configurations are needed for this, I'd suggest mainlining [this PR](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3513) back into openPMD instead of introducing new API calls.

guj:
> * [ ]  Before merging this, can you check that this does what you need? @guj

Thanks Franz. The json file does pass the info along and adios receives the parameter when decompresses.

ax3l:
> If dataset-specific configurations are needed for this, I'd suggest mainlining this PR back into openPMD instead of introducing new API calls.

Hi Franz, yes alos dataset-specific configs as in write would be good.

What do you mean with _mainlining_ that PR?  :) It's already merged and thus part of mainline.

test/SerialIOTest.cpp
```diff
@@ -3592,6 +3592,16 @@ this = "should not warn"
     "engine": {
       "type": "bp3",
       "unused": "parameter"
+    },
+    "dataset": {
+      "operators": [
+        {
+          "type": "blosc",
+          "parameters": {
+            "nthreads": 8
```
```suggestion
            "nthreads": 2
```

test/SerialIOTest.cpp
```diff
@@ -3602,6 +3612,16 @@ this = "should not warn"
     "engine": {
       "type": "bp4",
       "unused": "parameter"
+    },
+    "dataset": {
+      "operators": [
+        {
+          "type": "blosc",
+          "parameters": {
+            "nthreads": 8
```
```suggestion
            "nthreads": 2
```

ax3l (commented):
Let's use only 2 thread, CI resources are limited to 2 virtual cores.

ax3l:
We need to document this feature as well in user-facing docs / examples and note that this only works with ADIOS 2.8.0+, please.

franzpoeschel:
> > If dataset-specific configurations are needed for this, I'd suggest mainlining this PR back into openPMD instead of introducing new API calls.
> 
> Hi Franz, yes alos dataset-specific configs as in write would be good.
> 
> What do you mean with _mainlining_ that PR? :) It's already merged and thus part of mainline.

That PR is part of the PIConGPU mainline, but not of the openPMD-api mainline. It would give us dataset-specific JSON configurations without needing to use any API calls (which for the read-side API we would first need to add).

> We need to document this feature as well in user-facing docs / examples and note that this only works with ADIOS 2.8.0+, please.

Will do

franzpoeschel:
> Hi Franz, yes alos dataset-specific configs as in write would be good.

I have had a look at how we could implement this at the current time and in my opinion we should not do it (yet).

In writing, the workflow for dataset-specific configurations looks like this:
```cpp
auto E_x = iteration.meshes["E"]["x"];
Dataset ds{...};
ds.options = R"({"my": {"json": "config"}})";
E_x.resetDataset(ds);
// E_x has not been created in the backend yet
// it will be created upon flushing,
// taking everything specified so far into consideration
```

In reading however:
```cpp
auto E_x = iteration.meshes["E"]["x"];
// E_x already exists in the backend, has already been parsed and opened
```

Letting users specify a JSON config after this line means that we have to keep track of JSON configs after already opening a dataset. This would be inconsistent, since we don't keep track of JSON configs after creating a dataset either. Our current logic relies on users specifying configs ahead of time.

The problem is that in writing, the dataset is created only after the user had the chance to specify everything needed, while in reading the dataset is opened before the user gets the chance to specify anything.

If we really want to do this, I think the best solution would be to add an explicit call `RecordComponent::setConfig()` that in reading mode must be called between opening the dataset and the first time reading from it. Runtime checks are needed to (1) check that it is not called at the wrong time, (2) check that `resetDataset()` is not used at all in read mode, (3) check that `m_dataset` is not edited manually. We should think about deprecating direct field access of `m_dataset` and replacing it with `RecordComponent::dataset() const`.
We should be careful to keep `RecordComponent::setConfig()` somewhat consistent with write mode.

Pattern-based configurations as in PIConGPU would allow us to have everything specified truly ahead of time:
```cpp
std::string config = R"(
{
  "adios2": {
    "dataset": [
      {
        "pattern": "/meshes/E/.*",
        "cfg": {
          "my": {
            "operators": "here"
          }
        }
      },
      {
        "cfg": {
          "default": "config"
        }
      }
    ]
  }
}
)";
Series read("dataset.bp", Access::READ_ONLY, config);
```

Once we have added these pattern-based configurations, we could think about deprecating `Dataset::config` again to avoid having too many options overriding each other.

franzpoeschel:
I've updated the documentation.

docs/source/details/backendconfig.rst
```diff
@@ -35,6 +35,7 @@ Parameters that are directly passed through to an external library and not inter
 
 The configuration string may refer to the complete ``openPMD::Series`` or may additionally be specified per ``openPMD::Dataset``, passed in the respective constructors.
 This reflects the fact that certain backend-specific parameters may refer to the whole Series (such as storage engines and their parameters) and others refer to actual datasets (such as compression).
+Dataset-specific configurations are (currently) only available during dataset creation, but not when reading datasets.
```
franzpoeschel, line 38: We should try to solve this until the next release, so this note is temporary.
Solving this exceeds the scope of this PR as there are some design considerations involved.

ax3l (approved):
Thanks a lot! :rocket: :sparkles:

PR #1194, opened 2022-02-07
clang-tidy: performance-for-range-copy

ax3l:
Add a new check category to `clang-tidy`: [performance-for-range-copy](https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html)

Inspired by https://github.com/ECP-WarpX/WarpX/pull/2833

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,5 +2,5 @@
 # FIXME: all performance-* reports
 # FIXME: all cert-* reports
 # FIXME: all bugprone-* reports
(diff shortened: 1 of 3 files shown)


ax3l:
Hm, this converts things like:
```diff
-        for( auto iteration : series.readIterations() )
+        for( const auto& iteration : series.readIterations() )
```
in tests which is technically ok to avoid, but we want to ensure that such copies work.

Maybe adding some [ignore statements](https://clang.llvm.org/extra/clang-tidy/#suppressing-undesired-diagnostics).
On the other hand, it looks like there are still enough locations where we test Iterations by-copy...?

franzpoeschel (approved):

Issue #1195, opened 2022-02-07
ADIOS 1 interfering with WarpX mpi io


RTSandberg:
If I run WarpX locally on my machine with MPI enabled, then I get a segfault that is traced to the presence of ADIOS1 associated with an openpmd_api installation

**To Reproduce**
```console
# fails with:
cmake -S . -B build -DWarpX_DIMS=RZ -DCMAKE_BUILD_TYPE=Debug
cmake --build build
cd build/bin
./warpx ../../Examples/Physics_applications/laser_acceleration/inputs_rz amrex.throw_exception = 1 amrex.signal_handling = 0
```
Configuration output:
```
(warpx-dev) ryansand@m-krasny05 WarpX % cmake -S . -B build -DWarpX_DIMS=RZ -DCMAKE_BUILD_TYPE=Debug
-- Found CCache: /opt/anaconda3/envs/warpx-dev/bin/ccache
-- Downloading AMReX ...
-- AMReX repository: https://github.com/AMReX-Codes/amrex.git (22.02)
-- CMake version: 3.21.3
-- AMReX installation directory: /usr/local
-- Build type set by user to 'Debug'.
-- Building AMReX with AMReX_SPACEDIM = 2
-- Configuring AMReX with the following options enabled: 
--    AMReX_PRECISION = DOUBLE
--    AMReX_MPI
--    AMReX_MPI_THREAD_MULTIPLE
--    AMReX_OMP
--    AMReX_LINEAR_SOLVERS
--    AMReX_PARTICLES
--    AMReX_PARTICLES_PRECISION = DOUBLE
--    AMReX_TINY_PROFILE
-- Found MPI: TRUE (found version "3.1") found components: C CXX 
-- AMReX configuration summary: 
--    Build type               = Debug
--    Install directory        = /usr/local
--    C++ compiler             = /opt/anaconda3/envs/warpx-dev/bin/x86_64-apple-darwin13.4.0-clang++
--    C++ defines              = 
--    C++ flags                = -g -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -std=c++14 -fmessage-length=0 -isystem /opt/anaconda3/envs/warpx-dev/include -fopenmp=libomp
[...]

PR #1167, opened 2022-01-12
ADIOS1: Fix Serial Builds, CI: Clang 10->12

ax3l:
Update the Clang 10 build to Clang 12 in CI: commit 1 & 2.

Commit 3: ADIOS1 serial bugfixes: this is a regression of #1101 since around `CommonADIOS1IOHandler.cpp`

PR #1196, opened 2022-02-07
Fix: macOS ADIOS1 w/ cctools ld64 w/o XCode

ax3l:
Some compiler toolchains on macOS do not use the XCode tools. For instance, conda-forge uses vanilla clang to build packages combined with a cctools `ld64` (similar to the macOS system linker) but without XCode.

In that case, the symbol hiding instructions we used on macOS were not effective, leading to errors like #1195.

This also fixes regressions from #1101, similar to #1167.

The current patch checks if the toolchain understands GCC/LLVM/ld64-ish linker flags on macOS and if it does, it adds them.

### Attention

- [ ] `llvm-strip` < 14 is also broken, which is used in Conda-Forge `compilers` - `CMAKE_STRIP` should use `cctools`' `strip` command (needs to be set manually): https://github.com/AMReX-Codes/pyamrex/issues/35

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -459,15 +459,19 @@ set(IO_SOURCE
 set(IO_ADIOS1_SEQUENTIAL_SOURCE
         src/Error.cpp
         src/auxiliary/Filesystem.cpp
+        src/auxiliary/JSON.cpp
         src/ChunkInfo.cpp
         src/IO/ADIOS/CommonADIOS1IOHandler.cpp
-        src/IO/ADIOS/ADIOS1IOHandler.cpp)
+        src/IO/ADIOS/ADIOS1IOHandler.cpp
+        src/IO/IOTask.cpp)
 set(IO_ADIOS1_SOURCE
         src/Error.cpp
         src/auxiliary/Filesystem.cpp
+        src/auxiliary/JSON.cpp
         src/ChunkInfo.cpp
         src/IO/ADIOS/CommonADIOS1IOHandler.cpp
-        src/IO/ADIOS/ParallelADIOS1IOHandler.cpp)
+        src/IO/ADIOS/ParallelADIOS1IOHandler.cpp
+        src/IO/IOTask.cpp)
 
 # library
 if(openPMD_BUILD_SHARED_LIBS)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -563,6 +563,11 @@ if(openPMD_HAVE_ADIOS1)
         set_target_properties(openPMD.ADIOS1.Serial PROPERTIES
             LINK_FLAGS "-Wl,--exclude-libs,ALL")
     elseif("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+        check_linker_flag(CXX "-Wl,--exclude-libs,ALL" DARWIN_WO_XCODE)
```
ax3l: This might be a CMake 3.18+ feature:
- https://gitlab.kitware.com/cmake/cmake/-/merge_requests/4777
- https://cmake.org/cmake/help/latest/module/CheckLinkerFlag.html

Thus:
- [ ] need to bump CMake dependency from 3.15+ to 3.18+

ax3l: Also: should use `LINKER:`:
Expands to `-Wl,--...` on GCC and to `-Xlinker --...` on Clang.

CMakeLists.txt
```diff
@@ -459,6 +459,7 @@ set(IO_SOURCE
 set(IO_ADIOS1_SEQUENTIAL_SOURCE
         src/Error.cpp
         src/auxiliary/Filesystem.cpp
+        src/auxiliary/JSON.cpp
```
ax3l, line 462: Needed because `TracingJSON` is now used in `CommonADIOS1IOHandler.cpp`

CMakeLists.txt
```diff
@@ -459,15 +459,19 @@ set(IO_SOURCE
...
         src/ChunkInfo.cpp
         src/IO/ADIOS/CommonADIOS1IOHandler.cpp
-        src/IO/ADIOS/ADIOS1IOHandler.cpp)
+        src/IO/ADIOS/ADIOS1IOHandler.cpp
+        src/IO/IOTask.cpp)
```
ax3l, line 466: Needed because we use `openPMD::internal::operationAsString` in `CommonADIOS1IOHandler.cpp`

PR #1201, opened 2022-02-15
[WIP] More Emscripten/pyiode Python

ax3l:
Follow-up to #1199

More potential updates related to https://github.com/pyodide/pyodide/pull/2169

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "setuptools>=42",
+    "setuptools>=57.4",
     "wheel",
     "cmake>=3.15.0,<4.0.0",
     "pybind11>=2.6.2,<3.0.0"
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,18 @@ class CMakeBuild(build_ext):
         for ext in self.extensions:
             self.build_extension(ext)
 
+    def get_ext_filename(self, ext_name):
+        r"""Convert the name of an extension (eg. "foo.bar") into the name
+        of the file from which it will be loaded (eg. "foo/bar.so", or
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-    "setuptools>=42",
+    "setuptools>=57.4",
```
ax3l, line 3: Introduces the env var [SETUPTOOLS_EXT_SUFFIX](https://github.com/pypa/setuptools/blob/v60.9.0/setuptools/command/build_ext.py#L106-L112) ([ref](https://github.com/pypa/setuptools/issues/2722))

setup.py
```diff
@@ -35,6 +35,18 @@ def run(self):
         for ext in self.extensions:
             self.build_extension(ext)
 
+    def get_ext_filename(self, ext_name):
```
ax3l, line 38: Overwritten temporarily since it looks like the `build_wheel` step does not have the same env vars in pyoide than the earlier steps during build.

Wheel packaging still does not pick up the `openpmd_api_cxx` shared lib though...

setup.py
```diff
@@ -186,7 +195,7 @@ def build_extension(self, ext):
         'Source': 'https://github.com/openPMD/openPMD-api',
         'Tracker': 'https://github.com/openPMD/openPMD-api/issues',
     },
-    ext_modules=[CMakeExtension('openpmd_api_cxx')],
+    ext_modules=[CMakeExtension('openpmd_api.openpmd_api_cxx')],
```
ax3l, line 198: Playing with declaring the full sub-module here does not help either with the install pick-up-to-wheel issue.

PR #1207, opened 2022-02-24
 ADIOS2: Flush to disk within a step

franzpoeschel:
This was [recently added](https://github.com/ornladios/ADIOS2/pull/2807) in ADIOS2 (~~I think~~ only for BP5) and we need it to save memory e.g. in PIConGPU

Also it could be used to enable the IO pattern used by HiPace that so far only works in HDF5.

TODO
- [x] Prepare infrastructure in openPMD: pass-through for flushing parameters
    I have now removed flushing parameters from the AbstractIOHandler and instead created a struct for flushing parameters that is passed all the way down through our flushing logic
    This is a somewhat heavy change, but it makes our flushing logic stateless, meaning that we have to do less of these `try...catch...throw` tricks to restore state.
- [x] Python bindings
- [x] Implement the new logic in ADIOS2 (only for BP5 probably)
- [x] Maybe add a JSON option to turn it on/off as needed
- [x] Testing, documentation
- [x] More fine-grained user control
- [x] Documentation as mentioned in https://github.com/openPMD/openPMD-api/issues/986#issuecomment-1111990641
- [x] Update documentation
- [x] Add `preferred_flush_target="disk_override"/"buffer_override"` that takes precedence over the normal options
- [x] ~~Iteration::close??~~ We should it support as soon as it's actually needed, there is currently no use case for this API addition
- [x] ~~Drain m_attributeWrites in ADIOS2 backend upon this~~ No, data is not visible to a reader after PerformDataWrites() anyway, so writing the attributes early brings no advantage, but the disadvantage that they cannot be overwritten late


include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -57,6 +57,12 @@ class unsupported_data_error : public std::runtime_error
     {}
 };
 
+enum class FlushMode : bool
```
ax3l: As discussed:
https://openpmd-api.readthedocs.io/en/0.14.4/details/mpi.html

Flush is always declared collective.

franzpoeschel:
@psychocoderHPC asked about more fine-grained control for this to avoid flushing to disk when only some KB are in the buffers. There are several possible solutions that I can imagine:

1. Add a flushing threshold to our ADIOS2 backend, e.g. `adios2.engine.flushing_threshold = "2GiB"`. If the ADIOS2 backend notices that more than 2GiB were flushed, then it will store data to disk, otherwise use `PerformPuts()`.
    Problem -> parallelism
2. Turn "flush_during_step" into a dataset-specific option
2. Add JSON configuration to the `flush()` API, e.g. `series.flush(R"(adios2.preferred_flush_target = "disk")");`

test/SerialIOTest.cpp
```diff
@@ -30,6 +30,10 @@
 #include <tuple>
 #include <vector>
 
+#if __GNUC__ >= 11
+#include <filesystem>
```
ax3l: Can we use an auxiliary function for this?
Filesystem is tricky and sometimes needs also extra link flags like `-lstdc++fs`, that I don't yet want to get into.

Or did this get better with GCC 11+?

ax3l: Note that the `__GNUC__` macro **does not mean GCC** specifically and is also set by clang, icpc and others.

franzpoeschel: Ah, good catch about the `__GNUC__` thing, will fix

I can have a look if I can do the same test with `fstat` or similar as well, might also go with this. 
I only need this test to ensure that the right API call is chosen by the ADIOS2 backend, so restricting it to a single compiler set that implements `<filesystem>` well should be ok as an alternative

docs/source/details/backendconfig.rst
```diff
@@ -119,6 +122,12 @@ Explanation of the single keys:
   Please refer to the `official ADIOS2 documentation <https://adios2.readthedocs.io/en/latest/engines/engines.html>`_ for the available engine parameters.
   The openPMD-api does not interpret these values and instead simply forwards them to ADIOS2.
 * ``adios2.engine.usesteps``: Described more closely in the documentation for the :ref:`ADIOS2 backend<backends-adios2>` (usesteps).
+* ``adios2.engine.preferred_flush_target`` Only relevant for BP5 engine, possible values are ``"disk"`` and ``"buffer"`` (default: ``"disk"``).
```
franzpoeschel, line 125: I've called this `preferred_flush_target` since it is only implemented by BP5. Otherwise, we will see user requests `"I specified flush_target=disk, why is BP4 buffering the data anyway?"`.
I don't really want to explicitly use the BP5 name, e.g. `bp5_flush_target=disk` because other engines might in future implement this too.

ax3l:
@franzpoeschel please needs a small rebase

ax3l:
@franzpoeschel please needs a small rebase

include/openPMD/IO/FlushParametersInternal.hpp
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "openPMD/IO/AbstractIOHandler.hpp"
+#include "openPMD/auxiliary/JSON_internal.hpp"
```
franzpoeschel, line 25: Cannot include this in public headers, hence the necessary distinction between `FlushParams` and `ParsedFlushParams`, even if both structs are used only internally.
Unifying both structs requires removing `AbstractIOHandler.hpp` from public headers.

docs/source/details/backendconfig.rst
```diff
@@ -119,6 +122,12 @@ Explanation of the single keys:
...
+
+  * If ``"disk"``, data will be moved to disk on every flush.
+  * If ``"buffer"``, then only upon ending an IO step or closing an engine.
+
+  This behavior can be overridden on a per-flush basis by specifying this JSON/TOML key as an optional parameter to the ``Series::flush()`` or ``Attributable::flushSeries()`` methods.
```
franzpoeschel:
```suggestion
  This behavior can be overridden on a per-flush basis by specifying this JSON/TOML key as an optional parameter to the ``Series::flush()`` or ``Attributable::seriesFlush()`` methods.
```

include/openPMD/IO/FlushParametersInternal.hpp
ax3l: Good question why we need `AbstractIOHandler.hpp` in public headers actually.

ax3l:
Can you please post the multi-KB `.png` images as attachment to the PR description?
Then you can link to that image (click on it, get link) via this `.rst` syntax: https://github.com/ECP-WarpX/impactx/pull/173
```rst
.. figure:: https://user-images.githubusercontent.com/1353258/180332191-f9ce11fc-8c56-4713-a91a-2ad12ab09805.png
   :alt: Chicane beam width and emittance evolution
```

This allows us to not have to check them into the source code and avoid that our source repo gets too large (we clone openPMD-api very often in CMake superbuilds).

include/openPMD/IO/FlushParametersInternal.hpp
franzpoeschel: Because we use it heavily in the `private:` sections of our public classes, which are still defined in the headers. We would have to move the entire implementation logic out of the public headers to avoid including it, which would be a very heavy restructuring (and most likely end up in a solution similar what ADIOS2 does by having a separate source tree only for the public APIs)

franzpoeschel:
BP5 without intermittent flushes (old behavior, now opt-in):
memory_bp5_no_flush
BP5 with intermittent flushes (new behavior)
memory_bp5_yes_flush
BP4 file-based with specified InitialBufferSize as ~6GB:
memory_filebased
Catastrophic behavior: BP4 without using steps, no intermittent output possible
memory_groupbased_nosteps
Streaming with a slow reader, writer gets congested:
memory_sst_congested
Streaming with fast reader, InitialBufferSize specified:
memory_sst_easy
Variable-based encoding in BP4 with InitialBufferSize specified, buffer is reused:
memory_variablebased
Variable-based encoding in BP4 without InitialBufferSize specified, exponential probing phase in the beginning (memory usage >= twice the actually needed memory size due to reallocation)
memory_variablebased_initialization

franzpoeschel:
> Can you please post the multi-KB `.png` images as attachment to the PR description? Then you can link to that image (click on it, get link) via this `.rst` syntax: [ECP-WarpX/impactx#173](https://github.com/ECP-WarpX/impactx/pull/173)
> 
> ```
> .. figure:: https://user-images.githubusercontent.com/1353258/180332191-f9ce11fc-8c56-4713-a91a-2ad12ab09805.png
>    :alt: Chicane beam width and emittance evolution
> ```
> 
> This allows us to not have to check them into the source code and avoid that our source repo gets too large (we clone openPMD-api very often in CMake superbuilds).

Good idea, yeah
Last commit does that

include/openPMD/IO/FlushParametersInternal.hpp
ax3l: Same impression, would not have this added to our priorities right now.

ax3l (approved):
Awesome, thanks a lot :tada:

PR #1208, opened 2022-02-24
clang-tidy: add `pro-type-cstyle-cast`

ax3l:
Add `cppcoreguidelines-pro-type-cstyle-cast`:

> This check flags all use of C-style casts that perform a static_cast downcast, const_cast, or reinterpret_cast.
>
> Use of these casts can violate type safety and cause the program to access a variable that is actually of type X to be accessed as if it were of an unrelated type Z. Note that a C-style (T)expression cast means to perform the first of the following that is possible: a const_cast, a static_cast, a static_cast followed by a const_cast, a reinterpret_cast, or a reinterpret_cast followed by a const_cast. This rule bans (T)expression only when used to perform an unsafe cast.

https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-cstyle-cast.html

For an example how a c-style cast can accidentally cause a `const` cast, see: https://github.com/ECP-WarpX/WarpX/pull/2889

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,5 +2,5 @@
 # FIXME: all performance-* reports
 # FIXME: all cert-* reports
 # FIXME: all bugprone-* reports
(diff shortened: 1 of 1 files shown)


ax3l:
I wonder what I do wrong that locally, this does not apply fixes:
```bash
share/openPMD/download_samples.sh build
chmod u-w build/samples/git-sample/*.h5

cmake -S . -B build           -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy-10);-system-headers=0;--fix" -DopenPMD_USE_INVASIVE_TESTS=ON
cmake --build build -- VERBOSE=ON
```
(I also tried the (older?) `-fix`.) Looks to me like this check should have a fixit: https://reviews.llvm.org/D14096

I see the warnings, but it does not fix the code automatically...

PR #1209, opened 2022-02-24
Patch MSVC pybind11 debug bug

franzpoeschel:
Our MSVC CI seems to be [broken](https://github.com/openPMD/openPMD-api/runs/5323849404?check_suite_focus=true) currently due to an include order error in pybind11

This PR patches our internal version of pybind11 based on [this comment](https://github.com/microsoft/onnxruntime/issues/9735#issuecomment-970718821) to fix the issue.

TODO:

- [ ] Maybe try if upgrading pybind11 to a recent version will fix it instead? (According to the linked thread, it won't): #1220

### DIFF
--- a/share/openPMD/thirdParty/pybind11/include/pybind11/pybind11.h
+++ b/share/openPMD/thirdParty/pybind11/include/pybind11/pybind11.h
@@ -10,6 +10,24 @@
 
 #pragma once
 
+/*
+ * https://github.com/microsoft/onnxruntime/issues/9735#issuecomment-970718821
+ * The following block patches an MSVC bug.
+ */
+
+#if defined(_MSC_VER)
+#    if (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION < 4)
+#        define HAVE_ROUND 1
+#    endif
+#    include <corecrt.h>
+#    pragma warning(push)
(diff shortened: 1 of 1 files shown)


ax3l:
Updated now upstream as https://github.com/pybind/pybind11/pull/3784

share/openPMD/thirdParty/pybind11/include/pybind11/pybind11.h
```diff
@@ -10,6 +10,24 @@
 
 #pragma once
 
+/*
+ * https://github.com/microsoft/onnxruntime/issues/9735#issuecomment-970718821
+ * The following block patches an MSVC bug.
+ */
+
```
ax3l:
```suggestion
/* https://github.com/microsoft/onnxruntime/issues/9735#issuecomment-970718821
 * The following block patches MSVC debug builds:
 * Include Python header, disable linking to pythonX_d.lib on Windows in debug mode.
 */
```

ax3l (dismissed):

ax3l:
Hm, cannot reproduce an issue upstream.

ax3l:
Upgrade attempt in #1220

ax3l:
We can patch this on the 0.14.* release line, but if #1220 passes we will take an upgrade for `dev`/`0.15` instead :)

PR #1210, opened 2022-02-25
Toml11: Don't install internally shipped library

franzpoeschel:
toml11 adds itself to the install targets when using `add_subdirectory`. Using `EXCLUDE_FROM_ALL` ensures that it doesn't.
Also use the chance to hide the toml11 CMake options by marking them as advanced.

Related: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3986

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -211,7 +211,19 @@ target_link_libraries(openPMD::thirdparty::nlohmann_json
 
 # external library: toml11
 if(openPMD_USE_INTERNAL_TOML11)
-    add_subdirectory("${openPMD_SOURCE_DIR}/share/openPMD/thirdParty/toml11")
+    # EXCLUDE_FROM_ALL ensures that toml11 is not part of
+    # make install.
+    # The library is header-only and linking it against
+    # PIConGPU targets is sufficient.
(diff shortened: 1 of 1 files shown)


ax3l:
Waiting for https://github.com/ToruNiina/toml11/pull/185 and a new toml11 release to potentially replace this :)

CMakeLists.txt
```diff
@@ -211,7 +211,19 @@ target_link_libraries(openPMD::thirdparty::nlohmann_json
 
 # external library: toml11
 if(openPMD_USE_INTERNAL_TOML11)
-    add_subdirectory("${openPMD_SOURCE_DIR}/share/openPMD/thirdParty/toml11")
+    # EXCLUDE_FROM_ALL ensures that toml11 is not part of
+    # make install.
+    # The library is header-only and linking it against
```
ax3l, line 216: To be more precise: the library is private and not used in our public API interfaces (aka no types from it are used in our header files) :)

ax3l:
PR was merged. Waiting for a new release after the latest `3.7.0` now :)
https://github.com/ToruNiina/toml11/issues/184

ax3l:
Fixed via #1227 :tada:

PR #1211, opened 2022-02-28
added support for encoding support (f/g/v)

guj:
Added encoding support to the benchmark 8a. 
Default is variableEncoding. 

To change use the keyword "encoding", values are f/g/v for file/group/variable based, respectively.


franzpoeschel:
Something seems to have gone wrong with the JSON formatting? @ax3l

PR #1213, opened 2022-03-02
Fix GCC 11 Compiler Warnings

franzpoeschel:
1. Compile warnings in Attribute.hpp:
    g++ 11 does not understand that these variables are not primitive types
    and complains about possibly uninitialized values. Use the {}
    constructor to fix this.
2. We still had two instances of AccessType in the tests, replace this with Access

### DIFF
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -122,6 +122,7 @@ public:
 template <typename T, typename U>
 auto doConvert(T *pv) -> U
 {
+    (void)pv;
     if constexpr (std::is_convertible_v<T, U>)
     {
         return static_cast<U>(*pv);
@@ -199,7 +200,6 @@ auto doConvert(T *pv) -> U
             "getCast: no scalar to vector conversion possible.");
     }
 
-    (void)pv;
     throw std::runtime_error("getCast: no cast possible.");
 }
 
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -199,8 +199,10 @@ auto doConvert(T *pv) -> U
         throw std::runtime_error(
             "getCast: no scalar to vector conversion possible.");
     }
-
-    throw std::runtime_error("getCast: no cast possible.");
+    else
+    {
+        throw std::runtime_error("getCast: no cast possible.");
+    }
 }
 
 /** Retrieve a stored specific Attribute and cast if convertible.
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -132,7 +132,7 @@ auto doConvert(T *pv) -> U
                           typename T::value_type,
                           typename U::value_type>)
         {
-            U res;
+            U res{};
             res.reserve(pv->size());
             std::copy(pv->begin(), pv->end(), std::back_inserter(res));
             return res;
@@ -149,7 +149,7 @@ auto doConvert(T *pv) -> U
                           typename T::value_type,
                           typename U::value_type>)
         {
-            U res;
+            U res{};
             res.reserve(pv->size());
             std::copy(pv->begin(), pv->end(), std::back_inserter(res));
             return res;
@@ -167,7 +167,7 @@ auto doConvert(T *pv) -> U
                           typename T::value_type,
                           typename U::value_type>)
         {
-            U res;
+            U res{};
             if (res.size() != pv->size())
             {
                 throw std::runtime_error(
@@ -189,7 +189,7 @@ auto doConvert(T *pv) -> U
     {
         if constexpr (std::is_convertible_v<T, typename U::value_type>)
         {
-            U res;
+            U res{};
             res.reserve(1);
             res.push_back(static_cast<typename U::value_type>(*pv));
             return res;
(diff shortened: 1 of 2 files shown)


test/SerialIOTest.cpp
```diff
@@ -421,8 +421,7 @@ TEST_CASE("multiple_series_handles_test", "[serial]")
     {
         std::unique_ptr<openPMD::Series> series_ptr;
         {
-            openPMD::Series series(
-                "sample%T.json", openPMD::AccessType::CREATE);
```
ax3l: We left those intentionally to test that the old type still works.

We could silence this at one last location with a local suppressor?

ax3l:
PR rebased.

test/SerialIOTest.cpp
franzpoeschel: I was under the impression that they sneaked in because this test comes from a long-standing pull request (the bug tested by this took a long time to fix).
But we can definitely use the chance to test for existence of AccessType with this

include/openPMD/backend/Attribute.hpp
```diff
@@ -199,7 +200,6 @@ auto doConvert(T *pv) -> U
             "getCast: no scalar to vector conversion possible.");
     }
 
-    (void)pv;
     throw std::runtime_error("getCast: no cast possible.");
```
franzpoeschel: Need to put these into an else branch, ICC seems to determine unreachable statements after resolving if constexpr

test/SerialIOTest.cpp
```diff
@@ -414,6 +414,16 @@ TEST_CASE("available_chunks_test_json", "[serial][json]")
 
 TEST_CASE("multiple_series_handles_test", "[serial]")
 {
+#if defined(__INTEL_COMPILER)
+#pragma warning(disable : 2282)
+#endif
+    /*
+     * clang also understands these pragmas.
+     */
+#if defined(__GNUC__) || defined(__clang__)
```
ax3l: `__GNUC__` is quite often set (GCC, Clang, ICC, etc.)
but the version is specific to GCC:
https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

include/openPMD/backend/Attribute.hpp
```diff
@@ -180,27 +185,42 @@ auto doConvert(T *pv) -> U
...
+    }
+#if defined(__INTEL_COMPILER)
+/*
+ * ICPC has trouble with if constexpr, thinking that return statements are
+ * missing afterwards. Deactivate the warning.
+ * Note that putting a statement here will not help to fix this since it will
+ * then complain about unreachable code.
+ * https://community.intel.com/t5/Intel-C-Compiler/quot-if-constexpr-quot-and-quot-missing-return-statement-quot-in/td-p/1154551
+ */
+#pragma warning(disable : 1011)
```
ax3l, line 222: Since this is a `.hpp` file, do you need to `pop` this again?

franzpoeschel: Yes, I need to reset it to default again after this

include/openPMD/backend/Attribute.hpp
ax3l: ah classic. a range of NVCC 11 minor releases does that as well

include/openPMD/backend/Attribute.hpp
ax3l: done in latest commit :)

test/SerialIOTest.cpp
```diff
@@ -414,6 +414,13 @@ TEST_CASE("available_chunks_test_json", "[serial][json]")
 
 TEST_CASE("multiple_series_handles_test", "[serial]")
 {
+    /*
+     * clang also understands these pragmas.
+     */
+#if defined(__GNUC__MINOR__) || defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
ax3l, line 422: With this one set, you can remove `-Wno-deprecated-declarations` from quite a few CI scripts in `.github/workflows/`

ax3l (approved):

test/SerialIOTest.cpp
ax3l: -> #1246

test/SerialIOTest.cpp
```diff
@@ -414,6 +414,13 @@ TEST_CASE("available_chunks_test_json", "[serial][json]")
 
 TEST_CASE("multiple_series_handles_test", "[serial]")
 {
+    /*
+     * clang also understands these pragmas.
+     */
+#if defined(__GNUC__MINOR__) || defined(__clang__)
```
ax3l, line 420: Typo fixed in #1255: `__GNUC_MINOR__`

test/SerialIOTest.cpp
```diff
@@ -456,6 +463,12 @@ TEST_CASE("multiple_series_handles_test", "[serial]")
          */
         series_ptr->flush();
     }
+#if defined(__INTEL_COMPILER)
+#pragma warning(disable : 2282)
+#endif
+#if defined(__GNUC__MINOR__) || defined(__clang__)
```
ax3l, line 469: same here

ax3l:
I have problems applying this and the following PRs to the `release-0.14.5` branch

ax3l:

PR #1214, opened 2022-03-03
Clearly fail when users select a wrong backend

franzpoeschel:
Original issue: https://github.com/ComputationalRadiationPhysics/picongpu/issues/4004

When selecting a wrong backend, it might currently happen that openPMD activates a dummy backend, resulting in error messages like
```
[~Series] An error occurred: [Series] Closed iteration has not been written. This is an internal error.
Unhandled exception of type 'St13runtime_error' with message '[Series] Closed iteration has not been written. This is an internal error.', terminating
```

This fixes that

### DIFF
--- a/src/IO/AbstractIOHandlerHelper.cpp
+++ b/src/IO/AbstractIOHandlerHelper.cpp
@@ -53,6 +53,7 @@ namespace
                 "backend '" +
                 backendName + "'.");
         }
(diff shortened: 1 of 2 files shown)


ax3l (approved):
:+1:

ax3l:
I have problems backporting this to the `release-0.14.5` branch

PR #1215, opened 2022-03-03
Fixes for BP5 engine

franzpoeschel:
1. See [this comment](https://github.com/ornladios/ADIOS2/issues/2649#issuecomment-1058095101), the Span-based API is still very much relevant for performance in BP5.
2. The BP5 engine requires that every interaction with it happens inside an active IO step, take care of this

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -605,6 +605,7 @@ void ADIOS2IOHandlerImpl::writeAttribute(
         auto prefix = filePositionToString(pos);
 
         auto &filedata = getFileData(file, IfFileNotOpen::ThrowError);
+        filedata.requireActiveStep();
         filedata.invalidateAttributesMap();
         m_dirty.emplace(std::move(file));
 
@@ -703,7 +704,13 @@ void ADIOS2IOHandlerImpl::getBufferView(
     Writable *writable, Parameter<Operation::GET_BUFFER_VIEW> &parameters)
 {
     // @todo check access mode
-    if (m_engineType != "bp4" && m_engineType != "bp5")
+    std::string optInEngines[] = {"bp4", "bp5", "file", "filestream"};
+    if (std::none_of(
+            begin(optInEngines),
+            end(optInEngines),
+            [this](std::string const &engine) {
+                return engine == this->m_engineType;
+            }))
     {
         parameters.out->backendManagedBuffer = false;
         return;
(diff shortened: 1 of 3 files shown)


src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2189,7 +2200,8 @@ namespace detail
             {
                 isStreaming = true;
                 optimizeAttributesStreaming =
```
franzpoeschel, line 2210: We can think about removing this optimization, it becomes irrelevant in variable-based iteration encoding and there is no harm in using that encoding for streaming.
Not relevant for this PR though.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2511,9 +2522,22 @@ namespace detail
     adios2::Engine &BufferedActions::requireActiveStep()
     {
         adios2::Engine &eng = getEngine();
+        /*
+         * If streamStatus is Parsing, do NOT open the step.
```
franzpoeschel, line 2540: Should have added this comment earlier already, just took the opportunity now

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2715,28 +2744,31 @@ namespace detail
...
-            adios2::StepStatus adiosStatus = m_lastStepStatus;
+            adios2::StepStatus adiosStatus{};
 
-            // Step might have been opened implicitly already
-            // by requireActiveStep()
-            // In that case, streamStatus is DuringStep and Adios
-            // return status is stored in m_lastStepStatus
             if (streamStatus != StreamStatus::DuringStep)
             {
-                flush(
```
franzpoeschel, line 2726: Flushing here was wrong but harmless before. This dormant bug has woken up with this PR, so change it.

test/SerialIOTest.cpp
```diff
@@ -5145,6 +5145,26 @@ TEST_CASE("iterate_nonstreaming_series", "[serial][adios2]")
                 backend.extension,
             false,
             backend.jsonBaseConfig());
+#if openPMD_HAVE_ADIOS2 &&                                                     \
+    ADIOS2_VERSION_MAJOR * 1000000000 + ADIOS2_VERSION_MINOR * 100000000 +     \
+            ADIOS2_VERSION_PATCH * 1000000 + ADIOS2_VERSION_TWEAK >=           \
+        2701001223
```
franzpoeschel, line 5151: Note that our CI does not yet have any ADIOS2 versions that satisfy these conditions

franzpoeschel:
Reminder to myself: ReadRandomAccess

ax3l (approved):
Thank you!

Did I see correctly that we should backport this to `0.14.5`?

ax3l:
As discussed: targeting BP5 patches for the 0.15.0 release only (no backports to 0.14.*)

PR #1218, opened 2022-03-04
ADIOS2: more fine-grained control for file endings

### LINKED ISSUE #1205: Distinguish BP versions in file endings
**Is your feature request related to a problem? If so, please describe.**
BP3 and BP4 are already used in production, BP5 is coming up. The formats are incompatible with each other, but we use the same endings for them so far.
BP5 is not going to replace BP4, so we should go for a solution that will let people use both formats side by side without too much confusion.

Instead of `simData.bp`, write files like `simData.bp3`, `simData.bp4`, `simData.bp5`.
Optionally even enforce version numbers if users don't specify them, e.g. `Series("simData.bp", Access::CREATE)` might create `simData.bp4`. Only do this if we find a solution that does not break existing workflows.

**Describe alternatives you've considered**
Automatic version detection in ADIOS2 if using the generic ending `.bp`
Being explicit about versions is still good

franzpoeschel:
Close #1205

TODO:

- [x] Frontend implementation
- [x] Backend implementation
- [x] Ensure that ideally the filename on disk is the same as specified by the user in most cases
- [x] Testing
- [x] Merge #1214 first (they would conflict else)
- [x] What to do for `Series("asdf.bp5", Access::CREATE, "adios2.engine.type = \"file\"")`?
    Current implementation: Accept this in the backend, but warn the user that the `file` engine will decide the BP version independent of the extension: `[ADIOS2] Specified explicit ending '.bp5' in combination with generic file engine 'file'. ADIOS2 will pick a default file ending independent of specified suffix. (E.g. 'simData.bp5' might actually be written as a BP4 dataset.)`
- [x] Test that `Series("asdf.bp5", Access::CREATE, "backend = \"adios2\"")` is handled correctly
- [x] Merge #1215 first and delete those commits from this branch
- [x] Documentation
- [x] Merge #1262 first
- [x] extend `openPMD::getFileExtensions()`
- [x] Merge #1275 first

### DIFF
--- a/.github/workflows/tooling.yml
+++ b/.github/workflows/tooling.yml
@@ -52,7 +52,7 @@ jobs:
         python3 -m pip install -U numpy
         sudo .github/workflows/dependencies/install_spack
     - name: Build
-      env: {CC: mpicc, CXX: mpic++, OMPI_CC: clang-10, OMPI_CXX: clang++-10, CXXFLAGS: -Werror, OPENPMD_HDF5_CHUNKS: none, OPENPMD_TEST_NFILES_MAX: 100}
+      env: {CC: mpicc, CXX: mpic++, OMPI_CC: clang-10, OMPI_CXX: clang++-10, CXXFLAGS: -Werror, OPENPMD_HDF5_CHUNKS: none, OPENPMD_TEST_NFILES_MAX: 100, SKIP_ADIOS2_FILE_ENDINGS_TEST: 0}
       run: |
         eval $(spack env activate --sh .github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad1_ad2/)
         spack install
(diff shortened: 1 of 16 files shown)


franzpoeschel:
Current approach: Each ADIOS2 engine has a set of accepted file endings and a default file ending. If the user specifies any of the accepted endings, it will be used. If no ending is specified or any other ending is specified, the default ending will be used.
E.g.: Engine "file": default ending "bp", accepted endings "bp", "bp3", "bp4", "bp5"
Engine "bp4": default ending "bp4", accepted endings "bp", "bp3", "bp4", "bp5"

franzpoeschel:
This PR now has some BP5 fixes that I'll need to factor out.

ax3l:
@franzpoeschel can you please rebase this? Looks like all predicate PRs are merged now :)

ax3l:
CI looks good, but I'll run it again to make sure it's up-to-date.

ax3l (approved):
Looks great, thanks a lot! :+1:

PR #1223, opened 2022-03-09
Don't safeguard empty strings while reading

franzpoeschel:
This is a workaround/fix for #1221

### DIFF
--- a/include/openPMD/backend/Attributable.hpp
+++ b/include/openPMD/backend/Attributable.hpp
@@ -90,21 +90,37 @@ namespace internal
         A_MAP m_attributes;
     };
 
+    enum class SetAttributeMode : char
+    {
+        WhileReadingAttributes,
+        FromPublicAPICall
+    };
+
     /** Verify values of attributes in the frontend
      *
      * verify string attributes are not empty (backend restriction, e.g., HDF5)
      */
     template <typename T>
-    inline void attr_value_check(std::string const /* key */, T /* value */)
+    inline void attr_value_check(
+        std::string const /* key */, T /* value */, SetAttributeMode)
     {}
 
     template <>
-    inline void attr_value_check(std::string const key, std::string const value)
+    inline void attr_value_check(
+        std::string const key, std::string const value, SetAttributeMode mode)
     {
-        if (value.empty())
-            throw std::runtime_error(
-                "[setAttribute] Value for string attribute '" + key +
-                "' must not be empty!");
+        switch (mode)
+        {
+        case SetAttributeMode::FromPublicAPICall:
+            if (value.empty())
+                throw std::runtime_error(
+                    "[setAttribute] Value for string attribute '" + key +
+                    "' must not be empty!");
+            break;
+        case SetAttributeMode::WhileReadingAttributes:
+            // no checks while reading
+            break;
+        }
     }
 
     template <typename>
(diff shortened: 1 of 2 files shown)


include/openPMD/backend/Attributable.hpp
```diff
@@ -90,21 +90,37 @@ namespace internal
         A_MAP m_attributes;
     };
 
+    enum class SetAttributeMode : char
+    {
+        WhileReadingAttributes,
+        FromPublicAPICall
```
ax3l, line 96: Is that public API call both read and write? Might need a doxygen string, even though its readable already :)

ax3l: Ah, `Set`... so write

include/openPMD/backend/Attributable.hpp
```diff
@@ -410,11 +433,28 @@ OPENPMD_protected
...
+}
+
+inline bool
+Attributable::setAttribute(std::string const &key, char const value[])
+{
+    return setAttributeImpl(
+        key, value, internal::SetAttributeMode::FromPublicAPICall);
+}
+
+// TODO explicitly instantiate Attributable::setAttribute for all T in Datatype
```
ax3l: This is done in `src/backend/Attributable.cpp`
```suggestion
// note: we explicitly instantiate Attributable::setAttributeImpl for all T in Datatype in Attributable.cpp
```

ax3l (approved):
Thank you, LGTM :)

ax3l:
Will fix merge conflict from the merge of #1226

src/backend/Attributable.cpp
```diff
@@ -400,23 +487,38 @@ void Attributable::readAttributes(ReadMode mode)
...
         case DT::VEC_CDOUBLE:
-            setAttribute(att, a.get<std::vector<std::complex<double> > >());
+            setAttributeImpl(
+                att,
+                a.get<std::vector<std::complex<double> > >(),
+                internal::SetAttributeMode::WhileReadingAttributes);
             break;
         case DT::VEC_CLONG_DOUBLE:
             setAttribute(
                 att, a.get<std::vector<std::complex<long double> > >());
```
ax3l: Is this one missing?
```suggestion
            setAttributeImpl(
                att,
                a.get<std::vector<std::complex<long double> > >(),
                internal::SetAttributeMode::WhileReadingAttributes);
```

ax3l (approved):

src/backend/Attributable.cpp
ax3l: Added to last commit.

include/openPMD/backend/Attributable.hpp
ax3l: added to 2nd commit.

include/openPMD/backend/Attributable.hpp
franzpoeschel: It's write-only in the public API, but during parsing, we emulate read_write mode and use `setAttribute` also internally

src/backend/Attributable.cpp
franzpoeschel: Looks like it, good catch

ax3l:
I have trouble backporting this to the `release-0.14.5` branch due to changes in `include/openPMD/backend/Attributable.hpp`

ax3l:

PR #1197, opened 2022-02-08
CI: Add ADIOS1 (BP3) Python Test

ax3l:
Make sure we test ADIOS1 in Python tests.

We have known issues here that we need to fix.

PR #1224, opened 2022-03-09
Fix use after free in ADIOS1IOHandler

franzpoeschel:
Extracted from #1197

There is currently a (low-key) use after free in ADIOS1, happening if a user calls `loadChunk()` without storing the buffer. #1197 will add a test.


ax3l (approved):
:rocket: :sparkles:

ax3l:
I have problems backporting this to `release-0.14.5` branch due to the change in `CommonADIOS1IOHandler.hpp/cpp`

ax3l:

PR #1225, opened 2022-03-09
Increase reference count also in other load_chunk overload

franzpoeschel:
Don't know if it is necessary, but looks like we forgot it back when we added this reference count trick.

Extracted from #1197.

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -539,6 +539,10 @@ void load_chunk(
         }
     }
 
+    // here, we increase a reference on the user-passed data so that
+    // temporary and lost-scope variables stay alive until we flush
+    // note: this does not yet prevent the user, as in C++, to build
+    //       a race condition by manipulating the data they passed
     auto load_data =
         [&r, &buffer, &buffer_info, &offset, &extent](auto cxxtype) {
             using CXXType = decltype(cxxtype);
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -539,53 +539,51 @@ void load_chunk(
         }
     }
 
+    auto load_data =
+        [&r, &buffer, &buffer_info, &offset, &extent](auto cxxtype) {
+            using CXXType = decltype(cxxtype);
+            buffer.inc_ref();
+            // buffer_info.inc_ref();
+            void *data = buffer_info.ptr;
+            std::shared_ptr<CXXType> shared(
+                (CXXType *)data, [buffer](CXXType *) { buffer.dec_ref(); });
+            r.loadChunk(std::move(shared), offset, extent);
+        };
+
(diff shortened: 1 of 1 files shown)


src/binding/python/RecordComponent.cpp
```diff
@@ -539,53 +539,51 @@ void load_chunk(
         }
     }
 
+    auto load_data =
```
ax3l, line 546: Let's add the same comment we add in store so we know what's going on here :)
```suggestion
    // here, we increase a reference on the user-passed data so that
    // temporary and lost-scope variables stay alive until we flush
    // note: this does not yet prevent the user, as in C++, to build
    //       a race condition by manipulating the data they passed
    auto load_data =
```

ax3l (approved):
Thx for catching that!

x-ref: #912 #833

ax3l:
In #912, I noted that we did not yet have an in-memory overload for `load_chunk` in Python.

Do we have one yet that can trigger the issue and shall we add a test similar to #912?

franzpoeschel:
> In #912, I noted that we did not yet have an in-memory overload for `load_chunk` in Python.
> 
> Do we have one yet that can trigger the issue and shall we add a test similar to #912?

~~This PR did not change the public Python API, just fixed a potential memory bug~~

I misunderstood your post, I'll need to have a look where this function is actually called

ax3l:
Yes, I think we don't have a test in `test/python/unittest/API/APITest.py` for pre-allocated arrays yet, but let's add it :)

We also need to add a `pybind11` signature to store into a pre-allocated python array (we can check its extends when it is passed to us in `load_chunk`).

franzpoeschel:
> We also need to add a `pybind11` signature to store into a pre-allocated python array (we can check its extends when it is passed to us in `load_chunk`).

It seems that this exists already
```cpp
        .def(
            "load_chunk",
            [](RecordComponent &r,
               py::buffer buffer,
               Offset const &offset_in,
               Extent const &extent_in) {
                uint8_t ndim = r.getDimensionality();

                // default arguments
                //   offset = {0u}: expand to right dim {0u, 0u, ...}
                Offset offset = offset_in;
                if (offset_in.size() == 1u && offset_in.at(0) == 0u)
                    offset = Offset(ndim, 0u);

                //   extent = {-1u}: take full size
                Extent extent(ndim, 1u);
                if (extent_in.size() == 1u && extent_in.at(0) == -1u)
                {
                    extent = r.getExtent();
                    for (uint8_t i = 0u; i < ndim; ++i)
                        extent[i] -= offset[i];
                }
                else
                    extent = extent_in;

                std::vector<bool> flatten(ndim, false);
                load_chunk(r, buffer, offset, extent);
            },
            py::arg("pre-allocated buffer"),
            py::arg_v(
                "offset", Offset(1, 0u), "np.zeros(Record_Component.shape)"),
            py::arg_v("extent", Extent(1, -1u), "Record_Component.shape"))
```

PR #1228, opened 2022-03-14
[WIP] Pre-Commit: Clang-Format CXX Endings

ax3l:
`identify` used to detect our files does not catch all our file endings.

Refs.:
- https://pre-commit.com/#pre-commit-configyaml---hooks
- https://pre-commit.com/#filtering-files-with-types
  - https://pre-commit.com/#regular-expressions
- https://github.com/pre-commit/mirrors-clang-format
- https://github.com/ssciwr/clang-format-wheel#use-from-pre-commit
- https://github.com/pre-commit/identify/issues/119
- https://github.com/pybind/pybind11/blob/v2.9.1/.pre-commit-config.yaml#L96-L103

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -68,6 +68,8 @@ repos:
   rev: v13.0.1
   hooks:
   - id: clang-format
+    types: [file]
+    files: '.*\.(c|cpp|cxx|h|hpp|tpp)(\.in)?$'
 
 # Autoremoves unused Python imports
 - repo: https://github.com/hadialqattan/pycln
--- a/.github/workflows/clang-format/clang-format.sh
+++ b/.github/workflows/clang-format/clang-format.sh
@@ -7,6 +7,6 @@ else
     # received no arguments, find files on our own
     find include/ src/ test/ examples/ \
             -regextype egrep \
-            -type f -regex '.*\.(hpp|cpp|hpp\.in)$' \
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -68,6 +68,8 @@ repos:
   rev: v13.0.1
   hooks:
   - id: clang-format
+    types: [file]
+    files: '.*\.(c|cpp|cxx|h|hpp|tpp)(\.in)?$'
```
ax3l, line 72: This does not yet seem to work :thinking: 
Should be python regex: https://pre-commit.com/#regular-expressions

ax3l, line 72: This does not yet seem to work :thinking:

ax3l: Once this works, I to also need to add `.json` here as we use this for json formatting as well #1222

ax3l: Ah, could be python regex: https://pre-commit.com/#regular-expressions

franzpoeschel:
You were right, `.tpp` files are not yet considered by the manual script https://github.com/openPMD/openPMD-api/blob/dev/.github/workflows/clang-format/clang-format.sh

franzpoeschel:
Otherwise, this PR has been mainly obsoleted by https://github.com/openPMD/openPMD-api/pull/1403

PR #1229, opened 2022-03-14
Deprecate shareRaw

### LINKED ISSUE #1216: ADIOS2 BP5 optimization: Allow users to "give up" their data into openPMD
With BP5, using PerformPuts has a performance implication unlike in BP4 ([here](https://github.com/ornladios/ADIOS2/issues/2649#issuecomment-1057105950) is a detailed explanation). 

Users can currently avoid PerformPuts in two ways:

1. Use the span-based API
2. Use the streaming API and never flush()

A third way could theoretically be by allowing users to release their hold onto data:
```cpp
recordComponent.storeChunk( std::move( mySharedPointer ), offset, extent );
```

The ADIOS2 backend could then detect by using `std::shared_ptr::use_count()` that is the unique owner of the passed buffer and avoid calling PerformPuts().

**Big caveat:** This would be a breaking change since it would require openPMD to assume that the shared_ptr actually manages the memory. It would break things like this:
```cpp
std::vector<int> v{0,1,2,3,4,5};
recordComponent.storeChunk( std::shared_ptr<int>(v.data(), [](auto const * ptr){}), offset, extent);
```

This is not too bad because we already have an overload that takes vectors which most people use:
```cpp
std::vector<int> v{0,2,3,4,5};
recordComponent.storeChunk(v, offset, extent);
```
So we can use that overload for checking that situation.

A raw pointer overload should maybe be added for other cases.

franzpoeschel:
Idea: Instead of `rc.storeChunk(shareRaw(ptr), …)`, have users write `rc.storeChunkRaw(ptr, …)` which is a bit more transparent in what is happening. Partially resolves #1216 by introducing a clearer API that distinguishes between owned and non-owned resources.

TODO

- [x] Documentation
- [x] The diff on examples, tests and Python bindings shows the proposed API changes. Discuss if they are ok like this. (Especially the changes in `determineDatatype`)
- [ ] Maybe revert some usage changes, so the CI can test if shareRaw continues working
- [x] Introduce an internal version of shareRaw, so every creation of a non-owning shared pointer goes through the same function. Ideally, in the end a shared pointer is either owning or has been created by that function.

### DIFF
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -49,7 +49,7 @@ int main(int argc, char *argv[])
     cout << "Created a scalar mesh Record with all required openPMD "
             "attributes\n";
 
-    Datatype datatype = determineDatatypeContiguous(global_data);
+    Datatype datatype = determineDatatype(global_data.data());
     Extent extent = {size, size};
     Dataset dataset = Dataset(datatype, extent);
     cout << "Created a Dataset of size " << dataset.extent[0] << 'x'
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -49,7 +49,7 @@ int main(int argc, char *argv[])
     cout << "Created a scalar mesh Record with all required openPMD "
             "attributes\n";
 
-    Datatype datatype = determineDatatype(global_data.data());
+    Datatype datatype = determineDatatypeContiguous(global_data);
     Extent extent = {size, size};
     Dataset dataset = Dataset(datatype, extent);
     cout << "Created a Dataset of size " << dataset.extent[0] << 'x'
--- a/examples/3_write_serial.cpp
+++ b/examples/3_write_serial.cpp
@@ -49,7 +49,7 @@ int main(int argc, char *argv[])
     cout << "Created a scalar mesh Record with all required openPMD "
             "attributes\n";
 
-    Datatype datatype = determineDatatype(shareRaw(global_data));
+    Datatype datatype = determineDatatype(global_data.data());
     Extent extent = {size, size};
     Dataset dataset = Dataset(datatype, extent);
     cout << "Created a Dataset of size " << dataset.extent[0] << 'x'
@@ -63,7 +63,7 @@ int main(int argc, char *argv[])
     cout << "File structure and required attributes have been written\n";
 
     Offset offset = {0, 0};
-    rho.storeChunk(shareRaw(global_data), offset, extent);
+    rho.storeChunk(global_data, offset, extent);
     cout << "Stored the whole Dataset contents as a single chunk, "
             "ready to write content\n";
 
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -283,26 +283,58 @@ inline constexpr Datatype determineDatatype(std::shared_ptr<T>)
 }
 
 template <typename T>
-inline constexpr Datatype determineDatatypeRaw(T const *)
+inline constexpr Datatype determineDatatype(T *)
 {
     return determineDatatype<T>();
 }
 
+/*
+ * Catch-all overload for unsupported types, with static_assert errors
+ * triggered at compile-time.
+ */
 template <typename T_ContiguousContainer>
-inline constexpr Datatype determineDatatypeContiguous(T_ContiguousContainer &&)
+inline constexpr Datatype determineDatatype(T_ContiguousContainer &&)
 {
     using T_ContiguousContainer_stripped =
         std::remove_reference_t<T_ContiguousContainer>;
     if constexpr (auxiliary::IsContiguousContainer_v<
                       T_ContiguousContainer_stripped>)
     {
-        return determineDatatype<
-            typename T_ContiguousContainer_stripped::value_type>();
+        static_assert(auxiliary::dependent_false_v<T_ContiguousContainer>, R"(
+Error: Passed a contiguous container type to determineDatatype<>().
+These types are not directly supported due to colliding semantics.
+Assuming a vector object `std::vector<float> vec;`,
+use one of the following alternatives:
+
+1) If what you want is a direct openPMD::Datatype equivalent
+   of the container type, use:
+   `determineDatatype<decltype(vec)>()`
+   OR
+   `determineDatatype<std::vector<float>>()`.
+   The result will be `Datatype::VECTOR_FLOAT`.
+2) If what you want is the openPMD::Datatype equivalent of the *contained type*,
+   use the raw pointer overload by:
+   `determineDatatype(vec.data())`
+   The result will be `Datatype::FLOAT`.
+   This is the variant that you likely wish to use if intending to write data
+   from the vector via `storeChunk()`, e.g. `storeChunk(vec, {0}, {10})`.
+        )");
     }
     else
     {
-        return Datatype::UNDEFINED;
+        static_assert(auxiliary::dependent_false_v<T_ContiguousContainer>, R"(
+Error: Unknown datatype passed to determineDatatype<>().
+For a direct translation from C++ type to the openPMD::Datatype enum, use:
+`auto determineDatatype<T>() -> Datatype`.
+
+For detecting the contained datatpye of a pointer type (shared or raw pointer),
+use either of the following overloads:
+`auto determineDatatype<T>(std::shared_ptr<T>) -> Datatype` or
+`auto determineDatatype<T>(T *) -> Datatype`.
+        )");
     }
+    // Unreachable, but C++ does not know it
+    return Datatype::UNDEFINED;
 }
 
 /** Return number of bytes representing a Datatype
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -275,6 +275,12 @@ inline constexpr Datatype determineDatatype()
 }
 
 template <typename T>
+inline constexpr Datatype determineDatatype(T const *)
+{
+    return determineDatatype<T>();
+}
+
+template <typename T>
 inline constexpr Datatype determineDatatype(std::shared_ptr<T>)
 {
     using DT = Datatype;
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -20,6 +20,8 @@
  */
 #pragma once
 
+#include "openPMD/auxiliary/TypeTraits.hpp"
+
 #include <array>
 #include <climits>
 #include <complex>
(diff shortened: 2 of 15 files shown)


examples/3_write_serial.cpp
```diff
@@ -49,7 +49,7 @@ int main(int argc, char *argv[])
     cout << "Created a scalar mesh Record with all required openPMD "
             "attributes\n";
 
-    Datatype datatype = determineDatatype(shareRaw(global_data));
+    Datatype datatype = determineDatatype(global_data.data());
```
franzpoeschel, line 52: determineDatatype now also accepts (normal) pointers as input.

include/openPMD/backend/PatchRecordComponent.hpp
```diff
@@ -169,6 +172,14 @@ inline void PatchRecordComponent::load(std::shared_ptr<T> data)
     rc.m_chunks.push(IOTask(this, dRead));
 }
 
+template <typename T>
+inline void PatchRecordComponent::loadRaw(T *data)
+{
+    // @todo maybe use an internal shareraw version to avoid overlooking
+    // this instance in future
+    load<T>(std::shared_ptr<T>{data, [](auto const *) {}});
```
franzpoeschel: notice todo

examples/3_write_serial.cpp
ax3l: Since we are `shared_ptr` geeks in the public API, I would also add a `shared_ptr` overload to this.

ax3l: Let's also add a new `T_ContiguousContainer` overload for this one, for symmetry with `storeChunk`?

ax3l (commented):
This looks good to me and as discussed, let's proceed.

Only thing I would add is another overload for contiguous containers to `determineDatatype` for symmetry with `storeChunk`.

include/openPMD/backend/PatchRecordComponent.hpp
ax3l: done via `openPMD/auxiliary/ShareRawInternal.hpp` include

examples/3_write_serial.cpp
franzpoeschel: `T_ContiguousContainer` would conflict with our vector and array datatypes such as `VEC_INT`. Something like `determineDatatypeContiguousContainer` would work (or something less verbose..)

ax3l:
@franzpoeschel pls resolve the merge conflict :)

franzpoeschel:
> @franzpoeschel pls resolve the merge conflict :)

Done

include/openPMD/Datatype.hpp
```diff
@@ -277,161 +279,30 @@ inline constexpr Datatype determineDatatype()
...
+template <typename T>
+inline constexpr Datatype determineDatatypeRaw(T const *)
+{
+    return determineDatatype<T>();
+}
+
+template <typename T_ContiguousContainer>
+inline constexpr Datatype determineDatatypeContiguous(T_ContiguousContainer &&)
+{
+    using T_ContiguousContainer_stripped =
```
ax3l: For `determineDatatype`, I tend to just use one API function name with overloads for pointers and cont. containers.
There is no lifetime question here and additional naming is not really needed, adding unnecessary API calls to learn.

What do you think?

include/openPMD/RecordComponent.hpp
```diff
@@ -225,29 +197,118 @@ class RecordComponent : public BaseRecordComponent
...
+     *               is the unique owner).
+     *               For raw pointers, use loadChunkRaw().
+     * @param offset Offset within the dataset. Set to {0u} for full selection.
+     * @param extent Extent within the dataset, counted from the offset.
+     *               Set to {-1u} for full selection.
+     *               If offset is non-zero and extent is {-1u} the leftover
+     *               extent in the record component will be selected.
+     */
+    template <typename T>
+    void loadChunk(std::shared_ptr<T> data, Offset offset, Extent extent);
```
ax3l, line 222: Do we need a `std::shared_ptr<T[]>`, too?

include/openPMD/RecordComponent.hpp
```diff
@@ -225,29 +197,118 @@ class RecordComponent : public BaseRecordComponent
...
+     *               It is in the user of this API call's responsibility to
+     *               ensure that the lifetime of the container exceeds the next
+     *               <a
+     * href="https://openpmd-api.readthedocs.io/en/latest/usage/workflow.html#deferred-data-api-contract">
+     *               flush point</a>.
+     * @param offset Offset within the dataset.
+     * @param extent Extent within the dataset, counted from the offset.
+     */
     template <typename T_ContiguousContainer>
     typename std::enable_if<
```
ax3l: you can use `enable_if_t` :)

include/openPMD/Datatype.hpp
franzpoeschel: The reason I did it like this is a slight inconsistency that we would get:

```cpp
determineDatatype<std::vector<float>>(); // evaluates to Datatype::VECTOR_FLOAT
determineDatatype(std::vector<float>()); // evaluates to Datatype::FLOAT, interpreting the vector as a contiguous container
```

As long as we don't add the following overload, this would technically work, but still be extremely confusing imo:
```cpp
template <typename T>
inline constexpr Datatype determineDatatype(T)
{
     // ignoring reference decaying for this example now
    return determineDatatype<T>();
}
```

I see the appeal of not introducing new API calls left and right, but we should be aware of the possible source for confusion.

franzpoeschel: Checking this again, `determineDatatypeContiguous` is the only place where it is strictly *necessary* to have a new API instead of just an overload.
For `store/loadChunkRaw()`, it's just a good design decision imo (to indicate non-ownership of data), and for `determineDatatypeRaw` it's more a matter of consistency then.

franzpoeschel: I checked the PR again, and we could theoretically do the entire thing only with new overloads, instead of new API calls (even replace `storeChunkRaw` with `storeChunk`). So this is entirely a design decision:

* `storeChunkRaw()`: I like making it explicit that we are dealing with non-owned, raw memory
* `determineDatatypeContiguous()`: Avoids confusion, as described above
* `determineDatatypeRaw()`: Would then be consistent with the change above

include/openPMD/RecordComponent.hpp
franzpoeschel: Yeah, I forgot that in #1296. Should I open a separate PR or just add it into this one?

include/openPMD/RecordComponent.hpp
franzpoeschel: This is old code, just moved around. But I'll update it

include/openPMD/RecordComponent.hpp
franzpoeschel: Eh, I updated it now

include/openPMD/Datatype.hpp
ax3l: I have a slide tendency to overloads, simply because it's C++ pattern (vs. C).

ax3l: For this case, we could make it consistently evaluate to `Datatype::VECTOR_FLOAT`:
```C++
determineDatatype<std::vector<float>>();
determineDatatype(std::vector<float>());
```
and ask the user to call via `::value_type` or `.data()` to get the fundamental type aka `Datatype::FLOAT`.
```C++
determineDatatype<std::vector<float>::value_type>();
determineDatatype(std::vector<float>().data());
```

I think I overdid the original detection support here for STL :)

If this changes existing behavior, then let us note it in `NEWS.rst` pls.

include/openPMD/RecordComponent.hpp
ax3l: Just saw it with the other part in the enable if being so modern :D

include/openPMD/Datatype.hpp
franzpoeschel: I would maybe take your suggestion, but make it a little bit stricter:

If you want a direct translation between C++ types and our `openPMD::Datatype` enum, then use `determineDatatype<T>()` without argument (exactly as it is before this PR).

We provide overloads for pointer types, i.e. `determineDatatype<T>(std::shared_ptr<T>)` (already exists) and `determineDatatype<T>(T const *)` (new).

If a user passes a `std::vector` or a `std::array`, we don't even start guessing what is meant, but throw a detailed compile-time error:
```
Error: Passed a contiguous container type to determineDatatype<>().                                 [10/1843]
These types are not directly supported due to colliding semantics.                                           
Assuming a vector object `std::vector<float> vec;`,                                                          
use one of the following alternatives:                                                                       
                                                                                                             
1) If what you want is a direct openPMD::Datatype equivalent of the container                                
   type, use:                                                                                                
   `determineDatatype<decltype(vec)>()`                                                                      
   OR                                                                                                        
   `determineDatatype<std::vector<float>>()`.                                                                
   The result will be `Datatype::VECTOR_FLOAT`.                                                              
2) If what you want is the openPMD::Datatype equivalent of the *contained type*,
   use the raw pointer overload by:
   `determineDatatype(vec.data())`
   This is the variant that you likely wish to use if intending to write data
   from the vector via `storeChunk()`, e.g. `storeChunk(vec, {0}, {10})`.
```

Corresponding implementation:
```cpp
template <typename T_ContiguousContainer>
inline constexpr Datatype determineDatatype(T_ContiguousContainer &&)
{
    using T_ContiguousContainer_stripped =
        std::remove_reference_t<T_ContiguousContainer>;
    if constexpr (auxiliary::IsContiguousContainer_v<
                      T_ContiguousContainer_stripped>)
    {
        static_assert(auxiliary::dependent_false_v<T_ContiguousContainer>, R"(
Error: Passed a contiguous container type to determineDatatype<>().
These types are not directly supported due to colliding semantics.
Assuming a vector object `std::vector<float> vec;`,
use one of the following alternatives:

1) If what you want is a direct openPMD::Datatype equivalent of the container
   type, use:
   `determineDatatype<decltype(vec)>()`
   OR
   `determineDatatype<std::vector<float>>()`.
   The result will be `Datatype::VECTOR_FLOAT`.
2) If what you want is the openPMD::Datatype equivalent of the *contained type*,
   use the raw pointer overload by:
   `determineDatatype(vec.data())`
   This is the variant that you likely wish to use if intending to write data
   from the vector via `storeChunk()`, e.g. `storeChunk(vec, {0}, {10})`.
        )");
    }
    else
    {
        static_assert(auxiliary::dependent_false_v<T_ContiguousContainer>, R"(
Unknown datatype passed to determineDatatype<>().
For a direct translation from C++ type to the openPMD::Datatype enum, use:
`auto determineDatatype<T>() -> Datatype`.

For detecting the contained datatpye of a pointer type (shared or raw pointer),
use either of the following overloads:
`auto determineDatatype<T>(std::shared_ptr<T>) -> Datatype` or
`auto determineDatatype<T>(T const *) -> Datatype`.
        )")
    }
}
```

This would entirely avoid any API inconsistencies, without introducing unexpected behavior where we need to guess what the user wanted.

examples/3_write_serial.cpp
ax3l: Maybe we can keep it simple by documenting to use:
```C++
determineDatatype<decltype(global_data)::value_type>()
```

ax3l (approved):

PR #1231, opened 2022-03-16
added  support in 8a to use encodings

guj:


examples/8a_benchmark_write_parallel.cpp
```diff
@@ -393,6 +398,15 @@ void parse(TestInput &input, std::string line)
         return;
     }
 
+    if (vec[0].compare("encoding") == 0)
+    {
+        if (vec[1].compare("f") == 0)
+            input.m_Encoding = openPMD::IterationEncoding::fileBased;
+        else if (vec[1].compare("g") == 0)
```
ax3l: A bit safer due to runtime checks:
```suggestion
    if (vec.at(0).compare("encoding") == 0)
    {
        if (vec.at(1).compare("f") == 0)
            input.m_Encoding = openPMD::IterationEncoding::fileBased;
        else if (vec.at(1).compare("g") == 0)
```

ax3l (approved):

examples/8a_benchmark_write_parallel.cpp
```diff
@@ -1131,6 +1144,19 @@ unsigned int AbstractPattern::getNumBlocks()
...
+inline openPMD::Iteration
+AbstractPattern::GetIteration(Series &series, int const iteration) const
+{
+    if (m_Input.m_Encoding == openPMD::IterationEncoding::variableBased)
+    {
+        return series.writeIterations()[iteration];
+    }
+    else
+    {
+        return series.iterations[iteration];
```
franzpoeschel: Is there any reason to not use `writeIterations()` here? It should also work in group-based and file-based encoding

examples/8a_benchmark_write_parallel.cpp
```diff
@@ -627,42 +642,41 @@ void AbstractPattern::run()
...
                 Series series =
                     Series(filename, Access::CREATE, MPI_COMM_WORLD);
+                series.setIterationEncoding(m_Input.m_Encoding);
                 series.setMeshesPath("fields");
                 store(series, step);
             }
         }
     }
 
-#ifdef NEVER // runs into error for ADIOS. so temporarily disabled
```
franzpoeschel, line 644: Was this a bug? Is it resolved now?

franzpoeschel (commented):
Note that the JSON schema on the dev branch now also has options to pick the encoding:
```json
{"iteration_encoding": "variable_based"}
```
https://openpmd-api.readthedocs.io/en/latest/details/backendconfig.html#backend-independent-json-configuration
So this might not be strictly necessary. Otherwise, we can go forward with this, I just have two small questions.

examples/8a_benchmark_write_parallel.cpp
guj: It was an very old comment, and was resolved.

examples/8a_benchmark_write_parallel.cpp
guj: The file/group based  were not using begin/end steps that is why they were not using writeIterations.

As for json, it is convenient to put all the parameters in one input file. That's why I added these options.

franzpoeschel: > The file/group based were not using begin/end steps that is why they were not using writeIterations.

`readIterations()` should work even if you did not use `writeIterations()`, see [this test](https://github.com/openPMD/openPMD-api/blob/5a65f6df28a1f510d0ae242293d2d410aa32cda5/test/SerialIOTest.cpp#L5029). If it does not, then it's a bug.

> 
> As for json, it is convenient to put all the parameters in one input file. That's why I added these options.

That's reasonable yup

guj: Yes, readIterations() works.

ax3l: ok, then let's add it here, because then read works also with data generated with steps/streaming?

franzpoeschel: Can we remove the `series.iterations` branch then? Since this file is one of the openPMD-api examples, I'm reluctant to advertize that users need to switch between both APIs.

guj: > ok, then let's add it here, because then read works also with data generated with steps/streaming?

sorry, add what where?  yes read works with variable based files.

guj: > Can we remove the `series.iterations` branch then? Since this file is one of the openPMD-api examples, I'm reluctant to advertize that users need to switch between both APIs.

OK, let me see to it.

guj: > `readIterations()` should work even if you did not use `writeIterations()`, see [this test](https://github.com/openPMD/openPMD-api/blob/5a65f6df28a1f510d0ae242293d2d410aa32cda5/test/SerialIOTest.cpp#L5029). If it does not, then it's a bug.

If I just use writeIterations for all encodings, it does not work.

guj: I added code and make sure  BP5 must be associated with variable based iteration. Now can safely use  writeIteration() only.

examples/8a_benchmark_write_parallel.cpp
```diff
@@ -1131,6 +1150,13 @@ unsigned int AbstractPattern::getNumBlocks()
     return m_InRankMeshLayout.size();
 }
 
+inline openPMD::Iteration
+AbstractPattern::GetIteration(Series &series, int const iteration) const
```
ax3l: Do you still need that helper function now?

examples/8a_benchmark_write_parallel.cpp
ax3l: Ok, as a last step - do you still need `AbstractPattern::GetIteration` for the one-liner?

examples/8a_benchmark_write_parallel.cpp
guj: It won't hurt to leave it there

examples/8a_benchmark_write_parallel.cpp
guj: I think can leave it there. Might be flexible for later usage.

ax3l: we really dont need it anymore and this is an example, where we want to encourage best current usage to users.

we can add helpers when we really need them

examples/8a_benchmark_write_parallel.cpp
ax3l: It's unnecessary verbose and might confuse users. This is an example, after all

examples/8a_benchmark_write_parallel.cpp
guj: That's fine.

franzpoeschel: So the bug was only BP5 related? If yes, then it should be fixed with #1215, and I'd prefer that fix in favor of a workaround here

franzpoeschel (approved):

examples/8a_benchmark_write_parallel.cpp
guj: > So the bug was only BP5 related? If yes, then it should be fixed with #1215, and I'd prefer that fix in favor of a workaround here

yes, BP5 enforces begin/end steps()

guj: Can we merge this now?

franzpoeschel: I think so

franzpoeschel:
Pushed the last commit again to give the CI another chance. Will merge once it runs clean.

PR #1235, opened 2022-03-22
Add memory leak suppression: ps_make_timer_name_

franzpoeschel:
Noticed in my offline Clang Sanitizer builds. Can theoretically also be fixed by the following: 

```
leak:adios2::format::BP3Serializer::AggregateCollectiveMetadataIndices
leak:adios2::core::engine::BP5Writer::MarshalAttributes
```
But I guess, putting the offending function explicitly is better.

Might affect our CI once we upgrade to ADIOS 2.8.0.

### DIFF
--- a/.github/ci/sanitizer/clang/Leak.supp
+++ b/.github/ci/sanitizer/clang/Leak.supp
@@ -15,3 +15,4 @@ leak:adios_inq_var
 # ADIOS2
 leak:adios2::core::engine::SstReader::*
 leak:adios2::core::engine::SstWriter::*
+leak:ps_make_timer_name_


ax3l (approved):
Thanks!

Interesting, is that something used transitively in ADIOS2? Which lib ships that symbol?

franzpoeschel:
> Thanks!
> 
> Interesting, is that something used transitively in ADIOS2? Which lib ships that symbol?

The two places where the leak occurred for me were inside the two functions mentioned above, so during metadata collection. I guess that ADIOS2 uses this transitively for time tracing.

ax3l:
Thanks. Seems to be from `perfstubs`:
https://github.com/ornladios/ADIOS2/blob/v2.8.0/thirdparty/perfstubs/perfstubs/perfstubs_api/timer.h#L85
which is afaik a generic tracing interface.

PR #1237, opened 2022-03-25
Parsing logic: fail gracefully on unexpected input

franzpoeschel:
Reading failures should ideally affect only the innermost scope possible. When one record component fails parsing due to whatever error, the neighboring record components should remain readable. When one iteration fails parsing, skip it and go on to the next.
Current behavior: Parsing fails and the Series cannot be opened

Close #983 

TODO
- [x] Introduce `ReadError` error type that backends can throw to tell the frontend about recoverable failures
- [x] Throw `ReadError` inside ADIOS1 and ADIOS2 attribute reading routines
- [x] Frontend: Recover from `ReadError`s ~~(partially implemented so far)~~
- [x] ~~Other backends,~~ also consider dataset reading, ... UPDATE: Catching wrong attribute reads now possible in all backends, dataset reading and further things can come in a later update
- [x] Testing: ~~file-based, group-based, varialbe-based, deferred/eager~~, opening single iterations of a file-based series, ~~streaming API, skipping iteration~~
- [x] Merge #949 first to avoid merge conflicts
- [x] Correctly recover from deferred actions in ADIOS2
- [x] ADIOS2 / variable-based encoding: Backend reports wrong value for snapshot in iteration 10
- [x] ~~bp4_steps test seems to not write snapshot attribute~~ only written in new ADIOS2 schema as it should be
- [x] Avoid const_cast by using the parsing state

This whole PR is a bit of a stretch-goal as there are many different types and sources of errors to be recovered from.
@ax3l Can you name some specific error pattern (or even give me a test file) that the openPMD-api should handle more robustly?

I think that a backend failure should mean that the entire call to `AbstractIOHandler::flush()` has failed and the queue is cleared. For finer granularity error handling, the frontend must issue single `flush()` calls. So, I've changed error recovery procedures to clear the entire IO queue upon failure.

PR #1239, opened 2022-04-01
Fixes for opening an iteration

franzpoeschel:
Commit isolated from #1237

1. When opening a new step in the Streaming API, don't flush the Series before doing that. It seems that flushing at this stage seems to confuse HDF5 and lead to rare deadlocks in parallel contexts, seen in PIConGPU. This PR is able to fix that behavior in the observed instance.
2. When reading a variable-based Series, make sure to open an IO step before doing anything. Otherwise, attributes from later iterations might leak into the early iteration (RandomAccess mode in BP4).

TODO:
- [x] Testing

### DIFF
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -84,7 +84,11 @@ enum class FlushLevel : unsigned char
      * CREATE_DATASET tasks.
      * Attributes may or may not be flushed yet.
      */
-    SkeletonOnly
+    SkeletonOnly,
+    /**
+     * Only creates/opens files, nothing more
+     */
+    CreateOrOpenFiles
 };
 
 namespace internal
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -61,11 +61,9 @@ public:
  * @brief Determine what items should be flushed upon Series::flush()
  *
  */
-#ifdef __NVCOMPILER_MAJOR__
+// do not write `enum class FlushLevel : unsigned char` here since NVHPC
+// does not compile it correctly
 enum class FlushLevel
-#else
-enum class FlushLevel : unsigned char
-#endif
 {
     /**
      * Flush operation that was triggered by user code.
--- a/include/openPMD/IO/AbstractIOHandler.hpp
+++ b/include/openPMD/IO/AbstractIOHandler.hpp
@@ -61,7 +61,11 @@ public:
  * @brief Determine what items should be flushed upon Series::flush()
  *
  */
+#ifdef __NVCOMPILER_MAJOR__
+enum class FlushLevel
+#else
 enum class FlushLevel : unsigned char
+#endif
 {
     /**
      * Flush operation that was triggered by user code.
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -72,6 +72,7 @@ namespace internal
          * containing this iteration.
          */
         std::string filename;
+        bool beginStep = false;
     };
 
     class IterationData : public AttributableData
@@ -249,14 +250,16 @@ private:
     void flush(internal::FlushParams const &);
     void deferParseAccess(internal::DeferredParseAccess);
     /*
-     * Control flow for read(), readFileBased(), readGroupBased() and
-     * read_impl():
-     * read() is called as the entry point. File-based and group-based
+     * Control flow for runDeferredParseAccess(), readFileBased(),
+     * readGroupBased() and read_impl():
+     * runDeferredParseAccess() is called as the entry point.
+     * File-based and group-based
      * iteration layouts need to be parsed slightly differently:
      * In file-based iteration layout, each iteration's file also contains
      * attributes for the /data group. In group-based layout, those have
      * already been parsed during opening of the Series.
-     * Hence, read() will call either readFileBased() or readGroupBased() to
+     * Hence, runDeferredParseAccess() will call either readFileBased() or
+     * readGroupBased() to
      * allow for those different control flows.
      * Finally, read_impl() is called which contains the common parsing
      * logic for an iteration.
@@ -265,10 +268,10 @@ private:
      * Calling it on an Iteration not yet parsed is an error.
      *
      */
-    void read();
     void reread(std::string const &path);
-    void readFileBased(std::string filePath, std::string const &groupPath);
-    void readGorVBased(std::string const &groupPath);
+    void readFileBased(
+        std::string filePath, std::string const &groupPath, bool beginStep);
+    void readGorVBased(std::string const &groupPath, bool beginStep);
     void read_impl(std::string const &groupPath);
 
     /**
@@ -278,7 +281,7 @@ private:
      *
      * @return AdvanceStatus
      */
-    AdvanceStatus beginStep();
+    AdvanceStatus beginStep(bool reread);
 
     /**
      * @brief End an IO step on the IO file (or file-like object)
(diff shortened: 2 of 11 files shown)


ax3l:
restarting CI

ax3l:
It looks like NVHPC compilers (formerly PGI) fail on the `SerialIOTest` for 2 test right now.

franzpoeschel:
> It looks like NVHPC compilers (formerly PGI) fail on the `SerialIOTest` for 2 test right now.

I can reproduce this locally, will have a look

franzpoeschel:
The NVIDIA compilers have a lot of test failures for me locally..

ax3l:
discussed today: likely a glitch that the read toggles into write mode at some point.
will be addressed with a separate PR.

franzpoeschel:
When defining:
```cpp
namespace std
{
inline string to_string(openPMD::FlushLevel const &level)
{
    switch (level)
    {
    case openPMD::FlushLevel::UserFlush:
        return "UserFlush";
    case openPMD::FlushLevel::InternalFlush:
        return "InternalFlush";
    case openPMD::FlushLevel::SkeletonOnly:
        return "SkeletonOnly";
    case openPMD::FlushLevel::CreateOrOpenFiles:
        return "CreateOrOpenFiles";
    }
}
} // namespace std
```

and adding to `Iteration::flushGroupBased` the lines
```
    std::cout << "Flushlevel is: " << std::endl;
    std::cout << std::to_string(flushParams.flushLevel) << std::endl;
```
the result is this

```
Flushlevel is: 
 _ xPpphh!@_!_pp؅ptime
                      P
                       `\pdte
                             !?
                               "p
imeUnitSI

         ?
          1usage informationqPI vectorЏ`ofstringsÝ =p+вBpвBp0P-l
                                                                --list-tests!"Bp-tp
                                                                                   --list-tags"Bp0P0-s  --success1"Bp0P-bp--break"Bp P0-e   --nothrowA"Bp(tabs, newlines)0P-i`
                                          --invisibles"Bp-o--outq"Bp(defaults to con0Pp-r0 
--reporter"Bp-n --name"Bp P@ -a --abort1"Bp P -x`--abortx"Bp-w!--warnq"Bp Pp!-d "
                                                                                 --durations`t durations for tests taking at least the given number of seconds%`"↑↑└␋┼↑␍┤⎼▒├␋⎺┼1◆B⎻▮P"↑°P#
                                                      ↑↑␋┼⎻┤├↑°␋┌␊◆B⎻▮P#↑##1◆B⎻⎻$B⎻@P#↑␌◆$      ↑↑⎽␊␌├␋⎺┼◆B⎻$┤├ ┴␊⎼␉⎺@P $↑┴$
                                                                                                                            ↑↑┴␊⎼␉⎺⎽
␋├≤D▮▮◆#◆B⎻⎽↑⎺┼┌≤!%@0%`Bp 0 &e order (defaults to decP0%--orderP@ Fecific seed for random numbers@`BPP˞P˞pPd@dH'``˞˞˞pP@'˞̞̞pЇ8( ̞""P̞̞̞pВ(̞͞͞p() ͞@͞͞͞p0 )͞  ͞ΞΞpp`filename8*output filenameΞPΞPΞp0 *name`Ξ%%Ξ+name(+
suite name Ϟ ϞpƝ+0ϞPϞϞϞp ĝĝ+
                            no. failuresϞϞООpĝp,
                                                warning name,enable warningsОPОPОp,yes|no`ООООpp`-secondsII0ўpўpўp`Pfilenameў""ўpP.ҞҞpҞpҞp.p..
L3pppPd@d3p@pP@p4pЇ4""0pppВ`5p5 ``p0 P6p  pp`filename6output filename00p0 @7name @%%pp7nameP7
suite namepƝ08880ppp ĝĝ8
                        no. failurespĝ 9
                                        warning name@9enable warnings00p9yes|no1@`pp:seconds1II`Pfilename2`;3@3@`A`Ap.p.x;
                                                                                                                          section na'time'|numberC&&CDDpPm@mH>yes|no DAborted (core dumped)
```
:D

I'll try passing by value instead of reference

franzpoeschel:
I think that the last push should fix it. Looks like a compiler issue to me, but it works now.

However, I noticed that builds with MPI or ADIOS2 have lots of failure with NVC++, we should maybe try to extend our tests there and see if it can be fixed. The thing seems to do lots of very funny optimization that may or may not result in correct binaries.

franzpoeschel:
NVHPC is now running fine again :)

ax3l:
> Looks like a compiler issue to me, but it works now.

Actually, in C++, adding anything in the `namespace std` that is not a (class) template specialization is _undefined behavior_, defined by the standard. (re: https://github.com/openPMD/openPMD-api/pull/1239#issuecomment-1090144507)

To fix the UB in the code, rely on [ADL](https://en.cppreference.com/w/cpp/language/adl) and define the `to_string` overload in the `openPMD` namespace:
- https://stackoverflow.com/a/50555639/2719194
- https://en.cppreference.com/w/cpp/language/extending_std
> It is undefined behavior to add declarations or definitions to namespace std or to any namespace nested within std, with a few exceptions noted below. [...]

franzpoeschel:
> > Looks like a compiler issue to me, but it works now.
> 
> Actually, in C++, adding anything in the `namespace std` that is not a (class) template specialization is _undefined behavior_, defined by the standard. (re: [#1239 (comment)](https://github.com/openPMD/openPMD-api/pull/1239#issuecomment-1090144507))
> 
> To fix the UB in the code, rely on [ADL](https://en.cppreference.com/w/cpp/language/adl) and define the `to_string` overload in the `openPMD` namespace:
> 
>     * https://stackoverflow.com/a/50555639/2719194
> 
>     * https://en.cppreference.com/w/cpp/language/extending_std
> 
> 
> > It is undefined behavior to add declarations or definitions to namespace std or to any namespace nested within std, with a few exceptions noted below. [...]

Ah yeah, true

On the other hand, it was just some temporary debugging output and the memory issue that I saw came from somewhere else

franzpoeschel:
> I think that the last push should fix it. Looks like a compiler issue to me, but it works now.
> 
> However, I noticed that builds with MPI or ADIOS2 have lots of failure with NVC++, we should maybe try to extend our tests there and see if it can be fixed. The thing seems to do lots of very funny optimization that may or may not result in correct binaries.

The Python bindings don't compile at all

ax3l:
@franzpoeschel I think `clang-tidy` fails on this one?

ax3l:
restarting CI

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -61,7 +61,11 @@ class unsupported_data_error : public std::runtime_error
  * @brief Determine what items should be flushed upon Series::flush()
  *
  */
+#ifdef __NVCOMPILER_MAJOR__
```
ax3l, line 64: Is it worth adding this work-around instead of generally using `enum class` without a type?

include/openPMD/IO/AbstractIOHandler.hpp
```diff
@@ -61,7 +61,11 @@ class unsupported_data_error : public std::runtime_error
...
+enum class FlushLevel
+#else
 enum class FlushLevel : unsigned char
+#endif
```
ax3l: Is it worth adding this work-around instead of generally using `enum class` without a type?
```suggestion
enum class FlushLevel
```

ax3l:
@franzpoeschel one tiny suggestion inline for your consideration

include/openPMD/IO/AbstractIOHandler.hpp
franzpoeschel: We can also do that, but we should add a comment why we do it
```suggestion
// do not write `enum class FlushLevel : unsigned char` here since NVHPC
// does not compile it correctly
enum class FlushLevel
```

ax3l:
restarting CI since I merged a couple PRs in between

ax3l (approved):

ax3l:
Similar to https://github.com/openPMD/openPMD-api/pull/1253#issuecomment-1137531794 - tricky to bkacport to `release-0.14.5`

ax3l:
Backported via https://github.com/ax3l/openPMD-api/pull/5 to #1282 `0.14.5`

PR #1240, opened 2022-04-01
use CUDA ptrs in 8a when it is enabled

guj:
Added option in cmake "openPMD_USE_CUDA" to detect and compile with Cuda libraries.

8a allocates from GPU if CUDA is a supported option.


CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ openpmd_option(HDF5           "HDF5 backend (.h5 files)"                  AUTO)
 openpmd_option(ADIOS1         "ADIOS1 backend (.bp files)"                AUTO)
 openpmd_option(ADIOS2         "ADIOS2 backend (.bp files)"                AUTO)
 openpmd_option(PYTHON         "Enable Python bindings"                    AUTO)
+openpmd_option(CUDA           "Use CUDA"                                  AUTO)
```
ax3l: This is great, but only needed for tests/downstream (does not influence the API's feature set).

Let us name this `CUDA_TESTS` and set it to `OFF` by default, to avoid complications of building the API defaults (e.g., if CUDA is found but its not working properly). We can set it to ON explicitly for the workflows that we run on GPUs.

CMakeLists.txt
```diff
@@ -180,6 +181,15 @@ if(openPMD_MPI_LINK_C)
     set(openPMD_MPI_TARGETS MPI::MPI_C ${openPMD_MPI_TARGETS})
 endif()
 
+if(openPMD_USE_CUDA STREQUAL AUTO)
```
ax3l: Let's wrap this into `if(openPMD_BUILD_EXAMPLES)`, that's the only time this is needed so far and thus will simplify build requirements.

CMakeLists.txt
```diff
@@ -869,6 +879,11 @@ if(openPMD_BUILD_EXAMPLES)
         if(${examplename} MATCHES ".+parallel$")
             if(openPMD_HAVE_MPI)
                 add_executable(${examplename} examples/${examplename}.cpp)
+                if (openPMD_HAVE_CUDA)
+                  enable_language(CUDA)
+                  target_link_libraries(${examplename} PUBLIC CUDA::cudart CUDA::cuda_driver)
+                  set_target_properties(${examplename} PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
```
ax3l: A few more things that we need:
```suggestion
                  set_source_files_properties(examples/${examplename}.cpp PROPERTIES
                    LANGUAGE CUDA )
                  target_compile_features(${examplename} PUBLIC cuda_std_17)
                  set_target_properties(${examplename} PROPERTIES
                    CUDA_SEPARABLE_COMPILATION ON
                    CUDA_EXTENSIONS OFF
                    CUDA_STANDARD_REQUIRED ON
                  )
```

and please but the `openpmd_cxx_required` in an `else` branch.

Also: is `CUDA_SEPARABLE_COMPILATION` really needed (e.g., does ADIOS2 already build with it and can you try if it works without it)?
I am asking, because separable compilation can lead to less efficient device code and is harder on the compiler toolchain (read as: often buggy).
In case separable compilation does not yet work, revisit if you use global (device) variables in your example and try to refactor them out.

ax3l (commented):
Awesome idea to test this ADIOS 2.8.0 capability!

I added a first review :)

CMakeLists.txt
guj: OK, changed

CMakeLists.txt
guj: Changed.

CMakeLists.txt
guj: I got linking errors when CUDA_SEPARABLE_COMPILATION is off, like this: 
/tmp/tmpxft_00159479_00000000-6_8a_benchmark_write_parallel.cudafe1.stub.c:14: undefined reference to `__cudaRegisterLinkedBinary_62_tmpxft_00159479_00000000_7_8a_benchmark_write_parallel_cpp1_ii_d7f2f159'

guj: other suggestions are incorporated in the latest commit

CMakeLists.txt
```diff
@@ -865,12 +866,31 @@ if(openPMD_BUILD_CLI_TOOLS)
...
                 add_executable(${examplename} examples/${examplename}.cpp)
-                openpmd_cxx_required(${examplename})
-                target_link_libraries(${examplename} PRIVATE openPMD)
+                if (openPMD_HAVE_CUDA_TESTS)
+                   enable_language(CUDA)
+                   set_source_files_properties(examples/${examplename}.cpp PROPERTIES LANGUAGE CUDA )
+                   target_compile_features(${examplename} PUBLIC cuda_std_17)
+                   set_target_properties(${examplename} PROPERTIES CUDA_SEPARABLE_COMPILATION ON
+                                                        CUDA_EXTENSIONS OFF
+                                                        CUDA_STANDARD_REQUIRED ON)
```
```suggestion
                   set_source_files_properties(examples/${examplename}.cpp PROPERTIES
                       LANGUAGE CUDA )
                   target_compile_features(${examplename} PUBLIC cuda_std_17)
                   set_target_properties(${examplename} PROPERTIES
                       CUDA_SEPARABLE_COMPILATION ON
                       CUDA_EXTENSIONS OFF
                       CUDA_STANDARD_REQUIRED ON)
```

CMakeLists.txt
```diff
@@ -84,6 +84,7 @@ openpmd_option(HDF5           "HDF5 backend (.h5 files)"                  AUTO)
 openpmd_option(ADIOS1         "ADIOS1 backend (.bp files)"                AUTO)
 openpmd_option(ADIOS2         "ADIOS2 backend (.bp files)"                AUTO)
 openpmd_option(PYTHON         "Enable Python bindings"                    AUTO)
+openpmd_option(CUDA_TESTS     "Use CUDA"                                  OFF)
```
```suggestion
openpmd_option(CUDA_TESTS     "Use CUDA in tests"                         OFF)
```

ax3l (approved):

PR #1159, opened 2021-12-17
Follow-up to redesign of frontend classes: Documentation and cleanup

franzpoeschel:
See the comments [here](https://github.com/openPMD/openPMD-api/pull/1115#pullrequestreview-833507585).

PR #1253, opened 2022-04-21
Fix Zero Pattern Issue

### LINKED ISSUE #1173: Pattern Problem Returns in 0.14.3
@camille12225 reports that even with [our pattern fix](https://github.com/openPMD/openPMD-api/pull/1118) ported into 0.14.3, she still sees issues with inconsistent padding in files.

Here iterations are of the form:
```
openPMD_00000.bp
...
openPMD_08000.bp
openPMD_09000.bp
openPMD_10000.bp
openPMD_11000.bp
...
openPMD_98000.bp
openPMD_99000.bp
openPMD_110000.bp
openPMD_111000.bp
openPMD_112000.bp
```
and she needs to remove the files before 10,000 to read the series without failing:
```
diags/diag1/openpmd_8000.bp/md.idx, in call to POSIX open: errno = 2: No such file or directory: iostream error: iostream error
```
(the file on disk is called `diags/diag1/openpmd_08000.bp/md.idx`)

Interesting enough, no problem around 100,000 now.

The data is read with openPMD-viewer, using the general `%T` wildcard:
https://github.com/openPMD/openPMD-viewer/blob/1.2.0/openpmd_viewer/openpmd_timeseries/data_reader/data_reader.py#L99-L102

ax3l:
Follow-up to #1118. The pattern issue returned :scream: 

- [x] add demonstrator: `[JSON] Failed opening a file '../samples/varying_pattern_0.json'`
- [x] fix #1173

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -5659,3 +5659,47 @@ TEST_CASE("late_setting_of_iterationencoding", "[serial]")
     REQUIRE(
         auxiliary::file_exists("../samples/change_name_and_encoding_10.json"));
 }
+
+void varying_pattern(std::string const file_ending)
+{
+    {
+        std::string filename = "../samples/varying_pattern_%06T." + file_ending;
+        ::openPMD::Series series =
+            ::openPMD::Series(filename, ::openPMD::Access::CREATE);
+
+        for (auto i : {0, 8000, 10000, 100000, 2000000})
+        {
+            auto it = series.iterations[i];
+            it.setAttribute("my_step", i);
+        }
+        series.flush();
+    }
+    {
+        std::string filename = "../samples/varying_pattern_%T." + file_ending;
+        ::openPMD::Series series =
+            ::openPMD::Series(filename, ::openPMD::Access::READ_ONLY);
+
+        REQUIRE(series.iterations.size() == 5);
+        for (auto const [step, it] : series.iterations)
+        {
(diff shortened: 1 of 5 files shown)


franzpoeschel:
Inconsistent iteration paddings are correctly read, but they are currently not stored after first parsing.
When accessing an iteration after parsing, it needs to be opened again, using its filename. Currently, that filename is then reconstructed on the fly by using the iteration padding which is wrong.
I've added a commit that stores the filenames and uses the stored ones in favor of reconstructed ones.

test/SerialIOTest.cpp
```diff
@@ -5658,3 +5658,47 @@ TEST_CASE("late_setting_of_iterationencoding", "[serial]")
...
+        }
+        series.flush();
+    }
+    {
+        std::string filename = "../samples/varying_pattern_%T." + file_ending;
+        ::openPMD::Series series =
+            ::openPMD::Series(filename, ::openPMD::Access::READ_ONLY);
+
+        REQUIRE(series.iterations.size() == 5);
+        for (auto const [step, it] : series.iterations)
```
```suggestion
        for (auto const & [step, it] : series.iterations)
```

ax3l (commented):
Thank you so much, that's awesome :)

ax3l:
To apply this fix (especially the third commit) on the `release-0.14.5` branch, we problably also need to backport first:
- #1115
- #1159

ax3l:
Ok I think I managed to backport this in #1282 now :tada:

PR #1259, opened 2022-04-29
Revert SerialIOTest: Fix GCC Pragma Check

ax3l:
Revert #1255, which broke icc.


franzpoeschel:
Do I understand it correctly that icc actually defines `__GNUC_MINOR__` :eyes:

ax3l:
Yeah, that was... unexpected :laughing:

PR #1260, opened 2022-04-29
SerialIOTest: Fix GCC Pragma Check

ax3l:
There was a typo in `__GNUC_MINOR__`, causing the test not to work on all GCC versions.

Follow-up to #1213

Redo of #1255


ax3l:

PR #1262, opened 2022-04-29
BP5 tests: Only run them if ADIOS2 has BP5

### LINKED ISSUE #1257: ADIOS BP5 Support: Conditional Compile Variant
Unfortunately, we cannot rely on ADIOS 2.8.0+ to provide BP5 unconditionally.

https://github.com/ornladios/ADIOS2/blob/v2.8.0/CMakeLists.txt#L155

We need to do more checks if that compile option was actually set.

X-ref: https://github.com/ornladios/ADIOS2/issues/3182

franzpoeschel:
Fix #1257

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -88,7 +88,7 @@ install:
   # Configure the VM.
   - cmd: conda install -n root --quiet --yes numpy cmake hdf5 python=%CONDA_PY%
   # ADIOS2 build only for 64bit Windows
-  - cmd: if "%TARGET_ARCH%"=="x64" conda install -n root --quiet --yes adios2==2.7.1 python=%CONDA_PY%
+  - cmd: if "%TARGET_ARCH%"=="x64" conda install -n root --quiet --yes adios2 python=%CONDA_PY%
 
 before_build:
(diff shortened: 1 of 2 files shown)


ax3l:
Thanks a lot!

Do you also want to update the "bp5" locations in `src/IO/ADIOS/ADIOS2IOHandler.cpp`?

I also looked at `src/config.cpp` for `openPMD::getFileExtensions()` - but there we only list `bp` right now.

franzpoeschel:
We don't currently check for available engines in the ADIOS2 backend at all (on Windows, there's no SST, but we don't check for that)
Maybe we should do it? But if so, then that should maybe go into a different PR. Currently, there will be an ADIOS2 error if selecting an unknown engine (e.g. "This version didnt compile with…" or "Unknown engine type …")



> I also looked at `src/config.cpp` for `openPMD::getFileExtensions()` - but there we only list `bp` right now.

Good point, this is a Todo for #1218

ax3l (approved):
Sounds good to me, thank you :)

ax3l:
As discussed: BP5 patches will first land in 0.15.0

PR #1264, opened 2022-05-02
Fix premature flushing (performance bug)

### LINKED ISSUE #1263: Performance bug: Premature IO flushing when using scalar components
See [this line](https://github.com/openPMD/openPMD-api/blob/e5bc793cd2a701bfb3258231629af85187d73b1b/src/Mesh.cpp#L233), similarly in `Record.cpp`.

In some backends, there are performance benefits from avoiding flushes before closing an iteration:

* BP4 engine of ADIOS2: The internal ADIOS2 buffer will be allocated to the correct size upon `Engine::EndStep()`. If flushing before that, it needs to be allocated before ending the step and might need to be resized later.
* BP5 engine of ADIOS2: Flushing in BP5 might cause unnecessary copies into internal buffers. If using deferred Put() and EndStep() only, BP5 can avoid copies entirely.

Calling `IOHandler()->flush()` at those place ruins this and should be replaced with something more fine-grained. 
It's currently essentially impossible to avoid these unnecessary copies into BP5 buffers while using scalar components.

franzpoeschel:
Fix #1263 

TODO:

- [x] Fix premature flushes in context of scalar components
- [x] Document the new IO task

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -469,6 +469,7 @@ set(CORE_SOURCE
         src/benchmark/mpi/OneDimensionalBlockSlicer.cpp
         src/helper/list_series.cpp)
 set(IO_SOURCE
+        src/IO/AbstractIOHandlerImpl.cpp
         src/IO/AbstractIOHandlerHelper.cpp
         src/IO/DummyIOHandler.cpp
         src/IO/IOTask.cpp
--- a/docs/source/dev/design.rst
+++ b/docs/source/dev/design.rst
@@ -23,7 +23,7 @@ Therefore, enabling users to handle hierarchical, self-describing file formats w
 
 .. literalinclude:: IOTask.hpp
    :language: cpp
-   :lines: 44-60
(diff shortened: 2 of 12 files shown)


ax3l:
ready to rebase :)

ax3l (approved):

franzpoeschel:
This PR introduced a subtle bug: In scalar constant record components, some superfluous attributes are created

```
  uint64_t  /data/1/particles/e/positionOffset/positionOffset/shape  attr
  int32_t   /data/1/particles/e/positionOffset/positionOffset/value  attr
  uint64_t  /data/1/particles/e/positionOffset/shape                 attr
  float     /data/1/particles/e/positionOffset/timeOffset            attr
  double    /data/1/particles/e/positionOffset/unitDimension         attr
  double    /data/1/particles/e/positionOffset/unitSI                attr
  int32_t   /data/1/particles/e/positionOffset/value                 attr
```

PR #1268, opened 2022-05-04
Avoid copying std::string in for loop

### LINKED ISSUE #1266: GCC 11.2 Warning About String in IOTask.cpp
GCC 11.2.0 on Perlmutter for the current `dev`:
```
/global/homes/a/ahuebl/src/openPMD-api/src/IO/IOTask.cpp: In static member function 'static void openPMD::Parameter<openPMD::Operation::CREATE_DATASET>::warnUnusedParameters(TracingJSON&, const string&, const string&) [with TracingJSON = openPMD::json::TracingJSON; std::string = std::__cxx11::basic_string<char>]':
/global/homes/a/ahuebl/src/openPMD-api/src/IO/IOTask.cpp:45:29: warning: loop variable 'key' of type 'const string&' {aka 'const std::__cxx11::basic_string<char>&'} binds to a temporary constructed from type 'const char* const' [-Wrange-loop-construct]
   45 |     for (std::string const &key : {"resizable"})
      |                             ^~~
/global/homes/a/ahuebl/src/openPMD-api/src/IO/IOTask.cpp:45:29: note: use non-reference type 'const string' {aka 'const std::__cxx11::basic_string<char>'} to make the copy explicit or 'const char* const&' to prevent copying
```

Is that a legit warning, I wonder?

franzpoeschel:
Fix #1266

### DIFF
--- a/src/IO/IOTask.cpp
+++ b/src/IO/IOTask.cpp
@@ -42,7 +42,7 @@ void Parameter<Operation::CREATE_DATASET>::warnUnusedParameters<
      * Fake-read non-backend-specific options. Some backends don't read those
      * and we don't want to have warnings for them.
      */
-    for (std::string const &key : {"resizable"})
+    for (char const *key : {"resizable"})
     {
         config[key];
     }


ax3l (approved):

ax3l:
Does not apply to `release-0.14.5`?

ax3l:

PR #1272, opened 2022-05-10
CI: Expand Read-Only Permission Tests

ax3l:
Expand to:
- thetaMode h5 files
- issue sample h5 files
- ADIOS bp files


franzpoeschel:
As discussed in VC: this is a good idea to automatically have this in every Unix CI run as well as in local dev setups

franzpoeschel (approved):

PR #1275, opened 2022-05-16
Distinguish all char types in ADIOS2 backend 

### LINKED ISSUE #1273: BP5 char datatype
BP5 now distinguishes between `char`, `unsigned char` and `signed char`, the ADIOS2 backend must treat all three datatypes as distinct

franzpoeschel:
ADIOS2 now distinguishes between `char`, `signed char` and `unsigned char`. Especially BP5 is unforgiving about this.

Until now, the ADIOS2 backend uses the public `openPMD::Datatype` enum, which has already caused some issues in the past. This PR introduces a new internal enum `ADIOS2Datatype` which is used to fully support ADIOS2 datatypes, especially char types. It is purely internal so we can change it as we please without breaking the API.

Close #1273

Tests will mainly come in #1218 where BP5 will be a default backend for tests

### DIFF
--- a/include/openPMD/Datatype_internal.hpp
+++ b/include/openPMD/Datatype_internal.hpp
@@ -1,3 +1,23 @@
+/* Copyright 2022 Franz Poeschel, Axel Huebl
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * openPMD-api is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
 #pragma once
 
 #include "openPMD/auxiliary/TypeTraits.hpp"
--- a/include/openPMD/Datatype_internal.hpp
+++ b/include/openPMD/Datatype_internal.hpp
@@ -32,7 +32,20 @@ namespace detail
             {
                 return DoDetermineDatatype::template call<T>();
             }
+#if defined(__INTEL_COMPILER)
+/*
+ * ICPC has trouble with if constexpr, thinking that return statements are
+ * missing afterwards. Deactivate the warning.
+ * Note that putting a statement here will not help to fix this since it will
+ * then complain about unreachable code.
+ * https://community.intel.com/t5/Intel-C-Compiler/quot-if-constexpr-quot-and-quot-missing-return-statement-quot-in/td-p/1154551
+ */
+#pragma warning(disable : 1011)
         }
+#pragma warning(default : 1011)
+#else
+        }
+#endif
 
         constexpr static char const *errorMsg =
             "basicDatatype: received unknown datatype.";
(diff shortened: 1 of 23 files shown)


include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
```diff
@@ -36,6 +36,219 @@
...
+    VEC_USHORT,
+    VEC_UINT,
+    VEC_ULONG,
+    VEC_ULONGLONG,
+    VEC_FLOAT,
+    VEC_DOUBLE,
+    VEC_LONG_DOUBLE,
+    VEC_CFLOAT,
+    VEC_CDOUBLE,
+    VEC_CLONG_DOUBLE,
```
franzpoeschel: Maybe remove `VEC_CLONG_DOUBLE`?

franzpoeschel:
Note: We should think about introducing a copied Datatype enum in the JSON backend as well. In the JSON backend, the string representations of the datatypes are part of the written dataset on disk, so having that JSON-specific will help not accidentally introducing a breaking change.

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
franzpoeschel: -> comment out

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
```diff
@@ -36,6 +36,219 @@
 
 namespace openPMD
 {
+/** Concrete datatype of an object available at runtime.
+ * Since ADIOS2 distinguishes unsigned char, signed char and char, this enum
+ * has all three of them (unlike the public Datatype enum)
```
ax3l: Update: let's add this to the fundamental Datatypes we support.

Also x-ref the type here in Doxygen, so it's easy to "click" through.

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
ax3l: Yes, let's comment this here since ADIOS2 does not support it.

franzpoeschel:
I now removed the ADIOS2Datatype thing again in the ADIOS2 backend since the benefits of keeping both enums separate were less than I'd thought in the beginning. I kept some of the refactoring, putting some code in private headers away from public includes.

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
franzpoeschel: `signed char` is now a fundamental supported datatype. The ADIOS2Datatype enum is no more, so the Doxygen thing is probably outdated?

src/IO/ADIOS/ADIOS2Auxiliary.cpp
```diff
@@ -230,7 +233,8 @@ Datatype attributeInfo(
                 {
                     errorMsg << std::to_string(ext) << ", ";
                 }
-                errorMsg << "] of type " << datatypeToString(basicType);
+                errorMsg
```
franzpoeschel: Note to self: revert this

include/openPMD/Datatype_internal.hpp
```diff
@@ -0,0 +1,94 @@
+#pragma once
```
ax3l, line 21:
```suggestion
/* Copyright 2022 Franz Poeschel, Axel Huebl
 *
 * This file is part of openPMD-api.
 *
 * openPMD-api is free software: you can redistribute it and/or modify
 * it under the terms of of either the GNU General Public License or
 * the GNU Lesser General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * openPMD-api is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License and the GNU Lesser General Public License
 * for more details.
 *
 * You should have received a copy of the GNU General Public License
 * and the GNU Lesser General Public License along with openPMD-api.
 * If not, see <http://www.gnu.org/licenses/>.
 */
#pragma once
```

include/openPMD/Datatype_internal.hpp
```diff
@@ -0,0 +1,94 @@
...
+                return DoDetermineDatatype::template call<
+                    typename T::value_type>();
+            }
+            else
+            {
+                return DoDetermineDatatype::template call<T>();
+            }
+#if defined(__INTEL_COMPILER)
+/*
+ * ICPC has trouble with if constexpr, thinking that return statements are
```
ax3l, line 57: I think NVCC 11.0 has a similar issue, potentially check and add a similar suppression?
Around 11.5+, the false warning is solved.

include/openPMD/Datatype_internal.hpp
```diff
@@ -0,0 +1,94 @@
...
+                return DoDetermineDatatype::template call<
+                    std::vector<typename T::value_type>>();
+            }
+            else
+            {
+                return DoDetermineDatatype::template call<std::vector<T>>();
+            }
+#if defined(__INTEL_COMPILER)
+/*
+ * ICPC has trouble with if constexpr, thinking that return statements are
```
ax3l, line 97: Pls see above note on NVCC

ax3l (approved):
:rocket: :sparkles:

ax3l (approved):

PR #1277, opened 2022-05-17
[WIP] Use JSON/TOML template for defining openPMD metadata in a config file

franzpoeschel:
Not relevant for next release

* **Until now:** Simulations specify their metadata in-code via API calls
* **With this PR:** In some workflows (e.g. experiments) there is no omniscient simulation, but metadata is instead input by the experimentors via configuration files, using the API is not a good workflow for that

Idea: We already have a JSON backend, use an openPMD-conforming JSON dataset to define only metadata. With this, the configuration file will be just another openPMD dataset.
Then, add some functionality to initialize an empty Series from such a metadata file.

TODO:

- [x] Add a `template` mode to the JSON backend that (1) does not pre-fill datasets, (2) does not allow writing to datasets
- [x] Use TOML as an alternative target for the JSON backend (better because it allows comments)
- [x] Add a basic `intitializeFromTemplate()` functionality
- [ ] Check how this fits experiments workflows, adapt
- [x] Make JSON/TOML maybe a bit easier to write, e.g. make the `datatype` field optional
- [ ] Maybe find another way to distinguish groups from datasets, so that not even `dtype` and `extent` are required
- [ ] Maybe find a way to make default attributes optional
- [x] Merge #1218 first
- [x] Introduce a good workflow to use a variable-based single iteration Series such that the single iteration represents all iterations
- [ ] Check interplay with Read Mode changes
- [x] Maybe automatically detect template mode in reading
- [x] Merge https://github.com/openPMD/openPMD-api/pull/1278 first
- [ ] documentation
- [ ] Merge #1493 first
- [ ] Will probably need an update to work together with #1432

https://github.com/franzpoeschel/openPMD-api/compare/topic-json-short-modes..topic-json-template

PR #1278, opened 2022-05-19
Add Attribute::getOptional<T>() and use to add some more dynamic datatype conversions at read time

franzpoeschel:
Currently, we have `Attribute::get<T>()` that throws a `std::runtime_error()` if no conversion is possible.
Adding `Attribute::getOptional<T>()` is helpful to check if a conversion is possible without having to resort to catching exceptions.

This PR also uses that new API method while reading a Series in. This became necessary in #1277 from where I have extracted this PR (that PR adds a less verbose mode to the JSON backend that does not explicitly annotate datatypes and might hence report different numerical types at read times).

### DIFF
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -34,6 +34,7 @@
 #include <string>
 #include <type_traits>
 #include <utility>
+#include <variant>
 #include <vector>
 
 namespace openPMD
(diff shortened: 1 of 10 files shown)


include/openPMD/backend/Attribute.hpp
```diff
@@ -251,4 +370,16 @@ U Attribute::get() const
     return getCast<U>(Variant::getResource());
```
franzpoeschel: We could remove `getCast` here in a similar way. Slightly API-breaking though since this is in a public header.

include/openPMD/backend/Attribute.hpp
```diff
@@ -226,6 +241,110 @@ auto doConvert(T *pv) -> U
 }
 #endif
 
+template <typename T, typename U>
```
franzpoeschel: Use `std::variant<T, std::string>` to combine both functions into one and still have error messages

include/openPMD/backend/Attribute.hpp
```diff
@@ -117,6 +118,20 @@ class Attribute
...
+     *
+     * @note This performs a static_cast and might introduce precision loss if
+     *       requested. Check dtype explicitly beforehand if needed.
+     *
+     * @tparam  U   Type of the object to be casted to.
+     * @return  Copy of the retrieved object, casted to type U.
+     *          An empty std::optional if no conversion is possible.
+     */
+    template <typename U>
+    std::optional<U> getOptional() const;
```
ax3l, line 137: Please add in Python, too (pybind11 returns `py:None` for `std::nullopt`)

franzpoeschel: Python does not use this call at all:
```c++
        // C++ pass-through API: Getter
        .def(
            "get_attribute",
            [](Attributable &attr, std::string const &key) {
                auto v = attr.getAttribute(key);
                return v.getResource();
                // TODO instead of returning lists, return all arrays (ndim > 0)
                // as numpy arrays?
            })
```

There is no explicit type requesting as `get<T>()` does, in Python you just get the type that there is

include/openPMD/backend/Attribute.hpp
franzpoeschel: done

include/openPMD/backend/Attribute.hpp
franzpoeschel: I did this API break now. All implementation details are now moved to the `detail` namespace.

include/openPMD/backend/Attribute.hpp
ax3l: Yes, I was wondering if we want to add such an option in Python to cast?

ax3l (approved):

PR #1290, opened 2022-06-08
Python: grid_spacing & time_offset `double`

ax3l:
In `pybind11`, overloads on types are order-dependent (first wins).
https://github.com/pybind/pybind11/issues/1512

We specialize `double` here generically and cast in read if needed (see #345 #1137).

Later on, we could add support for 1D numpy arrays with distinct type.

### DIFF
--- a/src/binding/python/Mesh.cpp
+++ b/src/binding/python/Mesh.cpp
@@ -71,18 +71,16 @@ void init_Mesh(py::module &m)
             [](Mesh &mesh, char d) { mesh.setDataOrder(Mesh::DataOrder(d)); },
             "Data Order of the Mesh (deprecated and set to C in openPMD 2)")
         .def_property("axis_labels", &Mesh::axisLabels, &Mesh::setAxisLabels)
-        .def_property(
-            "grid_spacing",
-            &Mesh::gridSpacing<float>,
-            &Mesh::setGridSpacing<float>)
+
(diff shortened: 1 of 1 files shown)


franzpoeschel:
Is there a reason not to specify `long double` generically instead of `double`? With this PR, don't you get a narrowing conversion if the Series stores these attributes as a long double type?

ax3l:
So in Python there is only a dynamic-bitness builtin float type. We could use `long double` here to case up to the platform-specific largest type, but I think I was conservative in my design here because `long double` differs from platform to platform.

I think the most general thing that we could do for the getter:
- write a lambda switch loop that checks at runtime the type of the attribute
- return a `std::variant<float, double, long double>`
and for the setter:
- overload for generic Python double type and
- numpy scalar

For now, this feels like a bit too detailed since there is most likely a conversation when going to Python builtin float anyways.

franzpoeschel:
With the current status of this PR, I think the following situation might happen:

```
openPMD dataset        ---> openPMD Python API               ---> Python script
(contains long double)      (uses Mesh::gridSpacing<double>)      (platform supports long double, but value was cast to double)
```

Your point of cross-platform consistency is valid, so I think we can go forward with this PR. We should just maybe be aware that this leaves stored `long double` values accessible only via C++.

ax3l:
Yes, that's the problem we need to generalize :+1:

PR #1291, opened 2022-06-09
New Access type READ_LINEAR

franzpoeschel:
Situation:
Using ADIOS2 steps in openPMD was always a bit challenging due to slightly incompatible IO workflows.
Traditionally in openPMD, we parse a Series upon opening it and users expect to use random-access for navigating through the Series. Alternatively, in ADIOS2 one can parse each iteration only after opening its corresponding IO step. With an IO step opened, ADIOS2 will only show the data pertaining to exactly that step and no other data (except for attributes that are designed to be constant). With the novel BP5 engine, this distinction is now mandatory to make since it has two access modes: `adios2::Mode::Read` and `adios2::Mode::ReadRandomAccess`.

Also adds the missing Append mode to the Python bindings.

This PR adds a new read mode `READ_LINEAR` and slightly restricts the usage of `READ_ONLY`:

* `READ_ONLY` corresponds to our traditional workflow of parsing everything up-front and then not again. It is implemented in the backend via `adios2::Mode::ReadRandomAccess`. 
    New in this PR: If the engine supports it at all, the Series **will** be parsed before opening any IO step. In variable-based iteration encoding, this means that data from a later iteration might leak into the first.
    Iterations that were parsed up front will not be parsed again upon opening their corresponding IO step.
   Optionally: Stop using IO steps altogether in this access mode in front- and backend. That would be API breaking for streaming (all other API breakage of this PR only affects the experimental new ADIOS2 schema), but with streaming, the other read mode makes more sense anyway.
* `READ_LINEAR`: Don't parse anything upon opening the Series. To get any data, `Series::readIterations()` must be used.

TODO

- [ ] Don't use steps at all in READ_ONLY (-> breaking change since this makes READ_LINEAR required for streaming workflows)
    → Probably best for the next release: Give a runtime warning when using READ_ONLY in contexts that should require READ_LINEAR (such as streaming), and disable the functionality altogether in the release after that
- [x] Memory optimization: Delete old iterations in READ_LINEAR mode (also ok for a future PR)
- [x] Merge #1237 first
- [x] ADIOS 2.7 compatibility
- [x] Actually test BP5, especially with old iteration encoding
- [x] Documentation
- [x] Test that Streaming keeps working with READ_ONLY mode for now
- [x] What happens if writing a Series with /data/snapshot attribute and then reading in random-access mode
    -> /data/snapshot gets properly ignored, bp4_steps tests this
- [x] Ignore /data/snapshot attribute if the backend reports that it prefers upfront parsing

### Commits:

**Basic implementation**
9a7fab780 Prepare scaffolding for Access::ReadLinearly
6f0e74c0f Adapt frontend to linear read mode
ed4083717 Make OPEN_FILE task parameters writable
73bae08ec ADIOS2 implementation

**Testing and bindings**
0b64fa2e1 Necessary fixes for tests
8dad84524 Add and adapt Python bindings
0ff801b65 Testing

**Documentation**
b06d257f6 Documentation: C++ and Python
fbca5e814 rst documentation

**Allow parsing global attributes before opening first iteration¹**
c841156e6 Make SeriesIterator into a handle
357e48723 Reading attributes without needing to open the first step

**Workaround for ADIOS2 aggregation issue**
891cba0d1 Don't write attributes too early

**Delete old iterations from data structures** ²
dd4228a41 Delete old iterations in READ_LINEAR mode
87d2004c3 DEREGISTER task

**¹)** In READ_LINEAR, the `Series` object is completely empty after constructing it. This is weird when we want to read global attributes:
```cpp
Series s("data.bp", Access::READ_LINEAR);
// no attributes available
for(auto iteration: s.readIterations())
{
    // Attributes are now readable
    // It's weird having to open the first iteration for this
}
```
Solution:
```cpp
Series s("data.bp", Access::READ_LINEAR);
// no attributes available
s.readIterations();
// global attributes can now be read, no iterations yet accessible
for(auto iteration: s.readIterations()) ...
```

**²)** In READ_LINEAR mode, iterations can only be accessed via `series.readIterations()`. We should not store them into `series.iterations` (except the currently active one), to avoid (1) users getting confused why old iterations are still there and (2) slowly rising memory usage in long-running simulation workflows.

### DIFF
--- a/include/openPMD/IO/Access.hpp
+++ b/include/openPMD/IO/Access.hpp
@@ -27,10 +27,53 @@ namespace openPMD
 enum class Access
 {
     READ_ONLY, //!< open series as read-only, fails if series is not found
+    READ_RANDOM_ACCESS = READ_ONLY,
+    READ_LINEAR,
     READ_WRITE, //!< open existing series as writable
     CREATE, //!< create new series and truncate existing (files)
     APPEND //!< write new iterations to an existing series without reading
 }; // Access
+
+namespace access
+{
+    inline bool readOnly(Access access)
+    {
+        switch (access)
+        {
+        case Access::READ_LINEAR:
+        case Access::READ_ONLY:
+            return true;
+        case Access::READ_WRITE:
+        case Access::CREATE:
+        case Access::APPEND:
+            return false;
+        }
+    }
+
+    inline bool write(Access access)
+    {
+        return !readOnly(access);
+    }
+
+    inline bool writeOnly(Access access)
+    {
+        switch (access)
+        {
+        case Access::READ_LINEAR:
+        case Access::READ_ONLY:
+        case Access::READ_WRITE:
+            return false;
+        case Access::CREATE:
+        case Access::APPEND:
+            return true;
+        }
+    }
+
+    inline bool read(Access access)
+    {
+        return !writeOnly(access);
+    }
+} // namespace access
 
 // deprecated name (used prior to 0.12.0)
 // note: "using old [[deprecated(msg)]] = new;" is still badly supported, thus
--- a/include/openPMD/IO/Access.hpp
+++ b/include/openPMD/IO/Access.hpp
@@ -20,6 +20,8 @@
  */
 #pragma once
 
+#include <stdexcept>
+
 namespace openPMD
 {
 /** File access mode to use during IO.
@@ -48,6 +50,7 @@ namespace access
         case Access::APPEND:
             return false;
         }
+        throw std::runtime_error("Unreachable!");
     }
 
     inline bool write(Access access)
@@ -67,6 +70,7 @@ namespace access
         case Access::APPEND:
             return true;
         }
+        throw std::runtime_error("Unreachable!");
     }
 
     inline bool read(Access access)
--- a/include/openPMD/IO/Access.hpp
+++ b/include/openPMD/IO/Access.hpp
@@ -66,7 +66,10 @@ enum class Access
     READ_RANDOM_ACCESS = READ_ONLY, //!< more explicit alias for READ_ONLY
     /*
      * Open Series as read-only, fails if Series is not found.
-     * This access mode requires use of Series::readIterations()
+     * This access mode requires use of Series::readIterations().
+     * Global attributes are available directly after calling
+     * Series::readIterations(), Iterations and all their corresponding data
+     * become available by use of the returned Iterator, e.g. in a foreach loop.
      * See Access::READ_ONLY for when to use this.
      */
     READ_LINEAR,
--- a/include/openPMD/IO/Access.hpp
+++ b/include/openPMD/IO/Access.hpp
@@ -28,10 +28,53 @@ namespace openPMD
  */
 enum class Access
 {
-    READ_ONLY, //!< open series as read-only, fails if series is not found
-    READ_RANDOM_ACCESS = READ_ONLY,
+    /**
+     * Open Series as read-only, fails if Series is not found.
+     * When to use READ_ONLY or READ_LINEAR:
+     *
+     * * When intending to use Series::readIterations()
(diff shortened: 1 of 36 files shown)


src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2335,99 +2422,311 @@ namespace detail
...
+                     * Scenario: BP5 with old ADIOS2 schema and normal read
+                     * mode. Need to close the engine and reopen.
+                     */
+                    parsePreference = ParsePreference::RandomAccess;
+                    m_engine->Close();
+                    streamStatus = StreamStatus::NoStream;
+                    m_mode = adios2::Mode::ReadRandomAccess;
+                    // Overwrite the old IO object since it is stained with the
+                    // objects of the now closed engine
+                    create_IO();
```
franzpoeschel: Problem: This deletes all configuration

include/openPMD/IO/Access.hpp
```diff
@@ -20,18 +20,65 @@
...
+
 namespace openPMD
 {
 /** File access mode to use during IO.
  */
 enum class Access
 {
     READ_ONLY, //!< open series as read-only, fails if series is not found
+    READ_RANDOM_ACCESS = READ_ONLY,
+    READ_LINEAR,
```
franzpoeschel, line 75: Maybe deprecate:
```cpp
enum class Access
{
    READ_RANDOM_ACCESS, //!< open series as read-only, fails
                        //!< if series is not found
    READ_ONLY [[deprecated("Renamed as READ_RANDOM_ACCESS")]] =
        READ_RANDOM_ACCESS,
    READ_LINEAR,
    READ_WRITE, //!< open existing series as writable
    CREATE, //!< create new series and truncate existing (files)
    APPEND //!< write new iterations to an existing series without reading
}; // Access
```

Would then need to change every occurrence in the code base.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1308,8 +1317,15 @@ void ADIOS2IOHandlerImpl::availableChunks(
     std::string varName = nameOfVariable(writable);
     auto engine = ba.getEngine(); // make sure that data are present
     auto datatype = detail::fromADIOS2Type(ba.m_IO.VariableType(varName));
+    bool allSteps = m_handler->m_frontendAccess != Access::READ_LINEAR &&
+        ba.streamStatus == detail::BufferedActions::StreamStatus::NoStream;
```
franzpoeschel, line 1323: Need to use `AllStepsBlocksInfo` if `allSteps` is true. This fixes https://github.com/openPMD/openPMD-viewer/issues/350.

ax3l (approved):
This looks great to me :rocket: :sparkles: 

Thank you also for the clear structure for reviewing :pray:

test/SerialIOTest.cpp
```diff
@@ -4923,6 +4925,17 @@ TEST_CASE("bp4_steps", "[serial][adios2]")
     bp4_steps("../samples/nullcore.bp", nullcore, "");
     bp4_steps("../samples/bp4steps_default.bp", "{}", "{}");
 
+    // bp4_steps(
```
ax3l, line 4928: Just flagging: these tests are still commented - intentional?
Remove dead code in favor of a comment?

test/SerialIOTest.cpp
```diff
@@ -6850,46 +6960,24 @@ TEST_CASE("append_mode", "[serial]")
...
                 false,
                 ParseMode::WithSnapshot,
                 jsonConfigNew);
-            append_mode(
-                "../samples/append/append_variablebased." + t,
-                true,
-                ParseMode::WithSnapshot,
-                jsonConfigOld);
+            // This test config does not make sense
+            // append_mode(
```
ax3l, line 6976: Just flagging: this test is still commented - intentional?
Remove and just leave a comment?

PR #1294, opened 2022-06-24
Add unique_ptr overloads for storeChunk calls

franzpoeschel:
See https://github.com/openPMD/openPMD-api/issues/1216 for an explanation. The approach in this PR is to use `unique_ptr`s instead of relying on the refcount of `shared_ptr`s to make this optimization more explicit to the user.

TODO

- [x] Add a class `OpenpmdUniquePtr` that allows specifying dynamic destructors via `std::function<void(T*)>`, similar to `std::shared_ptr`. In the STL, `unique_ptr`s have a different type depending on their destructor, which is annoying for our use case, and the overhead of making this dynamic is negligible in our use case.
    This way, it's easily possible to e.g. pass ownership of pinned memory to openPMD and still let it correctly destroy it once used.
- [x] Extend the IOTask for `Write_Dataset` operations to accept `std::variant<std::shared_ptr<void const>, OpenpmdUniquePtr<void const>>` as a buffer.
    This requires some light refactoring because with this change, the IOTask is not copyable any more.
    This includes making the backends deal with both types correctly.
- [x] Add a `storeChunk` overload that accepts `OpenpmdUniquePtr`s
- [x] Actually implement the intended optimization in the ADIOS2 backend
- [x] testing, documentation
- [x] support for determineDatatype
- [x] Merge #1207 first
- [x] Check if it also works to let this be handled by `Engine::close()`.
- [x] I think it might be possible to implement this in such a way that makes `OpenpmdUniquePtr` not leak into the public API.

@ax3l This PR should not be too much work any more, so we might as well add this to 0.15.*?

### DIFF
--- a/include/openPMD/Datatype.hpp
+++ b/include/openPMD/Datatype.hpp
@@ -21,6 +21,7 @@
 #pragma once
 
 #include "openPMD/auxiliary/TypeTraits.hpp"
+#include "openPMD/auxiliary/UniquePtr.hpp"
 
 #include <array>
 #include <climits>
@@ -276,31 +277,26 @@ inline constexpr Datatype determineDatatype()
         return Datatype::UNDEFINED;
 }
 
-template <typename T>
-inline constexpr Datatype determineDatatype(std::shared_ptr<T>)
-{
-    return determineDatatype<T>();
-}
-
-template <typename T>
-inline constexpr Datatype determineDatatype(T *)
-{
-    return determineDatatype<T>();
-}
-
-/*
- * Catch-all overload for unsupported types, with static_assert errors
- * triggered at compile-time.
+/**
+ * @brief Determine datatype of passed value
+ *
+ * @param val Value whose type to evaluate
+ * @tparam T Type of the passed value
+ * @return If T is of a pointer type, then the type of the contained value.
+ *         Otherwise, a compile-time error detailing the use of this function.
  */
-template <typename T_ContiguousContainer>
-inline constexpr Datatype determineDatatype(T_ContiguousContainer &&)
+template <typename T>
+inline constexpr Datatype determineDatatype(T &&val)
 {
-    using T_ContiguousContainer_stripped =
-        std::remove_reference_t<T_ContiguousContainer>;
-    if constexpr (auxiliary::IsContiguousContainer_v<
-                      T_ContiguousContainer_stripped>)
+    (void)val; // don't need this, it only has a name for Doxygen
+    using T_stripped = std::remove_cv_t<std::remove_reference_t<T>>;
+    if constexpr (auxiliary::IsPointer_v<T_stripped>)
     {
-        static_assert(auxiliary::dependent_false_v<T_ContiguousContainer>, R"(
+        return determineDatatype<auxiliary::IsPointer_t<T_stripped>>();
+    }
+    else if constexpr (auxiliary::IsContiguousContainer_v<T_stripped>)
+    {
+        static_assert(auxiliary::dependent_false_v<T_stripped>, R"(
 Error: Passed a contiguous container type to determineDatatype<>().
 These types are not directly supported due to colliding semantics.
 Assuming a vector object `std::vector<float> vec;`,
@@ -322,15 +318,14 @@ use one of the following alternatives:
     }
     else
     {
-        static_assert(auxiliary::dependent_false_v<T_ContiguousContainer>, R"(
+        static_assert(auxiliary::dependent_false_v<T_stripped>, R"(
 Error: Unknown datatype passed to determineDatatype<>().
 For a direct translation from C++ type to the openPMD::Datatype enum, use:
 `auto determineDatatype<T>() -> Datatype`.
 
 For detecting the contained datatpye of a pointer type (shared or raw pointer),
-use either of the following overloads:
-`auto determineDatatype<T>(std::shared_ptr<T>) -> Datatype` or
-`auto determineDatatype<T>(T *) -> Datatype`.
+use this following template (i.e. `auto determineDatatype<T>(T &&) -> Datatype`)
+which accepts pointer-type parameters (raw, shared or unique).
         )");
     }
     // Unreachable, but C++ does not know it
--- a/include/openPMD/IO/IOTask.hpp
+++ b/include/openPMD/IO/IOTask.hpp
@@ -25,6 +25,7 @@
 #include "openPMD/IterationEncoding.hpp"
 #include "openPMD/Streaming.hpp"
 #include "openPMD/auxiliary/Export.hpp"
+#include "openPMD/auxiliary/Memory.hpp"
 #include "openPMD/auxiliary/Variant.hpp"
 #include "openPMD/backend/Attribute.hpp"
 #include "openPMD/backend/ParsePreference.hpp"
@@ -33,6 +34,7 @@
 #include <memory>
 #include <string>
 #include <utility>
+#include <variant>
 #include <vector>
 
 namespace openPMD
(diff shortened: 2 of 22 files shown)


include/openPMD/auxiliary/Memory.hpp
```diff
@@ -163,5 +165,102 @@ namespace auxiliary
         return allocatePtr(dtype, numPoints);
     }
 
+    template <typename T>
+    struct DecayArray
```
franzpoeschel: This is available in STL as std::remove_extent

franzpoeschel:
I can only continue work on this once #1207 is merged since the features will interfere with each other in the ADIOS2 backend.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2665,32 +2850,77 @@ namespace detail
...
+            case FlushTarget::Buffer:
+            case FlushTarget::Buffer_Override:
+                performDataWrite = false;
+                break;
+            }
+            performDataWrite = performDataWrite && m_engineType == "bp5";
+
+            if (performDataWrite)
+            {
+                // should we really do this now?
```
```suggestion
```

Should we write attributes before PerformDataWrites()?

1. Yes: Then setting another value for the attribute later won't work (see e.g SerialIOTest:)
    ```cpp
        o.setOpenPMDextension(
            1); // this happens intentionally "late" in this test
    ```
2. No: Then attributes would only be written very late, i.e. before `EndStep()`.

Personally, I think if `preferred_flush_target = disk` is selected, then everything that can be written to disk should be, so we should leave it as it is.

include/openPMD/RecordComponent.hpp
```diff
@@ -387,6 +404,9 @@ class RecordComponent : public BaseRecordComponent
      */
     RecordComponent &makeEmpty(Dataset d);
 
+    void storeChunk(
+        auxiliary::WriteBuffer buffer, Datatype datatype, Offset o, Extent e);
+
```
franzpoeschel, line 424: private, no new API

include/openPMD/IO/ADIOS/ADIOS2Auxiliary.hpp
```diff
@@ -1,3 +1,4 @@
+
```
franzpoeschel:
```suggestion
```

include/openPMD/Datatype.hpp
```diff
@@ -276,31 +277,26 @@ inline constexpr Datatype determineDatatype()
...
+ *
+ * @param val Value whose type to evaluate
+ * @tparam T Type of the passed value
+ * @return If T is of a pointer type, then the type of the contained value.
+ *         Otherwise, a compile-time error detailing the use of this function.
  */
-template <typename T_ContiguousContainer>
-inline constexpr Datatype determineDatatype(T_ContiguousContainer &&)
+template <typename T>
+inline constexpr Datatype determineDatatype(T &&val)
```
franzpoeschel, line 289: This function is now aware of `std::unique_ptr` as well as `OpenpmdUniquePtr`, additionally to the previous raw and shared pointers. It's now implemented via `if constexpr` (previously overloads). Otherwise, the API is unchanged.

include/openPMD/RecordComponent.hpp
```diff
@@ -291,6 +293,36 @@ class RecordComponent : public BaseRecordComponent
...
+     *               The unique pointer must own and manage the buffer.
+     *               Optimizations might be implemented based on this
+     *               assumption (e.g. further deferring the operation if the
+     *               backend is the unique owner).
+     *               For raw pointers, use storeChunkRaw().
+     * @param offset Offset within the dataset.
+     * @param extent Extent within the dataset, counted from the offset.
+     */
+    template <typename T>
+    void storeChunk(OpenpmdUniquePtr<T> data, Offset offset, Extent extent);
```
franzpoeschel: We could theoretically delete this overload, as `OpenpmdUniquePtr`s are also accepted by the `std::unique_ptr` overload. However, `OpenpmdUniquePtr` is useful for specifying custom destructors, and having its own overload will promote its usage.
Apart from this, `OpenpmdUniquePtr` is needed internally.

include/openPMD/auxiliary/UniquePtr.hpp
```diff
@@ -0,0 +1,161 @@
...
+ * for GPU buffers.
+ *
+ * If not specifying a custom deleter explicitly, this class emulates the
+ * behavior of a std::unique_ptr with std::default_delete.
+ * This also means that array types are supported as expected.
+ *
+ * @tparam T The pointer type, as in std::unique_ptr.
+ */
+template <typename T>
+class OpenpmdUniquePtr
```
franzpoeschel: Maybe rename `DynamicUniquePtr`?

ax3l: `UniquePtrWithLambda`

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l: I think we can revisit this once we change the defaults of frontends and avoid setting defaults of attributes too early. Writing attributes after datasets are defined can help with sensible defaults, e.g., `x,y,z` for 3D but only `x,y` for 2D mesh axis labels, etc.

include/openPMD/Datatype.hpp
ax3l: Ok. Could be a good candidate for a C++ concept at some point.

include/openPMD/IO/IOTask.hpp
```diff
@@ -704,9 +677,10 @@ struct OPENPMDAPI_EXPORT Parameter<Operation::DEREGISTER>
     Parameter(Parameter const &) : AbstractParameter()
```
ax3l, line 677: CodeQL: Looks like `Parameter` generally lacks the copy assignment operator (independent of this PR):
```C++
Parameter& operator=(const Parameter& other)
```

https://github.com/openPMD/openPMD-api/security/code-scanning/498

include/openPMD/RecordComponent.hpp
ax3l: (could nonetheless have Doxygen block for us ;-) )

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -3355,24 +3418,19 @@ namespace detail
...
-                {
-                    m_alreadyEnqueued.emplace_back(
-                        std::unique_ptr<BufferedAction>{
-                            new BufferedAttributeWrite{
-                                std::move(task.second)}});
-                }
-                m_attributeWrites.clear();
+                throw error::Internal(
+                    "ADIOS2 backend: Flush of late writes was requested at the "
+                    "wrong time.");
```
ax3l, line 3425: This error message is not hard to understand for me :)

ax3l (approved):
This look really good to me. I have a few inline comments that can be addressed as a follow-up :)

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: To be fair, this is an `error::Internal` that should only happen when there is a bug somewhere in our code. So if this error occurs, it's not really supposed to be understood by the user, but to help them report the issue in more detail.

`error::Internal` is always thrown in combination with a request to report it:
```cpp
    Internal::Internal(std::string const &what)
        : Error(
              "Internal error: " + what +
              "\nThis is a bug. Please report at ' "
              "https://github.com/openPMD/openPMD-api/issues'.")
    {}
```

ax3l: That is great!

I mean for us (devs) to remember, I would extend the half sentence about the "wrong time" a bit :D

PR #1296, opened 2022-06-28
storeChunk: Add an overload for shared_ptr<T[]>

franzpoeschel:
Noticed while working on #1294: With C++17, if you use `shared_ptr<T[]>`, it will automatically select the right destructor, i.e. `delete[]`.
Note that our old version of `storeChunk<>()` already supports this more or less on accident, since we use `std::decay` inside `determineDatatype` then cast the shared_ptr down to void type. With giving this an explicit overload, we make the support more explicit to the user and also to ourselves, so we don't accidentally break this functionality in future.

### DIFF
--- a/examples/10_streaming_write.cpp
+++ b/examples/10_streaming_write.cpp
@@ -26,7 +26,8 @@ int main()
     constexpr unsigned long length = 10ul;
     Extent global_extent = {length};
     Dataset dataset = Dataset(datatype, global_extent);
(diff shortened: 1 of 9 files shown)


franzpoeschel:
Clang-6 with libcpp does not seem to know about this? We can revert the examples and tests again and just leave the overload?

ax3l (approved):
C++17 ftw :)

PR #1297, opened 2022-07-12
Avoid LD_PRELOAD in ASAN UBSAN script, use LD_LIBRARY_PATH instead

franzpoeschel:
Doesn't really matter for the CI, but I often use these scripts as a blueprint for other stuff and it's good to avoid `LD_PRELOAD` if possible.


franzpoeschel:
Alright, it does seem to be necessary in the Python bindings as they seem to work with LD_PRELOAD themselves?
Asan requires that if LD_PRELOAD is used, then Asan must be first in list.

PR #1302, opened 2022-07-14
Fix append mode double attributes

franzpoeschel:
When appending to an existing group/variable-based Series, don't define top-level attributes again (except `basePath` and `openPMD`, these are needed internally and ADIOS2 doesn't care as long as they are duplicate but equivalent)
This fixes an Append mode bug that is already present on dev and has a low chance of making CI runs fail.

- [x] Merge #1218 first
- [x] Add `CHECK_FILE` task for other backends too (maybe not for ADIOS1)
- [x] ~~maybe merge `CHECK_FILE` and `CREATE_FILE` tasks~~, no this would require executing `CREATE_FILE` very early
- [x] cleanup
- [x] Support file creation via Append mode in group-based iteration encoding
- [x] ~~This seems to trigger an ADIOS2 bug?:~~ I get this on dev already, will need to see what's going on
    ```
    append_mode                                                                                                                                                                                                     
    /home/franzpoeschel/git-repos/openPMD-api/test/SerialIOTest.cpp:6486                                                                                                                                            
                                                                                                                                                                                                                    
    /home/franzpoeschel/git-repos/openPMD-api/test/SerialIOTest.cpp:6486: FAILED:                                                                                                                                   
      [Tue Sep 20 14:18:56 2022] [ADIOS2 EXCEPTION] <Engine>                                                                                                                                                        
      <BP5Reader> <ParseMetadataIndex> : Record 5 (id = 115) has invalid length 48.                                                                                                                                 
      We parsed 2195488 bytes for this record      
    ```


include/openPMD/IO/AbstractIOHandlerImpl.hpp
```diff
@@ -198,6 +204,17 @@ class AbstractIOHandlerImpl
                     break;
                 }
             }
+            catch (std::exception const &e)
```
franzpoeschel: Note to self: remove this branch again

src/Series.cpp
```diff
@@ -502,15 +502,18 @@ namespace
         int padding;
         uint64_t iterationIndex;
         std::set<int> paddings;
-        for (auto const &entry : auxiliary::list_directory(directory))
```
franzpoeschel, line 505: This fixes another subtle bug: Creating a Series via Append mode should be possible. But so far, this routine failed if the directory was not present.

include/openPMD/auxiliary/Mpi.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+#if openPMD_HAVE_MPI
+#include <mpi.h>
+#endif
+
+namespace openPMD::auxiliary
+{
+#if openPMD_HAVE_MPI
+
+using Mock_MPI_Comm = MPI_Comm;
```
ax3l: As discussed: Mocking in shipped code is dangerous and the reason we have to do a full linker-wrapping for ADIOS1.

Let's avoid this if we can write `#ifdef`s instead.

franzpoeschel: I've removed these changes now.
The lines that are now still in the new `Mpi.hpp` file are moved there from `Filesystem.cpp`, I think it makes sense to keep that change.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2616,17 +2679,28 @@ namespace detail
     {
         if (!m_engine)
         {
+            auto tempMode = m_mode;
             switch (m_mode)
             {
             case adios2::Mode::Append:
+#ifdef _WIN32
+                /*
+                 * On Windows, ADIOS2 Append mode only works with existing
```
ax3l: Let's mention the ADIOS2 version here and link an issue?

include/openPMD/auxiliary/Mpi.hpp
ax3l: Ok, but the line `using Mock_MPI_Comm = MPI_Comm;` still needs to be cleaned out pls :)
```suggestion
```

include/openPMD/auxiliary/Mpi.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "openPMD/config.hpp"
+
+#include <optional>
+#include <stdexcept>
+#include <type_traits>
+#include <utility>
```
ax3l: Please remove unused includes :)

src/IO/ADIOS/ADIOS2IOHandler.cpp
ax3l:
```suggestion
                 * On Windows, ADIOS2 2.8 Append mode only works with existing
```

include/openPMD/Series.hpp
```diff
@@ -82,6 +82,8 @@ namespace internal
          * the same instance.
          */
         std::optional<WriteIterations> m_writeIterations;
+
+        Attributable mockedDefaults;
```
ax3l: Unused:
```suggestion
```

include/openPMD/auxiliary/Mpi.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace openPMD::auxiliary
+{
+#if openPMD_HAVE_MPI
+
+using Mock_MPI_Comm = MPI_Comm;
+
+namespace
+{
+    template <typename>
+    struct MPI_Types;
```
ax3l: Future refactoring: since this is an openPMD internal type trait, we should consider renaming this so it does not collide with the `MPI_....` prefix used as convention in the MPI C libraries for all its functions.

Suggestions: `MPI_Types` -> `openPMD_MPI_Type`

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: I'm not sure on affected ADIOS2 versions. I can test again next week, and then report an issue.

include/openPMD/auxiliary/Mpi.hpp
franzpoeschel: Looking at this, I'm asking myself if this code is even correct. If an MPI application runs on a heterogeneous system with differing CPU architectures, there will be undefined behavior, no?
Maybe, we should just use the largest numeric type for those cases? Or is there a better way?

ax3l: This depends a bit and would require more work, but I think not in the location here.

> heterogeneous system with differing CPU architectures

The questions here is not really the CPU (micro) architecture but the data model: https://en.cppreference.com/w/cpp/language/types Essentially all modern systems you would put on an HPC system are `LP64`.

If you would like to run with MPI on a system with heterogeneous hosts, you would have to compile your binaries for each of them and spawn them with MPI. That's the easy part, the hard part would be if someone ever implemented an MPI that would allow to communicate between LP64 and, let's say ILP32, or LP64 with big and little endian.

Either way, it would not change the local code in this location.

ax3l: I think we can actually simplify this by using `static constexpr MPI_Datatype const value = ...;`

ax3l: We can probably even make this a simpler `constexpr` function now, which might also benefit compile time.

franzpoeschel: `constexpr` will probably not work fully since the MPI datatypes are not `constexpr` in every implementation, but we can still replace the specialized class template with `if constexpr`, yeah.

> If you would like to run with MPI on a system with heterogeneous hosts, you would have to compile your binaries for each of them and spawn them with MPI. That's the easy part, the hard part would be if someone ever implemented an MPI that would allow to communicate between LP64 and, let's say ILP32, or LP64 with big and little endian.

> Either way, it would not change the local code in this location.

Fair enough, thanks for clarifying!

src/IO/ADIOS/ADIOS2IOHandler.cpp
franzpoeschel: Done

ax3l (approved):
Splendid, thanks a lot! :)

PR #1303, opened 2022-07-19
Fix file existence check in parallel tests

franzpoeschel:
The parallel_adios2_json_config test currently checks for existence of SerialIOTests files, which always exist since the parallel tests run after the serial ones. Fix that.


ax3l (approved):
:+1:

PR #1305, opened 2022-07-26
Wheels: Fix HDF5 on Windows (0.14.5.post1)

ax3l:
Fix the missing HDF5 dependency on Windows.
Update the build logic to error out if ADIOS2 or HDF5 are not found (instead of using auto-accept or disable logic).

- [x] bump version to `.post1`

## Note

`Build x86_64 wheel on macos-10.15`
This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on August 30th, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

Not a big problem, we can restart after merge and do not really need this platform the be rebuilt.

### DIFF
--- a/library_builders.bat
+++ b/library_builders.bat
@@ -1,6 +1,6 @@
 set CURRENTDIR="%cd%"
 
-:: BUILD_PREFIX="${BUILD_PREFIX:-/usr/local}"
+set BUILD_PREFIX="C:/Program Files (x86)"
 set CPU_COUNT="2"
 
 echo "CFLAGS: %CFLAGS%"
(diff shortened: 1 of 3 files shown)


library_builders.bat
```diff
@@ -1,6 +1,6 @@
 set CURRENTDIR="%cd%"
 
-:: BUILD_PREFIX="${BUILD_PREFIX:-/usr/local}"
+set BUILD_PREFIX="C:/Program Files/"
```
ax3l: Potentially `set BUILD_PREFIX="C:/Program Files (x86)/"` for 32bit builds...

PR #1313, opened 2022-09-15
Remove caching cmake vars

nischild:
If openPMD is included into a CMake project as a subproject it modifies variables in or adds variables to the CACHE.
This can lead to conflicts in the parent project since these variables are overwritten in or set for the parent project.

Not adding variables to the [CACHE limits the scope of variables](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-variables) to the openPMD project which makes it safer for other developers to add openPMD as a subproject.

This is a less invasive approach to the pull request #1312 and it is more complete since in #1312 only `CMAKE_RUNTIME_OUTPUT_DIRECTORY` is substituted by a new openPMD specific variable.

List of variables handled in this pull request:
```
CMAKE_ARCHIVE_OUTPUT_DIRECTORY
CMAKE_LIBRARY_OUTPUT_DIRECTORY
CMAKE_RUNTIME_OUTPUT_DIRECTORY
CMAKE_INSTALL_CMAKEDIR
CMAKE_BUILD_TYPE
```

Close #1312

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -740,16 +740,16 @@ if(openPMD_HAVE_PYTHON)
 
     if(WIN32)
         set(CMAKE_INSTALL_PYTHONDIR_DEFAULT
-            "${CMAKE_INSTALL_LIBDIR}/site-packages"
-        )
+            "${CMAKE_INSTALL_LIBDIR}/site-packages")
     else()
         set(CMAKE_INSTALL_PYTHONDIR_DEFAULT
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you very much for the fix!

I never noticed this, because I set all these vars conditionally also in my super projects. But this makes sense, thx!

ax3l:
@DerNils-git  do you mind adding your IPP affiliation to your [GitHub profile](https://github.com/settings/profile)? That just helps me to keep track of who is who :)
I sent you an invite to https://github.com/openPMD :)

PR #1314, opened 2022-09-29
Deprecate ADIOS1 backend

franzpoeschel:
* print deprecation warning when selecting ADIOS1 backend
* deactivate ADIOS1 build by default

TODO
- [x] We might want to manually specify ADIOS1=ON in CI and package managers

### DIFF
--- a/src/IO/AbstractIOHandlerHelper.cpp
+++ b/src/IO/AbstractIOHandlerHelper.cpp
@@ -20,6 +20,8 @@
  */
 #include "openPMD/IO/AbstractIOHandlerHelper.hpp"
 
+#include "openPMD/config.hpp"
+
 #include "openPMD/Error.hpp"
 #include "openPMD/IO/ADIOS/ADIOS1IOHandler.hpp"
 #include "openPMD/IO/ADIOS/ADIOS2IOHandler.hpp"
@@ -28,7 +30,12 @@
 #include "openPMD/IO/HDF5/HDF5IOHandler.hpp"
 #include "openPMD/IO/HDF5/ParallelHDF5IOHandler.hpp"
 #include "openPMD/IO/JSON/JSONIOHandler.hpp"
+#include "openPMD/auxiliary/Environment.hpp"
 #include "openPMD/auxiliary/JSON_internal.hpp"
+
+#if openPMD_HAVE_MPI
+#include <mpi.h>
+#endif
 
 #include <memory>
 #include <utility>
@@ -65,7 +72,12 @@ namespace
     Please consider switching to ADIOS2.
     We recommend checking your ADIOS1 datasets for compatibility with ADIOS2.
     Conversion of data from one backend to another may optionally be achieved
-    by using the `openpmd-pipe` tool.)";
+    by using the `openpmd-pipe` tool.)
+
+    Suppress this warning via `export OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED=1`.)";
+
+    constexpr char const *suppressAdios1DeprecationWarning =
+        "OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED";
 } // namespace
 
 #if openPMD_HAVE_MPI
@@ -86,7 +98,15 @@ std::unique_ptr<AbstractIOHandler> createIOHandler<json::TracingJSON>(
         return constructIOHandler<ParallelHDF5IOHandler, openPMD_HAVE_HDF5>(
             "HDF5", path, access, comm, std::move(options));
     case Format::ADIOS1:
-        std::cerr << adios1Deprecation << std::endl;
+        if (auxiliary::getEnvNum(suppressAdios1DeprecationWarning, 0) == 0)
+        {
+            int rank;
+            MPI_Comm_rank(comm, &rank);
+            if (rank == 0)
+            {
+                std::cerr << adios1Deprecation << std::endl;
+            }
+        }
         return constructIOHandler<ParallelADIOS1IOHandler, openPMD_HAVE_ADIOS1>(
             "ADIOS1", path, access, std::move(options), comm);
     case Format::ADIOS2_BP:
@@ -156,7 +176,10 @@ std::unique_ptr<AbstractIOHandler> createIOHandler<json::TracingJSON>(
         return constructIOHandler<HDF5IOHandler, openPMD_HAVE_HDF5>(
             "HDF5", path, access, std::move(options));
     case Format::ADIOS1:
-        std::cerr << adios1Deprecation << std::endl;
+        if (auxiliary::getEnvNum(suppressAdios1DeprecationWarning, 0) == 0)
+        {
+            std::cerr << adios1Deprecation << std::endl;
+        }
         return constructIOHandler<ADIOS1IOHandler, openPMD_HAVE_ADIOS1>(
             "ADIOS1", path, access, std::move(options));
     case Format::ADIOS2_BP:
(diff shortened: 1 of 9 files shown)


ax3l (approved):
I took a look at CI script and am actually already setting `=ON` for all backends when we expect them. This was already a safety measure to make sure that issues in the environment, e.g., a missing dependency, are noted quickly :)

I pushed a minor cleanup where ADIOS1 was actually intentionally off, but the CI job name did not reflect this yet.

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -75,6 +86,7 @@ std::shared_ptr<AbstractIOHandler> createIOHandler<json::TracingJSON>(
         return constructIOHandler<ParallelHDF5IOHandler, openPMD_HAVE_HDF5>(
             "HDF5", path, access, comm, std::move(options));
     case Format::ADIOS1:
+        std::cerr << adios1Deprecation << std::endl;
```
ax3l: I am not sure if this is the cleanest way to warn users, because it cannot be suppressed. This might also be a problem in MPI-parallel contexts.

Maybe we can add an env var to turn it off or something.

franzpoeschel:
I have pushed a commit that will print the warning on rank 0 only, and makes it suppressible via `OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED`. 
(Note that we mostly don't consider MPI in other warnings that we print)

In parallel: 

```
Singularity> mpirun -n 2 bin/ParallelIOTests adios_write_test
Filters: adios_write_test
Filters: 
[Deprecation warning]
    Development on the ADIOS1 IO library has ceased.
    Support for ADIOS1 in the openPMD-api has been deprecated
    and will be removed in a future version.

    Please consider switching to ADIOS2.
    We recommend checking your ADIOS1 datasets for compatibility with ADIOS2.
    Conversion of data from one backend to another may optionally be achieved
    by using the `openpmd-pipe` tool.


    Suppress this warning via `export OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED=1`.
adios_write_test
WARN : MPI_AMR method: No OST to use. Set num_ost=NNN in the adios config xml file.
test cases: 1 | ===============================================================================
1 passed
assertions: - none -

test cases: 1 | 1 passed
assertions: - none -

Singularity> OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED=1 mpirun -n 2 bin/ParallelIOTests adios_write_test
Filters: adios_write_test
Filters: adios_write_test
WARN : MPI_AMR method: No OST to use. Set num_ost=NNN in the adios config xml file.
test cases: 1 | 1 passed
assertions: - none -
test cases: 1 | 1
 passed
assertions: - none -
```

Serial:
```
[Deprecation warning]                                                                                                                
    Development on the ADIOS1 IO library has ceased.                                                                                 
    Support for ADIOS1 in the openPMD-api has been deprecated                                                                        
    and will be removed in a future version.

    Please consider switching to ADIOS2.
    We recommend checking your ADIOS1 datasets for compatibility with ADIOS2.
    Conversion of data from one backend to another may optionally be achieved
    by using the `openpmd-pipe` tool.


    Suppress this warning via `export OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED=1`.
openPMD series: out
openPMD standard: 1.1.0
openPMD extensions: 1

data author: unknown
data created: 2018-02-06 09:40:21 -0800
data backend: ADIOS1
generating machine: unknown
generating software: warp (version: 4)
generating software dependencies: unknown

number of iterations: 5 (groupBased)
  all iterations: 100 200 300 400 500 

number of meshes: 2
  all meshes:
    E
    rho

number of particle species: 1
  all particle species:
    electrons
Singularity> OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED=1 !!
OPENPMD_WARNING_SUPPRESS_ADIOS1_DEPRECATED=1 openpmd-ls out.bp 
openPMD series: out
openPMD standard: 1.1.0
openPMD extensions: 1

data author: unknown
data created: 2018-02-06 09:40:21 -0800
data backend: ADIOS1
generating machine: unknown
generating software: warp (version: 4)
generating software dependencies: unknown
[…]
```

ax3l:
Splendid, I added it to docs and renamed it to `OPENPMD_ADIOS_SUPPRESS_DEPRECATED_WARNING` for consistency with other ADIOS1 env variables.

ax3l (dismissed):

test/CoreTest.cpp
```diff
@@ -1356,6 +1360,7 @@ TEST_CASE("unavailable_backend", "[core]")
 #endif
 
 #if openPMD_HAVE_MPI
+    MPI_Init(nullptr, nullptr);
```
ax3l: If we need MPI init here, then I have the fear we also need the logic in serial user code that is coded against an MPI-enabled openPMD-api, no?

src/IO/AbstractIOHandlerHelper.cpp
```diff
@@ -75,6 +98,15 @@ std::shared_ptr<AbstractIOHandler> createIOHandler<json::TracingJSON>(
         return constructIOHandler<ParallelHDF5IOHandler, openPMD_HAVE_HDF5>(
             "HDF5", path, access, comm, std::move(options));
     case Format::ADIOS1:
+        if (auxiliary::getEnvNum(suppressAdios1DeprecationWarning, 0) == 0)
+        {
+            int rank;
+            MPI_Comm_rank(comm, &rank);
```
ax3l, line 104: Maybe guard with an [MPI_Initialized](https://www.mpich.org/static/docs/v3.1/www3/MPI_Initialized.html) check?
If false, then just print. if true, then only print from rank 0

test/CoreTest.cpp
franzpoeschel: The line only runs if `Series` has been constructed with an MPI communicator. openPMD-api should be able to assume that MPI has been initialized if a communicator has been used.

src/IO/AbstractIOHandlerHelper.cpp
franzpoeschel: I was not aware that such a check is possible. We can do that, but I think that we don't need to (see comment above).

test/CoreTest.cpp
franzpoeschel: > we also need the logic in serial user code that is coded against an MPI-enabled openPMD-api

(Note: this situation is tested in every MPI-enabled CI run by `SerialIOTests`)

ax3l: On Slack: ok, so its an actual bug here because we pass `MPI_COMM_WORLD` 5 lines below :)

ax3l: No, sorry - MPI context handling for Catch2 should all go in `test/CatchRunner.cpp`
https://github.com/openPMD/openPMD-api/blob/dev/test/CatchRunner.cpp

Is the wrong runner used here?

Maybe we need to split out the non-MPI core tests from the MPI Core tests to stay clean.

src/IO/AbstractIOHandlerHelper.cpp
ax3l: Added a comment above: I think we use the wrong (serial) Catch runner for the MPI Core tests and should split them.

franzpoeschel: I've moved the parallel part to the ParallelIOTests now

ax3l (approved):
I also updated the docs - LGTM now :)

franzpoeschel:
Good catch, thanks!

PR #1316, opened 2022-10-04
RecordComponent: Properly handle uninitialized datasets

franzpoeschel:
We currently have a non-functional check during `flush()` that a dataset has been initialized.
The check tests that the `Datatype` is not `UNDEFINED`, but it is never undefined since the constructor initializes it as `CHAR`.

We can use a `std::optional` to have a trustworthy check. This gives users a better feedback if they forget this call.

TODO

- [x] Merge #1278 first
- [x] The newly-added check uncovers that the CoreTests are mostly wrong, since they don't initialize the dataset. Since the CoreTests usually don't need to actually interact with the resulting dataset, they still run fine, but show a warning in the destructor. Maybe fix them.
    ```
    [~Series] An error occurred: Wrong API usage: [RecordComponent] Must specify dataset type and extent before flushing (see RecordComponent::resetDataset()).
    [~Series] An error occurred: Wrong API usage: [RecordComponent] Must specify dataset type and extent before flushing (see RecordComponent::resetDataset()).
    [~Series] An error occurred: Wrong API usage: [RecordComponent] Must specify dataset type and extent before flushing (see RecordComponent::resetDataset()).
    [~Series] An error occurred: Wrong API usage: [RecordComponent] Must specify dataset type and extent before flushing (see RecordComponent::resetDataset()).
    [~Series] An error occurred: Wrong API usage: [RecordComponent] Must specify dataset type and extent before flushing (see RecordComponent::resetDataset()).
    [~Series] An error occurred: Wrong API usage: [RecordComponent] Must specify dataset type and extent before flushing (see RecordComponent::resetDataset()).
    ```
- [ ] Maybe turn this into a warning, and default to `{CHAR, {1}}` instead
- [x] Wait for release 0.15 before merging this. This might uncover bugs in downstream code, and they should have time to fix it.
- [x] Still problems at `myPath` and `structure_test` of CoreTests
- [x] Maybe check if the JSONIOHandlerImpl fix should be applied to other backends, too: The other backends only free their resources in the destructor, but don't flush

Notes:

* This is also useful for things like #1277, since that PR actually gives a dataset with UNDEFINED datatype a meaning (inside TOML templates). Distinguishing between "forgotten to specify" and "explicitly specified as UNDEFINED" is useful here.
  This is an argument in favor of **not** defaulting to `{CHAR, {1}}`
* This seems to uncover some such bugs in our own test suite.

### DIFF
--- a/examples/7_extended_write_serial.cpp
+++ b/examples/7_extended_write_serial.cpp
@@ -83,8 +83,9 @@ int main()
                 {{io::UnitDimension::M, 1}});
             electrons["displacement"]["x"].setUnitSI(1e-6);
             electrons.erase("displacement");
-            electrons["weighting"][io::RecordComponent::SCALAR].makeConstant(
-                1.e-5);
+            electrons["weighting"][io::RecordComponent::SCALAR]
+                .resetDataset({io::Datatype::FLOAT, {1}})
+                .makeConstant(1.e-5);
         }
 
         io::Mesh mesh = cur_it.meshes["lowRez_2D_field"];
--- a/examples/7_extended_write_serial.py
+++ b/examples/7_extended_write_serial.py
@@ -90,7 +90,9 @@ if __name__ == "__main__":
     electrons["displacement"].unit_dimension = {Unit_Dimension.M: 1}
     electrons["displacement"]["x"].unit_SI = 1.e-6
     del electrons["displacement"]
-    electrons["weighting"][SCALAR].make_constant(1.e-5)
+    electrons["weighting"][SCALAR] \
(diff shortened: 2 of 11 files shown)


ax3l:
This looks good! Anything on the open checks in the PR description we shall discuss tomorrow? :)

franzpoeschel:
> This looks good! Anything on the open checks in the PR description we shall discuss tomorrow? :)

I've pushed a little fix that makes this whole PR a lot less invasive by just skipping any datasets, that have neither a dataset definition nor written chunks. That makes a flush directly after first access possible, in my opinion eliminating all remaining questions.

ax3l (approved):
:+1:

PR #1324, opened 2022-11-03
Add Series::close() call

franzpoeschel:
We currently often use `del series` in Python which is not ideal for a garbage collected language, as you cannot rely on the garbage collector to actually trigger application logic.
So, add a call `Series::close` whose effects are equivalent to destroying all `Series` instances.

TODO
- [x] Documentation
- [x] Keep at least some `del series` instances to ensure that this keeps working during the next release cycle


ax3l (approved):
Excellent :+1:

PR #1336, opened 2022-11-24
Add switches --mpi and --no-mpi to openpmd-pipe

franzpoeschel:
The default behavior of `openpmd-pipe` is described by its help message:
```
By default, openpmd-pipe will use MPI if all of the following conditions
are fulfilled:
1) The mpi4py package can be imported.
2) The openPMD-api has been built with support for MPI.
3) The MPI size is greater than 1.
   By default, the openPMD-api will be initialized without an MPI communicator
   if the MPI size is 1. This is to simplify the use of the JSON backend
   which is only available in serial openPMD.
```

This mostly works fine, but sometimes, it's necessary to run it with MPI even if the world size is only 1, conversely it is sometimes necessary to not even import `mpi4py`.  This PR keeps the default behavior, but adds switches to enforce either behavior.

### DIFF
--- a/src/binding/python/openpmd_api/pipe/__main__.py
+++ b/src/binding/python/openpmd_api/pipe/__main__.py
@@ -73,7 +73,7 @@ Examples:
 
 args = parse_args(sys.argv[0])
 # MPI is an optional dependency
-if (args.mpi is None or args.mpi) and io.variants['mpi']:
+if io.variants['mpi'] and (args.mpi is None or args.mpi):
     try:
         from mpi4py import MPI
         HAVE_MPI = True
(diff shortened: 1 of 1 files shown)


src/binding/python/openpmd_api/pipe/__main__.py
```diff
@@ -72,10 +60,45 @@ def parse_args(program_name):
...
+            print("""
+    openPMD-api was built with support for MPI,
+    but mpi4py Python package was not found.
+    Will continue in serial mode.""",
+                  file=sys.stderr)
+            HAVE_MPI = False
+else:
+    HAVE_MPI = False
+
+debug = False
```
ax3l, line 93: Unrelated: we could make debug a cmd line option, too :)

ax3l (approved):

PR #1337, opened 2022-11-25
Document why m_backendAccess and m_frontendAccess are distinguished

franzpoeschel:
I can never remember why these are distinct, so I wrote it down. Also, it seems that its meaning/use has even changed from then to now.


ax3l (approved):
Thx :sparkling_heart:

PR #1355, opened 2023-01-05
Unify exception types

### LINKED ISSUE #1354: Python bindings for all custom exception types
**Is your feature request related to a problem? If so, please describe.**
For our new exception types in `Error.hpp`, we have bindings for Python inside https://github.com/openPMD/openPMD-api/blob/dev/src/binding/python/Error.cpp. For older ones, such as `no_such_attribute_error`, we don't. This means that it is currently impossible in Python to catch an exception thrown e.g. by `get_attribute`.

1. Add those missing bindings
2. Maybe take the chance to move those exceptions into `Error.hpp` and into the `error` namespace.

franzpoeschel:
This PR:

1. Removes `ParseError` from `Error.hpp`. It's a leftover from #1237 that I ended up not using.
2. Uses `error::OperationUnsupportedInBackend` instead of `unsupported_data_error` and removes `unsupported_data_error` (only used in ADIOS1 implementation)
3. Puts `no_such_attribute_error` into `Error.hpp`, moves it into the `error` namespace as `error::NoSuchAttribute`, creates a compatibility alias for it and creates a Python binding for it.

Fix #1354.

### DIFF
--- a/include/openPMD/Error.hpp
+++ b/include/openPMD/Error.hpp
@@ -106,6 +106,12 @@ namespace error
             std::optional<std::string> backend_in,
             std::string description_in);
     };
+
+    class NoSuchAttribute : public Error
+    {
+    public:
+        NoSuchAttribute(std::string attributeName);
+    };
 } // namespace error
 
 /**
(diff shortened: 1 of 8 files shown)


ax3l (approved):
Thx! :)

PR #1357, opened 2023-01-10
Fix dtype_from_numpy

franzpoeschel:
Use char comparison to detect type equality.
I had an error today where a double type was not recognized by the old `dt.is()` implementation and switching to char comparison fixed that.


ax3l:
Oh, how did you find `pybind11::dtype::char_`? I was not aware of this...
https://github.com/pybind/pybind11/pull/2864

ax3l (approved):

franzpoeschel:
To be fair, I just browsed the header files for something that looked useful as a type representation :D

Issue #1363, opened 2023-01-20
HDF5 cross-platform compatibility of attributes


franzpoeschel:
**Describe the bug**
Some HDF5 Datasets written on x86 platforms cannot be opened on ARM and PowerPC architectures.

**To Reproduce**

Use either:
[Be_dens.zip](https://github.com/openPMD/openPMD-api/files/10467642/Be_dens.zip)

or:
https://github.com/openPMD/openPMD-example-datasets/blob/draft/example-femm-3d.h5

and try opening it on a PowerPC system (e.g. Summit) or an ARM system (e.g. modern Apple devices) with `openpmd-ls`.

**Expected behavior**

```
> openpmd-ls Be_dens.h5 
openPMD series: Be_dens
openPMD standard: 1.1.0
openPMD extensions: 0

data author: ...
data created: 2023-01-04 15:19:57 +0100
data backend: HDF5
generating machine: unknown
generating software: MALA (version: 1.1.0)
generating software dependencies: unknown

number of iterations: 1 (groupBased)
  all iterations: 0 

number of meshes: 1
  all meshes:
    Density

number of particle species: 1
  all particle species:
    atoms
```

**Actual behavior**
```
> openpmd-ls ../Be_dens.h5 
openPMD series: Be_dens
openPMD standard: 1.1.0
openPMD extensions: 0

data author: ...
data created: 2023-01-04 15:19:57 +0100
data backend: HDF5
generating machine: unknown
generating software: MALA (version: 1.1.0)
generating software dependencies: unknown

number of iterations: 1 (groupBased)
  all iterations: [AbstractIOHandlerImpl] IO Task READ_ATT failed with exception. Removing task from IO queue and passing on the exception.
An error occurred while opening the specified openPMD series!
[HDF5] Unsupported simple attribute type
```

**Software Environment**
Recent conda, pip installs.

**Additional context**

PR #1364, opened 2023-01-20
HDF5IOHandler: Support for float128 on ARM64/PPC64

skalarproduktraum:
This PR (hopefully) addresses #1363 by adding h5py-compatible float128 data types in HDF5IOHandler. The original issue seems to originate from a AMD64 vs. ARM64/PPC64 difference in the handling of `long double` types. While on AMD64, 128bit float with 80 bits of precision directly maps to H5T_NATIVE_LDOUBLE, it does not on ARM64/PPC64.

For the time being, I have also added some more debug output to the else branch where an unknown type is encountered, such that the HDF5 type info can be more easily recovered.

The PR currently still has a `Heap corruption detected, free list is damaged` error, the attributes that failed to read correctly originally read correctly now, though. Maybe @franzpoeschel has an idea there 👍

### DIFF
--- a/src/IO/HDF5/HDF5Auxiliary.cpp
+++ b/src/IO/HDF5/HDF5Auxiliary.cpp
@@ -98,7 +98,7 @@ hid_t openPMD::GetH5DataType::operator()(Attribute const &att)
         return H5Tcopy(H5T_NATIVE_DOUBLE);
     case DT::LONG_DOUBLE:
     case DT::VEC_LONG_DOUBLE:
-        return H5Tcopy(H5T_NATIVE_LDOUBLE);
+        return H5Tcopy(m_userTypes.at(typeid(long double).name()));
     case DT::CFLOAT:
     case DT::VEC_CFLOAT:
         return H5Tcopy(m_userTypes.at(typeid(std::complex<float>).name()));
--- a/src/IO/HDF5/HDF5Auxiliary.cpp
+++ b/src/IO/HDF5/HDF5Auxiliary.cpp
@@ -98,7 +98,7 @@ hid_t openPMD::GetH5DataType::operator()(Attribute const &att)
         return H5Tcopy(H5T_NATIVE_DOUBLE);
     case DT::LONG_DOUBLE:
     case DT::VEC_LONG_DOUBLE:
-        return H5Tcopy(m_userTypes.at(typeid(long double).name()));
+        return H5Tcopy(H5T_NATIVE_LDOUBLE);
     case DT::CFLOAT:
     case DT::VEC_CFLOAT:
         return H5Tcopy(m_userTypes.at(typeid(std::complex<float>).name()));
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -2242,21 +2242,37 @@ void HDF5IOHandlerImpl::readAttribute(
         }
         else if (H5Tequal(attr_type, m_H5T_LONG_DOUBLE_80_BE))
         {
-            // worst case, sizeof(long double) is only 8, so allocate enough memory to fit 16 bytes per member
-            std::vector<long double> vld80be(dims[0]*2, 0);
+            // worst case, sizeof(long double) is only 8, so allocate enough
+            // memory to fit 16 bytes per member
+            std::vector<long double> vld80be(dims[0] * 2, 0);
             status = H5Aread(attr_id, attr_type, vld80be.data());
-            H5Tconvert(attr_type, H5T_NATIVE_LDOUBLE, dims[0],
-                vld80be.data(), nullptr, H5P_DEFAULT);
-            std::cout << attr_name << " size " << dims[0] << ", " << vld80be[0] << " " << vld80be[1] << " " << vld80be[2] << std::endl;
+            H5Tconvert(
+                attr_type,
+                H5T_NATIVE_LDOUBLE,
+                dims[0],
+                vld80be.data(),
+                nullptr,
+                H5P_DEFAULT);
+            std::cout << attr_name << " size " << dims[0] << ", " << vld80be[0]
+                      << " " << vld80be[1] << " " << vld80be[2] << std::endl;
             a = Attribute(vld80be);
         }
         else if (H5Tequal(attr_type, m_H5T_LONG_DOUBLE_80_LE))
         {
-            // worst case, sizeof(long double) is only 8, so allocate enough memory to fit 16 bytes per member
-            std::vector<long double> vld80le(dims[0]*2, 0);
+            // worst case, sizeof(long double) is only 8, so allocate enough
+            // memory to fit 16 bytes per member
+            std::vector<long double> vld80le(dims[0] * 2, 0);
             status = H5Aread(attr_id, attr_type, vld80le.data());
-            H5Tconvert(attr_type, H5T_NATIVE_LDOUBLE, dims[0], vld80le.data(), nullptr, H5P_DEFAULT);
-            std::cout << attr_name << sizeof(long double) <<  " size " << dims[0] << ", " << vld80le[0] << " " << vld80le[1] << " " << vld80le[2] << std::endl;
+            H5Tconvert(
+                attr_type,
+                H5T_NATIVE_LDOUBLE,
+                dims[0],
+                vld80le.data(),
+                nullptr,
+                H5P_DEFAULT);
+            std::cout << attr_name << sizeof(long double) << " size " << dims[0]
+                      << ", " << vld80le[0] << " " << vld80le[1] << " "
+                      << vld80le[2] << std::endl;
             a = Attribute(vld80le);
         }
         else if (H5Tget_class(attr_type) == H5T_STRING)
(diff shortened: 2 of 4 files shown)


skalarproduktraum:
Figured to work around the heap corruption. Issue is that on macOS/ARM64, `sizeof(long double)` is 8, which I did not anticipate, but apparently has its reasons in ABI compatibility...

franzpoeschel:
Hey Ulrik, 
thank you for this! The issue on Summit seems to be another one, the dataset is still not legible after this PR, but I can have a look at that separately. I'll add a test for this PR and add support for float128 to other places in the HDF5 backend where it's relevant.

franzpoeschel:
Aaah, monday has caught me again. After using the correct branch, this also solves the issue on Summit/PPC64.

franzpoeschel:
Hmm, we'll probably also need complex 128bit double types then..

skalarproduktraum:
This is actually quite an interesting issue. I had completely forgotten that x86 uses 80bit precision for `long double`. So would it make actually more sense to not use `H5T_NATIVE_LDOUBLE` at all?

franzpoeschel:
So, you would suggest instead to do something like `H5Tset_size(m_H5T_LONG_DOUBLE_VARLEN_BE, sizeof(long double));`? 
Also, is there any reason to have both the big and little endian versions, or could we just use the endianness of the platform instead? (I have not written the HDF5 backend, so I don't know the details of HDF5 here too well..)

franzpoeschel:
I opened a PR with suggestions https://github.com/skalarproduktraum/openPMD-api/pull/1

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -2184,6 +2208,47 @@ void HDF5IOHandlerImpl::readAttribute(
...
+                nullptr,
+                H5P_DEFAULT);
+            std::cout << attr_name << " size " << dims[0] << ", " << vld80be[0]
+                      << " " << vld80be[1] << " " << vld80be[2] << std::endl;
+            a = Attribute(vld80be);
+        }
+        else if (H5Tequal(attr_type, m_H5T_LONG_DOUBLE_80_LE))
+        {
+            // worst case, sizeof(long double) is only 8, so allocate enough
+            // memory to fit 16 bytes per member
```
franzpoeschel: We should maybe use malloc here to avoid running into aliasing issues?

skalarproduktraum: Yep, that makes much more sense. This was just my stop-gap solution for the prototype.

ax3l:
Yes, I think that generally cross-platform `long double` varies, also see x86-64 Windows...
- https://en.wikipedia.org/wiki/Long_double
- https://en.cppreference.com/w/cpp/language/types
> long double - extended precision floating-point type. Matches [IEEE-754 binary128 format](https://en.wikipedia.org/wiki/Quadruple-precision_floating-point_format) if supported, otherwise matches [IEEE-754 binary64-extended format](https://en.wikipedia.org/wiki/Extended_precision) if supported, otherwise matches some non-IEEE-754 extended floating-point format as long as its precision is better than binary64 and range is at least as good as binary64, otherwise matches IEEE-754 binary64 format.
>
> - binary128 format is used by some HP-UX, SPARC, MIPS, ARM64, and z/OS implementations.
> - The most well known IEEE-754 binary64-extended format is 80-bit x87 extended precision format. It is used by many x86 and x86-64 implementations (a notable exception is MSVC, which implements long double in the same format as double, i.e. binary64).

So far, we did avoid converting (casting) data during I/O and kept it platform-specific. I am open to suggestions to enable platform-portability, but we need to check if that works on all platforms and is worth the performance hit (ideally, opt-in).

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -73,6 +73,8 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(
     , m_H5T_CFLOAT{H5Tcreate(H5T_COMPOUND, sizeof(float) * 2)}
     , m_H5T_CDOUBLE{H5Tcreate(H5T_COMPOUND, sizeof(double) * 2)}
     , m_H5T_CLONG_DOUBLE{H5Tcreate(H5T_COMPOUND, sizeof(long double) * 2)}
+    , m_H5T_LONG_DOUBLE_80_BE{H5Tcopy(H5T_IEEE_F64BE)}
```
ax3l: BE is not really used: virtually all PPC machines in HPC use LE these days :)

franzpoeschel:
@ax3l I have reduced this to little endian and added a comment that this conversion fix specifically targets the AMD64 -> ARM64/PPC64 special case. The workaround is used only if the double type is not recognized as a native type, so I'd say that the opt-in criterium is sufficiently fulfilled.
Should be ready for review :) (Reviewing should ideally include testing if an ARM Mac can open the datasets linked here #1363, I don't have one)

src/IO/HDF5/HDF5Auxiliary.cpp
```diff
@@ -98,7 +98,7 @@ hid_t openPMD::GetH5DataType::operator()(Attribute const &att)
         return H5Tcopy(H5T_NATIVE_DOUBLE);
     case DT::LONG_DOUBLE:
     case DT::VEC_LONG_DOUBLE:
-        return H5Tcopy(H5T_NATIVE_LDOUBLE);
+        return H5Tcopy(m_userTypes.at(typeid(long double).name()));
```
franzpoeschel:
```suggestion
        return H5Tcopy(H5T_NATIVE_LDOUBLE);
```
I think that the change can be reverted again. In writing procedures, we can always use the native datatype anyway. In `HDF5IOHandlerImpl::readDataset()` we need to manually check if the dataset can be read as native `long double` anyway before attempting the workaround with the manually defined `long double`. As a result, it's fine to return native `long double` here in `::openDataset()`. Other reading procedures don't use this anyway.

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -107,6 +110,24 @@ HDF5IOHandlerImpl::HDF5IOHandlerImpl(
     H5Tinsert(m_H5T_CLONG_DOUBLE, "r", 0, H5T_NATIVE_LDOUBLE);
     H5Tinsert(m_H5T_CLONG_DOUBLE, "i", sizeof(long double), H5T_NATIVE_LDOUBLE);
 
+    H5Tset_size(m_H5T_LONG_DOUBLE_80_LE, 16);
```
ax3l, line 117:
```suggestion
    // inline comment as above for this block :)
    H5Tset_size(m_H5T_LONG_DOUBLE_80_LE, 16);
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -1760,6 +1793,28 @@ void HDF5IOHandlerImpl::readDataset(
...
     });
     hid_t dataType = getH5DataType(a);
+    if (H5Tequal(dataType, H5T_NATIVE_LDOUBLE))
+    {
+        // We have previously determined in openDataset() that this dataset is
+        // of type long double.
+        // We cannot know if that actually was H5T_NATIVE_LDOUBLE or if it was
+        // the worked-around m_H5T_LONG_DOUBLE_80_LE.
+        // Check this.
+        hid_t checkDatasetTypeAgain = H5Dget_type(dataset_id);
```
ax3l, line 1807: I think every `H5Dget_type` needs a matching `H5Tclose(checkDatasetTypeAgain)`.

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -2210,6 +2288,42 @@ void HDF5IOHandlerImpl::readAttribute(
             status = H5Aread(attr_id, attr_type, vcld.data());
             a = Attribute(vcld);
         }
+        else if (H5Tequal(attr_type, m_H5T_CLONG_DOUBLE_80_LE))
+        {
+            // worst case, sizeof(long double) is only 8, so allocate enough
+            // memory to fit 16 bytes per member
+            auto *tmpBuffer =
+                static_cast<long double *>(malloc(16 * 2 * dims[0]));
```
ax3l: In C++, I would use `new` and `delete` over `malloc` and `free`.

ax3l (changes requested):

ax3l:
ping @franzpoeschel @skalarproduktraum pls see review comments to finalize :)

We would like to cut the 0.15.2 release soon, let me know if this works to be updated :)

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: This will need a `reinterpret_cast` then though, as you can't `new` a void pointer. But I'll push a commit.

ax3l: We found today: can be `new long double` :sweat_smile:

ax3l (changes requested):
Looks awesome, we just want to update the `new long double`.

src/IO/HDF5/HDF5IOHandler.cpp
franzpoeschel: It turns out, we cannot 🙃 
This branch is active when `sizeof(long double) == 8`, but the dataset on disk has 16-byte doubles. There is no way to allocate a correctly-sized buffer using native float types, so `new char[]` it is.

franzpoeschel:
> Looks awesome, we just want to update the `new long double`.

I added an update and made a comment a bit clearer

ax3l (approved):
Excellent, good point!

PR #1368, opened 2023-02-02
Deactivate warning when opening single file of file-based Series

franzpoeschel:
Deactivated warning: `"Series constructor called with explicit iteration suggests loading a single file with groupBased iteration encoding. Loaded file is fileBased."`

Opening a single file of a file-based Series is a valid workflow, warnings are not necessary here. The warning can currently not be deactivated and is annoying.
The PR leaves the old warning as a comment, because we might want to add this back in again if we add some kind of verbosity level specification or logging.


ax3l (approved):
Agreed, that's a valid workflow and could be added in verbosity logging.

PR #1369, opened 2023-02-02
Iteration.time and Iteration.dt: Prefer double in Python setters

franzpoeschel:
When using Pybind to define multiple overloads of a setter for numeric types, e.g. float, double and long double, Python does not actually let you specify the overload to use.
This PR uses double as a default.

### DIFF
--- a/src/binding/python/Iteration.cpp
+++ b/src/binding/python/Iteration.cpp
@@ -44,30 +44,31 @@ void init_Iteration(py::module &m)
                 return ss.str();
             })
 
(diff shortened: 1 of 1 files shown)


ax3l (approved):

ax3l:
Cycling CI to fix pre-commit failure.

ax3l (approved):
Thank you, good for now!

Better solution: https://github.com/pybind/pybind11/issues/191#issuecomment-217754152

PR #1378, opened 2023-02-23
Some fixes for char signs

franzpoeschel:
* HDF5 backend did not yet fully support signed char, making tests fail on Summit
* Numpy bindings assumed that `char` is signed

Follow-up to #1275


ax3l (approved):
Phew! :+1:

PR #1381, opened 2023-02-27
Python bindings: Release GIL during IO wait operations

### LINKED ISSUE #1380: Python: GIL is not released during blocking read from ADIOS2 stream
**Describe the bug**
When waiting to read from an `openpmd_api.Series.read_iterations` that was created from an ADIOS2 stream, the GIL is never released.
This creates a race condition when asynchronously reading via `threading.Thread` and adding to a `queue.Queue`: if data is not consumed fast enough and a read out of the stream ever starts to block because the `Queue` is full, the Python process cannot continue execution because the GIL is forever kept on the openPMD side.

Using `multiprocessing.{Process,Queue}` is a way around this problem.

**Pseudocode**
```python
from queue import Queue
from threading import Thread
import time

import openpmd_api

def read_queued(stream, queue):
    for x in iter(stream.read_iterations()):
        queue.put(x)

stream = openpmd_api.Series('stream.sst', openpmd_api.Access_Type.read_only)
queue = Queue(maxsize=3)

reader_thread = Thread(target=read_queued, args=(stream, queue))
reader_thread.start()

for i in range(20):
    queue.get()
    # This duration must be larger than what reading a data point from the stream takes.
    time.sleep(5)
print('I'll never be reached!')
reader_thread.join()
```

**Expected behavior**
If the read blocks while waiting for stream data, the GIL needs to be released. ([Related Python doc](https://docs.python.org/3/c-api/init.html#thread-state-and-the-global-interpreter-lock) and [PyBind11 doc](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#global-interpreter-lock-gil).)

**Software Environment**
 - version of openPMD-api: [0.15.0-dev](https://github.com/franzpoeschel/openPMD-api/commit/214b7eafe20025a37a1a3772a0dc8710f74d4871)
 - installed openPMD-api via: from source
 - operating system: Rocky Linux 8
 - machine: JUWELS Booster
 - name and version of Python implementation: CPython 3.9.6
 - version of ADIOS2: 3d498e5c423ec3830264b1cfdf7c0182dd0bddc7
 - name and version of MPI: ParaStationMPI 5.5.0

**Additional context**
Sorry for no example, don't know how to create a minimal ADIOS2 stream.

franzpoeschel:
Fix #1380

This releases the GIL basically in those calls that trigger `adios2::Engine::BeginStep()`. Releasing the GIL for `adios2::Engine::EndStep`, and during flushes is more difficult since those calls may imply write access to Python buffers.

For a follow-up PR, we could try adding an `async for` implementation for the `ReadIterations` API.

### DIFF
--- a/src/binding/python/Series.cpp
+++ b/src/binding/python/Series.cpp
@@ -53,6 +53,20 @@ struct openPMD_PyMPICommObject
 using openPMD_PyMPIIntracommObject = openPMD_PyMPICommObject;
 #endif
 
+struct SeriesIteratorPythonAdaptor : SeriesIterator
+{
+    SeriesIteratorPythonAdaptor(SeriesIterator it)
+        : SeriesIterator(std::move(it))
+    {}
+
+    /*
+     * Python iterators are weird and call `__next__()` already for getting the
+     * first element.
+     * In that case, no `operator++()` must be called...
+     */
+    bool first_iteration = true;
+};
+
 void init_Series(py::module &m)
 {
     py::class_<WriteIterations>(m, "WriteIterations")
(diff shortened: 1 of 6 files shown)


src/binding/python/Series.cpp
```diff
@@ -59,26 +59,66 @@ void init_Series(py::module &m)
         .def(
             "__getitem__",
             [](WriteIterations writeIterations, Series::IterationIndex_t key) {
+                py::gil_scoped_release release;
```
franzpoeschel, line 83: Need to remove this again / refine it a bit

ax3l (approved):
:+1:

PR #1382, opened 2023-02-27
Joined arrays in ADIOS2

### LINKED ISSUE #1374: Particles: Support Auto-Shape Counting
For particles, we need to investigate how we can deine openPMD Datatyopes that have only a local size, where their global size is determined at runtime.

This currently would only work with ADIOS2 2.9+ using `JoinedArrays` and would safe users from having to do collective All2All comms (reduce sum and a scan) for particle global sizes and offsets prior to I/O.

Related to: https://github.com/ornladios/ADIOS2/issues/3452

franzpoeschel:
This implements https://github.com/ornladios/ADIOS2/issues/3452 in openPMD and (partially) fixes #1374.

We will need a fallback implementation for backends other than ADIOS2. Quoting the discussion in #1374:

> Since this feature is relevant for basically all parallel code that writes particles, we should (as a second step) implement this also in openPMD, so that those codes still stay compatible with HDF5.
> Using MPI collectives to determine local block offsets inside the particle list is a really common pattern for using openPMD in such code, and this is a great chance to eliminate this pattern, while still staying compatible with all openPMD backends.

TODO
- [x] More extensive testing, parallel tests
- [x] documentation
- [x] Merge #1316 first
- [x] Don't use .joinedDimension, magic number instead
- [x] ~Needs a standard update: numParticlesOffset may optionally be skipped, particlePatches must then be in order so they can be reconstructed from numParticles~ Not updating the standard yet
- [x] Merge #1577 first

### DIFF
--- a/docs/source/usage/workflow.rst
+++ b/docs/source/usage/workflow.rst
@@ -25,6 +25,7 @@ Storing and reading chunks
    The chunk is then stored by specifying an empty offset vector ``{}``.
    The chunk extent vector must be equivalent to the global extent in all non-joined dimensions (i.e. joined arrays allow no further sub-chunking other than concatenation along the joined dimension).
    The joined dimension of the extent vector specifies the extent that this piece should have along the joined dimension.
+   In the Python API, the slice-based setter syntax can be used as an abbreviation since the necessary information is determined from the passed array, e.g. ``record_component[()] = local_data``.
    The global extent of the dataset along the joined dimension will then be the sum of all local chunk extents along the joined dimension.
 
    Since openPMD follows a struct-of-array layout of data, it is important not to lose correlation of data between components. E.g., joining an array must take care that ``particles/e/position/x`` and ``particles/e/position/y`` are joined in uniform way.
--- a/docs/source/usage/workflow.rst
+++ b/docs/source/usage/workflow.rst
@@ -3,6 +3,48 @@
 Workflow
 ========
 
+Storing and reading chunks
+--------------------------
+
+1. **Chunks within an n-dimensional dataset**
+
+   Most commonly, chunks within an n-dimensional dataset are identified by their offset and extent.
+   The extent is the size of the chunk in each dimension, NOT the absolute coordinate within the entire dataset.
+
+   In the Python API, this is modeled to conform to the conventional ``__setitem__``/``__getitem__`` protocol.
+
+2. **Joined arrays (write only)**
+
+   (Currently) only supported in ADIOS2 no older than v2.9.0 under the conditions listed in the `ADIOS2 documentation on joined arrays <https://adios2.readthedocs.io/en/latest/components/components.html#shapes>`_.
+
+   In some cases, the concrete chunk within a dataset does not matter and the computation of indexes is a needless computational and mental overhead.
+   This commonly occurs for particle data which the openPMD-standard models as a list of particles.
+   The order of particles does not matter greatly, and making different parallel processes agree on indexing is error-prone boilerplate.
+
+   In such a case, at most one *joined dimension* can be specified in the Dataset, e.g. ``{Dataset::JOINED_DIMENSION, 128, 128}`` (3D for the sake of explanation, particle data would normally be 1D).
+   The chunk is then stored by specifying an empty offset vector ``{}``.
+   The chunk extent vector must be equivalent to the global extent in all non-joined dimensions (i.e. joined arrays allow no further sub-chunking other than concatenation along the joined dimension).
+   The joined dimension of the extent vector specifies the extent that this piece should have along the joined dimension.
+   The global extent of the dataset along the joined dimension will then be the sum of all local chunk extents along the joined dimension.
+
+   Since openPMD follows a struct-of-array layout of data, it is important not to lose correlation of data between components. E.g., joining an array must take care that ``particles/e/position/x`` and ``particles/e/position/y`` are joined in uniform way.
+
+   The openPMD-api makes the **following guarantee**:
+
+   Consider a Series written from ``N`` parallel processes between two (collective) flush points. For each parallel process ``n`` and dataset ``D``, let:
+
+    * ``chunk(D, n, i)`` be the ``i``'th chunk written to dataset ``D`` on process ``n``
+    * ``num_chunks(D, n)`` be the count of chunks written by ``n`` to ``D``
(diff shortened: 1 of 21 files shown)


franzpoeschel:
Except for the documentation, this is now somewhat ready.
I see one problem: With this, it's practically impossible to write the `particlePatchesOffset` as that is determined by ADIOS2 and we don't know it before it's written.

test/ParallelIOTest.cpp
```diff
@@ -1865,4 +1865,161 @@ TEST_CASE("unavailable_backend", "[core][parallel]")
...
+        }
+        for (size_t i = 0; i < size * patches_per_rank; ++i)
+        {
+            REQUIRE(length_of_patch * i == numParticlesOffset.get()[i]);
+            REQUIRE(type(length_of_patch * i) == patchOffset.get()[i]);
+        }
+    }
+}
+
+TEST_CASE("joined_dim", "[serial]")
```
guj: I think you mean "parallel"?

test/ParallelIOTest.cpp
```diff
@@ -1865,4 +1865,161 @@ TEST_CASE("unavailable_backend", "[core][parallel]")
...
+        auto numParticlesOffset =
+            it.particles["e"]
+                .particlePatches["numParticlesOffset"][RecordComponent::SCALAR];
+        numParticles.resetDataset(numParticlesDS);
+        numParticlesOffset.resetDataset(numParticlesDS);
+
+        auto patchOffset = it.particles["e"].particlePatches["offset"]["x"];
+        auto patchExtent = it.particles["e"].particlePatches["extent"]["x"];
+        Dataset particlePatchesDS(determineDatatype<float>(), {0});
+        particlePatchesDS.joinedDimension = 0;
```
guj: Is a sanity check needed here? do we want to throw error if someone set an invalid value of joinedDimension?

guj (approved):

test/ParallelIOTest.cpp
franzpoeschel: Yep, good catch

test/ParallelIOTest.cpp
franzpoeschel: We will probably switch to the same API as ADIOS2, i.e. have some magic number `JOINED_DIMENSION=std::numeric_limits<>::max` and then use something like `Dataset ds(<type>, {JOINED_DIMENSION, 128 , 128})`. We should add a sanity check there.

franzpoeschel (commented):
see inline comments

include/openPMD/RecordComponent.hpp
```diff
@@ -434,6 +434,11 @@ class RecordComponent : public BaseRecordComponent
     std::shared_ptr<internal::RecordComponentData> m_recordComponentData{
         new internal::RecordComponentData()};
 
+    template <typename T>
+    void verifyChunk(Offset const &, Extent const &) const;
+
+    void verifyChunk(Datatype, Offset const &, Extent const &) const;
```
franzpoeschel: Earlier versions of this PR added additional calls to `RecordComponent::storeChunk`, so instead of duplicating the verification logic, I extracted it. 
Now no longer necessary, but it's still a useful refactoring.

test/ParallelIOTest.cpp
franzpoeschel: The constructor of `Dataset` now verifies this.

src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -260,6 +260,12 @@ void JSONIOHandlerImpl::createDataset(
             "[JSON] Creating a dataset in a file opened as read only is not "
             "possible.");
     }
+    if (parameter.joinedDimension.has_value())
+    {
+        error::throwOperationUnsupportedInBackend(
+            "ADIOS1", "Joined Arrays currently only supported in ADIOS2");
```
ax3l, line 266:
```suggestion
            "JSON", "Joined Arrays currently only supported in ADIOS2");
```

src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -452,6 +452,12 @@ void HDF5IOHandlerImpl::createDataset(
             "[HDF5] Creating a dataset in a file opened as read only is not "
             "possible.");
 
+    if (parameters.joinedDimension.has_value())
+    {
+        error::throwOperationUnsupportedInBackend(
+            "ADIOS1", "Joined Arrays currently only supported in ADIOS2");
```
ax3l, line 452:
```suggestion
            "HDF5", "Joined Arrays currently only supported in ADIOS2");
```

ax3l:
This looks great! Potentially can you add a Python test, to make sure the new enum exposed to the bindings works well?

docs/source/usage/workflow.rst
```diff
@@ -3,6 +3,49 @@
...
+   Most commonly, chunks within an n-dimensional dataset are identified by their offset and extent.
+   The extent is the size of the chunk in each dimension, NOT the absolute coordinate within the entire dataset.
+
+   In the Python API, this is modeled to conform to the conventional ``__setitem__``/``__getitem__`` protocol.
+
+2. **Joined arrays (write only)**
+
+   (Currently) only supported in ADIOS2 no older than v2.9.0 under the conditions listed in the `ADIOS2 documentation on joined arrays <https://adios2.readthedocs.io/en/latest/components/components.html#shapes>`_.
+
+   In some cases, the concrete chunk within a dataset does not matter and the computation of indexes is a needless computational and mental overhead.
```
ax3l, line 20:
```suggestion
   In some cases, the concrete chunk within a dataset does not matter and the computation of indexes is a needless computational, (MPI) communication and mental overhead.
```

examples/5_write_parallel.py
```diff
@@ -64,7 +76,15 @@
...
 
     # example shows a 1D domain decomposition in first index
-    mymesh[comm.rank*10:(comm.rank+1)*10, :] = local_data
+
+    if USE_JOINED_DIMENSION:
+        # explicit API
+        # mymesh.store_chunk(local_data, [], [10, 300])
+        mymesh[:, :] = local_data
+        # or short:
+        # mymesh[()] = local_data
```
ax3l, line 85: Fancy!
Good examples even in the commented code.

ax3l (approved):
Looks great!!

PR #1386, opened 2023-03-06
Close HFD5 handles in availableChunks task

franzpoeschel:
I forgot this back then. When using `availableChunks` in parallel code, this leads to errors when closing HDF5.

Follow up to #802 and #1035


ax3l (approved):
Thanks! :)

PR #1384, opened 2023-03-01
CMake: Multi-Config Generator

ax3l:
Set target properties for multi-config generators.

Fix #1383

Regression to #1326

- [x] test in https://github.com/ax3l/cmake-easyinstall/pull/12

PR #1326, opened 2022-11-03
CMake: Fix Build & Install Option Names

ax3l:
We should not overwrite `CMAKE_` variables but instead set our own. This corrects this.

This also fixes that the cmake install path was unset in superbuilds failing the configure steps unless set by the calling project did set `CMAKE_INSTALL_CMAKEDIR` (or turned `openPMD_INSTALL=OFF`):
```
CMake Error at openpmd-src/CMakeLists.txt:1080 (install):
  install EXPORT given no DESTINATION!

CMake Error at openpmd-src/CMakeLists.txt:1085 (install):
  install FILES given no DESTINATION!
```
seen with https://github.com/fnalacceleratormodeling/synergia2/pull/100

Follow-up to #1313

cc @DerNils-git @s-sajid-ali

PR #1393, opened 2023-03-26
CMake: Fix Python Install Directory

ax3l:
Fix regressions in 0.15.0 that showed up during packaging.

Regression to #1384 #1326 #1313


ax3l:
Tested in  #1392 :heavy_check_mark:

PR #1395, opened 2023-03-26
macOS AppleClang12 Fixes

ax3l:
## CI: macOS-10.15

CI already runs on macOS-12. Adds an older runner.
Needs to update to `macOS-11` soon.

## Fix: Older libc++ (e.g. AppleClang, LLVM Clang <7)

```
    In file included from /Users/runner/work/openPMD-api/openPMD-api/src/include/openPMD/RecordComponent.hpp:467:
    /Users/runner/work/openPMD-api/openPMD-api/src/include/openPMD/RecordComponent.tpp:381:20: error: no matching constructor for initialization of 'std::shared_ptr<char []>'
                return std::shared_ptr< T[] >{ new T[ size ] };
                       ^                     ~~~~~~~~~~~~~~~~~
```

Issue seen with AppleClang 12.0 and not seen with 14.0.
Assume 13 might be affected, too.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -51,9 +51,8 @@ jobs:
     - uses: actions/checkout@v3
     - name: Install
       run: |
-        rm -rf /usr/local/bin/2to3
         set +e
-        python3 -m pip install -U mpi4py numpy pandas
+        python3 -m pip install -U numpy pandas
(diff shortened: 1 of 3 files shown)


ax3l:
tested here with new CI runner and in #1392 :heavy_check_mark:

PR #1396, opened 2023-03-26
ADIOS1 on macOS

ax3l:
## CI: macOS ADIOS1

Ensure symbol exports are properly set.

## Fix

Disable runtime error recovery in ADIOS1, include missing symbols, avoid globals in `JSON.cpp` options.

First seen on https://github.com/conda-forge/openpmd-api-feedstock/pull/99

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -43,16 +43,24 @@ jobs:
         cmake --build build --parallel 2
         ctest --test-dir build --verbose
 
-  appleclang12_py:
+  appleclang12_py_ad1:
     runs-on: macos-10.15
     # next: macOS-11
     if: github.event.pull_request.draft == false
     steps:
     - uses: actions/checkout@v3
     - name: Install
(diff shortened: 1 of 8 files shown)


ax3l:
triggering CI

ax3l:
The `Linux / clang14_py311_nompi_h5_ad1_ad2` test seems to reproducibly fail on this.

franzpoeschel:
The issue is caused by serial builds with ADIOS1

franzpoeschel:
Ok, for some reason, when building serial openPMD-api with ADIOS1, linking the parallel ADIOS1 stub library against JSON.cpp causes this issue... ??

franzpoeschel:
Aaah..
We have a global in that file:
```cpp
std::vector<std::string> backendKeys{"adios1", "adios2", "json", "hdf5"};
```

And linking that against both ADIOS1 implementations makes them destroy that twice

ax3l:
Great catch, thank you for your help!
I'll squash this for easier backportability.

PR #1403, opened 2023-03-27
Enable clang-format also for .tpp files by using a regex instead of a predefined filter

franzpoeschel:
This makes the precommit bot recognize the file. Easier than trying to figure out how to make it accept `.tpp` too...

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,6 +69,8 @@ repos:
   rev: v16.0.0
   hooks:
   - id: clang-format
+    'types_or': []
+    files: .*\.(tpp|hpp|cpp)$
 
 # Autoremoves unused Python imports
 - repo: https://github.com/hadialqattan/pycln
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,6 +69,12 @@ repos:
   rev: v16.0.0
   hooks:
   - id: clang-format
+    # By default, the clang-format hook configures:
+    # 'types_or': [c++, c, c#, cuda, java, javascript, json, objective-c, proto, textproto]
+    # Unfortunately, the c++ option does not recognize .tpp files, so we need to do this manually
+    # Since file filters in pre-commit work by logical AND, it's only possible to narrow the filter definition
+    # So, we add a regex for the type of file that we want and additionally disable the 'types_or'
+    # option entirely.
     'types_or': []
     files: .*\.(tpp|hpp|cpp)$
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -76,7 +76,7 @@ repos:
     # So, we add a regex for the type of file that we want and additionally disable the 'types_or'
     # option entirely.
     'types_or': []
-    files: .*\.(tpp|hpp|cpp)$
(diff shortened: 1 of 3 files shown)


ax3l:
> makes the precommit bot recognize the file

You mean for `clang-format`? There is probably an option we can set for additional, custom file endings?

franzpoeschel:
Ok, I tried deactivating the automatic pre-commit file type recognition and instead using a regex for the same purpose:
```yaml
    'types_or': []
    files: .*\.(tpp|hpp|cpp)$
```
Locally, that works, but the bot is not reacting?

ax3l:
It looks like it is working but does not push the change back on the PR :thinking:

ax3l:
> push (skipped)
> a conflict occurred when applying fixes to the pr. try pulling the upstream branch

Maybe try to rebase against `dev`. Will do that now.

ax3l:
Here we go :)

.pre-commit-config.yaml
```diff
@@ -69,6 +69,8 @@ repos:
   rev: v16.0.0
   hooks:
   - id: clang-format
+    'types_or': []
```
franzpoeschel, line 78:
```suggestion
    # By default, the clang-format hook configures:
    # 'types_or': [c++, c, c#, cuda, java, javascript, json, objective-c, proto, textproto]
    # Unfortunately, the c++ option does not recognize .tpp files, so we need to do this manually
    # Since file filters in pre-commit work by logical AND, it's only possible to narrow the filter definition
    # So, we add a regex for the type of file that we want and additionally disable the 'types_or' 
    # option entirely.
    'types_or': []
```

.pre-commit-config.yaml
```diff
@@ -69,6 +69,14 @@ repos:
...
   hooks:
   - id: clang-format
+    # By default, the clang-format hook configures:
+    # 'types_or': [c++, c, c#, cuda, java, javascript, json, objective-c, proto, textproto]
+    # Unfortunately, the c++ option does not recognize .tpp files, so we need to do this manually
+    # Since file filters in pre-commit work by logical AND, it's only possible to narrow the filter definition
+    # So, we add a regex for the type of file that we want and additionally disable the 'types_or' 
+    # option entirely.
+    'types_or': []
+    files: .*\.(tpp|hpp|cpp)$
```
franzpoeschel: Are there any other file extensions that we should add here? json?

franzpoeschel:
~~Ah wait, some commits got lost during rebase~~
Fixed that now

ax3l:
Yes, I removed & squashed them intentionally as they were triggers and the old experiment.

.pre-commit-config.yaml
ax3l: Doesn't it already find json? I seem to remember it did. Let's check again if it still does with the filter set.

franzpoeschel: It previously did, but this new configuration entirely replaces the old one, as there is no way to just extend it.
So, if we want JSON (do we?), we need to add it to the regex.

ax3l: Did a quick search:
```suggestion
    files: .*\.(tpp|h|hpp|hpp\.in|cpp|cxx|js|json)$
```

PR #1404, opened 2023-03-27
Fix docs post 0.15

franzpoeschel:
1. Typos in JSON/TOML configs for ADIOS1
2. Update best practises for ADIOS2


ax3l (approved):
Thank you :)

PR #1408, opened 2023-03-29
CI: Test on Ubuntu 20.04

ax3l:
Transition the Ubuntu OS from 18 to 20 LTS.
Time to finally drop clang 6 :tada: (as already advertised in our requirements) - a matching libc++ is not available in Ubuntu 20.04 anymore.

GH Actions will remove Ubuntu 18 on April 1st, 2023.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -7,26 +7,26 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  clang6_nopy_nompi_h5_libcpp:
-    runs-on: ubuntu-18.04
+  clang7_nopy_nompi_h5_libcpp:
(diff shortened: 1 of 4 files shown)


.github/workflows/linux.yml
```diff
@@ -40,20 +40,20 @@ jobs:
         ctest --test-dir build --output-on-failure
 
   clang6_nopy_ompi_h5_ad1_ad2_bp3_libcpp:
```
ax3l:
```suggestion
  clang7_nopy_ompi_h5_ad1_ad2_bp3_libcpp:
```

ax3l:
Wow, libc++ in LLVM7 also affected by the shared ptr issue.
Fix via #1409

PR #1426, opened 2023-04-18
Add JSON schema

franzpoeschel:
TODO:

- [x] Define particlePatches
- [x] Fill in things like "description", "type", "title", ...
- [x] ~~Maybe add support for abbreviated JSON representations~~ Not now
- [x] Dataset layout

I have added a CLI tool `openpmd-convert_json_toml`, exposing internally existing functionality to convert between TOML and JSON. I think it's better to define and maintain the schema in terms of TOML (it can be commented and is more legible), and just compile it to JSON for verification.

x-ref #1436


.github/workflows/linux.yml
```diff
@@ -253,6 +253,22 @@ jobs:
         cmake --build build --parallel 2
         ctest --test-dir build --output-on-failure
 
+        python3 -m pip install jsonschema
```
ax3l: Extra section might be cleaner:
```suggestion
    - name: Schema Check
      run: |
        python3 -m pip install jsonschema
```

.github/workflows/linux.yml
```diff
@@ -253,6 +253,22 @@ jobs:
...
+        cd share/openPMD/json_schema
+        PATH="../../../build/bin:$PATH" make -j 2
+        # We need to exclude the thetaMode example since that has a different
+        # meshesPath and the JSON schema needs to hardcode that.
```
ax3l, line 282: Are we able to patch this in `check.py`?

ax3l (commented):
Great work, thank you!

Wondering a bit if validation of an XML backend would be easier.... not sure.

franzpoeschel:
> Wondering a bit if validation of an XML backend would be easier.... not sure.

The major issue of the JSON schema approach is that it is not a predictive parser (https://en.wikipedia.org/wiki/Recursive_descent_parser), meaning that it deals with `{"anyOf": [<option1>, <option2>, ...]}`-like statements very naively. If there is an error somewhere deep down the hierarchy, the JSON schema just sees that all options have failed and cannot give any more details where the failure was. 

If we want to seriously consider XML as an alternative, my opinion is that it must at least solve this particular issue .. which I doubt it can properly, since this involves a bit of parser theory and grammar transformation.

This job is better done by handwritten parsers, which the openPMD-api and the openPMD-validator are.

However, if we add an XML backend at some point, an XML schema is definitely at least worth a consideration.

.github/workflows/linux.yml
franzpoeschel: Not very easily. The JSON schema is on the file system and the single .json files refer to each other by their file names. Changing this would require (1) traversing the entire JSON schema and overriding the meshes path, the particles path and the references and (2) somehow setting up python-jsonschema to cross-reference in-memory schemas which I don't even know if it supports that, both at runtime of `check.py`.

share/openPMD/json_schema/README.md
```diff
@@ -0,0 +1,47 @@
...
+### Generating the JSON schema
+
+For improved readability, maintainability and documentation purposes, the JSON schema is written in `.toml` format and needs to be "compiled" to `.json` files first before usage.
+To do this, the openPMD-api installs a tool named `openpmd-convert-json-toml` which can be used to convert between JSON and TOML files in both directions, e.g.:
+
+```bash
+openpmd_convert-json-toml @series.toml > series.json
+```
+
+A `Makefile` is provided in this folder to simplify the application of this conversion tool.
```
ax3l: Makefile? Isn't this just built with CMake as one of the CLI tools?

ax3l: Oh, I see. The makefile does the conversion of the schema toml files

ax3l:
```suggestion
A `Makefile` is provided in this folder to simplify the application of this conversion tool to the `.toml` files in this folder.
```

CMakeLists.txt
```diff
@@ -685,11 +685,12 @@ set(openPMD_TEST_NAMES
 # command line tools
 set(openPMD_CLI_TOOL_NAMES
     ls
+    convert-json-toml
 )
 set(openPMD_PYTHON_CLI_TOOL_NAMES
     pipe
 )
-set(openPMD_PYTHON_CLI_MODULE_NAMES ${openPMD_CLI_TOOL_NAMES})
+set(openPMD_PYTHON_CLI_MODULE_NAMES ls)
```
ax3l, line 693: This line change looks like a hack?

CMakeLists.txt
```diff
@@ -685,11 +685,12 @@ set(openPMD_TEST_NAMES
 # command line tools
 set(openPMD_CLI_TOOL_NAMES
     ls
+    convert-json-toml
```
ax3l: Also add this to `setup.py` -> `entry_points` -> `console_scripts`?

share/openPMD/json_schema/Makefile
```diff
@@ -0,0 +1,15 @@
+convert := openpmd-convert-json-toml
```
ax3l: Naming it `openpmd-convert-toml-json` seems to be more intuitive for what we use it here, but you mentioned anyway that it works bi-directional?

share/openPMD/json_schema/README.md
```diff
@@ -0,0 +1,47 @@
...
+./check.py path/to/my/dataset.json
+```
+
+For further usage notes check the documentation of the script itself `./check.py --help`.
+
+## Caveats
+
+The openPMD standard is not entirely expressible in terms of a JSON schema:
+
+* Many semantic dependencies, e.g. that the `position/x` and `position/y` vector of a particle species be of the same size, or that the `axisLabels` have the same dimensionality as the dataset itself, will go unchecked.
```
ax3l:
```suggestion
* Many semantic dependencies, e.g., that the `position/x` and `position/y` vectors of a particle species need to be of the same size, or that the `axisLabels` have the same dimensionality as the dataset itself, will go unchecked.
```

share/openPMD/json_schema/README.md
```diff
@@ -0,0 +1,47 @@
...
+* The `meshesPath` is assumed to be `meshes/` and the `particlesPath` is assumed to be `particles/`. This dependency cannot be expressed.
+
+While a large part of the openPMD standard can indeed be verified by checking against a JSON schema, the standard is generally large enough to make this approach come to its limits. Verification of a JSON schema is similar to the use of a naive recursive-descent parser. Error messages will often be unexpectedly verbose and not very informative.
```
ax3l: Might clarify a bit:
```suggestion
While a large part of the openPMD standard can indeed be verified by checking against a static JSON schema, the standard is generally large enough to make this approach come to its limits. Verification of a JSON schema is similar to the use of a naive recursive-descent parser. Error messages will often be unexpectedly verbose and not very informative.
```

share/openPMD/json_schema/README.md
```diff
@@ -0,0 +1,47 @@
...
+The layout of attributes is assumed to be that which is created by the JSON backend of the openPMD-api, e.g.:
+
+```json
+"meshesPath": {
+  "datatype": "STRING",
+  "value": "meshes/"
+}
+```
+
+Support for an abbreviated notation such as `"meshesPath": "meshes/"` is currently not (yet) available.
```
ax3l: In this section, you might want to advertise & link openPMD-validator again.

CMakeLists.txt
franzpoeschel: Not really. Until now, both `openPMD_CLI_TOOL_NAMES` and `openPMD_PYTHON_CLI_MODULE_NAMES` were identical since both contained only `openpmd-ls`. 
But they're not identical in general; and now, we additionally have `openpmd-convert-json-toml` as a CLI tool, but it is not written in Python.

share/openPMD/json_schema/Makefile
franzpoeschel: It is bi-directional, but we can flip the name, since that's the intended use here.

share/openPMD/json_schema/README.md
franzpoeschel: Maybe a bit clearer: A `Makefile` is provided in this folder to automate generating the needed JSON files from the TOML files.
?

share/openPMD/json_schema/README.md
franzpoeschel: Will do. We might additionally advertise the JSON schema in the validator. Adding the JSON validation to our CI brought up a number of bugs after all.

CMakeLists.txt
franzpoeschel: This is not a Python module, but written in C++, so I doubt that will work?

ax3l (approved):
LGTM :+1:

PR #1432, opened 2023-05-03
Custom Hierarchies

franzpoeschel:
The openPMD standard works by defining "what must be there", but does not impose restrictions as to "what must not be there". By this principle, openPMD is an extensible standard.
So far, standard extensions relied mostly on defining additional metadata in terms of attributes, e.g. for storing the name of the employed field solver for the ED-PIC extension. Custom hierarchies and custom n-dimensional datasets ("heavy" data in comparison to lightweight metadata) have not been employed so far despite the theoretical possibility to do so, granted by the openPMD standard. The major hindrance to such data organization has been the lacking support at the level of the openPMD-api, i.e. the implementation of the standard.

As the first part of this PR, the openPMD-api now supports writing custom-defined hierarchies and datasets within the basepath, i.e. within Iterations. This change is entirely independent from the standard as it makes use of the already existing liberty within the standard's conception as explained in the introduction.

This alone finds useful applications already:

* Data that has been marked up according to another standard can be embedded side-by-side with openPMD-formatted particle-mesh data. A [short example](https://github.com/franzpoeschel/openPMD-api/blob/topic-custom-hierarchies/examples/14_custom_hierarchy.py) is given as part of this PR that writes an openPMD-formatted temperature mesh side by side with a simple NeXus example. The resulting dataset is shown below:
    ```
      string       /basePath                                                attr   = "/data/%T/"
      string       /date                                                    attr   = "2024-08-12 16:58:01 +0200"
      string       /iterationEncoding                                       attr   = "groupBased"
      string       /iterationFormat                                         attr   = "/data/%T/"
      string       /meshesPath                                              attr   = "meshes/"
      string       /openPMD                                                 attr   = "1.1.0"
      uint32_t     /openPMDextension                                        attr   = 0
      string       /software                                                attr   = "openPMD-api"
      string       /softwareVersion                                         attr   = "0.16.0-dev"
      double       /data/100/dt                                             attr   = 1
      double       /data/100/time                                           attr   = 0
      double       /data/100/timeUnitSI                                     attr   = 1
      string       /data/100/Scan/NX_class                                  attr   = "NXentry"
      string       /data/100/Scan/data/NX_class                             attr   = "NXdata"
      string       /data/100/Scan/data/axes                                 attr   = {"two_theta"}
      int64_t      /data/100/Scan/data/counts                               {15} = 0 / 0
      string       /data/100/Scan/data/counts/long_name                     attr   = "photodiode counts"
      string       /data/100/Scan/data/counts/units                         attr   = "counts"
      string       /data/100/Scan/data/signal                               attr   = "counts"
      double       /data/100/Scan/data/two_theta                            {15} = 0 / 0
      string       /data/100/Scan/data/two_theta/long_name                  attr   = "two_theta (degrees)"
      string       /data/100/Scan/data/two_theta/units                      attr   = "degrees"
      uint8_t      /data/100/Scan/data/two_theta_indices                    attr   = {0}
      string       /data/100/Scan/default                                   attr   = "data"
      double       /data/100/meshes/temperature                             {5, 5} = 0 / 0
      string       /data/100/meshes/temperature/axisLabels                  attr   = {"x", "y"}
      string       /data/100/meshes/temperature/dataOrder                   attr   = "C"
      string       /data/100/meshes/temperature/geometry                    attr   = "cartesian"
      double       /data/100/meshes/temperature/gridGlobalOffset            attr   = {0, 0}
      double       /data/100/meshes/temperature/gridSpacing                 attr   = {1, 1}
      double       /data/100/meshes/temperature/gridUnitSI                  attr   = 1
      long double  /data/100/meshes/temperature/position                    attr   = {0.5, 0.5}
      float        /data/100/meshes/temperature/timeOffset                  attr   = 0
      double       /data/100/meshes/temperature/unitDimension               attr   = {0, 0, 1, 0, 0, 0, 0}
      double       /data/100/meshes/temperature/unitSI                      attr   = 1
    ```
* Embedding non-physical information into output files. An example is the particle-in-cell simulation PIConGPU that uses openPMD for regular output as well as for checkpoint-restart output. In the case of checkpoint-restart, internal program state must be serialized along with the physical state of the simulation, currently only possible by pretending that the internal state is a mesh which confuses many post-processing tools such as visualizers. PIConGPU has been adapted to make use of this change on [this Git tree](https://github.com/franzpoeschel/picongpu/tree/topic-custom-hierarchies), check [here](https://github.com/franzpoeschel/picongpu/compare/ad02340ee5cfed7fe3e015af6291f3277b150b29..87b0f999400e446a8326d40b923ed974f79e7123) for a diff. A shortened example output is pasted below, demonstrating that internal state information is now cleanly separated from physical data:
    ```
      float     /data/100/fields/E/x                                      {192, 1024, 192}
      float     /data/100/fields/E/y                                      {192, 1024, 192}
      float     /data/100/fields/E/z                                      {192, 1024, 192}
      float     /data/100/particles/e/momentum/x                          {71958528}
      float     /data/100/particles/e/momentum/y                          {71958528}
      float     /data/100/particles/e/momentum/z                          {71958528}
      float     /data/100/particles/e/position/x                          {71958528}
      float     /data/100/particles/e/position/y                          {71958528}
      float     /data/100/particles/e/position/z                          {71958528}
      int32_t   /data/100/particles/e/positionOffset/x                    {71958528}
      int32_t   /data/100/particles/e/positionOffset/y                    {71958528}
      int32_t   /data/100/particles/e/positionOffset/z                    {71958528}
      float     /data/100/particles/e/weighting                           {71958528}
      char      /data/100/picongpu_internal/RNG/RNGProvider3XorMin        {48, 128, 147456}
      uint64_t  /data/100/picongpu_internal/idProvider/nextId             {1, 1, 1}
      uint64_t  /data/100/picongpu_internal/idProvider/startId            {1, 1, 1}
    ```

Building on top of this, the other logical component of this PR consists in the support of [this standard extension](https://github.com/openPMD/openPMD-standard/pull/282). While the PR as described so far brings custom hierarchies and datasets to the openPMD-api in a way that is transparent to the standard itself, the purpose of this next standard extension is to now make the standard aware of these hierarchies by embedding openPMD markup within them.

The schematic idea behind this is pictured below:
267274652-a4a4a4ac-636f-4349-bc14-c4e4a2cc36a1

With this, the data organization can step back into openPMD markup from anywhere within a custom-defined hierarchy. This further extends the use of this PR to:

* Using openPMD markup within another standard, rather than merely beside it. This is currently being applied exploratively in [this script](https://github.com/NeXus-for-HELPMI/POLARIS_ExpData_Example/blob/add_openpmd_layer/add_openpmd_layer.py) for a sample dataset collected in the [POLARIS](https://de.wikipedia.org/wiki/Polaris_(Laser)) laboratory.
* For more complex setups, this permits a better organization of output data. As an example, meshes can be of different kinds such as 3-dimensional physical fields or 2-dimensional images; also there might be similar kinds of dependencies between particle data. It is desirable to group such data in a way that reflects the logical adjacencies and interdependencies between them.
* A particular instance of the above is mesh refinement, currently proposed in [a standard extension](https://github.com/openPMD/openPMD-standard/pull/252) as a suffix-based naming scheme. Switching to an approach based on custom hierarchies, [this comment](https://github.com/openPMD/openPMD-standard/issues/115#issuecomment-1253991631) details a more natural and more easily parsed approach at mesh refinement. A mesh-refined dataset of this type might be structured as follows:
    ```
    /data/0/refined_mesh_levels/0/meshes/E
    /data/0/refined_mesh_levels/0/meshes/B
    /data/0/refined_mesh_levels/1/meshes/E
    /data/0/refined_mesh_levels/1/meshes/B
    /data/0/refined_mesh_levels/2/meshes/E
    /data/0/refined_mesh_levels/2/meshes/B
    +++++++ ––––––––––––––––––––– ++++++++
    standard        custom        standard
    
    /data/0/simulation_internal/some_checkpointing_info
    +++++++ –––––––––––––––––––––––––––––––––––––––––––
    standard                  custom
    ```

TODO

- [x] Merge first: #1154 
- [x] Await Pybind11 release that has merged this fix: https://github.com/pybind/pybind11/pull/4623
- [x] Implement custom groups at the Iteration level that can hold custom attributes
- [x] Implement custom datasets inside custom hierarchy 
- [x] Implement openPMD-defined meshes/particles-data from anywhere in the hierarchy
- [x] Implement extended meshesPath/particlesPath
- [ ] Update the openPMD standard, see https://github.com/openPMD/openPMD-standard/issues/115#issuecomment-1253991631
- [x] Lenient parsing in CustomHierarchy class
- [ ] Maybe lazy parsing of the custom hierarchy?
- [ ] Use the new `SharedAttributableData` pattern to better implement variable-based encoding (where `series.iterations` and `series.iterations[0]` are the same backend objects)
- [ ] Replace `Iteration::meshes` with `Iteration::mesh("subdir/E")` and `Iteration::allMeshes() -> std::map<std::string, Mesh>`, similar `Iteration::species("subdir/e")` and `Iteration::allSpecies() -> std::map<std::string, ParticleSpecies>`. But should it be `species("subdir/particles/e")` or `species("subdir/e")`?
- [ ] Generalize to `Attributable::openAsCustomHierarchy()`?

Diff: https://github.com/franzpoeschel/openPMD-api/compare/topic-remove-scalar-component..topic-custom-hierarchies

### DIFF
--- a/src/CustomHierarchy.cpp
+++ b/src/CustomHierarchy.cpp
@@ -1,0 +1,672 @@
+/* Copyright 2023 Franz Poeschel
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * openPMD-api is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License and the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "openPMD/CustomHierarchy.hpp"
+
+#include "openPMD/Dataset.hpp"
+#include "openPMD/Error.hpp"
+#include "openPMD/IO/AbstractIOHandler.hpp"
+#include "openPMD/IO/Access.hpp"
+#include "openPMD/IO/IOTask.hpp"
+#include "openPMD/Mesh.hpp"
+#include "openPMD/ParticleSpecies.hpp"
+#include "openPMD/RecordComponent.hpp"
+#include "openPMD/Series.hpp"
+#include "openPMD/auxiliary/StringManip.hpp"
+#include "openPMD/backend/Attributable.hpp"
+#include "openPMD/backend/BaseRecord.hpp"
+#include "openPMD/backend/MeshRecordComponent.hpp"
+#include "openPMD/backend/Writable.hpp"
+
+#include <algorithm>
+#include <deque>
+#include <initializer_list>
+#include <iostream>
+#include <iterator>
+#include <map>
+#include <optional>
+#include <regex>
+#include <sstream>
+#include <stdexcept>
+#include <string>
+#include <tuple>
+#include <variant>
+
+// @todo add handselected choice of [:punct:] characters to this
+// using a macro here to make string interpolation simpler
+#define OPENPMD_LEGAL_IDENTIFIER_CHARS "[:alnum:]_"
+#define OPENPMD_SINGLE_GLOBBING_CHAR "%"
+#define OPENPMD_DOUBLE_GLOBBING_CHAR "%%"
+
+namespace
+{
+template <typename Iterator>
+std::string
+concatWithSep(Iterator &&begin, Iterator const &end, std::string const &sep)
+{
+    if (begin == end)
+    {
+        return "";
+    }
+    std::stringstream res;
+    res << *(begin++);
+    for (; begin != end; ++begin)
+    {
+        res << sep << *begin;
+    }
+    return res.str();
+}
+
+std::string
+concatWithSep(std::vector<std::string> const &v, std::string const &sep)
+{
+    return concatWithSep(v.begin(), v.end(), sep);
+}
+
+// Not specifying std::regex_constants::optimize here, only using it where
+// it makes sense to.
+constexpr std::regex_constants::syntax_option_type regex_flags =
+    std::regex_constants::egrep;
+
+template <typename OutParam>
+void setDefaultMeshesParticlesPath(
+    std::vector<std::string> const &meshes,
+    std::vector<std::string> const &particles,
+    OutParam &writeTarget)
+{
+    std::regex is_default_path_specification(
+        "[" OPENPMD_LEGAL_IDENTIFIER_CHARS "]+/",
+        regex_flags | std::regex_constants::optimize);
+    constexpr char const *default_default_mesh = "meshes";
+    constexpr char const *default_default_particle = "particles";
+    for (auto [vec, defaultPath, default_default] :
+         {std::make_tuple(
+              &meshes, &writeTarget.m_defaultMeshesPath, default_default_mesh),
+          std::make_tuple(
+              &particles,
+              &writeTarget.m_defaultParticlesPath,
+              default_default_particle)})
+    {
+        bool set_default = true;
+        /*
+         * The first eligible path in meshesPath/particlesPath is used as
+         * the default, "meshes"/"particles" otherwise.
+         */
+        for (auto const &path : *vec)
+        {
+            if (std::regex_match(path, is_default_path_specification))
+            {
+                *defaultPath = openPMD::auxiliary::replace_last(path, "/", "");
+                set_default = false;
+                break;
+            }
+        }
+        if (set_default)
+        {
+            *defaultPath = default_default;
+        }
+    }
+}
+
+bool anyPathRegexMatches(
+    std::regex const &regex, std::vector<std::string> const &path)
+{
+    std::string pathToMatch = '/' + concatWithSep(path, "/") + '/';
+    return std::regex_match(pathToMatch, regex);
+}
+} // namespace
+
+namespace openPMD
+{
+namespace internal
+{
+    namespace
+    {
+        std::string globToRegexLongForm(std::string const &glob)
+        {
+            return auxiliary::replace_all(
+                auxiliary::replace_all(
+                    glob,
+                    OPENPMD_DOUBLE_GLOBBING_CHAR,
+                    "([" OPENPMD_LEGAL_IDENTIFIER_CHARS "/]*)"),
+                OPENPMD_SINGLE_GLOBBING_CHAR,
+                "([" OPENPMD_LEGAL_IDENTIFIER_CHARS "]*)");
+        }
+
+        std::string globToRegexShortForm(std::string const &glob)
+        {
+            return "[" OPENPMD_LEGAL_IDENTIFIER_CHARS "/]*/" + glob;
+        }
+    } // namespace
+
+    MeshesParticlesPath::MeshesParticlesPath(
+        std::vector<std::string> const &meshes,
+        std::vector<std::string> const &particles)
+    {
+        /*
+         * /group/meshes/E is a mesh if the meshes path contains:
+         *
+         * 1) '/group/meshes/' (absolute path to mesh container)
+         * 2) 'meshes/' (relative path to mesh container)
+         *
+         * All this analogously for particles path.
+         */
+
+        // regex for detecting option 1)
+        // e.g. '/path/to/meshes/': The path to the meshes. Mandatory slashes at
+        //                          beginning and end, possibly slashes in
+        //                          between. Mandatory slash at beginning might
+        //                          be replaced with '%%' to enable paths like
+        //                          '%%/path/to/meshes'.
+        // resolves to: `(/|%%)[[:alnum:]_%/]+/`
+        std::regex is_legal_long_path_specification(
+            "(/|" OPENPMD_DOUBLE_GLOBBING_CHAR
+            ")[" OPENPMD_LEGAL_IDENTIFIER_CHARS OPENPMD_SINGLE_GLOBBING_CHAR
+            "/]+/",
+            regex_flags | std::regex_constants::optimize);
+
+        // Regex for detecting option 2)
+        // e.g. 'meshes/': The name without path. One single mandatory slash
+        //                 at the end, no slashes otherwise.
+        // resolves to `[[:alnum:]_]+/`
+        std::regex is_legal_short_path_specification(
+            "[" OPENPMD_LEGAL_IDENTIFIER_CHARS "]+/",
+            regex_flags | std::regex_constants::optimize);
+
+        for (auto [target_regex, vec] :
+             {std::make_tuple(&this->meshRegex, &meshes),
+              std::make_tuple(&this->particleRegex, &particles)})
+        {
+            std::stringstream build_regex;
+            // neutral element: empty language, regex doesn't match anything
+            build_regex << "(a^)";
+            for (auto const &entry : *vec)
+            {
+                if (std::regex_match(entry, is_legal_short_path_specification))
+                {
+                    build_regex << "|(" << globToRegexShortForm(entry) << ')';
+                }
+                else if (std::regex_match(
+                             entry, is_legal_long_path_specification))
+                {
+                    build_regex << "|(" << globToRegexLongForm(entry) << ')';
+                }
+                else
+                {
+                    std::cerr
+                        << "[WARNING] Not a legal meshes-/particles-path: '"
+                        << entry << "'. Will skip." << std::endl;
+                }
+            }
+            auto regex_string = build_regex.str();
+            // std::cout << "Using regex string: " << regex_string << std::endl;
+            *target_regex = std::regex(
+                regex_string, regex_flags | std::regex_constants::optimize);
+        }
+        setDefaultMeshesParticlesPath(meshes, particles, *this);
+    }
+
+    ContainedType MeshesParticlesPath::determineType(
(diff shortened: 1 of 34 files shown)


franzpoeschel:
comment removed, updated version in comments below

src/CustomHierarchy.cpp
```diff
@@ -626,4 +596,95 @@ template auto CustomHierarchy::asContainerOf<RecordComponent>()
...
+                    embeddedAccessor(it->second)->m_containerData.get() !=
+                        alias.m_containerData.get())
+                {
+                    /*
+                     * This might happen if a user first creates a custom group
+                     * "fields" and sets the default meshes path as "fields"
+                     * only later.
+                     * If the CustomHierarchy::meshes alias carries no data yet,
+                     * we can just redirect it to that group now.
+                     * Otherwise, we need to fail.
```
franzpoeschel: Scrap this, we should always fail. We can redirect the internal `meshes` handle, but not ones that the user might have as a variable. The joys of dealing with a public member field..

franzpoeschel:
For the meshesPath (equivalently for particlesPath), I have now implemented a prototype that does the following:

A path `/data/0/custom/group/meshes/E` is a mesh if the `meshesPath` contains any of the following:

1. **Full path to the group containing the mesh**: `/custom/group/meshes/`
2. ~~**Full path to the mesh itself**: `/custom/group/meshes/E`~~ No longer supported
3. **Shorthand notation**: `meshes/`

The underlying rule: Full paths are denoted by a leading slash and are based on the data path (`/data/%T`)

Remark: The shorthand notation achieves backwards compatibility with old openPMD files

franzpoeschel:
One nontrivial design question is how to deal with the traditional openPMD hierarchy, especially with the paths `/data/%T/meshes` and `/data/%T/particles`. There is no definition of any form of physical data for those groups in the openPMD standard, a normal openPMD file contains no attributes `/data/%T/meshes/<attr_name>`. 

This suggests to me that in the extended openPMD standard with custom hierarchies these paths should be treated as "nothing special". Rather, they become the canonical, but not mandatory layout/organization of a simple openPMD dataset.

Two somewhat tricky consequences from this point of view:

**1. There might be more than 1 meshes paths in the same group** 
E.g. the paths `/data/%T/meshes` and `/data/%T/images` might exist side by side. In the openPMD standard, this is no problem, in the openPMD-api this becomes challenging. 
The problem is with the member `Iteration::meshes` (made even worse by the fact that it's not a getter method, but a data member). Should it point to `/data/%T/meshes`? To a union of both? What about writing?

Imo, the best solution is to consider `Iteration::meshes` a shorthand API that should not be used in more complex setups. Rather, since `/data/%T/meshes` is now just another normal path in the custom Iteration hierarchy, one should access `iteration["meshes"].asContainerOf<Mesh>()` for clarity.

`Iteration::meshes` will point to the first user-specified meshes path that takes the form of a shorthand notation. E.g., after `series.setMeshesPath({"fields/"})`, the call `iteration.meshes` will be the same as `iteration["fields"].asContainerOf<Mesh>()`. This ensures backwards compatibility.

(Note: Since `Iteration::meshes` is unfortunately a member and not a method, this means that the meshes path must be set before creating or opening any Iteration. And it was enough fighting with pointers to get things to that state.)

~~**2. There might be custom data inside `/data/%T/meshes`**
This is not really a problem, but could be unexpected. When setting `series.setMeshesPath({"/meshes/E"})`, you state that only the E field is a mesh. Since `/data/%T/meshes` is otherwise "just a regular group" with no special meaning, there might be other data in there, too, e.g. `/data/%T/meshes/custom/hierarchy`. It's the job of the user to create a meaningful data layout here.~~ 
With the more restricted definition of meshesPath and particlesPath, this is no longer supported.

pgrete:
What's the status/schedule here?
I'm asking as we probably need this for a similar use case (storing additional data for checkpoint files) or I need a different hint with regard to our use case.
More specifically, I want to store some parameters (scalars and vectors of int, float, double,...) that may contain to different "packages".
Our current (in other output format) paradigm has been to store them in attributes called `Params/PACKAGE_NAME/PARAM_NAME`.
As far as I understand `Params` and `PACKAGE_NAME` could be a "group" within the standard (which is logically consistent with out data model).
Writing these with the openpmd-api also works (I see the data in the output files).
However, reading does not work 
```
Reading 'Params/tracers/t_lookback' with type: St6vectorIdSaIdEE
terminate called after throwing an instance of 'openPMD::error::NoSuchAttribute'
  what():  Params/tracers/t_lookback
```
I assume that this is because the attributes in "groups" are not parsed by default.
Here's what OpenPMD sees (print the `iteration->attributes()`:
```
Contains attribute: BlocksPerPE
Contains attribute: BoundaryConditions
Contains attribute: Coordinates
Contains attribute: IncludesGhost
Contains attribute: InputFile
Contains attribute: MaxLevel
Contains attribute: MeshBlockSize
Contains attribute: Multilevel
Contains attribute: NBDel
Contains attribute: NBNew
Contains attribute: NCycle
Contains attribute: NGhost
Contains attribute: NumDims
Contains attribute: NumMeshBlocks
Contains attribute: Refine
Contains attribute: RootGridDomain
Contains attribute: RootGridSize
Contains attribute: RootLevel
Contains attribute: WallTime
Contains attribute: derefinement_count
Contains attribute: dt
Contains attribute: loc.level-gid-lid-cnghost-gflag
Contains attribute: loc.lx123
Contains attribute: time
Contains attribute: timeUnitSI
```
and here's what's in the file
```
$ bpls ../parthenon.opmd.00002.bp -A 
  string    /author                                                                attr
  string    /basePath                                                              attr
  uint8_t   /bla                                                                   attr
  string    /comment                                                               attr
  int32_t   /data/2/BlocksPerPE                                                    attr
  string    /data/2/BoundaryConditions                                             attr
  string    /data/2/Coordinates                                                    attr
  int32_t   /data/2/IncludesGhost                                                  attr
  string    /data/2/InputFile                                                      attr
  int32_t   /data/2/MaxLevel                                                       attr
  int32_t   /data/2/MeshBlockSize                                                  attr
  int32_t   /data/2/Multilevel                                                     attr
  int32_t   /data/2/NBDel                                                          attr
  int32_t   /data/2/NBNew                                                          attr
  int32_t   /data/2/NCycle                                                         attr
  int32_t   /data/2/NGhost                                                         attr
  int32_t   /data/2/NumDims                                                        attr
  int32_t   /data/2/NumMeshBlocks                                                  attr
  double    /data/2/Params/Hydro/AdiabaticIndex                                    attr
  uint8_t   /data/2/Params/Hydro/calc_c_h                                          attr
  uint8_t   /data/2/Params/Hydro/calc_dt_hyp                                       attr
  double    /data/2/Params/Hydro/cfl                                               attr
  double    /data/2/Params/Hydro/cfl_diff                                          attr
  double    /data/2/Params/Hydro/dt_diff                                           attr
  uint8_t   /data/2/Params/Hydro/first_order_flux_correct                          attr
  double    /data/2/Params/Hydro/max_dt                                            attr
  int32_t   /data/2/Params/Hydro/nhydro                                            attr
  int32_t   /data/2/Params/Hydro/nscalars                                          attr
  uint8_t   /data/2/Params/Hydro/pack_in_one                                       attr
  int32_t   /data/2/Params/Hydro/scratch_level                                     attr
  double    /data/2/Params/Hydro/turbulence/accel_rms                              attr
  int32_t   /data/2/Params/Hydro/turbulence/inject_n_blobs                         attr
  int32_t   /data/2/Params/Hydro/turbulence/inject_once_at_cycle                   attr
  double    /data/2/Params/Hydro/turbulence/inject_once_at_time                    attr
  uint8_t   /data/2/Params/Hydro/turbulence/inject_once_on_restart                 attr
  double    /data/2/Params/Hydro/turbulence/kpeak                                  attr
  int32_t   /data/2/Params/Hydro/turbulence/rescale_once_at_cycle                  attr
  double    /data/2/Params/Hydro/turbulence/rescale_once_at_time                   attr
  uint8_t   /data/2/Params/Hydro/turbulence/rescale_once_on_restart                attr
  double    /data/2/Params/Hydro/turbulence/rescale_to_rms_Ms                      attr
  uint32_t  /data/2/Params/Hydro/turbulence/rseed                                  attr
  double    /data/2/Params/Hydro/turbulence/sol_weight                             attr
  double    /data/2/Params/Hydro/turbulence/t_corr                                 attr
  uint8_t   /data/2/Params/tracers/enabled                                         attr
  int32_t   /data/2/Params/tracers/n_lookback                                      attr
  double    /data/2/Params/tracers/num_tracers_per_cell                            attr
  int32_t   /data/2/Params/tracers/rng_seed                                        attr
  double    /data/2/Params/tracers/t_lookback                                      attr
  int32_t   /data/2/Refine                                                         attr
  double    /data/2/RootGridDomain                                                 attr
  int32_t   /data/2/RootGridSize                                                   attr
  int32_t   /data/2/RootLevel                                                      attr
  double    /data/2/WallTime                                                       attr
  int32_t   /data/2/derefinement_count                                             attr
  double    /data/2/dt                                                             attr
  int32_t   /data/2/loc.level-gid-lid-cnghost-gflag                                attr
  int64_t   /data/2/loc.lx123                                                      attr
  string    /data/2/meshes/acc_acc_0_lvl0/axisLabels                               attr
  string    /data/2/meshes/acc_acc_0_lvl0/dataOrder                                attr
  string    /data/2/meshes/acc_acc_0_lvl0/geometry                                 attr
...
```

Any short/long term recommendations?

src/Iteration.cpp
```diff
@@ -488,68 +560,20 @@ void Iteration::read_impl(std::string const &groupPath)
...
 
-    /* Find the root point [Series] of this file,
-     * meshesPath and particlesPath are stored there */
     Series s = retrieveSeries();
 
     Parameter<Operation::LIST_PATHS> pList;
+    IOHandler()->enqueue(IOTask(this, pList));
     std::string version = s.openPMD();
-    bool hasMeshes = false;
-    bool hasParticles = false;
```
franzpoeschel, line 582: TODO: unused

PR #1436, opened 2023-05-08
TOML Backend

franzpoeschel:
Extracted from #1277, since that PR contains thematically related, but clearly distinct items.

TODO:

- [x] Performance testing: TOML Parsing is quite slow in toml11. As this is more for lightweight data with user interaction, this is not a huge problem.
- [x] Reduce full test suite
- [x] Docs
- [x] Merge #1473 first

### DIFF
--- a/docs/source/backends/json.rst
+++ b/docs/source/backends/json.rst
@@ -8,11 +8,12 @@ The JSON and TOML backends are always available.
 
 .. note::
 
-   Both the JSON and the TOML backends are not intended for large-scale use.
+   Both the JSON and the TOML backends are not intended for large-scale data I/O.
 
    The JSON backend is mainly intended for prototyping and learning, or similar workflows where setting up a large IO backend such as HDF5 or ADIOS2 is perceived as obstructive. It can also be used for small datasets that need to be stored in text format rather than binary.
 
-   The TOML backend is intended for handwritten configuration files. The motivation for its addition to the openPMD-api was the handwritten specification for physical quantities along the openPMD standard within the context of experiment workflows.
+   The TOML backend is intended for exchanging the *structure* of a data series without its "heavy" data fields.
+   For instance, one can easily create and exchange human-readable, machine-actionable data configurations for experiments and simulations.
 
 
 JSON File Format
--- a/docs/source/backends/json.rst
+++ b/docs/source/backends/json.rst
@@ -1,10 +1,10 @@
 .. _backends-json:
 
-JSON
-====
+JSON/TOML
+=========
 
-openPMD supports writing to and reading from JSON files.
-The JSON backend is always available.
+openPMD supports writing to and reading from JSON and TOML files.
+The JSON and TOML backends are always available.
 
 
 JSON File Format
@@ -43,9 +43,17 @@ Every such attribute is itself a JSON object with two keys:
  * ``datatype``: A string describing the type of the value.
  * ``value``: The actual value of type ``datatype``.
 
+TOML File Format
+----------------
 
-Restrictions
-------------
+A TOML file uses the file ending ``.toml``. The TOML backend is chosen by creating a ``Series`` object with a filename that has this file ending.
+
+The TOML backend internally works with JSON datasets and converts to/from TOML during I/O.
+As a result, data layout and usage are equivalent to the JSON backend.
+
+
+JSON Restrictions
+-----------------
 
 For creation of JSON serializations (i.e. writing), the restrictions of the JSON backend are
 equivalent to those of the `JSON library by Niels Lohmann <https://github.com/nlohmann/json>`_
@@ -72,6 +80,20 @@ may have originally been +/-Infinity or beyond the supported value range.
 
 Upon reading ``null`` when expecting any other datatype, the JSON backend will
 propagate the exception thrown by Niels Lohmann's library.
+
+The (keys) names ``"attributes"``, ``"data"`` and ``"datatype"`` are reserved and must not be used for base/mesh/particles path, records and their components.
+
+A parallel (i.e. MPI) implementation is *not* available.
+
+TOML Restrictions
+-----------------
+
+Note that the JSON datatype-specific restrictions do not automatically hold for TOML, as those affect only the representation on disk, not the internal representation.
+
+TOML supports most numeric types, up to long double and long long types.
+Special floating point values such as NaN are also support.
+
+TOML does not support null values.
 
 The (keys) names ``"attributes"``, ``"data"`` and ``"datatype"`` are reserved and must not be used for base/mesh/particles path, records and their components.
 
--- a/docs/source/backends/json.rst
+++ b/docs/source/backends/json.rst
@@ -90,7 +90,7 @@ TOML Restrictions
 
 Note that the JSON datatype-specific restrictions do not automatically hold for TOML, as those affect only the representation on disk, not the internal representation.
 
-TOML supports most numeric types, up to long double and long long types.
+TOML supports most numeric types, with the support for long double and long integer types being platform-defined.
 Special floating point values such as NaN are also support.
 
 TOML does not support null values.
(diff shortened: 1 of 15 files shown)


franzpoeschel:
We will not be able to activate TOML for our broad test suite as TOML serialization and deserialization are both quite slow. 
For the intended usage, this is not a large problem as TOML is intended for small, handwritten datasets.

The following example is the `write_and_read_many_iterations` test that writes a file-based Series of 1030 iterations per backend. This first profile (created with google-perftools) is without the TOML backend activated:

prof_before

With TOML backend activated, most time is spent parsing ~1000 TOML files:

prof

src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -54,9 +58,82 @@ namespace openPMD
...
+            {
+                return typename T::value_type{};
+            }
+            else
+            {
+                return T{};
+            }
+#if defined(__INTEL_COMPILER)
+/*
+ * ICPC has trouble with if constexpr, thinking that return statements are
```
ax3l, line 78: I think this is likely with EDG frontents in general and might also affect nvcc. I remember we saw this there as well and reported it at some point (should be fixed in newer >CUDA 12 versions).

src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -205,20 +278,23 @@ void JSONIOHandlerImpl::createDataset(
...
-            complexExtent.push_back(2);
-            dset["data"] = initializeNDArray(complexExtent);
+            extent.push_back(2);
             break;
         }
         default:
-            dset["data"] = initializeNDArray(parameter.extent);
             break;
         }
+        // TOML does not support nulls, so initialize with zero
```
ax3l, line 293: Interesting discussion: https://github.com/toml-lang/toml/issues/30

test/python/unittest/API/APITest.py
```diff
@@ -359,7 +359,8 @@ def attributeRoundTrip(self, file_ending):
         # c_types
         self.assertEqual(series.get_attribute("byte_c"), 30)
         self.assertEqual(series.get_attribute("ubyte_c"), 50)
-        if file_ending != "json":  # TODO: returns [100] instead of 100 in json
+        # TODO: returns [100] instead of 100 in json
```
ax3l: same issue in toml as well?

test/python/unittest/API/APITest.py
```diff
@@ -1824,7 +1825,7 @@ def testIterator(self):
                 self.makeIteratorRoundTrip(b, backend_filesupport[b])
 
     def makeAvailableChunksRoundTrip(self, ext):
-        if ext == "h5":
+        if ext == "h5" or ext == "toml":
```
ax3l, line 1828: What does not round-trip for toml (and h5)?

src/IO/JSON/JSONIOHandlerImpl.cpp
franzpoeschel: Probably the central bit of that discussion is "TOML is intended for configuration", which is exactly the use case for which we are adding the TOML backen.
If users insist on writing entire datasets with TOML, that's fine too, but it will be initialized with 0.
Otherwise, TOML is mostly intended for usage in conjunction with the follow-up #1493 which only writes the dataset metadata.

test/python/unittest/API/APITest.py
franzpoeschel: `available_chunks()` is a backend-specific feature. In HDF5, it returns the entire dataset as there is no *native* way to retrieve defined regions from an HDF5 dataset (or if there is: I don't know about it and we are not using it).
In the JSON backend, we use a [rather crude algorithm](https://github.com/openPMD/openPMD-api/blob/99daca78e9ef2207b78e922ad393cadf9a8ab019/src/IO/JSON/JSONIOHandlerImpl.cpp#L303) to recover written chunks, based on where the dataset is `null` and where it is defined. (This is what I was recently referring to when I said that @eschnett's region calculus might be relevant also in our internal code).

In TOML however, there is no null, so this does not work.

test/python/unittest/API/APITest.py
franzpoeschel: Yes. This issue is triggered by using `series.set_attribute("char_c", ctypes.c_char(100).value)` in Python which will always look like a 1D buffer  with one single element inside C++ (unlike other types that look like 0D scalars). As a result, a `VEC_CHAR` attribute write is requested from the backends.

As our HDF5 and ADIOS2 backends do not distinguish between scalars and single-element vectors, this has no consequence there. In JSON/TOML, we get a vector, not a single char.

src/IO/JSON/JSONIOHandlerImpl.cpp
franzpoeschel: So, we should change the macro to something more precise?
If yes, do you know what would be the right check?

ax3l: I think I was mumbling here and kept this for future reference, in case we see a warning from it. Nothing to do now.

src/IO/JSON/JSONIOHandlerImpl.cpp
ax3l: Agreed, that makes total sense. Should we be a bit more explicit in our guidance in `docs/source/backends/json.rst` to avoid that users mistake it as a full-fledged, high-performance data backend?

test/python/unittest/API/APITest.py
ax3l: Got it, thanks!

```suggestion
        # TODO: returns [100] instead of 100 in json/toml
```

test/python/unittest/API/APITest.py
ax3l: Ah right, this is chunk round-tripping. Sounds good!

src/IO/JSON/JSONIOHandlerImpl.cpp
franzpoeschel: Yes, I'll add something. Also, in #1493 we could think about enabling the abbreviated modes by default in TOML.

ax3l: Great idea about the default with #1493, yes! :)

docs/source/backends/json.rst
```diff
@@ -1,10 +1,18 @@
...
+=========
 
-openPMD supports writing to and reading from JSON files.
-The JSON backend is always available.
+openPMD supports writing to and reading from JSON and TOML files.
+The JSON and TOML backends are always available.
+
+.. note::
+
+   Both the JSON and the TOML backends are not intended for large-scale use.
```
ax3l: "Use" could imply "do not use it by a wide audience". Better is:
```suggestion
   Both the JSON and the TOML backends are not intended for large-scale data I/O.
```

docs/source/backends/json.rst
```diff
@@ -1,10 +1,18 @@
...
+
+   The JSON backend is mainly intended for prototyping and learning, or similar workflows where setting up a large IO backend such as HDF5 or ADIOS2 is perceived as obstructive. It can also be used for small datasets that need to be stored in text format rather than binary.
+
+   The TOML backend is intended for handwritten configuration files. The motivation for its addition to the openPMD-api was the handwritten specification for physical quantities along the openPMD standard within the context of experiment workflows.
```
ax3l:
```suggestion
   The TOML backend is intended for exchanging the *structure* of a data series without its "heavy" data fields.
   For instance, one can easily create and exchange human-readable, machine-actionable data configurations for experiments and simulations.
```

ax3l (commented):
Small suggestions on motivation.
Generalizing a bit and clarifying.

ax3l (approved):
This is great, thank you! :tada:

PR #1452, opened 2023-05-24
Python: Better error message when loading to a buffer with mismatched type

franzpoeschel:
In the Python API, when using `load_chunk()` to a preallocated buffer, the buffer is checked for being contiguous.
The calculations happen on the basis of the record component's datatype, not on the datatype of the buffer (the datatype of the buffer cannot be queried on a `py::buffer` object, just the item size).
If the datatypes are mismatched, the error message is very confusing (and in fact wrong): `[Record_Component::load_chunk()] Requires contiguous slab of memory.` 

This checks if the item size matches the datatype before doing and check for contiguous data.

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -23,6 +23,7 @@
 #include <pybind11/stl.h>
 
 #include "openPMD/DatatypeHelpers.hpp"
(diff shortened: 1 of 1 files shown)


ax3l (approved):
Thank you, that is a good check & error message.

PR #1460, opened 2023-06-13
added async write flag for ADIOS2

guj:
*please add a PR description, including links to ADIOS2 docs about the added feature*

This PR adds a new env variable OPENPMD_ADIOS2_ASYNC_WRITE. Default = 0. When set to 1,  it sets the ADIOS parameter "AsyncWrite" to "on".  In which case the BP5 engine will handle write calls asynchronously. 

See https://adios2.readthedocs.io/en/latest/engines/engines.html#bp5
for details of BP5 feature "AsyncWrite".

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -2997,7 +2997,7 @@ namespace detail
         }
         if (notYetConfigured("AsyncWrite"))
         {
-            if (1 == auxiliary::getEnvNum("OPENPMD_ADIOS2_ASYNC_WRITE", 1) &&
+            if (1 == auxiliary::getEnvNum("OPENPMD_ADIOS2_ASYNC_WRITE", 0) &&
                 notYetConfigured("AsyncWrite"))
             {
                 m_IO.SetParameter("AsyncWrite", "On");
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -2863,6 +2863,7 @@ namespace detail
         // set engine parameters
         std::set<std::string> alreadyConfigured;
         std::optional<bool> userSpecifiedUsesteps;
+        bool wasTheFlushTargetSpecifiedViaJSON = false;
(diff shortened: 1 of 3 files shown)


guj:
Hi @franzpoeschel @ax3l 

Any idea why the test fail with hdf5 on Conda_ompi_all fails on ?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2995,6 +2995,19 @@ namespace detail
                 m_IO.SetParameter("Profile", "Off");
             }
         }
+        if (notYetConfigured("AsyncWrite"))
+        {
+            if (1 == auxiliary::getEnvNum("OPENPMD_ADIOS2_ASYNC_WRITE", 1) &&
```
```suggestion
            if (1 == auxiliary::getEnvNum("OPENPMD_ADIOS2_ASYNC_WRITE", 0) &&
```

Your change would activate async write by default. This probably explains the failing CI run, too.

franzpoeschel (changes requested):
Can you please document the additional parameter in `adios2.rst`?

Also, note that in BP5, openPMD will flush to disk by default which makes no sense in combination with async I/O. So, if you add an environment variable for this, it probably makes sense to flip that default in this case.

guj:
> Can you please document the additional parameter in `adios2.rst`?
> 
Will do
> Also, note that in BP5, openPMD will flush to disk by default which makes no sense in combination with async I/O. So, if you add an environment variable for this, it probably makes sense to flip that default in this case.

Yes, I had to flip the  PerformDataWrite.  I wonder whether this should an option as well.

franzpoeschel:
> Yes, I had to flip the PerformDataWrite. I wonder whether this should an option as well.

I think that we don't need an extra option for this. I added a commit that sets the default flush target to the Buffer when using this env var.

franzpoeschel (approved):
LGTM
The failing docs PR run concerns all PRs currently and has nothing to do with this PR.

guj:
> LGTM
> The failing docs PR run concerns all PRs currently and has nothing to do with this PR.



> > Yes, I had to flip the PerformDataWrite. I wonder whether this should an option as well.
> 
> I think that we don't need an extra option for this. I added a commit that sets the default flush target to the Buffer when using this env var.

That will do. Thanks.

PR #1463, opened 2023-06-20
Add Python bindings for myPath

franzpoeschel:
This is somewhat available through `.__getstate__()` already, but not for all types and it should not be treated as a stable API.
Having this is very useful for debugging purposes.


ax3l (approved):
Great idea! :+1: 

Fixed CI in the other PR and restarted here.

PR #1467, opened 2023-06-27
Script: New Version

ax3l:
Adding a script to update all version notes programmatically for releases.

- [x] finish script
- [x] document how to use it in `Maintenance` section of manual, similar to https://warpx.readthedocs.io/en/latest/maintenance/release.html

### DIFF
--- a/new_version.sh
+++ b/new_version.sh
@@ -1,0 +1,175 @@
+#!/usr/bin/env python3
+#
+# Copyright 2021-2023 Axel Huebl
+#
+# This file is part of openPMD-api.
+#
+
+# This file is a maintainer tool to bump the versions inside openPMD-api's
+# source directory at all places where necessary.
+#
+import datetime
+from pathlib import Path
+import re
+import sys
+
+import requests
+
+try:
+    from configupdater import ConfigUpdater
+except ImportError:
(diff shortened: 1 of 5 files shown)


new_version.sh
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env python3
```
ax3l: More complete, latest script this is coming from:
https://github.com/ECP-WarpX/WarpX/blob/development/Tools/Release/newVersion.sh

update: oh wait, Python script is better than bash.
https://github.com/ECP-WarpX/WarpX/tree/development/Tools/Release

franzpoeschel:
I've adapted this to openPMD now. It's mostly text search and replace, I've entirely removed the dependency and config file management that WarpX uses since it does not really apply to our use cases.
Note that moving this to another directory means that the REPO_DIR definition must be changed too:
```python
REPO_DIR = Path(__file__).parent.absolute()
```

ax3l:
This script forgets to update:
- `docs/source/index.rst`
- `CITATION.cff`
- `docs/source/conf.py` <-- problem for RTD builds
- `setup.py` <-- problem for PyPI releases

PR #1469, opened 2023-06-30
HDF5: Handle unknown datatypes in datasets

franzpoeschel:
**First commit:** Throw `error::ReadError` in `HDF5IOHandlerImpl::openDataset()`. Upon encountering an error, the middle-end will know that something has gone wrong that it can then recover from, skipping the dataset.

**Second commit:** Sometimes, datasets use custom datatypes based on a native type. This can be supported by checking the parent datatypes if the actual datatypes is not recognized. See [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/4568#issuecomment-1613501627) for an example that uses enums to emulate booleans.

- [x] Merge first: #1482


ax3l:
Thanks for the patch!

This seems to fail on 32bit Windows right now @franzpoeschel

ax3l:
Does this need test coverage? :)

franzpoeschel:
The Windows tests look like a random failure. EDIT: Yep, after restarting the CI, it runs fine.
Testing this is a bit difficult since the datasets that are fixed by this cannot be created by the openPMD-api, so we would need to first manually create one and then add it to the sample datasets. Might still be worth it as we could use this to ensure read compatibility with old PIConGPU HDF5 files which we apparently did not so far.

franzpoeschel:
For testing, we'll need https://github.com/openPMD/openPMD-example-datasets/pull/20

ax3l (approved):
:tada:

PR #1470, opened 2023-06-30
Don't require unitSI when reading a patch record component

franzpoeschel:
The openPMD standard is not quite clear if `unitSI` is required for patch record components. At least for `numParticles` and `numParticlesOffset`, they don't really make sense.
The openPMD-api currently requires that these attributes are set and will fail at reading time if they are not (and subsequently skip the components). 

With this PR, the openPMD-api ignores if they are not present, but still writes them to maintain compatibility with versions of the openPMD-api that are not so lenient.


ax3l (approved):
Thanks! LGTM!

PR #1477, opened 2023-07-14
make it possible to manually set chunks when loading dask arrays

pordyna:
Make it possible to override the default chunking behavior when loading to `dask` arrays. For `PIConGPU` the default creates a chunk for each GPU with its local domain. When using static load balancing with `gridDist` in `PIConGPU`, this results in very uneven chunks. It looks like, for many operations, it makes sense to chunk along the outermost direction only (reading contiguous domains is faster). With this PR this can be achieved with `.to_daks_array(chunks={0 : 'auto', 1: -1,  2:  -1})`

### DIFF
--- a/docs/source/analysis/dask.rst
+++ b/docs/source/analysis/dask.rst
@@ -41,6 +41,9 @@ The central Python API calls to convert to DASK datatypes are the ``ParticleSpec
 
    # note: no series.flush() needed
 
+The ``to_dask_array`` method will automaticaly set dask array chunking based on the availible chunks.
+The default behaviour can be overriden by passing an additional keyword argument ``chunks``, see ``dask.array.from_array`` documentation for more details.
+For example, to chunk only along the outhermost axis in a 3D dataset using the default dask array chunk size call ``to_dask_array(chunks={0: 'auto', 1: -1,  2:  -1})``.
 
 Example
 -------
--- a/docs/source/analysis/dask.rst
+++ b/docs/source/analysis/dask.rst
@@ -41,9 +41,9 @@ The central Python API calls to convert to DASK datatypes are the ``ParticleSpec
 
    # note: no series.flush() needed
 
-The ``to_dask_array`` method will automaticaly set dask array chunking based on the availible chunks.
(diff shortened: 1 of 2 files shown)


src/binding/python/openpmd_api/DaskArray.py
```diff
@@ -1,8 +1,8 @@
 """
 This file is part of the openPMD-api.
 
-Copyright 2021 openPMD contributors
-Authors: Axel Huebl
+Copyright 2023 openPMD contributors
```
ax3l:
```suggestion
Copyright 2021-2023 openPMD contributors
```

docs/source/analysis/dask.rst
```diff
@@ -41,6 +41,9 @@ The central Python API calls to convert to DASK datatypes are the ``ParticleSpec
 
    # note: no series.flush() needed
 
+The ``to_dask_array`` method will automaticaly set dask array chunking based on the availible chunks.
+The default behaviour can be overriden by passing an additional keyword argument ``chunks``, see ``dask.array.from_array`` documentation for more details.
+For example, to chunk only along the outhermost axis in a 3D dataset using the default dask array chunk size call ``to_dask_array(chunks={0: 'auto', 1: -1,  2:  -1})``.
```
ax3l: A few typos and linking external docs in Dask:
```suggestion
The ``to_dask_array`` method will automatically set Dask array chunking based on the available chunks in the read data set.
The default behavior can be overridden by passing an additional keyword argument ``chunks``, see the `dask.array.from_array documentation <https://docs.dask.org/en/stable/generated/dask.array.from_array.html>`__ for more details.
For example, to chunk only along the outermost axis in a 3D dataset using the default Dask array chunk size, call ``to_dask_array(chunks={0: 'auto', 1: -1,  2:  -1})``.
```

ax3l (commented):
Cool, great idea!!

I added a few suggestions to finalize the PR :)

pordyna:
Thanks @ax3l, I applied your suggestions.

ax3l (approved):
Thank you! :)

PR #1490, opened 2023-08-01
CI fix: type comparison in openpmd-pipe

franzpoeschel:
"Source / style" CI runner seems to be picking this up recently


ax3l (approved):
Thx!

PR #1492, opened 2023-08-02
ADIOS2: Ensure that a step is always active at write time

franzpoeschel:
Even when not using steps, dump everything into a single large step. BP5 will fail otherwise. The PR modifies a BP5 test in a way that would fail before this PR.


ax3l (approved):
Discussed: w/o steps, we might currently create one big step. To be discussed next week.

PR #1497, opened 2023-08-08
Warn on BP5+Blosc in ADIOS2 v2.9 up to patch level 1

franzpoeschel:
Since we are about to ship ADIOS2 v2.9 with our Python Wheel, we should maybe try to avoid that users fall into [this trap](https://github.com/ornladios/ADIOS2/issues/3504).
I will add a second PR based on this that will warn on groupBased encoding. That second PR should probably not go into the current release cycle since the alternative (variableBased encoding) is not stable yet.

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -645,6 +645,41 @@ void ADIOS2IOHandlerImpl::createDataset(
             parameters.extent.begin(), parameters.extent.end());
 
         auto &fileData = getFileData(file, IfFileNotOpen::ThrowError);
+
+#define HAS_BP5_BLOSC2_BUG                                                     \
+    (ADIOS2_VERSION_MAJOR * 100 + ADIOS2_VERSION_MINOR == 209 &&               \
+     ADIOS2_VERSION_PATCH <= 1)
+#if HAS_BP5_BLOSC2_BUG
+        std::string engineType = fileData.getEngine().Type();
+        std::transform(
+            engineType.begin(),
+            engineType.end(),
+            engineType.begin(),
+            [](unsigned char c) { return std::tolower(c); });
+        if (!printedWarningsAlready.blosc2bp5 && engineType == "bp5writer")
+        {
+            for (auto const &op : operators)
+            {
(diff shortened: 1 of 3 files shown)


test/SerialIOTest.cpp
```diff
@@ -80,10 +80,24 @@ std::vector<std::string> testedFileExtensions()
...
-        allExtensions.begin(), allExtensions.end(), [](std::string const &ext) {
+        allExtensions.begin(),
+        allExtensions.end(),
+        []([[maybe_unused]] std::string const &ext) {
+#if openPMD_HAVE_ADIOS2
+#define HAS_ADIOS_2_9 (ADIOS2_VERSION_MAJOR * 100 + ADIOS2_VERSION_MINOR >= 209)
+#if HAS_ADIOS_2_9
+            // sst and ssc need a receiver for testing
+            // bp5 is already tested via bp
+            return ext == "sst" || ext == "ssc" || ext == "bp5";
```
ax3l, line 91: No "bp4" and "bp"?

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -645,6 +645,41 @@ void ADIOS2IOHandlerImpl::createDataset(
...
+                std::transform(
+                    operatorType.begin(),
+                    operatorType.end(),
+                    operatorType.begin(),
+                    [](unsigned char c) { return std::tolower(c); });
+                if (operatorType == "blosc")
+                {
+                    std::cerr << &R"(
+[Warning] Use BP5+Blosc with care in ADIOS2 v2.9.0 or v2.9.1.
+Unreadable data might be created, for further details see
```
```suggestion
[Warning] Use BP5+Blosc with care in ADIOS2 v2.9.0 and v2.9.1.
Unreadable data might be created, to mitigate either deactivate Blosc or use BP4+Blosc.
For further details see
```

test/SerialIOTest.cpp
franzpoeschel: This depends on the ADIOS2 version

ax3l: Ah, that's a test:
- ADIOS >=2.9: BP == BP5
- ADIOS <=2.9: BP == BP4

This return filters out duplicates.

ax3l (approved):

PR #1498, opened 2023-08-08
Adios2 warn groupbased encoding

franzpoeschel:
Should not go into 0.15 release cycle since it advises to use variable-based encoding

- [x] Merge #1497 first
- [x] ~Add some environment variable or similar to deactivate this~ The warning is now only printed for the new ADIOS2 encoding (group table), so this does not affect legacy workflows
- [x] ~~Maybe adapt tests~~ Postponed to when variable-based encoding becomes more useful
- [x] Entirely forbid usage without steps? Would make things easier
    EDIT: `adios2.usesteps` option is now removed, is automatically used if `WriteIterations` is used in the frontend
- [x] For a less abrupt transition, apply warnings only in new schema (i.e. group table enabled)

More context: [https://github.com/openPMD/openPMD-api/issues/1457](https://github.com/openPMD/openPMD-api/issues/1457)

The warning is printed only when using the (currently opt-in) group table feature, for the sake of a smoother transition for users. 
We should soon make group tables opt-out. A good time would be when we have better support for variable-based encoding.

Breaking because: It's no longer possible to open ADIOS2 group-based files with linear access mode if they don't use the group table feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1362,16 +1362,18 @@ if(openPMD_BUILD_TESTING)
                         ${MPI_TEST_EXE} ${Python_EXECUTABLE}                       \
                             ${openPMD_RUNTIME_OUTPUT_DIRECTORY}/openpmd-pipe       \
                             --infile ../samples/git-sample/data00000100.h5         \
-                            --outfile ../samples/git-sample/single_iteration.bp && \
+                            --outfile                                              \
+                                ../samples/git-sample/single_iteration_%T.bp &&    \
                                                                                    \
                         ${MPI_TEST_EXE} ${Python_EXECUTABLE}                       \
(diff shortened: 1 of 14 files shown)


ax3l:
@franzpoeschel ready for rebase :)

franzpoeschel:
@ax3l You recently mentioned a workflow that you have where you repeatedly close and reopen an Iteration to write data to it. Since this workflow implies that you cannot use variable-based encoding, I assume that you use file-based encoding instead?
I want to avoid discontinuing group-based encoding in ADIOS2 while there are workflows which actually use it.

franzpoeschel:
I would like to adapt all examples and tests to this change. Since this requires more features for file-based enconding (e.g. reopening a closed iteration) and for variable-based encoding (e.g. random-access) alike, I will postpone this PR for now.

ax3l (approved):

PR #1506, opened 2023-08-16
Python: Series to DataFrame

ax3l:
Add new helpers to create Pandas and cuDF dataframes for a single particle species, over all iterations at once.

- [x] add tests

### DIFF
--- a/src/binding/python/openpmd_api/DataFrame.py
+++ b/src/binding/python/openpmd_api/DataFrame.py
@@ -114,7 +114,7 @@ def iterations_to_dataframe(series, species_name):
     df = pd.concat(
         (
             series.iterations[i]
-            .particles["electrons"]
+            .particles[species_name]
             .to_df()
             .assign(iteration=i)
             for i in series.iterations
@@ -171,7 +171,7 @@ def iterations_to_cudf(series, species_name):
         (
             cudf.from_pandas(
                 series.iterations[i]
-                      .particles["electrons"]
+                      .particles[species_name]
                       .to_df()
                       .assign(iteration=i)
             )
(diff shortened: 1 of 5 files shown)


src/binding/python/openpmd_api/DataFrame.py
```diff
@@ -74,3 +74,111 @@ def particles_to_dataframe(particle_species, slice=None):
...
+    try:
+        import pandas as pd
+    except ImportError:
+        raise ImportError("pandas NOT found. Install pandas for DataFrame "
+                          "support.")
+
+    df = pd.concat(
+        (
+            series.iterations[i]
+            .particles["electrons"]
```
franzpoeschel:
```suggestion
            .particles[species_name]
```

src/binding/python/openpmd_api/DataFrame.py
```diff
@@ -74,3 +74,111 @@ def particles_to_dataframe(particle_species, slice=None):
...
+        import cudf
+    except ImportError:
+        raise ImportError("cudf NOT found. Install RAPIDS for CUDA DataFrame "
+                          "support.")
+
+    cdf = cudf.concat(
+        (
+            cudf.from_pandas(
+                series.iterations[i]
+                      .particles["electrons"]
```
franzpoeschel:
```suggestion
                      .particles[species_name]
```

franzpoeschel (changes requested):
There are two small hardcoded parameters that you probably meant to replace still.
Looks good otherwise, thank you!

src/binding/python/openpmd_api/DataFrame.py
ax3l: Good catch!

ax3l:
Ouch, good catch! Thank you, fixed now.

franzpoeschel (approved):

PR #1509, opened 2023-08-17
Replace openPMD_Datatypes global with function

franzpoeschel:
Forward-port from 0.15.2 bugfix release
Slightly API breaking since it replaces a barely-used global with a function.


ax3l (approved):
Thank you! :)

PR #1507, opened 2023-08-16
Release 0.15.2

ax3l:
https://github.com/openPMD/openPMD-api/milestone/10

### Python, ADIOS2 and HDF5 Fixes

This release fixed regressions in the Python frontend as well as the ADIOS2 and HDF5 backends.
Supported macOS versions are now 11.0+ and Python versions are 3.8+.

Thanks to @franzpoeschel, @axel, @dpgrote, @skalarproduktraum and @IlianCS for contributing to this release!

### Python

```python
import openpmd_api
print(openpmd_api.__version__)
```

### C++

```C++
#include <openPMD/openPMD.hpp>
#include <iostream>

int main() {
    // query compile-time macros:
    std::cout << OPENPMDAPI_VERSION_MAJOR << "."
              << OPENPMDAPI_VERSION_MINOR << "."
              << OPENPMDAPI_VERSION_PATCH << std::endl;
    // there is also: OPENPMDAPI_VERSION_LABEL (e.g. "alpha")

    // or query the runtime API (since 0.12.0):
    std::cout << openPMD::getVersion() << std::endl;
    return 0;
}
```

### CLI

```bash
# command line option (since 0.12.0):
openpmd-ls --version
```

PR #1510, opened 2023-08-17
Docs: Fix records length in "9_particle_write_serial.py"

IlianCS:
All records should have the same length (n_particles).


RemiLehe (approved):

ax3l (approved):
Good catch, thanks a lot!

ax3l:
I squashed all commits in this PR into a single one, so I can easily backport it to today's release #1507

PR #1512, opened 2023-08-18
CMake: Warn and Continue on Empty HDF5_VERSION

ax3l:
Seen on Conda-Forge for arm64 on macOS for HDF5 1.14.1

https://github.com/conda-forge/openpmd-api-feedstock/pull/107


rhaas80:
This still fails to detect the emtpy string on NERSC's Perlmutter (using CC). Would need to be `"${HDF5_VERSION}" STREQUAL ""` to detect this it seems (`HDF5_VERSION` is not even defined it seems).

PR #1516, opened 2023-08-19
CI: Linux aarch64/arm64

ax3l:
Add native Linux aarch64/arm64 runners with CircleCI.

Adding CI here first before adding an entry to `wheels` for building Python (Pip) Wheels.
Note: We already successfully build this architecture on conda-forge.


ax3l:
replaced by #1517

PR #1522, opened 2023-08-20
setup.py: Transitive ZLIB Static

ax3l:
Enable search for static transitive libs (in HDF5, ADIOS2 and/or C-Blosc2).
This might help with external static zlib packages in C-Blosc2, as seen in the last commit message of #1515.

- https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html#package-dependency-search-optimizations
- https://cmake.org/cmake/help/latest/module/FindZLIB.html#hints


franzpoeschel (approved):
LGTM, thank you!

PR #1528, opened 2023-09-08
Unknown openPMD version: Add upgrade hint

franzpoeschel:
Adding this now so users have an instructive upgrade message once we hit openPMD 2.0.

### DIFF
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -1277,7 +1277,13 @@ void Series::readOneIterationFileBased(std::string const &filePath)
     if (version == "1.0.0" || version == "1.0.1" || version == "1.1.0")
         pOpen.path = auxiliary::replace_first(basePath(), "/%T/", "");
     else
-        throw std::runtime_error("Unknown openPMD version - " + version);
+        throw error::ReadError(
+            error::AffectedObject::File,
+            error::Reason::UnexpectedContent,
+            std::nullopt,
+            "Unknown openPMD version - " + version +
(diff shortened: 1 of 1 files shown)


src/Series.cpp
```diff
@@ -1277,7 +1277,13 @@ void Series::readOneIterationFileBased(std::string const &filePath)
...
         pOpen.path = auxiliary::replace_first(basePath(), "/%T/", "");
     else
-        throw std::runtime_error("Unknown openPMD version - " + version);
+        throw error::ReadError(
+            error::AffectedObject::File,
+            error::Reason::UnexpectedContent,
+            std::nullopt,
+            "Unknown openPMD version - " + version +
+                ". Consider upgrading your installation of the openPMD-api "
+                "(https://openpmd-api.readthedocs.io).");
```
franzpoeschel, line 1286: Is there a way to link https://openpmd-api.readthedocs.io/en/0.15.2/install/install.html here without hardcoding the version or using latest? (latest is the dev)

franzpoeschel: I guess this idea does not make sense as the docs structure could change between releases.
Since we are trying to link to a newer version than $current for the purpose of upgrading, the root link should be fine.

ax3l:
LGTM :)

ax3l:
Btw, I have a define in `version.hpp` that knows the highest currently supported version numbers.

So, technically speaking, we could give this guidance:
- version too old: downgrade
- version too new: upgrade
- version corrupted format: print and reach out to data source

PR #1532, opened 2023-10-04
Add all the performance-* clang-tidy checks

lucafedeli88:
This PR adds all the `performance-*` checks to `clang-tidy` CI test, namely:

- [performance-faster-string-find](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-faster-string-find.html)
- [performance-for-range-copy](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-for-range-copy.html)
- [performance-implicit-conversion-in-loop](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-implicit-conversion-in-loop.html)
- [performance-inefficient-algorithm](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-inefficient-algorithm.html)
- [performance-inefficient-string-concatenation](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-inefficient-string-concatenation.html)
- [performance-inefficient-vector-operation](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-inefficient-vector-operation.html)
- [performance-move-const-arg](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-move-const-arg.html)
- [performance-move-constructor-init](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-move-constructor-init.html)
- [performance-no-automatic-move](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-no-automatic-move.html)
- [performance-no-int-to-ptr](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-no-int-to-ptr.html)
- [performance-noexcept-move-constructor](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-noexcept-move-constructor.html)
- [performance-trivially-destructible](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-trivially-destructible.html)
- [performance-type-promotion-in-math-fn](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-type-promotion-in-math-fn.html)
- [performance-unnecessary-copy-initialization](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-unnecessary-copy-initialization.html)
- [performance-unnecessary-value-param](https://releases.llvm.org/14.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-unnecessary-value-param.html)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,5 +2,5 @@
 # FIXME: all performance-* reports
 # FIXME: all cert-* reports
 # FIXME: all bugprone-* reports
-Checks: -*,bugprone-*,-bugprone-unhandled-self-assignment,-bugprone-parent-virtual-call,-bugprone-narrowing-conversions,-bugprone-exception-escape,-bugprone-string-literal-with-embedded-nul,cppcoreguidelines-slicing,mpi-*,readability-non-const-parameter,performance-for-range-copy,modernize-*,-modernize-use-trailing-return-type,-modernize-use-bool-literals,-modernize-avoid-c-arrays,-modernize-use-auto,-modernize-return-braced-init-list
+Checks: -*,bugprone-*,-bugprone-unhandled-self-assignment,-bugprone-parent-virtual-call,-bugprone-narrowing-conversions,-bugprone-exception-escape,-bugprone-string-literal-with-embedded-nul,cppcoreguidelines-slicing,mpi-*,readability-non-const-parameter,performance-*,modernize-*,-modernize-use-trailing-return-type,-modernize-use-bool-literals,-modernize-avoid-c-arrays,-modernize-use-auto,-modernize-return-braced-init-list
 HeaderFilterRegex: '((^(?!\/share\/openPMD\/).*)*include\/openPMD\/.+\.hpp|src\/^(?!binding).+\.cpp$)'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ exclude: '^share/openPMD/thirdParty'
 # See https://pre-commit.com/hooks.html for more hooks
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.5.0
     hooks:
     - id: trailing-whitespace
       args: [--markdown-linebreak-ext=md]
@@ -66,7 +66,7 @@ repos:
 # clang-format v13
 #   to run manually, use .github/workflows/clang-format/clang-format.sh
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v16.0.6
+  rev: v17.0.4
   hooks:
   - id: clang-format
     # By default, the clang-format hook configures:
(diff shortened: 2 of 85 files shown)


lucafedeli88:
I see that a test case is failing. I will investigate

lucafedeli88:
> I see that a test case is failing. I will investigate

The bug should be fixed!

franzpoeschel:
Currently looking through this. The checks are very useful, thank you for adding this!

It seems that in some places the suggested fixes by clang-tidy started an avalanche of recursive fixes and I'm trying to undo those. (Basic pattern: A `std::move()` was forgotten or wrongly applied, so clang-tidy suggests to use a `const &` argument and that change then propagates through the entire code base)

Feel free to ping either Axel or me when you need to have the CI started

franzpoeschel:
I've taken the freedom to push a commit on this branch that reverts some of the changes and tries to fix the warnings in another way. Most of these were "pass by value" turned into "pass by reference" because of:

* `std::move` was forgotten or wrongly applied. Fix that instead of changing the function signatures.
* API consistency: Some parameters are needed only in some configurations. clang-tidy then thinks that an unused parameter should be turned to a reference. Depending on the `#ifdef` logic, several implementations may be chosen. Here, we should optimize for the implementations that actually use the parameters since the others are often just there to satisfy the linker.
* Planned changes: A parameter is not yet used today, but will be in future. clang-tidy again thinks that unused parameters should be references.

The one thing that is a bit dangerous about clang-tidy imo is that it is very eager to change the API and ABI because of an implementation detail. This is mostly fine since those cases can mostly be handled case-by-case, but one should be aware of it.

One method that seems to successfully trick clang-tidy is to define something like `template<typename T> void forget(T){}` and passing unused parameters off into that template. I'll need to add a second commit that puts this in its own header, but not today.

Let's first see what the CI now thinks of this.

EDIT: I seem to have broken some things. Will investigate next week.

franzpoeschel:
Hey @lucafedeli88 
The CI is now passing. If you are fine with the changes I did (only in the code, I did not touch the checks), then I would merge this.

EDIT: I forgot one thing, CI needs to run again

franzpoeschel (approved):

franzpoeschel:
Merging this now, thank you for adding this!
Since many files are affected, this is likely to cause merge conflicts. But since each single change is mostly rather simple, the single merge conflicts should not be hard to resolve. In doubt, current changes can be replaced by incoming changes, and then clang-tidy suggestions can be reapplied.

PR #1493, opened 2023-08-04
JSON/TOML backend: introduce abbreviated IO modes

franzpoeschel:
Factored out of #1277, based on #1436

Introduces two options:

1. `json.dataset.mode = "template"` (default: `= "dataset"`), write just the dataset extent and not the actual content
2. `json.attribute.mode = "short"` (default: `= "long"`), write an attribute as `{"software": "openPMD-api"}` instead of `{"software": {"value": "openPMD-api", "type": "STRING"}}`

TODO:

- [x] Documentation
- [x] More testing: openpmd-pipe, erroring IO workflows – fixed, but add test
- [x] Merge #1436 first 
- [x] In TOML: Enable abbreviated modes by default: Done, but document this
- [x] Maybe add an option to skip the 'Trying to write data to a template dataset. Will skip.' warning

Diff: https://github.com/franzpoeschel/openPMD-api/compare/parallel-json..topic-json-short-modes

PR #1534, opened 2023-10-05
Non-spatial meshes

franzpoeschel:
Implements https://github.com/openPMD/openPMD-standard/pull/193

TODO:

- [x] Forward/backward compatibility? (of API + of created datasets): No compatibility
- [x] Python docu
- [x] Set openPMD version in written datasets as "2.0"
- [x] Merge #1551 first

### DIFF
--- a/src/Mesh.cpp
+++ b/src/Mesh.cpp
@@ -184,6 +184,20 @@ double Mesh::gridUnitSI() const
 Mesh &Mesh::setGridUnitSI(double gusi)
 {
     setAttribute("gridUnitSI", gusi);
+    if (auto series_opt = retrieveSeries_optional(); series_opt.has_value())
+    {
+        if (auto version = series_opt->openPMD(); version >= "2.")
+        {
+            std::cerr << "[Mesh::setGridUnitSI] Warning: Setting a scalar "
+                         "`gridUnitSI` in a file with openPMD version '" +
+                    version +
+                    "'. Consider specifying a vector instead in order to "
+                    "specify "
+                    "the gridUnitSI per axis (ref.: "
+                    "https://github.com/openPMD/openPMD-standard/pull/193)."
+                      << std::endl;
+        }
+    }
     return *this;
 }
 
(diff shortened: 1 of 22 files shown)


franzpoeschel:
This PR should be split into two parts at some point:

1. Introduction of openPMD 2.0, together with some helper functions
2. Introduction of non-spatial meshes

The first part might then also be relevant for other PRs where breaking changes are introduced (e.g. #1493 use of abbreviated attributes by default in JSON backend)

EDIT: Done #1551

src/Mesh.cpp
```diff
@@ -182,21 +183,161 @@ double Mesh::gridUnitSI() const
...
+         * Since some backends cannot distinguish between vector and
+         * scalar values, the most likely answer here is 1.
+         */
+        return 1;
+    }
+} // namespace
+
+std::vector<double> Mesh::gridUnitSIPerDimension() const
+{
+    return getAttribute("gridUnitSI").get<std::vector<double>>();
```
franzpoeschel: TODO: introduce the same backwards-compatible logic as for `gridUnitSI` (line 330)

ikbuibui:
I took a look over the PR and it seems okay to me. The interface fits for the binningPlugin use case. However the review isnt very thorough as I am unfamiliar with most of the code base and I don't know enough to follow some workflows.

ikbuibui (approved):

ikbuibui:
A small comment is that in the binningPlugin in PIConGPU i was using ``std::array<double, 7>`` to hold the dimensionality data. I see that the API requires vectors, even though it assumes the 7 SI Dimensions. Is it really required to be like this?

franzpoeschel:
Thank you for reviewing, Tapish!

> A small comment is that in the binningPlugin in PIConGPU i was using `std::array<double, 7>` to hold the dimensionality data. I see that the API requires vectors, even though it assumes the 7 SI Dimensions. Is it really required to be like this?

Can you point me to where exactly you mean? Within the context of this PR, `std::vector` is used for n-dimensional data, i.e. for a 3-dimensional mesh, you would specify three numbers in `Mesh &setGridUnitSI(std::vector<double> const &gridUnitSI)`.

Beyond that, we have always had two representations for Units: `std::array<double, 7>` and `std::map<UnitDimension, double>`. I have taken the chance of this PR to make this a bit more systematic and provide setters of both kinds, as well as conversion functionality.

With this PR,  `std::vector` may now additionally used to specify the unit per dimension: in `Mesh &setGridUnitDimension(unit_representations::AsArrays const &gridUnitDimension);`, the type `AsArrays` resolves to `std::vector<std::array<double, 7>>`. So, the use of `std::array` mostly stays.

ikbuibui:
Yes, you are right. It was my misunderstanding. Looks good!

PR #1535, opened 2023-10-06
Fix double mesh.read() call in parsing procedures

franzpoeschel:
First call was outside try{}, so errors would not be caught


ax3l:
@franzpoeschel thank you! 

Just for labeling: is this a bug in an existing release as well or purely on `dev`?

ax3l (approved):

franzpoeschel:
> @franzpoeschel thank you!
> 
> Just for labeling: is this a bug in an existing release as well or purely on `dev`?

Yes, it is also in the latest release

PR #1544, opened 2023-10-20
Add visit-like pattern for RecordComponent

### LINKED ISSUE #1372: A Variant-Based LoadChunk
It would be great if we can design load routines that are variant-based.

That way, the user-side logic is more concise than knowing the exact type upfront.

Similar for attribute reads (where we already support some flexible casts).

franzpoeschel:
Close #1372 

TODO:

- [x] Once we agree on a design, add some further examples

This adds:

1. `std::variant<...>` aliases for openPMD datatypes
2. A `RecordComponent::visit(...)` call that instantiates the passed Functor Template Object with the current type of the `RecordComponent`
3. Using 1. and 2., a `RecordComponent::loadChunkVariant()` call that returns a `std::variant<...>` of `shared_ptr` types

### DIFF
--- a/examples/10_streaming_read.cpp
+++ b/examples/10_streaming_read.cpp
@@ -11,7 +11,6 @@ using namespace openPMD;
 int main()
 {
 #if openPMD_HAVE_ADIOS2
-    using position_t = double;
     auto backends = openPMD::getFileExtensions();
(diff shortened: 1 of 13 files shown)


ax3l (approved):
Very nice!

PR #1545, opened 2023-10-23
Fix Attribute copy/move constructors

franzpoeschel:
Copy and move constructors of the `Attribute` class were accidentally custom-defined by having a constructor of the form:

```cpp
    template <typename T>
    Attribute(T &&val);
```

Hence, be a bit more careful:

```cpp
    template <typename T>
    Attribute(std::enable_if_t<
              // If T is `Attribute` or `Attribute const &`, this constructor
              // should not be used, but instead the move/copy constructors
              !std::is_same_v<
                  std::remove_cv_t<std::remove_reference_t<T>>,
                  Attribute>,
              T> &&val)
```

### DIFF
--- a/include/openPMD/backend/Attribute.hpp
+++ b/include/openPMD/backend/Attribute.hpp
@@ -21,6 +21,7 @@
 #pragma once
 
 #include "openPMD/Datatype.hpp"
+#include "openPMD/DatatypeMacros.hpp"
 #include "openPMD/auxiliary/TypeTraits.hpp"
 #include "openPMD/auxiliary/Variant.hpp"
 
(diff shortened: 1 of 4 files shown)


include/openPMD/backend/Attribute.hpp
```diff
@@ -107,7 +107,14 @@ class Attribute
...
      */
     template <typename T>
-    Attribute(T &&val) : Variant(resource(std::forward<T>(val)))
+    Attribute(std::enable_if_t<
+              // If T is `Attribute` or `Attribute const &`, this constructor
+              // should not be used, but instead the move/copy constructors
+              !std::is_same_v<
+                  std::remove_cv_t<std::remove_reference_t<T>>,
+                  Attribute>,
+              T> &&val)
```
ax3l: Would specializing be simpler/cleaner than enable if?

franzpoeschel: That might be better, yes. We would need to do this for every openPMD datatype then, though.
Probably best at that point to introduce something like the datatype macros that ADIOS2 has? `openPMD_FOREACH_DATATYPE` or so?

ax3l: Is that needed? The specialization could be something like
```C++
template <typename T>
Attribute(Attribute<T> const &) = default;

template <typename T>
Attribute(Attribute<T> &&) = default;
```

franzpoeschel: Template specialization doesn't work in this case: **With** `template`, we cannot use `default`, **without** `template`, it's not a specialization, but an overload that is never chosen. 
Using a specialization doesn't fix the underlying problem. In the current form, the constructor `template <typename T> Attribute(T &&val)` accepts anything, but is supposed to work only for all openPMD datatypes. That's a wrong interface.
Since the Intel Compiler doesn't understand this use of `enable_if`, I'd suggest using a macro.

franzpoeschel: The best would be a C++20 concept here though :D

ax3l: What I propose is no template specialization but an overload for the constructors. Did you try if that works?

ax3l: I see you did another approach with macros now ;-)

ax3l (approved):

include/openPMD/backend/Attribute.hpp
franzpoeschel: > What I propose is no template specialization but an overload for the constructors. Did you try if that works?

That's exactly what this PR does. Just, it doesn't write all ~30 constructors explicitly, but uses a macro for that instead.

PR #1551, opened 2023-11-03
Add openPMD version 2.0 to supported API versions

franzpoeschel:
This does not yet add any openPMD 2.0 functionality, but serves as a basis for PRs that add new functionality.

This does not yet set 2.0 as the default written version, hence `getStandard()` needs to be split into `getStandard()` and `getStandardDefault()`. 

- [x] Maybe add `getStandardMax()` additionally and deprecate `getStandard()`?
- [x] issue warning that 2.0 is in development

### DIFF
--- a/src/binding/python/Error.cpp
+++ b/src/binding/python/Error.cpp
@@ -9,6 +9,7 @@
 #include "openPMD/Error.hpp"
 
 #include "openPMD/binding/python/Common.hpp"
+#include <pybind11/pybind11.h>
 
 void init_Error(py::module &m)
 {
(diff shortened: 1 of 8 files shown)


ax3l:
Sounds good :+1: 

We could also separate into default/min/max standard for read and write.

franzpoeschel:
Marking this as WIP since I'm using this PR as a common ground for new openPMD 2.0 features and might want to extract more features into it in future.

src/binding/python/Error.cpp
```diff
@@ -9,6 +9,7 @@
 #include "openPMD/Error.hpp"
 
 #include "openPMD/binding/python/Common.hpp"
+#include <pybind11/pybind11.h>
```
franzpoeschel:
```suggestion
```

PR #1560, opened 2023-11-21
Remove `FindADIOS.cmake`

ax3l:
We removed ADIOS1 support in the upcoming 0.16 release series.

Follow-up to #1419


franzpoeschel (approved):
Thank you!

PR #1578, opened 2023-12-19
Add documentation for typical use cases of openpmd-pipe

franzpoeschel:


docs/source/analysis/pipe.rst
```diff
@@ -0,0 +1,141 @@
+.. _analysis-pipe:
+
+openpmd-pipe
+============
+
+Any Python-enabled openPMD-api installation with enabled CLI tools comes with a command-line tool named ``openpmd-pipe``.
+Naming and use are inspired from the piping concept known from UNIX shells.
```
ax3l:
```suggestion
Naming and use are inspired from the `piping concept <https://en.wikipedia.org/wiki/Pipeline_(Unix)>`__ known from UNIX shells.
```

ax3l (approved):

docs/source/analysis/pipe.rst
```diff
@@ -0,0 +1,141 @@
+.. _analysis-pipe:
```
ax3l: Let's move this to `Utilities > Command Line Tools`

ax3l (approved):

PR #1580, opened 2023-12-20
Initial support for HDF5 subfiling

franzpoeschel:
I'm following the documentation in [this presentation](https://www.hdfgroup.org/wp-content/uploads/2022/09/HDF5-Subfiling-VFD.pdf).

TODO:

- [x] parameterization
- [x] opt-in
- [x] documentation

### DIFF
--- a/docs/source/backends/hdf5.rst
+++ b/docs/source/backends/hdf5.rst
@@ -15,6 +15,15 @@ I/O Method
 
 HDF5 internally either writes serially, via ``POSIX`` on Unix systems, or parallel to a single logical file via MPI-I/O.
 
+Virtual File Drivers
+********************
+
+Rudimentary support for HDF5 VFDs (`virtual file driver <https://www.hdfgroup.org/wp-content/uploads/2021/10/HDF5-VFD-Plugins-HUG.pdf>`_) is available (currently only the *subfiling* VFD).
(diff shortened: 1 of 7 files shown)


ax3l (approved):
Awesome, thank you!

ax3l:
cc @guj and @jeanbez : just FYI, sub-filing support for HDF5 in openPMD-api now :)

ax3l:
We should check if Conda-Forge (hdf5 and h5py), h5py (pip builds) and Spack enable this feature in the HDF5 package already (non-default).

franzpoeschel:
Note that this needs to be enabled explicitly when configuring HDF5 `-DHDF5_ENABLE_SUBFILING_VFD=ON`.

guj:
Is  MPI_THREAD_MULTIPLE. still a requirement to use subfiles for HDF5? 
If so, needs to state  in the doc.

franzpoeschel:
> Is MPI_THREAD_MULTIPLE. still a requirement to use subfiles for HDF5? If so, needs to state in the doc.

Yes, it is a requirement. It's mentioned in the docs (`Note that the subfiling VFD needs to be enabled explicitly when configuring HDF5 and threaded MPI must be used.`), though the specific level is not mentioned yet (MPI_THREAD_MULTIPLE).

PR #1581, opened 2023-12-20
Draft: attribute_writing_ranks in HDf5

franzpoeschel:
Follow-up to #1542, this draft PR tries to implement that same feature for HDF5, too.

This is API-breaking, but only on the dev, since it turns `attribute_writing_ranks` into a global JSON option.

The results are a bit weird, since there always seems to be one rank from which HDF5 takes the attribute definitions. In my tests, when running with 14 ranks, this is always rank 12; when running with 28 ranks, it's always rank 25. This might mean that the current approach will not work.

### DIFF
--- a/examples/5_write_parallel.cpp
+++ b/examples/5_write_parallel.cpp
@@ -20,10 +20,13 @@
  */
 #include <openPMD/openPMD.hpp>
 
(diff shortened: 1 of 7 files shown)


franzpoeschel:
This does not work in the intended way.

PR #1583, opened 2024-01-06
Replace internal dependencies by FetchContent

nischild:
Replace internal dependencies management by CMake [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) commands.

Advantage: 
- Updating the dependencies is simply done by updating the version/hash used within `FetchContent_Declare` which makes it easier to stay at the project HEAD of dependencies for project maintainers
- Less CMake options that users have to take care about while keeping the same flexibility

Required changes: 
- Increase CMake minimum required version

Dependencies still have to be updated:
- TOML11 requires `CMAKE_CXX_STANDARD` to be set if the latest hash shall be used
- Catch2 requires to be moved from v2 to v3 which requires additional source changes.

If the pull request is accepted I would open issues to update the dependencies mentioned above and assign them to me.

(Sorry about the slightly chaotic git history. The change to FetchContent is done from [0ba14dc](https://github.com/openPMD/openPMD-api/pull/1583/commits/0ba14dc45a4f57ec0685faa889b4567fedd5650c) to [0084e9b](https://github.com/openPMD/openPMD-api/pull/1583/commits/0084e9b4b442d4aff13d36f7d9dbfbed166ee27f). The previous commits resulted from a previous try to move to FetchContent which failed and the commits are reverted)


.gitignore
```diff
@@ -96,6 +96,17 @@ Temporary Items
...
+#########
+# CMake #
+#########
+CMakeUserPresets.json
+
+# Directories which are created and filled by FetchContent
+share/openPMD/thirdParty/json/
+share/openPMD/thirdParty/toml11/
+share/openPMD/thirdParty/catch2/
+share/openPMD/thirdParty/pybind11/
```
franzpoeschel: We should be aware that with this, the build folder will modify the source tree.
What does the workflow look like switching between different openPMD versions that specify different dependency versions. Will FetchContent notice and update the dependencies' versions upon configuring?

CMakeLists.txt
```diff
@@ -259,27 +239,39 @@ endif()
...
-    find_package(toml11 3.7.1 CONFIG REQUIRED)
-    message(STATUS "toml11: Found version '${toml11_VERSION}'")
+find_package(toml11 CONFIG)
+if(NOT toml11_FOUND)
+    message(STATUS "Fetching toml11 from https://github.com/ToruNiina/toml11")
+    FetchContent_Declare(
+      toml11
+      GIT_REPOSITORY https://github.com/ToruNiina/toml11
+      # Migrate to the latest commit to remove CMake Warning which is not yet
+      # available in any official release.
```
franzpoeschel: You mean the warning below? Do I understand correctly that this has been fixed on toml11 dev and we are waiting for the next release?

```
CMake Deprecation Warning at share/openPMD/thirdParty/toml11/CMakeLists.txt:1 (cmake_minimum_required):                                                                                                                    
  Compatibility with CMake < 3.5 will be removed from a future version of                                                                                                                                                  
  CMake.                                                                                                                                                                                                                   
                                                                                                                                                                                                                           
  Update the VERSION argument <min> value or use a ...<max> suffix to tell                                                                                                                                                 
  CMake that the project does not need compatibility with older versions. 
```

CMakeLists.txt
```diff
@@ -550,18 +526,22 @@ target_include_directories(openPMD PUBLIC
 
 # Catch2 for unit tests
 if(openPMD_BUILD_TESTING)
-    add_library(openPMD::thirdparty::Catch2 INTERFACE IMPORTED)
-    if(openPMD_USE_INTERNAL_CATCH)
-        target_include_directories(openPMD::thirdparty::Catch2 SYSTEM INTERFACE
-            $<BUILD_INTERFACE:${openPMD_SOURCE_DIR}/share/openPMD/thirdParty/catch2/include>
+    find_package(Catch2 CONFIG)
+    if(NOT Catch2_FOUND)
+        message(STATUS "Fetching toml11 from https://github.com/catchorg/Catch2")
```
franzpoeschel:
```suggestion
        message(STATUS "Fetching catch2 from https://github.com/catchorg/Catch2")
```

README.md
```diff
@@ -261,15 +261,12 @@ CMake controls options with prefixed `-D`, e.g. `-DopenPMD_USE_MPI=OFF`:
...
-Additionally, the following libraries are shipped internally.
-The following options allow to switch to external installs:
-
-| CMake Option                    | Values     | Library       | Version  |
-|---------------------------------|------------|---------------|----------|
-| `openPMD_USE_INTERNAL_CATCH`    | **ON**/OFF | Catch2        | 2.13.10+ |
-| `openPMD_USE_INTERNAL_PYBIND11` | **ON**/OFF | pybind11      |  2.11.1+ |
-| `openPMD_USE_INTERNAL_JSON`     | **ON**/OFF | NLohmann-JSON |   3.9.1+ |
-| `openPMD_USE_INTERNAL_TOML11`   | **ON**/OFF | toml11        |   3.7.1+ |
+Additionally, the following libraries are shipped internally or, if the
```
franzpoeschel:
```suggestion
Additionally, the following libraries are shipped internally via FetchContent or, if the
```

If we're updating the text anyway, might as well make it explicit that there will be a Download.

.gitignore
nischild: That is true and might be a problem. I don't have experience with this.

Moving from lower to higher versions of OpenPMD should not be a problem. These directories have been deleted using `git rm` and should therefore be deleted when moving forward. FetchContent can then download into these directories w/o problem.

A problem might arise if one moves from a higher to a lower version. If the directories have been populated by FetchContent and git tries to add the old libraries again in the same directories I can not predict the behavior.

It would be possible to use a different directory with FetchContent until we are sure that it works as expected.

CMakeLists.txt
nischild: Yes.
But also moving to the last commit is not trivial with toml11 since that requires `CMAKE_CXX_STANDARD` to be set which I could not add to openPMD-api w/o breaking the CI.

README.md
nischild: I gave a more detailed explanation in the next commit

nischild:
Is it to be expected that [one of the CI stages](https://ci.appveyor.com/project/ax3l/openpmd-api/builds/48960419/job/f6lnsxa4apn30w2s) fails? The error in this step does not seem to be related to the changes of this pull request.

franzpoeschel:
> Is it to be expected that [one of the CI stages](https://ci.appveyor.com/project/ax3l/openpmd-api/builds/48960419/job/f6lnsxa4apn30w2s) fails? The error in this step does not seem to be related to the changes of this pull request.

The failure comes from the increased minimum CMake version:

```
CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
  CMake 3.24.0 or higher is required.  You are running version 3.22.1
```

The other "red text" seems to be unrelated (I see it on other PRs also), but does not cause a failure

Is version 3.22 also fine as a minimum? Otherwise, I'll push a commit that tries bumping the minimum CMake version of that CI run.

CMakeLists.txt
franzpoeschel: > since that requires `CMAKE_CXX_STANDARD` to be set which I could not add to openPMD-api w/o breaking the CI.

That's probably an issue unrelated to this PR which we should better solve separately?

.gitignore
franzpoeschel: Since @ax3l uses FetchContent in WarpX, let's wait until he is back to see what their approach looks like (relevant bits in their CMake config: https://github.com/ECP-WarpX/WarpX/blob/11e2a1722aac8929db0ed677f634673d235c1396/cmake/dependencies/openPMD.cmake#L29)

nischild:
Yes, sorry I was confused by the first red section in that action. 

CMake 3.24 was the lowest version which still works with the implementation that I provided here. Using a lower CMake Version requires to remove the `OVERIDE_FIND_PACKAGE` argument in FetchContent.

CMakeLists.txt
nischild: Yes, after finishing this PR I would try to Update catch2 and toml11 with separate issues as described above

franzpoeschel:
It seems like we need to consider saying good-bye to the Win32 CI runner. The last updates in https://repo.anaconda.com/pkgs/main/win-32/ were in 2022, the latest included CMake version is 3.22.1 from Dec 7, 2021, hence the failure.

franzpoeschel:
> Using a lower CMake Version requires to remove the `OVERIDE_FIND_PACKAGE` argument in FetchContent.

After an offline discussion, this might be the preferred way to go. Even if Win32 is end of life, the runners are still useful as they are a good way to check our handling of datatypes, also Win32 is still being used in lab settings.
Given that this is a program library. we should be careful about raising the required CMake version to something released only very recently, if it can be avoided.

nischild:
Is it possible to trigger the failed CI again. In the last run it seems that there was an internal error since the directory in which openPMD should have been placed already existed.

franzpoeschel:
> Is it possible to trigger the failed CI again. In the last run it seems that there was an internal error since the directory in which openPMD should have been placed already existed.

I think the Appveyor part of our CI has no way to trigger a selective restart
I force-pushed to give it a second chance

nischild:
@ax3l is there any chance to get direct access to the system on which the CI is failing? That would be easier than debugging this error by additional commits to this pull request.
Unfortunately I could not reproduce this error locally.

franzpoeschel:
I can reproduce the same error an a Linux system using CMake 3.22.1
This might be easier for investigating, I'll have a look

franzpoeschel:
The problem is that CMake 3.22 does not correctly interpret the `SOURCE_DIR` param on `FetchContent_Declare`, leaving empty directories. This leads ot a rather cryptic failure in `FetchContent_MakeAvailable`.
Seems fixed with CMake 3.23. I'd say, we can either use the `SOURCE_DIR` param only in CMake 3.23 or later, alternatively leave it out altogether.

nischild:
We could also just depricate the `SOURCE_DIR` keyword. I hoped that FetchContent skips the download if the directory is already populated. Unfortunately this does not work. Therefore adding the `SOURCE_DIR` does not give any advantage and we could also just use the CMake default directories for FetchContent.

.gitignore
ax3l: I'll move this to the build directory, as we do in other project.

ax3l:
Thanks a lot, @DerNils-git!

I'll merge this and do a few follow-up PRs:
- [x] remove the files in thirdparty separately #1666
- [x] add support for local source directories #1667
- [x] harmonize further with my usual logic ([[1]](https://github.com/AMReX-Codes/pyamrex/tree/development/cmake/dependencies), [[2]](https://github.com/ECP-WarpX/WarpX/tree/development/cmake/dependencies), [[3]](https://github.com/ECP-WarpX/impactx/tree/development/cmake/dependencies)) #1667

One thing I really miss in `FetchContent` is that it seems to lack support for shallow git clones, which can be a major time sink for large repos. If someone finds an option for this please let me know and we will add it (aka `git clone --depth 1`).

Alternatively, we could pull a tarball from the releases page of projects... #1668
https://cmake.org/cmake/help/latest/module/FetchContent.html#overview

PR #1586, opened 2024-01-11
Fix availableChunks for READ_LINEAR in ADIOS2

franzpoeschel:
Fix regression introduced by #1498, affects dev only

- [x] Testing


src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -33,6 +33,7 @@
 #include "openPMD/auxiliary/StringManip.hpp"
 #include "openPMD/auxiliary/TypeTraits.hpp"
 
+#include <adios2/common/ADIOSTypes.h>
```
```suggestion
```

PR #1594, opened 2024-02-13
Derive PatchComponent from RecordComponent

franzpoeschel:
`PatchRecordComponent` uselessly duplicates lots of logic from `RecordComponent` and lots of functionality from `RecordComponent` is missing in `ParticlePatches`, e.g. selective chunk loading. This is crucial in streaming setups where loading the entire particle patches information results in NxM communication patterns.

TODO:

- [x] ~~`BaseRecordComponent` is now useless and can be merged with `RecordComponent`~~ Keeping them separate for now keeps the diff smaller
- [x] ~~Alternatively, move a large part of the `RecordComponent` logic to `BaseRecordComponent` and keep only those parts in `RecordComponent` that really *should not* be part of `PatchRecordComponent`~~ That separation would imply having to separate the flush/read logic again.


ax3l (approved):
Nice! Looks great (incl Python update).

PR #1598, opened 2024-02-26
Fix dirtyRecursive() performance for Series with many steps

franzpoeschel:
~We currently do an error-check if users accidentally touched an Iteration that is currently closed. It seems like we'll need to stop doing that as it leads to O(n^2) runtime where n is the number of Iterations~ The check is still needed even in the regular logic, so the alternative was to make it more performant by tracking this information from the beginning.
- [x] finalize / clean up

Fix: Clear dirty files after flushing in ADIOS2

cc @ax3l

### DIFF
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -499,8 +499,6 @@ private:
     void storeChunk(
         auxiliary::WriteBuffer buffer, Datatype datatype, Offset o, Extent e);
 
-    bool containingIterationClosed() const;
-
     // clang-format off
 OPENPMD_protected
     // clang-format on
--- a/include/openPMD/RecordComponent.hpp
+++ b/include/openPMD/RecordComponent.hpp
@@ -69,6 +69,8 @@ namespace internal
          * Chunk reading/writing requests on the contained dataset.
          */
         std::queue<IOTask> m_chunks;
+
+        auto push_chunk(IOTask &&task) -> void;
         /**
          * Stores the value for constant record components.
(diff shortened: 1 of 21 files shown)


include/openPMD/RecordComponent.hpp
```diff
@@ -69,6 +69,8 @@ namespace internal
          * Chunk reading/writing requests on the contained dataset.
          */
         std::queue<IOTask> m_chunks;
+
+        auto push_chunk(IOTask &&task) -> void;
```
ax3l: Simpler?
```suggestion
        void push_chunk(IOTask &&task);
```

franzpoeschel: I've switched to using mostly the trailing style, even for return types as simple as this, because functions align much more nicely that way and declarations are easier to visually parse at a first glance, see [here](https://github.com/openPMD/openPMD-api/blob/19395b09479f75eb992c01bcaac0f1165a371ea3/include/openPMD/snapshots/ContainerImpls.hpp) for example.
Since the rest of this class does not follow that style, I'll use your suggestion anyway.

include/openPMD/RecordComponent.tpp
```diff
@@ -287,6 +287,11 @@ template <typename T, typename F>
 inline DynamicMemoryView<T>
 RecordComponent::storeChunk(Offset o, Extent e, F &&createBuffer)
 {
+    if (containingIterationClosed())
```
franzpoeschel: TODO: remove this check again

ax3l (approved):

PR #1601, opened 2024-03-11
ADIOS2 variable shape LocalValue

franzpoeschel:
Ref https://adios2.readthedocs.io/en/v2.9.2/components/components.html#shapes

This PR considers LocalValues in ADIOS2 merely an implementation detail, without actually exposing an API analogous to to how LocalValues work in ADIOS2.

The reasoning is:

1. LocalValues are useful not only for their API, but also since they give a hint to ADIOS2 that the variable should be handled as metadata and take part e.g. in metadata aggregation. With this PR, datasets such as the particlePatches in PIConGPU (one single value per MPI rank) can be given to the metadata system without changing a single line of code in PIConGPU. This can be crucial in streaming setups: When each reading rank loads the entire dataset naively (without this setting), the data plane of SST will open a connection between every writer and reader for just a single value each. Better to aggregate.
2. If needed at some point, a LocalValue-like API can still be exposed on top of this.

TODO

- [x] Documentation
- [ ] Check usage of `readOperators, m_operators`

### DIFF
--- a/examples/10_streaming_read.cpp
+++ b/examples/10_streaming_read.cpp
@@ -87,7 +87,8 @@ int main()
                 for (auto const &chunk : rc.second.availableChunks())
                 {
                     std::cout << "\n\tRank " << chunk.sourceID << "\t"
-                              << auxiliary::vec_as_string(chunk.offset) << "\t– "
+                              << auxiliary::vec_as_string(chunk.offset)
+                              << "\t– "
                               << auxiliary::vec_as_string(chunk.extent);
                 }
                 std::cout << std::endl;
--- a/examples/10_streaming_read.cpp
+++ b/examples/10_streaming_read.cpp
@@ -87,8 +87,8 @@ int main()
                 for (auto const &chunk : rc.second.availableChunks())
                 {
                     std::cout << "\n\tRank " << chunk.sourceID << "\t"
-                              << auxiliary::format_vec(chunk.offset) << "\t– "
-                              << auxiliary::format_vec(chunk.extent);
+                              << auxiliary::vec_as_string(chunk.offset) << "\t– "
+                              << auxiliary::vec_as_string(chunk.extent);
                 }
                 std::cout << std::endl;
             }
--- a/examples/10_streaming_read.cpp
+++ b/examples/10_streaming_read.cpp
@@ -18,8 +18,8 @@
  * and the GNU Lesser General Public License along with openPMD-api.
  * If not, see <http://www.gnu.org/licenses/>.
  */
-#include <openPMD/openPMD.hpp>
 #include <openPMD/auxiliary/StringManip.hpp>
+#include <openPMD/openPMD.hpp>
 
 #include <algorithm>
 #include <array>
--- a/examples/10_streaming_read.cpp
+++ b/examples/10_streaming_read.cpp
@@ -19,6 +19,7 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 #include <openPMD/openPMD.hpp>
+#include <openPMD/auxiliary/StringManip.hpp>
 
 #include <algorithm>
 #include <array>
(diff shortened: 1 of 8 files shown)


examples/10_streaming_write.cpp
```diff
@@ -49,13 +69,66 @@ int main()
...
+            pos.storeChunk(local_data, Offset{length * mpi_rank}, {length});
+        }
+
+        // Use the `local_value` ADIOS2 dataset shape to send a dataset not via
+        // the data plane, but the control plane of ADIOS2 SST. This is
+        // advisable for datasets where each rank contributes only a single item
+        // since the control plane performs data aggregation, thus avoiding
+        // fully interconnected communication meshes for data that needs to be
+        // read by each reader. A local value dataset can only contain a single
+        // item per MPI rank, forming an array of length equal to the MPI size.
```
ax3l, line 93: Maybe add link to https://adios2.readthedocs.io/en/v2.9.2/components/components.html#shapes again?
```suggestion
        // item per MPI rank, forming an array of length equal to the MPI size.
        // https://adios2.readthedocs.io/en/v2.9.2/components/components.html#shapes
```

src/IO/ADIOS/ADIOS2File.cpp
```diff
@@ -86,7 +87,23 @@ void WriteDataset::call(ADIOS2File &ba, detail::BufferedPut &bp)
                 adios2::Variable<T> var = ba.m_impl->verifyDataset<T>(
                     bp.param.offset, bp.param.extent, ba.m_IO, bp.name);
 
-                ba.getEngine().Put(var, ptr);
+                if (var.Shape() == adios2::Dims{adios2::LocalValueDim})
```
ax3l, line 114:
```suggestion
                // https://adios2.readthedocs.io/en/v2.9.2/components/components.html#shapes
                if (var.Shape() == adios2::Dims{adios2::LocalValueDim})
```

src/IO/ADIOS/ADIOS2File.cpp
```diff
@@ -143,7 +160,23 @@ struct RunUniquePtrPut
         auto ptr = static_cast<T const *>(bufferedPut.data.get());
         adios2::Variable<T> var = ba.m_impl->verifyDataset<T>(
             bufferedPut.offset, bufferedPut.extent, ba.m_IO, bufferedPut.name);
-        ba.getEngine().Put(var, ptr);
+        if (var.Shape() == adios2::Dims{adios2::LocalValueDim})
```
ax3l, line 203:
```suggestion
        // https://adios2.readthedocs.io/en/v2.9.2/components/components.html#shapes
        if (var.Shape() == adios2::Dims{adios2::LocalValueDim})
```

examples/10_streaming_read.cpp
```diff
@@ -1,3 +1,4 @@
+#include "openPMD/auxiliary/StringManip.hpp"
```
ax3l:
```suggestion
#include <openPMD/auxiliary/StringManip.hpp>
```

test/ParallelIOTest.cpp
```diff
@@ -422,6 +422,48 @@ void available_chunks_test(std::string const &file_ending)
         auto E_x = it0.meshes["E"]["x"];
         E_x.resetDataset({Datatype::INT, {mpi_size, 4}});
         E_x.storeChunk(data, {mpi_rank, 0}, {1, 4});
+
+        auto electrons = it0.particles["e"].particlePatches;
+        auto numParticles = electrons["numParticles"];
+        auto numParticlesOffset = electrons["numParticlesOffset"];
```
ax3l, line 453: Do you want to add an inline comment here somewhere what the block is functionally testing? :)

include/openPMD/auxiliary/StringManip.hpp
```diff
@@ -242,5 +242,24 @@ namespace auxiliary
         });
         return std::forward<S>(s);
     }
+
+    template <typename Vec>
+    auto format_vec(Vec const &v) -> std::string
```
ax3l: Pls add doxygen

PR #1614, opened 2024-04-24
Add define for ADIOS2_HAVE_BP5

franzpoeschel:
This was removed from ADIOS2 v2.9 to v2.10, causing test failures in #1613 since openPMD thinks that BP5 does not exist, but ADIOS2 creates BP5 files.

### DIFF
--- a/include/openPMD/IO/ADIOS/macros.hpp
+++ b/include/openPMD/IO/ADIOS/macros.hpp
@@ -19,11 +19,6 @@
 #define openPMD_HAS_ADIOS_2_9                                                  \
     (ADIOS2_VERSION_MAJOR * 100 + ADIOS2_VERSION_MINOR >= 209)
 
-#if !defined(ADIOS2_HAVE_BP5) && openPMD_HAS_ADIOS_2_9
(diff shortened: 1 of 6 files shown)


franzpoeschel:
I accidentally merged this somehow, @ax3l will revert the merge commit and we'll try this once more.

include/openPMD/IO/ADIOS/macros.hpp
```diff
@@ -19,6 +19,11 @@
 #define openPMD_HAS_ADIOS_2_9                                                  \
     (ADIOS2_VERSION_MAJOR * 100 + ADIOS2_VERSION_MINOR >= 209)
 
+#if !defined(ADIOS2_HAVE_BP5) && openPMD_HAS_ADIOS_2_9
+// ADIOS2 v2.10 no longer defines this
+#define ADIOS2_HAVE_BP5
```
franzpoeschel, line 24: todo: define our own macro instead of this

ax3l, line 24: Let's call this `openPMD_HAS_ADIOS_BP5` and add all logic here (and rename other uses).

PR #1592, opened 2024-01-29
Unify Random-Access API and Streaming API into Series::snapshots()

franzpoeschel:
So far, there were two distinct APIs for accessing Iterations:

1. random-access via `Series::iterations`
2. step-based access via `Series::writeIterations()` and `Series::readIterations()`

This PR introduces a new API `Series::snapshots()` that returns a unified container API. The unified container API can be implemented internally via **random-access** and via **synchronous access**, i.e. by using steps. 

This is part of an effort to bring both workflows and their functionalities closer together. While this redesigned container API does not yet go all the way along this road, it lays the fundament for such extensibility by (1) cleanly separating the API from its implementation and by (2) redesigning and cleaning up the implementation of the step-based workflow.

Preliminary/experimental support for reopening closed Iterations is now available, especially for file-based workflows.
Note that there are still caveats about reopening Iterations. Especially in ADIOS2, a step cannot be modified once closed, so writing new data is only possible into a new step. However, the Iteration handle currently does not delete old data definitions upon closing, meaning that there will be invalid data definitions in the handle.
In future, a closed Iteration will be removed from internal data structures (but users may keep it around for longer than that in their own data structures). 

 Future features to be built on top of this include:

1. Random-access for variable-based encoding based on `SetStepSelection()` in ADIOS2
2. Going backwards through the steps of an ADIOS2 file (by closing and reopening)
3. Dealing better with duplicate Iterations, arising from checkpoint-restart workflows (e.g. pre-scanning and eliminating duplicates beforehand)
4. Resource management, such as deleting Iteration handles upon closing them, and recreating them upon reopening

`writeIterations()` and `readIterations()` are now implemented as aliases/wrappers to `Series::snapshots()`. Due to the past somewhat unfortunate decision to introduce `IndexedIteration` as iteration type for `readIterations()`, the method name cannot be reused without breaking compatibility. `writeIterations()` however is just a wrapper for `snapshots()` with default arguments.

TODO:

- [x] Merge #1584 first
- [x] Better logic for READ_WRITE mode
- [x] backwards compatibility: `using` definitions, empty headers
- [x] documentation, change examples
- [x] maybe `snapshots(Snapshots::RandomAccess)` vs `snapshots(Snapshots::CollectiveAccess)` or so
* Follow-up: Container API: How to deal with erased Iterations? Answer: Iterations currently only being closed, not erased. Problem for later.
[...]

PR #1619, opened 2024-05-08
Workaround for independent writes to Iterations in parallel, better detection of BP5 which in turn uncovers more instances of the first issue

### LINKED ISSUE #1616: Parallel flushing cannot rely on dirty checks
**Describe the bug**
Currently, our flushing logic checks iterations if they are dirty, i.e. have any data that needs to be flushed, and flushes them only then. This does not work reliably in parallel since even an independent write must be flushed collectively.

**To Reproduce**
No need for a reproducer, the reason for the behavior is known.

**Expected behavior**
Since we are currently moving towards improving our support for more general `Iteration::open()`/`Iteration::close()` workflows, our flushing logic should flush exactly those Iterations that are currently open instead of trying to outsmart the user by just flushing what we think needs to be flushed.
Until then, an intermediate solution would be to expose sth like `Iteration::markDirty()`.

**Additional context**
Found by @psychocoderHPC

franzpoeschel:
This somewhat fixes #1616 until we add a better solution. With this PR: `seriesFlush()` will always flush the containing Iteration if called from within an Iteration (and will ignore missing `dirty` annotations).

At the same time, I added a better detection for BP5-specific features. Since this means that `adios2::Engine::PerformDataWrite()` is used automatically more often, this uncovers further parallel flushing bugs. So, these two items are treated together in this PR.

In a follow-up PR later on, as a more breaking change, we would also flush all open iterations in MPI-parallel contexts on `series.flush()`, but for this we will first need functionality to reopen iterations after close #1592.

TODO:

- [x] documentation
- [x] testing

### DIFF
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -133,7 +133,8 @@ void write_test_zero_extent(
     Series o = Series(
         filePath.append(".").append(file_ending),
         Access::CREATE,
-        MPI_COMM_WORLD);
+        MPI_COMM_WORLD,
+        "adios2.engine.preferred_flush_target = \"buffer\"");
 
     int const max_step = 100;
 
--- a/test/ParallelIOTest.cpp
+++ b/test/ParallelIOTest.cpp
@@ -133,8 +133,7 @@ void write_test_zero_extent(
     Series o = Series(
         filePath.append(".").append(file_ending),
         Access::CREATE,
-        MPI_COMM_WORLD,
-        "adios2.engine.preferred_flush_target = \"buffer\"");
+        MPI_COMM_WORLD);
 
     int const max_step = 100;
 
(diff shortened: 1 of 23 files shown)


test/ParallelIOTest.cpp
```diff
@@ -946,10 +946,16 @@ void hipace_like_write(std::string const &file_ending)
     int const last_step = 100;
     int const my_first_step = i_mpi_rank * int(local_Nz);
     int const all_last_step = last_step + (i_mpi_size - 1) * int(local_Nz);
+
+    bool participate_in_barrier = true;
```
franzpoeschel, line 1035: @ax3l Can you please check if this bug also affects Hipace? Currently, the sequence of barriers and flushes dont match from rank to rank. This was uncovered only now, since flushing is effectively not collective in many situations, but this test now uses `adios2::Engine::PerformDataWrite()` of BP5 which is a bit stricter there.

ax3l: Ok, great catch!

For HiPACE++, we changed the time stepping logic the other year, so that every MPI rank just writes to exactly one iteration. Thus, it cannot have this bug (anymore).

ax3l (approved):

PR #1624, opened 2024-05-24
Regexes: Sanitize user input

franzpoeschel:
For detecting files of file-based encoding, we use Regexes and emplace user input into them. This might lead to Regex injection when the file name contains a special character, e.g. `openpmd-ls simData_+10_1_%T.bp5`. Fix that.

### DIFF
--- a/src/Series.cpp
+++ b/src/Series.cpp
@@ -3035,12 +3035,12 @@ namespace
             std::smatch regexMatches;
             bool match = std::regex_match(filename, regexMatches, pattern);
             int processedPadding =
-                padding != 0 ? padding : (match ? regexMatches[1].length() : 0);
+                padding != 0 ? padding : (match ? regexMatches[2].length() : 0);
             return {
                 match,
                 processedPadding,
                 padding < 0 ? padding
-                    : match ? std::stoull(regexMatches[1])
+                    : match ? std::stoull(regexMatches[2])
                             : 0,
                 index_of_extension.has_value()
                     ? std::make_optional<std::string>(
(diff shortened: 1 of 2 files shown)


src/Series.cpp
```diff
@@ -3049,17 +3049,26 @@ namespace
         };
     }
 
+    namespace
+    {
+        auto sanitize_regex(std::string const &input) -> std::string
+        {
+            std::regex specialChars{R"([-[\]{}()*+?.,\^$|#\s])"};
```
franzpoeschel:
```suggestion
             // need to escape special characters reserved for regexes, see https://stackoverflow.com/questions/40195412/c11-regex-search-for-exact-string-escape
            std::regex specialChars{R"([-[\]{}()*+?.,\^$|#\s])"};
            // `$&` is the matched substring, see https://en.cppreference.com/w/cpp/regex/regex_replace
```

src/Series.cpp
```diff
@@ -3049,17 +3049,26 @@ namespace
...
+            return std::regex_replace(input, specialChars, R"(\$&)");
+        }
```
ax3l: Discussed today, let us maybe link:
- https://en.cppreference.com/w/cpp/regex/regex_replace
- https://stackoverflow.com/questions/40195412/c11-regex-search-for-exact-string-escape
- https://regex101.com/r/GDPK7E/2

src/Series.cpp
```diff
@@ -3049,17 +3049,30 @@ namespace
         };
     }
 
+    namespace
+    {
+        auto sanitize_regex(std::string const &input) -> std::string
+        {
+            // need to escape special characters reserved for regexes, see
+            // https://stackoverflow.com/questions/40195412/c11-regex-search-for-exact-string-escape
+            std::regex specialChars{R"([-[\]{}()*+?.,\^$|#\s])"};
```
ax3l, line 3058: Escaping backslashes https://regex101.com/r/GDPK7E/3
```suggestion
            std::regex specialChars{R"([-[\]{}()*+?.,\^$|#\s\\])"};
```

ax3l: Escaping backslashes, too
```suggestion
            // https://stackoverflow.com/questions/40195412/c11-regex-search-for-exact-string-escape
            // https://regex101.com/r/GDPK7E/3
            std::regex specialChars{R"([-[\]{}()*+?.,\^$|#\s\\])"};
```

ax3l (approved):

PR #1626, opened 2024-06-06
Fix particle patches flush api

franzpoeschel:
I don't understand our current particle patches flushing logic and it has led to surprising situations in the past often enough. For some reason, we initialize "numParticles" and "numParticlesOffset" (but not "offset" and "extent"???) upon creation of `ParticleSpecies`, but then flush the patches only if there are 3 or more records in the patches (since with numParticles and numParticlesOffset there are 2 already).....

Since #1594, we have all regular `storeChunk()` API calls in `PatchRecordComponent`, including the Span API which breaks under this logic, since it requires flushing the structure to the backend immediately internally and cannot handle being skipped by this item count.

Additionally, as a further follow-up to #1594 this erases special patch handling from `openpmd-pipe` since we can now just use the regular API calls.

### DIFF
--- a/examples/12_span_write.cpp
+++ b/examples/12_span_write.cpp
@@ -28,8 +28,28 @@ void span_write(std::string const &filename)
     for (size_t i = 0; i < 10; ++i)
     {
         Iteration iteration = iterations[i];
-        Record electronPositions = iteration.particles["e"]["position"];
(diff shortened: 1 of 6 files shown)


ax3l (approved):
LGTM

PR #1632, opened 2024-06-11
Flush target newstep

### LINKED ISSUE #1629: flush_target that ends/starts step
For boosted frame simulations, we plan to use a one-file-per-lab-frame mode with many adios steps in each.
During read, we flatten the adios steps into one [1].

In openPMD-api for writes, we need to add a third option for `flush_target` that calls `EndStep()/BeginStep()` instead of `PerformDataWrites()`.

ADIOS2 features:
- [1] https://github.com/ornladios/ADIOS2/pull/4147#discussion_r1583470124 -> https://github.com/ornladios/ADIOS2/pull/4160

franzpoeschel:
- [x] merge #1619 first
- [x] documentation
- [x] reading logic, e.g. `adios2.engine.mode = "what's the name of the mode that we need here does it even exist yet help"`
    It seems this needs no explicit reading logic, just `io.SetParameters("FlattenSteps=on");` on the **writer** side?
- [x] ~~Automatically add that option in file-based encoding~~ Not possible without breaking other workflows

Introduces a new flush target `"new_step"` used for creating readable output in file-based iteration encoding while the Iteration is not yet fully defined yet.

Close #1629

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -4437,6 +4437,7 @@ BufferChunkSize = 2147483646 # 2^31 - 2
 }
 #endif
 
+#if openPMD_HAVE_ADIOS2_BP5
 TEST_CASE("adios2_flush_via_step")
 {
     Series write(
@@ -4475,6 +4476,7 @@ TEST_CASE("adios2_flush_via_step")
     }
 #endif
 }
+#endif
 
 TEST_CASE("adios2_engines_and_file_endings")
 {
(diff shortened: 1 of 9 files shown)


test/SerialIOTest.cpp
```diff
@@ -2,6 +2,7 @@
 #include "openPMD/ChunkInfo_internal.hpp"
 #include "openPMD/Datatype.hpp"
 #include "openPMD/IO/Access.hpp"
+#include <adios2/cxx11/ADIOS.h>
```
franzpoeschel:
```suggestion
```

ax3l (approved):
This is awesome.

Let us also add a parallel test to cover serial and parallel I/O.

franzpoeschel:
> This is awesome.
> 
> Let us also add a parallel test to cover serial and parallel I/O.

Done as part of follow-up #1638

PR #1633, opened 2024-06-12
Pickle API: Avoid static Series hack, allow multiple Series

### LINKED ISSUE #1458: Pickle w/o Statics
For the first implementation of multi-process (multi-node) Dask, we pickle objects like the Record and RecordComponents + their series.

The series is unpickled into a static function member, to avoid:
- double open/parsing for each object
- having to express lifetime of the series on the unpickled side with respect to the pickled object.

This works as a hack until you need to work with two series at a time.

https://github.com/openPMD/openPMD-api/blob/0.15.1/include/openPMD/binding/python/Pickle.hpp#L73-L77

franzpoeschel:
Close #1458

Theoretically, it should be possible to modify our public API such that each handle, not only the Series handle would keep alive the Series, but that would require some greater internal restructuring.

Instead, this PR avoids the problem of finding a place to store the `Series` not by putting it in a static variable, but by hiding it inside the returned value's destructor. This does not lead to a reference cycle because the returned handle is a non-owning shared pointer whose destructor lambda owns all actual data. As soon as the handle is let go, the destructor is run and the owned objects are freed.

- [x] We could still use `static` (or better `thread_local`) to cache the latest opened Series for workflows that unpickle lots of single objects
- [x] Maybe take the chance to introduce pickling for `Series` and `Iteration`

@pordyna Can you check if this helps your use case?

### DIFF
--- a/include/openPMD/binding/python/Pickle.hpp
+++ b/include/openPMD/binding/python/Pickle.hpp
@@ -67,13 +67,9 @@ add_pickle(pybind11::class_<T_Args...> &cl, T_SeriesAccessor &&seriesAccessor)
             std::vector<std::string> const group =
                 t[1].cast<std::vector<std::string> >();
 
-            // Create a new openPMD Series and keep it alive.
-            // This is a big hack for now, but it works for our use
-            // case, which is spinning up remote serial read series
-            // for DASK.
-            static auto series = openPMD::Series(
+            openPMD::Series series(
                 filename, Access::READ_ONLY, "defer_iteration_parsing = true");
-            return seriesAccessor(series, group);
+            return seriesAccessor(std::move(series), group);
         }));
 }
 } // namespace openPMD
--- a/include/openPMD/binding/python/Pickle.hpp
+++ b/include/openPMD/binding/python/Pickle.hpp
@@ -71,7 +71,8 @@ add_pickle(pybind11::class_<T_Args...> &cl, T_SeriesAccessor &&seriesAccessor)
             // This is a big hack for now, but it works for our use
             // case, which is spinning up remote serial read series
             // for DASK.
-            static auto series = openPMD::Series(filename, Access::READ_ONLY);
+            static auto series = openPMD::Series(
+                filename, Access::READ_ONLY, "defer_iteration_parsing = true");
             return seriesAccessor(series, group);
         }));
 }
--- a/src/backend/Attributable.cpp
+++ b/src/backend/Attributable.cpp
@@ -514,12 +514,39 @@ namespace internal
     template <typename T>
     T &makeOwning(T &self, Series s)
     {
+        /*
+         * `self` is a handle object such as RecordComponent or Mesh (see
+         * instantiations below).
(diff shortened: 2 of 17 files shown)


pordyna (approved):
@franzpoeschel Just tested and it works. I can now change series within one kernel instance. 
Btw. I didn't new that, but one can install the api from a specific branch with `pip` like this `pip install git+https://github.com/franzpoeschel/openPMD-api.git@unpickle-multiple-series`, pretty cool.

franzpoeschel:
Nice, thank you for testing this!

> Btw. I didn't new that, but one can install the api from a specific branch with `pip` like this `pip install git+https://github.com/franzpoeschel/openPMD-api.git@unpickle-multiple-series`, pretty cool.

I didn't know that either. But I guess it means that you need to install ADIOS2 or HDF5 manually before that since they have no Pip packages? We usually just build them into the Wheel, but that doesn't help when compiling manually.

pordyna:
> I didn't know that either. But I guess it means that you need to install ADIOS2 or HDF5 manually before that since they have no Pip packages? We usually just build them into the Wheel, but that doesn't help when compiling manually.

I have installed adios2, hdf5, and blosc with conda. But see, https://pypi.org/project/adios2/ there is adios on pypi.

franzpoeschel:
> I have installed adios2, hdf5, and blosc with conda. But see, https://pypi.org/project/adios2/ there is adios on pypi.

Ah right, they actually have a package by now. Maybe we should add this as an optional dependency in our Pip-package @ax3l

src/backend/Attributable.cpp
```diff
@@ -488,4 +491,26 @@ void Attributable::linkHierarchy(Writable &w)
     writable().parent = &w;
     setDirty(true);
 }
+
+namespace internal
+{
+    template <typename T>
+    T &makeOwning(T &self, Series s)
```
ax3l, line 515: Let's add a nice doxygen string here :D

src/backend/Attributable.cpp
```diff
@@ -488,4 +491,26 @@ void Attributable::linkHierarchy(Writable &w)
     writable().parent = &w;
     setDirty(true);
 }
+
+namespace internal
+{
+    template <typename T>
```
franzpoeschel, line 514: maybe document the idea here with a little inline comment

include/openPMD/binding/python/Pickle.hpp
```diff
@@ -67,12 +68,40 @@ add_pickle(pybind11::class_<T_Args...> &cl, T_SeriesAccessor &&seriesAccessor)
             std::vector<std::string> const group =
                 t[1].cast<std::vector<std::string> >();
 
-            // Create a new openPMD Series and keep it alive.
-            // This is a big hack for now, but it works for our use
-            // case, which is spinning up remote serial read series
-            // for DASK.
-            static auto series = openPMD::Series(filename, Access::READ_ONLY);
-            return seriesAccessor(series, group);
+            /*
```
franzpoeschel: let's isolate this to a separate PR which goes into the 0.17 development branch

PR #1634, opened 2024-06-12
HDF5 independent flush config

### LINKED ISSUE #1513: HDF5 Option for OPENPMD_HDF5_INDEPENDENT
Although our API contract allows independent store/load calls (as in ADIOS), everything that is MPI-I/O based will essentially be performing better if the collective MPI-I/O calls are used.

Thus, the env control `OPENPMD_HDF5_INDEPENDENT` should be translated into an option for HDF5, so users that guarantee that they do collective `storeChunk` calls can activate it programmatically for performance.

Note that in MPI-I/O (and thus PHDF5) that means ranks with zero contribs need to issue zero-sized `storeChunk`s. The rationale behind that is that although an MPI rank might not contribute data, it might end up being a collection rank in MPI for collective data transport to disk.

franzpoeschel:
Close #1513

Todo:

- [x] ~~More extensive testing~~ suffices for now
- [x] Documentation
- [x] Check HDF5 version, older versions don't have this API


ax3l (approved):
Thank you!

PR #1638, opened 2024-06-21
ADIOS2: Override access mode

### LINKED ISSUE #1630: Append mode: Option to not truncate files in file-based iteration encoding
For boosted frame simulations, to implement checkpoint-restart workflows with the file-based, multi-steps-per-file output in #1629, we want to support to continue writing to an iteration after a checkpoint-restart interruption of an application.

We need an append mode in file-based iteration encoding that does not truncate files (openPMD iterations) that were already partially written.

As this is a very specific use case, adding a short-circuit style logic should be sufficient for this, e.g. `Series("output_%T.bp5", Access::APPEND, R"(adios2.engine.open_mode = "append")"` to override the ADIOS2 backend's access mode for ADIOS files. This would keep our frontend logic cleaner.
 
X-ref: #1629

franzpoeschel:
Close #1630. See there as well as the documentation and testing in the [diff](https://github.com/franzpoeschel/openPMD-api/compare/flush_target_newstep...adios2-override-access-type).

- [x] Merge #1632 first
- [x] Add a parallel test too

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -4475,6 +4475,58 @@ TEST_CASE("adios2_flush_via_step")
         }
     }
 #endif
+
+    /*
+     * Now emulate restarting from a checkpoint after a crash and continuing to
+     * write to the output Series. The semantics of openPMD::Access::APPEND
+     * don't fully fit here since that mode is for adding new Iterations to an
+     * existing Series. What we truly want to do is to continue writing to an
+     * Iteration without replacing it with a new one. So we must use the option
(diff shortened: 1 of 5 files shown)


test/SerialIOTest.cpp
```diff
@@ -4437,6 +4437,99 @@ BufferChunkSize = 2147483646 # 2^31 - 2
 }
 #endif
 
+#if openPMD_HAVE_ADIOS2_BP5
```
franzpoeschel: TODO: Add a parallel test too

ax3l (approved):
Looks great, thank you!

PR #1637, opened 2024-06-17
CI fixes and documentation for Pybind v2.12.0 and Numpy 2.0

franzpoeschel:
With the Pybind upgrade to v2.12.0 and the Numpy upgrade to v2.0, some previous CI setup warnings turned into errors, leading to broken Python imports. Fix by (1) removing conflicting Python versions or (2) using Python virtualenvs.

- [x] merge #1640 first

#1640 and this (#1637) are two separate PRs since I don't want to squash the commits, #1640 has another author (third-party lib).

PR #1640, opened 2024-06-24
Update Pybind to v2.12.0

franzpoeschel:
Factored out of #1637 since this commit has a different author and should not be squashed.
Pybind v2.12.0 is needed for Numpy 2.0 support.

CI will fail since the subsequent CI upgrades from #1637 are still needed.


ax3l:
@franzpoeschel can you do a follow-up PR to update the docs and CMake logic to require pybind11 v2.12.0+?

ax3l:
See it in #1637 :)

PR #1644, opened 2024-07-03
HDF5 filters (compression)

franzpoeschel:
Adds support for filters (e.g. compression) in HDF5.

TODO:

- [x] Documentation
- [x] Testing
- [x] Merge #1634 first
- [x] Add dataset-specific config also to HDF5, add a full example workflow also in Python, using https://pypi.org/project/hdf5plugin/ ~~maybe add some helpers~~
- [x] how to verify if a filter is actually effective? --> flags=mandatory
- [x] Add this to more CI runs, install hdf5plugin for Python tests in some CI runs
- [x] Merge #1775 first

* For a follow-up: Query available filters/operators

~~Diff: https://github.com/franzpoeschel/openPMD-api/compare/hdf5-independent-flush-config...hdf5-filters~~


src/IO/HDF5/HDF5IOHandler.cpp
```diff
@@ -609,6 +810,71 @@ void HDF5IOHandlerImpl::createDataset(
...
+{
+    if (access::readOnly(m_handler->m_backendAccess))
+        throw std::runtime_error(
+            "[HDF5] Creating a dataset in a file opened as read only is not "
+            "possible.");
+
+    if (parameters.joinedDimension.has_value())
+    {
+        error::throwOperationUnsupportedInBackend(
+            "ADIOS1", "Joined Arrays currently only supported in ADIOS2");
```
franzpoeschel: ADIOS1...

.github/workflows/dependencies/install_hdf5_blosc2
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
```
franzpoeschel:
```suggestion
set -eu -o pipefail
set -x
```

examples/15_compression.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+#include <openPMD/openPMD.hpp>
+
+/*
+ * If installed into a folder known to HDF5, then HDF5 will find the filter on
+ * its own. In other contexts, it might become necessary to manually register
+ * the filter into HDF5. For this, link the application against
+ * libblosc2_filter.so and set the below define to true.
+ */
+#define OPENPMD_INIT_BLOSC2_FILTER_MANUALLY false
```
ax3l: Is that valid pre-processor logic?
```suggestion
#define OPENPMD_INIT_BLOSC2_FILTER_MANUALLY 0
```

Not sure
```C++
#if false
```
is generally valid pre-processor...?

examples/15_compression.cpp
```diff
@@ -0,0 +1,328 @@
+/* Copyright 2025 Franz Poeschel
```
ax3l, line 1: Do you want to split this file in a "compressor" and "plugin/advanced" compression or HDF5 plugins?

ax3l: Or at least move the advanced part of blosc down to the end?

examples/15_compression.cpp
franzpoeschel: Note to self: Pull this logic down in the file, this should not be the first thing users see

ax3l (approved):
Looks great!

The C++ example is a bit involved and could be moved a bit, as discussed.

examples/15_compression.cpp
ax3l: Might be ok in C++ :sweat_smile: 
https://stackoverflow.com/a/69146085/2719194

ax3l (approved):

examples/15_compression.cpp
franzpoeschel: The advanced parts are now moved down

franzpoeschel:
Something is up with the Nvidia / NVHPC@25.1 CI job at the moment. That job ran cleanly before, and only commits affecting other CI jobs were added to the PR since then, so I am going to merge.
Will deal with the Nvidia CI job separately (first by hoping that it will be back on its own by next week).

PR #1646, opened 2024-07-10
Dataset-specific JSON/TOML configuration

franzpoeschel:
This was [initially prototyped](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3513) in PIConGPU, search for "extended format for JSON options" on [this documentation page](https://picongpu.readthedocs.io/en/latest/usage/plugins/openPMD.html).

This moves that functionality over into openPMD. 
Look at the diff on `examples/13_write_dynamic_configuration.cpp` for an example use case.
Bringing this to openPMD makes this functionality available in codes other than PIConGPU, and makes it available from TOML as well as from JSON.

TODO:

- [x] Document
- [x] Test
- [x] The NVIDIA compiler cant reliably parse TOML, don't let it run example 13
- [x] Check Python bindings for MyPath::openPMDPath()
- [x] Check JSON matching for OPEN_DATASET in ADIOS2
- [x] Merge #1706 first

### DIFF
--- a/docs/source/details/backendconfig.rst
+++ b/docs/source/details/backendconfig.rst
@@ -292,8 +292,8 @@ Dataset-specific configuration
 ------------------------------
 
 Some configuration options should be applicable on a per-dataset basis.
-All dataset-specific configuration options supported by the openPMD-api are additionally backend-specific, being format-specific serialization instructions such as compression or chunking.
-Dataset-specific configuration is hence specified under the key path ``<backend>.dataset``, e.g.:
+Most dataset-specific configuration options supported by the openPMD-api are additionally backend-specific, being format-specific serialization instructions such as compression or chunking.
+Such dataset-specific and backend-specific configuration is hence specified under the key path ``<backend>.dataset``, e.g.:
 
 .. code-block:: json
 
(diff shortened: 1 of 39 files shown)


docs/source/details/backendconfig.rst
```diff
@@ -287,3 +287,76 @@ Explanation of the single keys:
   In "template" mode, only the dataset metadata (type, extent and attributes) are stored and no chunks can be written or read (i.e. write/read operations will be skipped).
 * ``json.attribute.mode`` / ``toml.attribute.mode``: One of ``"long"`` (default in openPMD 1.*) or ``"short"`` (default in openPMD 2.* and generally in TOML).
   The long format explicitly encodes the attribute type in the dataset on disk, the short format only writes the actual attribute as a JSON/TOML value, requiring readers to recover the type.
+
+Dataset-specific configuration
+------------------------------
+
+Some configuration options should be applicable on a per-dataset basis.
```
ax3l: This might be a bit clearer:
```suggestion
Sometimes it is beneficial to set configuration options for specific datasets.
```

docs/source/details/backendconfig.rst
```diff
@@ -287,3 +287,76 @@ Explanation of the single keys:
...
+Most dataset-specific configuration options supported by the openPMD-api are additionally backend-specific, being format-specific serialization instructions such as compression or chunking.
+Such dataset-specific and backend-specific configuration is hence specified under the key path ``<backend>.dataset``, e.g.:
```
ax3l: The initial examples are not really showing dataset specific (filtered/selected) examples, so I reworded this a little to avoid confusion:
```suggestion

All dataset-specific and backend-specific configuration is specified under the key path ``<backend>.dataset``.
Without filtering by dataset name (see the ``select``` key below) this looks like:
```

examples/13_write_dynamic_configuration.cpp
```diff
@@ -10,13 +10,15 @@ using namespace openPMD;
...
-    if (!getVariants()["adios2"])
+    if (!getVariants()["hdf5"])
     {
         // Example configuration below selects the ADIOS2 backend
         return 0;
     }
 
     using position_t = double;
+
+#if !__NVCOMPILER // see https://github.com/ToruNiina/toml11/issues/205
```
ax3l: Do you want to filter by NVHPC version, since the issue is fixed in newer releases?
https://github.com/ToruNiina/toml11/issues/205#issuecomment-1487381207

Also, I think you meant to write
```suggestion
#if !defined(__NVCOMPILER)  // see https://github.com/ToruNiina/toml11/issues/205
```

ax3l (approved):
LGTM - thank you! :tada: 

Please review the changes to docs I made.
Please see last inline comment on NVHPC.

examples/13_write_dynamic_configuration.cpp
franzpoeschel: I'll replace it with the macro used in `src/config.cpp`:
```cpp
#if !defined(__NVCOMPILER_MAJOR__) || __NVCOMPILER_MAJOR__ >= 23
```

PR #1650, opened 2024-07-16
Parse lazily in linear access mode

franzpoeschel:
There's no reason not to, it makes things much more efficient by default.


ax3l (approved):
Makes great sense.

franzpoeschel:
The CI failure is currently seen in all PRs. Might be fixed by silencing compiler warnings.

PR #1652, opened 2024-07-19
Opt-in support for unknown ADIOS2 engines

franzpoeschel:
The ADIOS2 backend has a set of known and supported ADIOS2 engines. The consequence is that new, unknown or unsupported engines cannot be used without changing the code of the ADIOS2 backend.
This PR adds the backend key `adios2.engine.treat_unsupported_engine_like` in order to allow users to try alternative backends at their own risk.

e.g.:

```toml
[adios2.engine]
type = "dataman"
treat_unsupported_engine_like = "sst"
```

TODO:
- [x] Documentation

Notes to myself for documenting advanced features not supported by this:
QueueLimit
filename extension handling: default extensions, mandatory extensions in BP3 and SST
Span API
PerformDataWrite

### DIFF
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -48,6 +48,10 @@ Exceptions to this are the BP3 and SST engines which require their endings ``.bp
 
 For file engines, we currently leverage the default ADIOS2 transport parameters, i.e. ``POSIX`` on Unix systems and ``FStream`` on Windows.
 
+.. tip::
+
+  Use the ``adios2.engine.treat_unsupported_engine_as`` :ref:`JSON/TOML parameter <backendconfig-adios2>` for experimentally interacting with an unsupported ADIOS2 engine.
+
 Steps
 -----
 
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -50,7 +50,7 @@ For file engines, we currently leverage the default ADIOS2 transport parameters,
 
 .. tip::
 
-  Use the ``adios2.engine.treat_unsupported_engine_as`` :ref:`JSON/TOML parameter <backendconfig-adios2>` for experimentally interacting with an unsupported ADIOS2 engine.
+   Use the ``adios2.engine.treat_unsupported_engine_as`` :ref:`JSON/TOML parameter <backendconfig-adios2>` for experimentally interacting with an unsupported ADIOS2 engine.
 
 Steps
 -----
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -85,6 +85,7 @@ environment variable                  default    description
 ``OPENPMD_ADIOS2_HAVE_METADATA_FILE`` ``1``      Online creation of the adios journal file (``1``: yes, ``0``: no).
 ``OPENPMD_ADIOS2_NUM_SUBSTREAMS``     ``0``      Number of files to be created, 0 indicates maximum number possible.
 ``OPENPMD_ADIOS2_ENGINE``             ``File``   `ADIOS2 engine <https://adios2.readthedocs.io/en/latest/engines/engines.html>`_
+``OPENPMD_ADIOS2_PRETEND_ENGINE``     *empty*    Pretend that an (unknown) ADIOS2 engine is in fact another one (also see the ``adios2.pretend_engine`` :ref:`parameter <backendconfig-adios2>`).
 ``OPENPMD2_ADIOS2_USE_GROUP_TABLE``   ``0``      Use group table (see below)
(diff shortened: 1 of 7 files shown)


guj:
Is this PR intending to allow new additional adios engine types  (other than the ones defined in the list)? 

It seems to me that the function fileSuffix() still throws error when engine type (from either definitions) is not in the defined list.

franzpoeschel:
> Is this PR intending to allow new additional adios engine types (other than the ones defined in the list)?

Yes. The idea is that new engines which we don't support (like the currently experimental Campaign engine, or other engines like Dataman) can be experimentally used by telling the ADIOS2 backend to treat the engine like one of the supported engines.
When ADIOS2 v2.11 comes out and brings something new, this means that we ideally don't need to publish a new release for supporting it.

> 
> It seems to me that the function fileSuffix() still throws error when engine type (from either definitions) is not in the defined list.

The logic of `fileSuffix()` is skipped when an unsupported engine is in use:

```cpp
std::string ADIOS2IOHandlerImpl::fileSuffix(bool verbose) const
{
    // ... some defines ....

    if (m_realEngineType.has_value())
    {
        // unknown engine type, use whatever ending the user specified
        return m_userSpecifiedExtension;
    }

    // ....
```
So this should not be a problem

guj:
Thanks Franz.  As the engine type is also able to be defined as env variable "OPENPMD_ADIOS_ENGINE", should there be an equivalent env for "treat_unsupported_engine_like" as well?  I understand the json config takes priority over env variables.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -262,6 +263,25 @@ void ADIOS2IOHandlerImpl::init(
...
             }
+
+            if (engineConfig.json().contains(
+                    adios_defaults::str_treat_unsupported_engine_like))
+            {
+                auto maybeEngine = json::asLowerCaseStringDynamic(
+                    engineConfig
+                        [adios_defaults::str_treat_unsupported_engine_like]
+                            .json());
+                if (!maybeEngine.has_value())
```
guj, line 278: I guess a boolean flag would also do? "e.g. accept_unspported_engine=true"?

guj (approved):

src/IO/ADIOS/ADIOS2IOHandler.cpp
guj: If true,  can always return ".bp" at realEngineType(). The actual value of the  treat_supported_engine expected to be one of the known adios engines, and  exactly which one is not relavent I think.

franzpoeschel: > I guess a boolean flag would also do? "e.g. accept_unspported_engine=true"?

That will not be enough because there is no "standard" way how an engine should be treated. Especially, most engines differ slightly in how steps should be treated (staging engines require steps, BP5 requires steps with normal Read mode, but cannot deal with steps in ReadRandomAccess, the Campaign engine fails when using steps in any way, other engines are mostly a mixture). 

I don't think that providing one generic implementation for an unknown future engine will work, it seems better to use the existing implementations and let users try which works best.

> If true, can always return ".bp" at realEngineType(). The actual value of the treat_supported_engine expected to be one of the known adios engines, and exactly which one is not relavent I think.

`realEngineType()` will return the Engine type that is actually passed on to ADIOS2. `m_engineType` determines the behavior of the ADIOS2 backend. Especially BP5 behaves slightly different from the other BP engines.
I don't see a reason to put extra work into restricting the options here.

franzpoeschel:
> Thanks Franz. As the engine type is also able to be defined as env variable "OPENPMD_ADIOS_ENGINE", should there be an equivalent env for "treat_unsupported_engine_like" as well? I understand the json config takes priority over env variables.

I can add that, yeah

src/IO/ADIOS/ADIOS2IOHandler.cpp
guj: Sure. Thanks

docs/source/backends/adios2.rst
```diff
@@ -48,6 +48,10 @@ Exceptions to this are the BP3 and SST engines which require their endings ``.bp
 
 For file engines, we currently leverage the default ADIOS2 transport parameters, i.e. ``POSIX`` on Unix systems and ``FStream`` on Windows.
 
+.. tip::
+
+  Use the ``adios2.engine.treat_unsupported_engine_as`` :ref:`JSON/TOML parameter <backendconfig-adios2>` for experimentally interacting with an unsupported ADIOS2 engine.
```
```suggestion
   Use the ``adios2.engine.treat_unsupported_engine_as`` :ref:`JSON/TOML parameter <backendconfig-adios2>` for experimentally interacting with an unsupported ADIOS2 engine.
```

ax3l (approved):
Awesome, LGTM :)

PR #1656, opened 2024-08-01
setshape() should not touch the joined dim

guj:
if a joined dim exists, setshape() should not change it.

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -854,7 +854,8 @@ namespace detail
             {
                 dims.push_back(ext);
             }
-            var.SetShape(dims);
+            if (var.ShapeID() != adios2::ShapeID::JoinedArray)
+                var.SetShape(dims);
         }
 
         static constexpr char const *errorMsg = "ADIOS2: extendDataset()";
@@ -2032,7 +2033,8 @@ namespace detail
         }
         else
         {
-            var.SetShape(shape);
+            if (var.ShapeID() != adios2::ShapeID::JoinedArray)
+                var.SetShape(shape);
             if (count.size() > 0)
             {
                 var.SetSelection({start, count});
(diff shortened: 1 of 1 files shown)


franzpoeschel:
Do you have a specific use case in mind for this?
If yes, can you add that as a test, please?

If there is no specific use case, should we maybe rather just check this and throw an error if the shape is changed in unconventional ways?

Also, I think that this logic is more adequately placed in the frontend at `RecordComponent::resetDataset()`, since there is nothing inherently ADIOS2-specific about it.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -2032,7 +2035,18 @@ namespace detail
         }
         else
         {
-            var.SetShape(shape);
```
franzpoeschel, line 2035: I don't think that the previous `shape` should be considered here at all.
This code is executed when the Variable was already part of the previous step. The Variable shape can be something completely new in the new step. If the user means to use a joined dimension again, `Dataset::JOINED_DIMENSION` should be specified again; we should not try to outsmart the user by trying to keep the joined dimension from the previous step.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -849,10 +849,13 @@ namespace detail
...
-            dims.reserve(newShape.size());
-            for (auto ext : newShape)
+            dims.assign(newShape.begin(), newShape.end());
+            // keep the joined dim intact
+            auto joined_dim = joinedDimension(var.Shape());
+            if (joined_dim.has_value())
             {
-                dims.push_back(ext);
+                auto idx = joined_dim.value();
+                dims[idx] = var.Shape()[idx];
```
franzpoeschel, line 858: Undefined behavior if the vector lengths are not the same

franzpoeschel (changes requested):

guj:
> Do you have a specific use case in mind for this? If yes, can you add that as a test, please?
> 
> If there is no specific use case, should we maybe rather just check this and throw an error if the shape is changed in unconventional ways?
> 
> Also, I think that this logic is more adequately placed in the frontend at `RecordComponent::resetDataset()`, since there is nothing inherently ADIOS2-specific about it.

Yes, this change was needed when testing joined dimension in warpx.  Let me see if I can come up with a test case.

ax3l:
@guj can you please add a tiny test case for this that covers joined arrays with constant record components?
That way, @franzpoeschel can review this easier and we can cover that this works going forward.

franzpoeschel:
This has been fixed with #1740, I think? Feel free to reopen, if not.

PR #1660, opened 2024-08-06
Independent writes: Make workaround more flexible

franzpoeschel:
Follow-up to #1619
This lets users decide to flush only the containing Iteration or everything with `Attributable::seriesFlush()`.

TODO:

- [x] ~~Maybe use `false` as a default even. Would be a breaking change, but I'd argue it's the more intuitive behavior and most usage is probably along that line.~~ They're two different functions now.

### DIFF
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -133,6 +133,7 @@ class Iteration : public Attributable
     friend class internal::AttributableData;
     template <typename T>
     friend T &internal::makeOwning(T &self, Series);
+    friend class Writable;
 
 public:
     Iteration(Iteration const &) = default;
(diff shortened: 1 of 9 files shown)


guj:
Thanks Franz. 

1) It seems to me that for an I/O library  that requires collective efforts, e.g. BP5, it is better to  use iteration->seriesFlush() instead of series->flush().  So, should the default behavior of series->flush() be flushing everything, and set a flag like "dirtyonly= true"  if optimization is desired.

2) naming is a bit confusing to me.  by calling iteration->seriesFlush(flush_entire_series=true),  it will flush all other iterations in the underlying series owning this iteration?   might as well have something like series->flush(dirtyonly=false)?   i.e. I think letting iteration/series flush itself only would be cleaner.

ax3l:
Good points. I agree, we could make it maybe more transparent by introducing `iteration->flush()` and keeping `iteration->seriesFlush()` to point to the series flush?

franzpoeschel:
In that case, I'd suggest to add `Attributable::iterationFlush()` instead of `Iteration::flush()`:

* more intuitively tells the meaning of `Attributable::seriesFlush()` in opposition to `Attributable::iterationFlush()`
* users don't need to carry around the `Iteration` for flushing

ax3l (approved):

PR #1661, opened 2024-08-07
Make shape an optional attribute for constant components

### LINKED ISSUE #1663: Constant Records for JoinedArrays: Skip `shape` for Particles
For ADIOS2 JoinedArray support, we need to address this regression:
https://github.com/openPMD/openPMD-standard/pull/289

Opening this issue to document our discussion results two weeks ago @franzpoeschel @guj @ax3l

franzpoeschel:
Fix #1663

TODO:

- [x] Actually test this
- [x] Link to standard update: https://github.com/openPMD/openPMD-standard/pull/289
- [x] Maybe only do this for standard version 2.0 - actually ok as a fallback to expose in 1.X files, too UPDATE: [This comment](https://github.com/openPMD/openPMD-standard/pull/289#issuecomment-2407263974) explains a corner case that makes me hesitate adding this for version 1.0 since it might break datasets in singular cases. UPDATE: Corner case not relevant enough, would imply misuse of reserved `value` attribute, so not strictly standard-breaking.
- [x] For particles only or for meshes too?
- [x] Test that the shape can be queried from every component by the reader, even if shape attribute is missing
- [x] Document compatibility option

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -7074,8 +7074,11 @@ void unfinished_iteration_test(
     auto tryReading = [&config, file, encoding](
                           Access access,
                           std::string const &additionalConfig = "{}") {
+        auto merged_config = json::merge(
+            json::merge(config, additionalConfig),
+            R"({"verify_homogeneous_extents": false})");
         {
-            Series read(file, access, json::merge(config, additionalConfig));
+            Series read(file, access, merged_config);
 
             std::vector<decltype(Series::iterations)::key_type> iterations;
             std::cout << "\n\n\nGoing to list iterations in " << file
@@ -7126,7 +7129,7 @@ void unfinished_iteration_test(
         if (encoding == IterationEncoding::fileBased &&
             access == Access::READ_ONLY)
         {
-            Series read(file, access, json::merge(config, additionalConfig));
+            Series read(file, access, merged_config);
             if (additionalConfig == "{}")
             {
                 // Eager parsing, defective iteration has already been removed
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -806,7 +806,9 @@ inline void empty_dataset_test(std::string const &file_ending)
     }
     {
         Series series(
-            "../samples/empty_datasets." + file_ending, Access::READ_ONLY);
+            "../samples/empty_datasets." + file_ending,
+            Access::READ_ONLY,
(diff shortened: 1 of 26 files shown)


guj:
Thanks I will give it a try!

guj (approved):
I just tried with joined array, it worked!

.clang-tidy
```diff
@@ -2,5 +2,5 @@
 # FIXME: all performance-* reports
 # FIXME: all cert-* reports
 # FIXME: all bugprone-* reports
-Checks: -*,bugprone-*,-bugprone-unhandled-self-assignment,-bugprone-parent-virtual-call,-bugprone-narrowing-conversions,-bugprone-exception-escape,-bugprone-string-literal-with-embedded-nul,cppcoreguidelines-slicing,mpi-*,readability-non-const-parameter,performance-*,modernize-*,-modernize-use-trailing-return-type,-modernize-use-bool-literals,-modernize-avoid-c-arrays,-modernize-use-auto,-modernize-return-braced-init-list,-bugprone-easily-swappable-parameters
+Checks: -*,bugprone-*,-bugprone-unhandled-self-assignment,-bugprone-parent-virtual-call,-bugprone-narrowing-conversions,-bugprone-exception-escape,-bugprone-string-literal-with-embedded-nul,cppcoreguidelines-slicing,mpi-*,readability-non-const-parameter,performance-*,-performance-avoid-endl,modernize-*,-modernize-use-trailing-return-type,-modernize-use-bool-literals,-modernize-avoid-c-arrays,-modernize-use-auto,-modernize-return-braced-init-list,-bugprone-easily-swappable-parameters,-bugprone-chained-comparison
```
ax3l, line 5: Please rebase :)

franzpoeschel: done

test/SerialIOTest.cpp
```diff
@@ -5856,21 +5864,21 @@ void variableBasedSeries(std::string const &file)
             // this tests changing extents across iterations
             // ADIOS2 does not support changing the dimensionality
             // (older versions used to somewhat support it, but not really)
-            auto E_y = iteration.meshes["E"]["y"];
+            auto B_y = iteration.meshes["B"]["y"];
```
ax3l, line 5867: Interesting example. I was thinking primarily about particle records, but yes fields also could be used to shapes of some components depend on others...

ax3l (approved):

PR #1668, opened 2024-09-06
CMake Superbuild: Tarball

ax3l:
Cloning whole repos can be quite slow and CMake does not support shallow clones yet. We can use release tarballs instead.

Follow-up to #1583

Configure from scratch time:
- Before: 50.2s
- Now: 9.7s

Todo list:
- [x] rebase after #1667
- [x] docs update
- [x] catch
- [x] json
- [x] pybind11
- [x] toml11

### DIFF
--- a/cmake/dependencies/catch.cmake
+++ b/cmake/dependencies/catch.cmake
@@ -7,22 +7,33 @@ function(find_catch2)
         if(NOT IS_DIRECTORY ${openPMD_catch_src})
             message(FATAL_ERROR "Specified directory openPMD_catch_src='${openPMD_catch_src}' does not exist!")
         endif()
+    elseif(openPMD_catch_tar)
+        message(STATUS "Downloading Catch2 ...")
+        message(STATUS "Catch2 source: ${openPMD_catch_tar}")
     elseif(openPMD_USE_INTERNAL_CATCH)
         message(STATUS "Downloading Catch2 ...")
         message(STATUS "Catch2 repository: ${openPMD_catch_repo} (${openPMD_catch_branch})")
-        include(FetchContent)
     endif()
     if(TARGET Catch2::Catch2)
         # nothing to do, target already exists in the superbuild
-    elseif(openPMD_USE_INTERNAL_CATCH OR openPMD_catch_src)
+    elseif(openPMD_USE_INTERNAL_CATCH OR openPMD_catch_src OR openPMD_catch_tar)
         if(openPMD_catch_src)
             add_subdirectory(${openPMD_catch_src} _deps/localCatch2-build/)
         else()
-            FetchContent_Declare(fetchedCatch2
-                GIT_REPOSITORY ${openPMD_catch_repo}
-                GIT_TAG        ${openPMD_catch_branch}
-                BUILD_IN_SOURCE 0
-            )
+            include(FetchContent)
+            if(openPMD_catch_tar)
+                FetchContent_Declare(fetchedCatch2
+                    URL             ${openPMD_catch_tar}
+                    URL_HASH        ${openPMD_catch_tar_hash}
+                    BUILD_IN_SOURCE OFF
+                )
+            else()
+                FetchContent_Declare(fetchedCatch2
+                    GIT_REPOSITORY ${openPMD_catch_repo}
+                    GIT_TAG        ${openPMD_catch_branch}
+                    BUILD_IN_SOURCE OFF
+                )
+            endif()
             FetchContent_MakeAvailable(fetchedCatch2)
 
             # advanced fetch options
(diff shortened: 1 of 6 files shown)


cmake/dependencies/toml11.cmake
```diff
@@ -52,6 +63,14 @@ set(openPMD_toml11_src ""
     CACHE PATH
     "Local path to toml11 source directory (preferred if set)")
 
+# tarball fetcher
+set(openPMD_toml11_tar "https://github.com/ToruNiina/toml11/archive/refs/tags/v3.7.1.tar.gz"
+        CACHE STRING
+        "Remote tarball link link to pull and build toml11 from if(openPMD_USE_INTERNAL_TOML11)")
+set(openPMD_toml11_tar_hash "SHA256=afeaa9aa0416d4b6b2cd3897ca55d9317084103077b32a852247d8efd4cf6068"
```
nischild: This version of the toml11 library generates the CMake warning
```
-- toml11 repository: https://github.com/ToruNiina/toml11.git (v3.7.1)
CMake Deprecation Warning at build/_deps/fetchedtoml11-src/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```
Maybe consider updating the version at least to `v4.0.0` which adds `cmake_minimum_required(VERSION 3.16)`.
If it does not break anything you could consider updating to the latest version `v4.2.0`.

cmake/dependencies/catch.cmake
```diff
@@ -7,22 +7,33 @@ function(find_catch2)
...
-                GIT_REPOSITORY ${openPMD_catch_repo}
-                GIT_TAG        ${openPMD_catch_branch}
-                BUILD_IN_SOURCE 0
-            )
+            include(FetchContent)
+            if(openPMD_catch_tar)
+                FetchContent_Declare(fetchedCatch2
+                    URL             ${openPMD_catch_tar}
+                    URL_HASH        ${openPMD_catch_tar_hash}
+                    BUILD_IN_SOURCE 0
```
nischild: I would prefer `OFF` or `FALSE` for bool options, but it is a personal preference.

nischild (commented):
Reducing the configure time by a factor of almost five is significant so an obvious improvement.
Just a small disadvantage. One has to remember to update two lines if the version of a dependency should be updated instead of one.

We are working on a  wrapper for `FetchContent_Declare` to add a few features which should improve usability if the `SOURCE_DIR` directory is defined. We want to reduce the amount of downloads. If you are interested we could discuss this separately.

cmake/dependencies/toml11.cmake
ax3l: Had some other issue with this version, but will revisit tomorrow.

cmake/dependencies/catch.cmake
ax3l: I think that was not supported in the past for this specific option, will check again if that changed.

ax3l:
> We are working on a wrapper for FetchContent_Declare to add a few features which should improve usability if the SOURCE_DIR directory is defined. We want to reduce the amount of downloads. If you are interested we could discuss this separately.

Thanks! That sounds very interesting as a follow-up, happy to take a look on what you are working on :+1: 

> Just a small disadvantage. One has to remember to update two lines if the version of a dependency should be updated instead of one.

I would not wait for this extra update for the `0.16.0` release, since it is not critical and we do change the superbuild dependencies very, very seldomly. The superbuild is really just a developer convenience feature and should be deactivated for non-development deployments in favor of a proper external package/module.

cmake/dependencies/catch.cmake
```diff
@@ -44,6 +55,14 @@ set(openPMD_catch_src ""
     CACHE PATH
     "Local path to Catch2 source directory (preferred if set)")
 
+# tarball fetcher
+set(openPMD_catch_tar "https://github.com/catchorg/Catch2/archive/refs/tags/v2.13.10.tar.gz"
```
franzpoeschel, line 59: Will `openPMD_catch_branch`, `openPMD_catch_repo` and `openPMD_catch_src` be removed?

ax3l: We will support both. The order now of precedence is:
- src
- tar, or if `openPMD_catch_tar` is set to [off/false/empty/etc.](https://cmake.org/cmake/help/latest/command/if.html#constant)
- git

cmake/dependencies/toml11.cmake
ax3l: Memo to myself: @franzpoeschel uses `4.0.3`

ax3l: I repeated the test with toml11 `v4.2.0` and had no issues. Will use this now (aka the latest version :) ).

cmake/dependencies/catch.cmake
ax3l: You are right, this is a `BOOL` and I can make it more readable! :)
- [latest](https://cmake.org/cmake/help/latest/module/ExternalProject.html#build-step-options)
- [CMake 3.22](https://cmake.org/cmake/help/v3.22/module/ExternalProject.html)

cmake/dependencies/catch.cmake
ax3l: Docs added that explain the precedence.

PR #1645, opened 2024-07-09
Compatibility with toml11 4.0

franzpoeschel:
toml11 4.0+ seems to be a breaking change. Especially, output formatting now works differently.

- [x] Add a toml11 4.0.3 CI run

PR #1673, opened 2024-10-01
CMake Tar: TOML11 3.7.1

ax3l:
Keep branch and tar to TOML11 v3.7.1 in sync.
We currently still see issues with, so let's not ship it by default in superbuilds.

From @franzpoeschel:
```
About to parse 'iteration_encoding = "variable_based"'
[fwk394:211086:0:211086] Caught signal 11 (Segmentation fault: Sent by the kernel at address (nil))
```

Follow-up to #1668 #1645

### DIFF
--- a/cmake/dependencies/toml11.cmake
+++ b/cmake/dependencies/toml11.cmake
@@ -64,10 +64,10 @@ set(openPMD_toml11_src ""
     "Local path to toml11 source directory (preferred if set)")
 
 # tarball fetcher
-set(openPMD_toml11_tar "https://github.com/ToruNiina/toml11/archive/refs/tags/v4.2.0.tar.gz"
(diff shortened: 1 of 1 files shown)


cmake/dependencies/toml11.cmake
```diff
@@ -64,10 +64,10 @@ set(openPMD_toml11_src ""
     "Local path to toml11 source directory (preferred if set)")
 
 # tarball fetcher
-set(openPMD_toml11_tar "https://github.com/ToruNiina/toml11/archive/refs/tags/v4.2.0.tar.gz"
```
ax3l, line 67: This was out-of-sync, we use `3.7.1` also in `openPMD_toml11_branch` and documentation.

ax3l:
@franzpoeschel @DerNils-git please open a ticket in https://github.com/ToruNiina/toml11 with a reproducer for this issue.

PR #1678, opened 2024-10-08
Fix: CMake Internal Control

ax3l:
The `openPMD_USE_INTERNAL_*` options are for general internal/external control. They were accidentally overwritten by new tar flags and sometimes incomplete.

Follow-up to #1668


franzpoeschel (commented):
looks good to me, thank you!

PR #1679, opened 2024-10-09
Revert "CMake Tar: TOML11 3.7.1 (#1673)"

franzpoeschel:
This reverts commit aea4b72388351e89214c0c3c3246acde2375e3ec.

The problem that I saw with versions >= 4.0 was due to an ill-configured software stack on my end, there is nothing that blocks us from defaulting to 4.2.0.


ax3l:
@franzpoeschel I am confused, we already merged #1673 for this.

ax3l:
Oh, I see. You bumped back up to 4+

PR #1677, opened 2024-10-08
Wheels: 0.16.0

ax3l:
Update the cibuildwheel reference tracking to the 0.16.0 release.

- [x] patch in #1680 or wait for 0.16.1 #1676 because of botched `setup.py` #https://github.com/openPMD/openPMD-api/pull/1648#issuecomment-2398527616
- [x] rebase once ADIOS2 v2.10.1 is building #1554

PR #1676, opened 2024-10-08
Release: 0.16.1

ax3l:
- [x] See https://github.com/openPMD/openPMD-api/pull/1648#issuecomment-2398527616
- [x] Fix CMake superbuild control #1678
- [x] ULONG example fix #1710
- [x] Update macOS CI #1703

### Python

```python
import openpmd_api
print(openpmd_api.__version__)
```

### C++

```C++
#include <openPMD/openPMD.hpp>
#include <iostream>

int main() {
    // query compile-time macros:
    std::cout << OPENPMDAPI_VERSION_MAJOR << "."
              << OPENPMDAPI_VERSION_MINOR << "."
              << OPENPMDAPI_VERSION_PATCH << std::endl;
    // there is also: OPENPMDAPI_VERSION_LABEL (e.g. "alpha")

    // or query the runtime API (since 0.12.0):
    std::cout << openPMD::getVersion() << std::endl;
    return 0;
}
```

### CLI

```bash
# command line option (since 0.12.0):
openpmd-ls --version
```

PR #1680, opened 2024-10-10
Fix 0.16.0 versions

ax3l:
Update versions in a single commit so we can use this as a patch in package managers, e.g., #1677.

Used the script from #1676


franzpoeschel:
The script did not fix all occurrences since 0.16.0 was already partially applied, and it cannot deal properly with such a broken state. I hope we caught all places now.

franzpoeschel:
Last commit added to fix the failing MacOS run due to a removed API in Python unittest, cherry-picked from #1681

PR #1683, opened 2024-10-10
Fix Leftover ADIOS1 Mentions

ax3l:
Leaves the doc page on ADIOS1 for now in, because it contains a migration guide on how to read legacy BP3 files.

Note: Examples 7 and 13 still mention adios1 options, which will need a separate PR to update.


franzpoeschel:
> Note: Examples 7 and 13 still mention adios1 options, which will need a separate PR to update.

I pushed a commit that does this, I think this PR is fine for it

franzpoeschel (approved):

franzpoeschel (approved):

PR #1401, opened 2023-03-27
Wheels: 0.15.0.post1

ax3l:
## Applied Post-Release Patches

- [x] #1395 (already in `0.15.0` wheel via #1392)
- [x] #1393 (already in `0.15.0` wheel via #1392)
- [x] #1400
- [x] bump version to `0.15.0.post1`

### Not Yet Ready

- [ ] #1396 

## Notes

- Travis-CI is also back for Linux aarch64 and ppc64le runners.

PR #1696, opened 2024-11-12
Wheels: Fix duplicate ID

franzpoeschel:
Follow-up to #1695

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,7 +106,9 @@ jobs:
       run: |
         cd src
         git apply --exclude=.github/workflows/windows.yml .patch/1680.patch
-        git apply .patch/1684.patch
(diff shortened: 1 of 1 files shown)


franzpoeschel:
` wheels / Build x86 wheel on windows-2019` seems to have been broken in #1677 already: https://github.com/openPMD/openPMD-api/actions/runs/11279883854/job/31371607907

franzpoeschel:
Just tested locally against the new wheel: The problem that I saw (see description in #1695) seems to be fixed with this.

ax3l:
You need to update `setup.py` to add a `.postN` suffix to the version for PyPI to accept the new patch release. Example: https://github.com/openPMD/openPMD-api/pull/1401

The reason is that PyPI lets one only upload a specific version once (after that, all we can do is yank it in case it is very broken.)

PR #1697, opened 2024-11-19
HDF5: Delete and re-create attribute when overwriting with diff. type

franzpoeschel:
Seen in a file that was originally create with `h5py` and the opened in R/W-mode with openPMD-api. Since h5py and openPMD-api apparently use different vec-string types, this leads to corrupted files.
Since we do support attributes with multiple possible types, the type of an existing attribute should be checked before reuse.

TODO:

- [x] Test

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -7522,6 +7522,8 @@ void groupbased_read_write(std::string const &ext)
 
         E_x.setAttribute("updated_in_run", 0);
         E_y.setAttribute("updated_in_run", 0);
+        E_y.setAttribute("changed_datatype_in_run", 0);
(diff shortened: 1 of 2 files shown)


test/SerialIOTest.cpp
```diff
@@ -7522,6 +7522,8 @@ void groupbased_read_write(std::string const &ext)
 
         E_x.setAttribute("updated_in_run", 0);
         E_y.setAttribute("updated_in_run", 0);
+        E_y.setAttribute("changed_datatype_in_run", 0);
+        write.close();
```
franzpoeschel, line 7526: `.close()` calls added for stricter error check, the destructors will swallow exceptions.

PR #1700, opened 2024-11-25
JSON: Selectively suppress unknown key warnings

franzpoeschel:
Example use case: This came up when in the last release we added the key `adios2.engine.attribute_writing_ranks`. When adding this key in downstream codes (came up in PIConGPU), then that code might either be compiled against the recent version which knows that key or against the older version which knows it not.
In the latter case, a runtime warning about an unknown key will be printed. To avoid that issue, this PR adds a flag to opt-out keys from warnings:

```json
{
  "dont_warn_unused_keys": [
    "key1",
    "key3"
  ],
  "key1": 5,
  "key2": {
    "dont_warn_unused_keys": [
      "key1"
    ],
    "key1": 5,
    "key2": 6
  },
  "key3": {
    "dont_warn_unused_keys": [
      "key1"
    ],
    "key1": 5,
    "key2": 6
  }
}
```

Will yield the warning:

[Series] The following parts of the global JSON config remains unused: 
```json
{
  "key2": {
    "dont_warn_unused_keys": [
      "key1"
    ],
    "key1": 5,
    "key2": 6
  },
  "key3": {
    "key2": 6
  }
}
```

TODO:

- [x] better errors upon parsing wrong configurations
- [x] documentation

### DIFF
--- a/include/openPMD/auxiliary/JSON_internal.hpp
+++ b/include/openPMD/auxiliary/JSON_internal.hpp
@@ -164,6 +164,11 @@ namespace json
             bool trace);
 
         void init();
+        /*
+         * Called upon each traced access of a location in the JSON value, along
+         * with the matching subtree of the shadow.
+         * This implements the `dont_warn_unused_keys` functionality.
+         */
         static void
         init(nlohmann::json const &original, nlohmann::json &shadow);
     };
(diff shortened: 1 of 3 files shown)


include/openPMD/auxiliary/JSON_internal.hpp
```diff
@@ -162,6 +162,9 @@ namespace json
             nlohmann::json *positionInShadow,
             SupportedLanguages originallySpecifiedAs,
             bool trace);
+
+        void init();
+        static void init(nlohmann::json &original, nlohmann::json &shadow);
```
ax3l: - Are both original and shadow modified or is one `const &`?
- Please add a doxygen

franzpoeschel: `original` is now const, docstring is now there.

ax3l (approved):

PR #1701, opened 2024-12-03
Conda: Pin HDF5 to 1.14.1 in CI workflow

franzpoeschel:
Version 1.14.4 has trouble: https://github.com/HDFGroup/hdf5/issues/5164
Until that is resolved, let's stick with a version that works.

### DIFF
--- a/conda.yml
+++ b/conda.yml
@@ -23,7 +23,7 @@ dependencies:
   - dask
   - doxygen
   - git
-  - hdf5=*=mpi_openmpi_*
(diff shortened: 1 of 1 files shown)


franzpoeschel:
Ok, 1.14.0 does not like being used in MPI contexts, let's use 1.14.1...

franzpoeschel:
If there's no quick fix for this, maybe instead fix this in CMake on our end by not relying on `HDF5_VERSION`, `find_package(hdf5 <version> ...)` seems to work.

PR #1716, opened 2025-02-04
Integrate openpmd-scipp into the main API

franzpoeschel:
This merges the code from https://github.com/pordyna/openpmd_scipp into the main API such that it can be used by `series.to_scipp()`.

TODO:

- [x] Documentation (fields)
- [ ]  Documentation (particles) COMMENT: The TODO there refers to particles not being even implemented yet, I think?
- [ ] Clean up the example
- [x] Have a look at the rest of what is in the repository, see what needs to / can be transferred
- [x] CI integration
- [ ] Licensing headers
- [ ] remove Jupyter Notebook
- [ ] check image sizes

### DIFF
--- a/docs/source/analysis/scipp.rst
+++ b/docs/source/analysis/scipp.rst
@@ -5,6 +5,10 @@ Scipp
 
 Load openpmd datasets to ``scipp`` ``DataArrays``.
 
+.. note::
+
+   This documentation page is also available as an interactively executable Jupyter Notebook in ``examples/15_scipp_loader.ipynb``.
+
 What is this good for?
 ~~~~~~~~~~~~~~~~~~~~~~
 
@@ -63,13 +67,12 @@ Opening series
 
 .. code:: python
 
-   path = "openPMD-example-datasets/example-3d/hdf5/data%T.h5"
-   path = "./data/" + path
+   path = "../samples/git-sample/data%T.h5"
 
 .. code:: python
 
    series = pmd.Series(path, pmd.Access.read_random_access)
-   data_loader = sereies.to_scipp()
+   data_loader = series.to_scipp()
    print(data_loader.iterations)
 
 ::
--- a/docs/source/analysis/scipp.rst
+++ b/docs/source/analysis/scipp.rst
@@ -1,0 +1,370 @@
+.. _analysis-scipp:
+
+Scipp
+=====
+
+Load openpmd datasets to ``scipp`` ``DataArrays``.
+
+What is this good for?
+~~~~~~~~~~~~~~~~~~~~~~
+
+`scipp <https://github.com/scipp/scipp>`__ is an alternative to
+`xarray <https://github.com/pydata/xarray>`__ and provides basically
+numpy arrays with axes description and units.
+
+* Automatically load axes and units with openPMD data.
+* Axes information is automatically updated when slicing, indexing, or filtering your data.
+* With ``scipp``\ ’s plotting library `plopp <https://github.com/scipp/plopp>`__ it becomes an alternative to ``openpmd-viewer``.
+* Many numpy and some scipy functions including all the basic algebraic operations on arrays are supported by ``scipp``. When using these, the units and coordinates are automatically taken care of.
+
+Limitations
+~~~~~~~~~~~
+
+-  ``scipp`` currently handles units with a library, that does not
+   support non-integer exponents for units. This can become problematic
+   in some calculations.
+
+Installation
+------------
+
+It can be easily installed with pip.
+
+.. code:: bash
+
+   git clone https://github.com/pordyna/openpmd_scipp.git
+   cd openpmd-scipp
+   pip install .
+
+Getting started
+---------------
+
+Get example data sets from the ``openPMD-example-datasets`` repository.
+
+.. code:: bash
+
+   git clone https://github.com/openPMD/openPMD-example-datasets.git
+   cd openPMD-example-datasets
+   tar -zxvf example-2d.tar.gz
+   tar -zxvf example-3d.tar.gz
+
+Opening series
+~~~~~~~~~~~~~~
+
+.. code:: python
+
+   import scipp as sc
+
+   import openpmd_scipp as pmdsc
+
+.. code:: python
+
+   path = "openPMD-example-datasets/example-3d/hdf5/data%T.h5"
+
+.. code:: python
+
+   path = ".data/" + path
+
+.. code:: python
+
+   data_loader = pmdsc.DataLoader(path)
+   print(data_loader.iterations)
+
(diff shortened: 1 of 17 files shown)


pordyna:
I just took a quick look. This looks quite nice and wouldn't change much of the interface. The other stuff in the repo is mostly just for creating the README and testing, formatting etc. so I guess the only thing you would need to take over are the requirements, but this is only `numpy` and `scipp`. Also, it would make sense to make the example an integration test. Running the Jupyter notebook without verifying any outputs was also the only test I had.

franzpoeschel:
> so I guess the only thing you would need to take over are the requirements, but this is only numpy and scipp

Also the Readme; I will try to integrate that into the Readthedocs page.
For the requirements, note that the code in this PR loads `scipp` lazily, so the main API can still be used without having that installed. `numpy` is already a hard requirement, but `scipp` should be at most an optional requirement (if there is a way to tell that to the Python packaging).

> Also, it would make sense to make the example an integration test. Running the Jupyter notebook without verifying any outputs was also the only test I had.

I'd suggest that we add a test separately as part of [APITest.py](https://github.com/openPMD/openPMD-api/blob/dev/test/python/unittest/API/APITest.py). The "target audience" for the examples are users who can use them as guidelines for their own code, so I would not clutter the example with testing code.

examples/15_scipp_loader.ipynb
```diff
@@ -0,0 +1,340 @@
+{
```
ax3l, line 1: Is the notebook important to add or does the `.py` script + README have the same details?

docs/source/analysis/scipp.rst
```diff
@@ -0,0 +1,346 @@
...
+   * y                         float64              [m]  (y)  [-1e-05, -9.2e-06, ..., 9.2e-06, 1e-05]
+     z                         float64              [m]  ()  1.4e-05
+   Data:
+                               float64            [V/m]  (y)  [3.46069e+07, -2.94134e+07, ..., 8.89353e+06, -4.32182e+07]
+
+.. code:: python
+
+   Ex_line.plot()
+
+.. figure:: README_17_0.svg
```
ax3l, line 228: Can you potentially upload the four `.svg` images to a public GH gist and use them by URL? That way, we do not add larger binary artifacts to our git repo here, which keeps it small on `git clone` for CMake superbuilds.

src/binding/python/openpmd_api/scipp/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+    into scipp DataArrays.
+
+See README.md for documentation
+
+Author:
+    Pawel Ordyna <p.ordyna@hzdr.de>
+
+License:
+GPL - 3.0 license. See LICENSE file for details.
+"""
```
ax3l, line 10: @pordyna for openPMD-api, we use the LGPL license. Please confirm we can change the license for files in this PR and then please update the headers accordingly @franzpoeschel .

pordyna: yes sure go for it!

examples/15_scipp_loader.ipynb
pordyna: @ax3l the  README from my repo is identical to the notebook since I was generating my `README.md`  from it with:
```bash
rm -rf README_files
python -m nbconvert --to notebook --execute README.ipynb --output README.executed.ipynb
python -m nbconvert --to markdown README.executed.ipynb --output README.md - TagRemovePreprocessor.enabled=True --TagRemovePreprocessor.remove_cell_tags remove_cell
rm -f README.executed.ipynb
```
so you don't need both

docs/source/analysis/scipp.rst
franzpoeschel: SVG files are not binary, these are text files with each ~800 lines, less than many of our source files.
I tried putting them on a [Gist](https://gist.github.com/franzpoeschel/de7e3ef4b7b837c45387ffafac10f961), but it seems that Github does not hand out usable raw links.

examples/15_scipp_loader.ipynb
franzpoeschel: I added both since having this documentation page alternatively as an executable and interactive Notebook would be a useful extra for learning how to use this. It's of course not ideal for maintenance..

pordyna:
Btw. when it comes to handling particles, I have an idea on what would make sense there, but haven't implemented anything yet.  Since with particle data people are mostly just creating histograms, I think reading a species as a `DataArray` of weighting with all other attributes as coordinates would work well with the workflows described here: https://scipp.github.io/user-guide/binned-data/binned-data.html. This should be probably a separate PR though.

pordyna:
And the particle part could be, probably, quite easily implemented with https://scipp.github.io/user-guide/reading-and-writing-files.html#Using-pandas. Just load to a data frame with `to_df()` first and use this to convert it.

src/binding/python/openpmd_api/__init__.py
```diff
@@ -16,6 +17,7 @@
 Record_Component.to_dask_array = record_component_to_daskarray  # noqa
 Series.to_df = iterations_to_dataframe  # noqa
 Series.to_cudf = iterations_to_cudf  # noqa
+Series.to_scipp = series_to_scipp # noqa
```
ax3l, line 20:
```suggestion
Series.to_scipp = series_to_scipp  # noqa
```

pordyna:
@franzpoeschel I discovered some problems with the implementation. I already fixed them, but we need to bring the changes here. For that, I just want to update my docstrings first. 

What was wrong:
 * the relay functionality was not working
     *  the full mesh array was always allocated -> I rewrote the `DataRelay` class
 * there was an extra copy when loading data, since `sc.array` always copies values -> switched to `sc.empty` + `load_chunk` with preallocated buffer 
 * the `isclose` contiguity check was not good enough, using `sc.islinspace` instead.

franzpoeschel:
> @franzpoeschel I discovered some problems with the implementation. I already fixed them, but we need to bring the changes here. For that, I just want to update my docstrings first.
> 
> What was wrong:
> 
>     * the relay functionality was not working
>       
>       * the full mesh array was always allocated -> I rewrote the `DataRelay` class
> 
>     * there was an extra copy when loading data, since `sc.array` always copies values -> switched to `sc.empty` + `load_chunk` with preallocated buffer
> 
>     * the `isclose` contiguity check was not good enough, using `sc.islinspace` instead.

Hey @pordyna, can you implement those changes on a branch so I can pull them into this PR?

pordyna:
> Hey @pordyna, can you implement those changes on a branch so I can pull them into this PR?

Yes, I can branch from your branch and add the changes.

PR #1721, opened 2025-02-20
CI: Replace Appveyor for Win32

ax3l:
The installations on AppVeyor are very old and slow. We currently test a default Win64 and Win32 build with MSVC. The former we also test on GH Actions already.

One thing we want to cover in CI are the different fundamental types of C++ (https://en.cppreference.com/w/cpp/language/types) and in particular the Windows 32bit build for some edge computers in laboratories (in particular, some older device drivers are win32 on some systems). This tries to move this to a GH Action runner, using 32bit libraries on a 64bit host: https://github.com/conda-incubator/setup-miniconda?tab=readme-ov-file#example-11-alternative-architectures


ax3l:
Ah, looks like HDF5 was not build in 32bit yet.

Issue #1655, opened 2024-07-25
file based bp5 writer hang 


guj:
**Describe the bug**
The recent optimization breaks a MPI use case when in file based mode. A minimal code is included below. One can use 2 ranks to see effect. 
In short, at the second flush, rank 1 has nothing to contribute, so it didn't call BP5 while rank 0 did.  In essence, BP5 write is collective.  So rank 0 hangs because inactivity of rank 1. 
If we use variable based, it looks like a flush to ADIOS is forced  (by openPMD-api? ) on all ranks and so it works. 

**To Reproduce**
c++ example: 
```C++
#include <openPMD/openPMD.hpp>
#include <mpi.h>
#include <iostream>
#include <memory>

using std::cout;
using namespace openPMD;

int main(int argc, char *argv[])
{
    int provided;
    MPI_Init_thread(&argc, &argv, MPI_THREAD_MULTIPLE, &provided);

    int mpi_size;
    int mpi_rank;

    MPI_Comm_size(MPI_COMM_WORLD, &mpi_size);
    MPI_Comm_rank(MPI_COMM_WORLD, &mpi_rank);

    auto const value = float(mpi_size*100+mpi_rank);
    std::vector<float> local_data(10 * 300, value);

    std::string filename = "ptl_%T.bp";
    //std::string filename = "ptl.bp";  //this is variable based and it works                                                                                                        

    Series series = Series(filename, Access::CREATE, MPI_COMM_WORLD);

    Datatype datatype = determineDatatype<float>();

    auto myptl = series.writeIterations()[1].particles["ion"];
    Extent global_ptl = {10ul * mpi_size * 300};
    Dataset dataset_ptl = Dataset(datatype, global_ptl, "{}");
    myptl["charge"].resetDataset(dataset_ptl);

    series.flush();

    if (mpi_rank == 0)     // only rank 0 adds data                                                                                                                                  
        myptl["charge"].storeChunk(local_data, {0}, {3000});

    series.flush(); // hangs here                                                                                                                                                    
    MPI_Finalize();

    return 0;
}
```
 

**Software Environment**
 - version of openPMD-api:  latest
 - machine:  Mac
 - version of ADIOS2: latest

**Additional context**
-  I used OPENPMD_ADIOS2_BP5_TypeAgg=EveryoneWritesSerial, but any choice of aggregation will fail.
-  run with 2 cores is enough to see file based has issue. 
- As far as I was aware, this use case  worked fine not long ago. 
[...]

PR #1728, opened 2025-03-04
Further workaround for #1655: Avoid hangup automatically when using writeIterations()

franzpoeschel:
When using `writeIterations()`, the currently active Iteration should be flushed regardless if it is dirty (unless it has been closed).

#1655, @guj

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -791,6 +791,7 @@ if(openPMD_BUILD_TESTING)
             list(APPEND ${out_list}
                 test/Files_ParallelIO/read_variablebased_randomaccess.cpp
                 test/Files_ParallelIO/iterate_nonstreaming_series.cpp
+                test/Files_ParallelIO/bug_1655_bp5_writer_hangup.cpp
             )
         elseif(${test_name} STREQUAL "Core")
             list(APPEND ${out_list}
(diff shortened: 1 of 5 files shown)


guj:
originally writeIterations()[timestep]  was created to serve  for adios streaming steps.  The file based  write iterations were still using series->iterations[timestep]. As we discussed yesterday, sounds like the distinction is not there anymore. I just want to double check. An iteration will either be from writeIterations() or readIterations()?

guj:
By the way, thanks for putting the test in.

ax3l (approved):
Thanks for the fix!

franzpoeschel:
> originally writeIterations()[timestep] was created to serve for adios streaming steps. The file based write iterations were still using series->iterations[timestep]. As we discussed yesterday, sounds like the distinction is not there anymore. I just want to double check. An iteration will either be from writeIterations() or readIterations()?

`writeIterations()` was always usable both for streaming and file writing, see [the documentation](https://openpmd-api.readthedocs.io/en/0.16.1/usage/streaming.html):

> The writing end of the streaming API enforces further restrictions that become necessary through the nature of streaming. It can be used to write any kind of openPMD-compatible dataset, stream-based and filesystem-based alike.

`writeIterations()` is necessary to create ADIOS2 output that uses steps; but it never mattered if that output were files or a stream. (Other backends can be used, too, where steps will then just be ignored)

guj (commented):
Out of curiosity, if I use iterations() instead of writeIterations, 
will this fix still work?

PR #1730, opened 2025-03-04
Followup: Bug fixes for random-access in variable encoding

franzpoeschel:
- For deferred read operations, the step selection from the time of enqueuing must be used rather than the step selection from the time of running
- Fix warning message

- [x] Test for this

@ax3l With this fix, the openPMD-viewer is able to read from variable-encoded datasets


franzpoeschel:
brownout of Ubuntu 20.04 runner
Bildschirmfoto vom 2025-03-05 10-45-29
Bildschirmfoto vom 2025-03-05 10-45-42

ax3l (approved):

PR #1732, opened 2025-03-05
BP5+groupbased: allow only up to 100 steps

franzpoeschel:
Check:

* https://github.com/openPMD/openPMD-api/discussions/1724
* https://github.com/openPMD/openPMD-api/issues/1457

In a test simulation with PIConGPU, this canceled the simulation after step 999, creating output that was still readable, but had already accumulated ~1GB of metadata.

TODO:

- [x] Add an environment variable to allow the user to set this

### DIFF
--- a/docs/source/backends/adios2.rst
+++ b/docs/source/backends/adios2.rst
@@ -91,7 +91,7 @@ environment variable                    default    description
 ``OPENPMD_ADIOS2_BP5_NumSubFiles``      ``0``      ADIOS2 BP5 engine: num of subfiles
 ``OPENPMD_ADIOS2_BP5_NumAgg``           ``0``      ADIOS2 BP5 engine: num of aggregators
 ``OPENPMD_ADIOS2_BP5_TypeAgg``          *empty*    ADIOS2 BP5 engine: aggregation type. (EveryoneWrites, EveryoneWritesSerial, TwoLevelShm)
-``OPENPMD_BP5_GROUPENCODING_MAX_STEPS`` ``1000``   ADIOS2 BP5 engine: max number of allowed output steps in group encoding.
+``OPENPMD_BP5_GROUPENCODING_MAX_STEPS`` ``100``    ADIOS2 BP5 engine: max number of allowed output steps in group encoding.
 ======================================= ========== ================================================================================
 
 Please refer to the `ADIOS2 documentation <https://adios2.readthedocs.io/en/latest/engines/engines.html>`_ for details on I/O tuning.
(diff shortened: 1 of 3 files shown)


docs/source/backends/adios2.rst
```diff
@@ -75,23 +75,24 @@ The ADIOS2 SST engine for streaming can be picked by specifying the ending ``.ss
...
+``OPENPMD2_ADIOS2_USE_GROUP_TABLE``     ``0``      Use group table (see below)
+``OPENPMD_ADIOS2_STATS_LEVEL``          ``0``      whether to generate statistics for variables in ADIOS2. (``1``: yes, ``0``: no).
+``OPENPMD_ADIOS2_ASYNC_WRITE``          ``0``      ADIOS2 BP5 engine: 1 means setting "AsyncWrite" in ADIOS2 to "on". Flushes will go to the buffer by default (see ``preferred_flush_target``).
+``OPENPMD_ADIOS2_BP5_BufferChunkMB``    ``0``      ADIOS2 BP5 engine: applies when using either EveryoneWrites or EveryoneWritesSerial aggregation
+``OPENPMD_ADIOS2_BP5_MaxShmMB``         ``0``      ADIOS2 BP5 engine: applies when using TwoLevelShm aggregation
+``OPENPMD_ADIOS2_BP5_NumSubFiles``      ``0``      ADIOS2 BP5 engine: num of subfiles
+``OPENPMD_ADIOS2_BP5_NumAgg``           ``0``      ADIOS2 BP5 engine: num of aggregators
+``OPENPMD_ADIOS2_BP5_TypeAgg``          *empty*    ADIOS2 BP5 engine: aggregation type. (EveryoneWrites, EveryoneWritesSerial, TwoLevelShm)
+``OPENPMD_BP5_GROUPENCODING_MAX_STEPS`` ``1000``   ADIOS2 BP5 engine: max number of allowed output steps in group encoding.
```
ax3l: Let's lower to 100 or 200 by default, to cover more meta-data heavy cases safer by default.

guj:
Is it possible to automatically switch to file based after this OPENPMD_BP5_GROUPENCODING_MAX_STEP ?

I will find it annoying to have to restart the simulation due to this limit.

franzpoeschel:
> I will find it annoying to have to restart the simulation due to this limit.

If you have simulations that run into this limit, that means that you are probably already creating amounts of metadata with a size orders of magnitude more than in a "good" configuration for BP5.
We printed warnings for this until now, but the behavior is serious enough to switch to a clear error. 

Switching to file encoding during a running simulation *might* be possible, but more effort than it is worth. The proper solution will be to use variable encoding for this kind of setup automatically and I prefer putting my time into making that work properly.

guj:
> > I will find it annoying to have to restart the simulation due to this limit.
> 
> If you have simulations that run into this limit, that means that you are probably already creating amounts of metadata with a size orders of magnitude more than in a "good" configuration for BP5. We printed warnings for this until now, but the behavior is serious enough to switch to a clear error.
> 
> Switching to file encoding during a running simulation _might_ be possible, but more effort than it is worth. The proper solution will be to use variable encoding for this kind of setup automatically and I prefer putting my time into making that work properly.

OK. Thanks

PR #1734, opened 2025-03-12
Wheels: More on GH Actions

ax3l:
Build aarch64/arm64 on native arm runners ([GH action feature preview](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)).

Build ppc64le on emulated x64 runners on GH action (Travis keeps cancelling our runners even though they need 0 credits; support always needs to unblock us). Related to https://github.com/BLAST-WarpX/warpx/pull/5759

Update Ubuntu host systems to 24.04: we anyway run in docker containers and the 20.04 host is about to be removed from GH actions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,28 +10,39 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-20.04
+          - os: ubuntu-24.04
             arch: "x86_64"
+            env:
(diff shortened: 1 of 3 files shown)


ax3l:
Ok, just got Travis CI support and building 0.16.1 now for Linux on ppc64le and arm64.

PR #1731, opened 2025-03-05
CI: Move Ubuntu 20.04 workflows to 22.04

franzpoeschel:
Ubuntu 20.04 runners shutting down April 1st

PR #1739, opened 2025-03-24
Reactivate MPICH runner (gcc13_py312_mpich_h5_ad2)

franzpoeschel:
Deactivated temporarily in #1731 due to broken package in Ubuntu 24.04. This builds MPICH manually as a workaround.
https://bugs.launchpad.net/ubuntu/+source/mpich/+bug/2072338


franzpoeschel:
Seems like something is wrong with the MPI installation

```
> mpirun -n 4 './ParallelIOTests'
SIZE: 1 RANK: 0
SIZE: 1 RANK: 0
SIZE: 1 RANK: 0
SIZE: 1 RANK: 0
```

franzpoeschel:
The MPICH package of Ubuntu 24.04 is broken: https://bugs.launchpad.net/ubuntu/+source/mpich/+bug/2072338
No fix has been released yet, so the only options seems like self-compilation…

franzpoeschel:
Runner uses 39min with MPICH from Spack. Next compare the time used when manually compilining.
UPDATE: Yeah, manual compilation takes only 24min since we don't need to pull ten thousand Spack dependencies of MPICH.

PR #1740, opened 2025-03-28
working around an unusual encounter when the joined_dim has actual value "max::size_t - 1"

guj:
from time to time, I see exception was caught for joined array cases here:

 for (unsigned int i = 0; i < actualDim; i++)
   {
                if (!(joinedDim.has_value() && *joinedDim == i) &&
                    offset[i] + extent[i] > shape[I])
    }
because offset is given as {} for joined array variable.

the shape of the joined var is 18446744073709551615

 joined array: 
- from application:  
      - particle variable is initialized to  openPMD::Dataset::JOINED_DIMENSION
- definition in openPMD: 
      - Dataset.hpp:        JOINED_DIMENSION = std::numeric_limits<std::uint64_t>::max() 
      - ADIOS2IOHandler.cpp: finding  if (dims[i] == adios2::JoinedDim) 
- from ADIOS:
      -  ADIOSTypes.h:              constexpr uint64_t JoinedDim = MaxU64 - 1;
   
The reason it reached here is because the joined dimension from ADIOS is 18446744073709551614,  one less than the supposed max of size_t,  18446744073709551615 , which is the shape of the variable.

This pull request checked the variable shape ID and acknowledge as joined array if adios2::ShapeID::JoinedArary is returned.

### DIFF
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -509,32 +509,34 @@ private:
         }
         else
         {
-           if (var.ShapeID() ==  adios2::ShapeID::JoinedArray)
-           {
-               // When you reach here due to a mysterious inconsistency observed from time to time
-               // e.g. adios2::JoinedDim=(~(size_t)0) which for 64 bits = 18446744073709551615
-               // but the particle shape,  when printed out,  turns out to be
-               // [18446744073709551614]
-               //
-               if (!offset.empty())
-               {
-                 throw std::runtime_error(
-                     "[ADIOS2] Offset must be an empty vector in case of joined "
-                     "array.");
-               }
-           }
-           else
-           {
-            for (unsigned int i = 0; i < actualDim; i++)
+            if (var.ShapeID() == adios2::ShapeID::JoinedArray)
             {
-                if (!(joinedDim.has_value() && *joinedDim == i) &&
-                    offset[i] + extent[i] > shape[i])
+                // When you reach here due to a mysterious inconsistency
+                // observed from time to time e.g.
+                // adios2::JoinedDim=(~(size_t)0) which for 64 bits =
+                // 18446744073709551615 but the particle shape,  when printed
+                // out,  turns out to be [18446744073709551614]
+                //
+                if (!offset.empty())
                 {
                     throw std::runtime_error(
-                        "[ADIOS2] Dataset access out of bounds.");
+                        "[ADIOS2] Offset must be an empty vector in case of "
+                        "joined "
+                        "array.");
                 }
             }
-	   }// else
+            else
+            {
+                for (unsigned int i = 0; i < actualDim; i++)
+                {
+                    if (!(joinedDim.has_value() && *joinedDim == i) &&
+                        offset[i] + extent[i] > shape[i])
+                    {
+                        throw std::runtime_error(
+                            "[ADIOS2] Dataset access out of bounds.");
+                    }
+                }
+            } // else
         }
 
         var.SetSelection(
--- a/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
+++ b/include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
@@ -542,7 +542,7 @@ private:
                     throw make_runtime_error("Dataset access out of bounds.");
                 }
             }
-        } // else
+        }
 
         var.SetSelection(
             {adios2::Dims(offset.begin(), offset.end()),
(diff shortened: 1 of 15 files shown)


include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -509,15 +509,34 @@ class ADIOS2IOHandlerImpl
...
+                // 18446744073709551615 but the particle shape,  when printed
+                // out,  turns out to be [18446744073709551614]
+                //
+                if (!offset.empty())
                 {
                     throw std::runtime_error(
-                        "[ADIOS2] Dataset access out of bounds.");
+                        "[ADIOS2] Offset must be an empty vector in case of "
+                        "joined "
+                        "array.");
```
franzpoeschel: This duplicates the condition checked in the if branch from line 492 to 509. 
Can't we just change the if condition in line 492 to:
```cpp
if (joinedDim.has_value() || var.ShapeID() == adios2::ShapeID::JoinedArray)
```
This would be a single-line change and do basically the same

guj: It will break the for loop at 500

guj (commented):
It will break the for loop at 500

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
franzpoeschel: That issue only comes up when `joinedDim.has_value()` is false, but `var.ShapeID() == adios2::ShapeID::JoinedArray` is true, which we should treat as an error anyway?
I think I'll add a check for that circumstance.

franzpoeschel:
I went through the implementation of joined arrays again and added some further safety guards. I think that I might have an idea on what's going on in your crashing simulations: Joined arrays were until now ignored in the `extendDataset()` task, meaning that `openPMD::Dataset::JOINED_DIMENSION` will be propagated as-is to the ADIOS2 variable. This would explain why you see `~(size_t)0` and not `~(size_t)0 - 1` (which would be `adios2::JoinedDim`.
However, I'm not sure ADIOS2 even supports extending joined array datasets. Also, since you are using joined arrays for particle output (which is 1D), there is nothing to be extended since the joined extent is calculated on the fly anyway.

franzpoeschel:
@guj Can you try if you still observe the crashes with the recent commits on this PR? I hope that I might have found and fixed the issue.
I will not merge this yet, since I found another bug on the way and still have to find the cause for that.

franzpoeschel:
> I will not merge this yet, since I found another bug on the way and still have to find the cause for that.

Found and fixed the bug
Please still check if this fixes the issue you had, I think this might have been the problem

franzpoeschel:
CI failures due to deployment of CMake 4.0 to CI runners, will be fixed with #1742. I'll select the bugfix for the double-write from unique-pointers into a separate PR as it has nothing to do with this PR. Will rebase this onto that then.

guj:
> @guj Can you try if you still observe the crashes with the recent commits on this PR? I hope that I might have found and fixed the issue. I will not merge this yet, since I found another bug on the way and still have to find the cause for that.

The test job was finally scheduled.  Yes this version works!

guj:
> > I will not merge this yet, since I found another bug on the way and still have to find the cause for that.
> 
> Found and fixed the bug Please still check if this fixes the issue you had, I think this might have been the problem

OK I will submit another test to double check.

franzpoeschel:
> The test job was finally scheduled. Yes this version works!

Nice, thanks for checking!

> OK I will submit another test to double check.

I think it should not alter your use case since that's an entirely different issue, but feel free to run a test anyway

guj:
> > > I will not merge this yet, since I found another bug on the way and still have to find the cause for that.
> > 
> > 
> > Found and fixed the bug Please still check if this fixes the issue you had, I think this might have been the problem
> 
> OK I will submit another test to double check.

This one passed too

include/openPMD/IO/ADIOS/ADIOS2IOHandler.hpp
```diff
@@ -514,11 +539,10 @@ class ADIOS2IOHandlerImpl
                 if (!(joinedDim.has_value() && *joinedDim == i) &&
                     offset[i] + extent[i] > shape[i])
                 {
-                    throw std::runtime_error(
-                        "[ADIOS2] Dataset access out of bounds.");
+                    throw make_runtime_error("Dataset access out of bounds.");
                 }
             }
-        }
+        } // else
```
franzpoeschel:
```suggestion
        }
```

PR #1741, opened 2025-03-28
Performance optimizations for interacting with many meshes / species types

franzpoeschel:
Some assorted performance optimizations notices when developing the openPMD plugin for CP2K. These mostly affect the performance of flushing Iterations.

TODO:

- [x] Check if we should check for `dirtyRecursive()` in other places, too, or when reading. Update: No need in reading, Iterations are not reparsed.
- [x] Merge #1644 first

### DIFF
--- a/include/openPMD/backend/BaseRecord.hpp
+++ b/include/openPMD/backend/BaseRecord.hpp
@@ -28,7 +28,6 @@
 #include "openPMD/backend/Container.hpp"
 
 #include <array>
-#include <optional>
 #include <stdexcept>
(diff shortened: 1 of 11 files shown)


include/openPMD/backend/BaseRecord.hpp
```diff
@@ -28,6 +28,7 @@
 #include "openPMD/backend/Container.hpp"
 
 #include <array>
+#include <optional>
```
franzpoeschel:
```suggestion
```

PR #1746, opened 2025-04-10
Variable encoding in ADIOS2: Support partial datasets in ReadRandomAccess

franzpoeschel:
This brings full support for ReadRandomAccess in variable encoding.

TODO:

- [x] Better testing
- [x] Detect if variable preparsing is necessary and skip it if not
- [x] Documentation

### DIFF
--- a/src/IO/ADIOS/ADIOS2PreloadVariables.cpp
+++ b/src/IO/ADIOS/ADIOS2PreloadVariables.cpp
@@ -1,3 +1,24 @@
+/* Copyright 2025 Franz Poeschel
+ *
+ * This file is part of openPMD-api.
+ *
+ * openPMD-api is free software: you can redistribute it and/or modify
+ * it under the terms of of either the GNU General Public License or
+ * the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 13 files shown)


src/IO/ADIOS/ADIOS2PreloadVariables.cpp
```diff
@@ -0,0 +1,99 @@
+#include "openPMD/config.hpp"
```
franzpoeschel, line 22:
```suggestion
/* Copyright 2025 Franz Poeschel
 *
 * This file is part of openPMD-api.
 *
 * openPMD-api is free software: you can redistribute it and/or modify
 * it under the terms of of either the GNU General Public License or
 * the GNU Lesser General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * openPMD-api is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License and the GNU Lesser General Public License
 * for more details.
 *
 * You should have received a copy of the GNU General Public License
 * and the GNU Lesser General Public License along with openPMD-api.
 * If not, see <http://www.gnu.org/licenses/>.
 */
 
#include "openPMD/config.hpp"
```

PR #1748, opened 2025-04-17
Fix error when reading in linear mode from file-based Series

franzpoeschel:
```
[~Series] An error occurred: fileBased output can not be written with no iterations.
```

This happens, since the new `series.snapshots()` logic will remove Iterations from the object model once they are closed by the user, leading to a situation where at the end of the Series the Iterations container might be empty.

### DIFF
--- a/src/snapshots/StatefulIterator.cpp
+++ b/src/snapshots/StatefulIterator.cpp
@@ -648,8 +648,6 @@ std::optional<StatefulIterator *> StatefulIterator::loopBody(Seek const &seek)
             else if (
                 series.IOHandler()->m_frontendAccess == Access::READ_LINEAR)
             {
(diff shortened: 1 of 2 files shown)


src/snapshots/StatefulIterator.cpp
```diff
@@ -648,6 +648,8 @@ std::optional<StatefulIterator *> StatefulIterator::loopBody(Seek const &seek)
             else if (
                 series.IOHandler()->m_frontendAccess == Access::READ_LINEAR)
             {
+                std::cout << "Closing Iteration " << *maybe_current_iteration
```
franzpoeschel: TODO: remove this bit again
just as a reminder for myself where this error is caused

PR #1754, opened 2025-05-06
Workaround for Hipcc error (reference to host function in host device function)

franzpoeschel:
Observed on Frontier:

```
/ccs/home/fpoeschel/git-repos/openPMD-api/src/snapshots/RandomAccessIterator.cpp:13:35: error: reference to __host__ function '~AbstractSeriesIterator' in _
_host__ __device__ function                                                                                                                                 
   13 | RandomAccessIterator<iterator_t>::RandomAccessIterator(

/ccs/home/fpoeschel/git-repos/openPMD-api/src/snapshots/RandomAccessIterator.cpp:16:35: error: reference to __host__ function '~AbstractSeriesIterator' in _
_host__ __device__ function                                                                                                                                 
   16 | RandomAccessIterator<iterator_t>::RandomAccessIterator(
```

These look like compiler bugs to me. Since this is an internal header, we can move the defaulted implementation to the header file without trouble. This fixes the observed error.

### DIFF
--- a/include/openPMD/snapshots/RandomAccessIterator.hpp
+++ b/include/openPMD/snapshots/RandomAccessIterator.hpp
@@ -66,10 +66,11 @@ public:
     RandomAccessIterator(RandomAccessIterator &&other) noexcept(
         noexcept(iterator_t(std::declval<iterator_t &&>())));
 
-    RandomAccessIterator &operator=(RandomAccessIterator const &other);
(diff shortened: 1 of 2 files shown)


ax3l:
@franzpoeschel what ROCm version and hipcc/amdclang++ version are these issues with?

Issue #1751, opened 2025-04-29
Skip PerformPut for BP5


guj:
**Is your feature request related to a problem? If so, please describe.**
PerformPut is useful for BP3/BP4, and it is not needed for BP5. 

For BP5, call either PerformDataWrite, Put(sync) Put (deferred), followed by EndStep will do.

guj:
<img width="1040" alt="Image" src="https://github.com/user-attachments/assets/f779e89d-bab2-4fcc-a41c-7eeb0bbaeb88" />

Plot on the right shows extra metadata overhead (marked by red)  caused by calling PP on Frontier

franzpoeschel:
I have added a prototype for this on #1756. Can you test that please?

PR #1756, opened 2025-05-08
[WIP] BP5 Put performance optimization

franzpoeschel:
According to #1751, BP5 performance takes a hit by calling PerformPuts(). Instead, for BP5, always use Put(Sync), so that we can skip PerformPuts().

@guj Can you check if this really improves performance?

### DIFF
--- a/src/IO/ADIOS/ADIOS2File.cpp
+++ b/src/IO/ADIOS/ADIOS2File.cpp
@@ -109,7 +109,12 @@ void WriteDataset::call(ADIOS2File &ba, detail::BufferedPut &bp)
                     std::nullopt,
                     ba.variables());
 
-                engine.Put(var, ptr);
+                // @todo cache this
+                auto is_bp5 = ba.m_impl->realEngineType() == "bp5" ||
+                    auxiliary::lowerCase(engine.Type()) == "bp5writer";
+                auto do_defer =
+                    is_bp5 ? adios2::Mode::Sync : adios2::Mode::Deferred;
+                engine.Put(var, ptr, do_defer);
             }
             else if constexpr (std::is_same_v<
                                    ptr_type,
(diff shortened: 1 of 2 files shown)


guj:
I checked it. Works!!

src/IO/ADIOS/ADIOS2File.cpp
```diff
@@ -109,7 +109,9 @@ void WriteDataset::call(ADIOS2File &ba, detail::BufferedPut &bp)
                     std::nullopt,
                     ba.variables());
 
-                engine.Put(var, ptr);
+                auto do_defer =
+                    ba.m_is_bp5 ? adios2::Mode::Sync : adios2::Mode::Deferred;
+                engine.Put(var, ptr, do_defer);
```
franzpoeschel, line 114: This incurs an overhead in EndStep() / PerformDataWrite(). Use Async in those cases.

guj:
@pnorbert @eisenhauer

ax3l:
@pnorbert @eisenhauer can you take a look at this?

@franzpoeschel and I wonder if this is a performance bug that rather should be fixed in ADIOS2.

eisenhauer:
> @pnorbert @eisenhauer can you take a look at this?
> 
> @franzpoeschel and I wonder if this is a performance bug that rather should be fixed in ADIOS2.

IMHO, PerformPuts was always kind of an odd thing.  Semantically it's the equivalent of "Remember when I did put deferred earlier? Forget about that, I really meant put sync so copy the data now so I can reuse those buffers".  BP4 is always going to copy the data into internal buffers at some point, so it didn't necessarily matter much if it happened in Put, PerformPuts, or EndStep (which itself calls PerformPuts in BP4).  On the other hand, BP5 at least _has_ the ability to not copy the data at all.  I.E. if you do Put deferred, we just keep that pointer and EndStep _can_ write the data from application memory directly to disk without any copies.  Note that I say "_can_".  For smaller data blocks BP5 copies at the time of Put whether you say deferred or sync.  Some aggregators may also end up copying the data into a contiguous block.  There are various other reasons why zero copy I/O might not happen, but at least for BP5 put deferred and not doing PerformPuts (ever) gives you a chance of zero copy with BP5.  

However, if you need to reuse buffers before EndStep, you've got to force the copy sometime.  I'm not sure it would make much of a difference if you do that in Put sync or PerformPuts.

PR #1757, opened 2025-05-15
CMake: Install toml11 and nlohmann_json for inclusion in downstream code

franzpoeschel:
This is an attempt at finding a solution for this issue https://github.com/ComputationalRadiationPhysics/picongpu/issues/5355

**Problem:** If a downstream code wants to use nlohmann_json / toml11 as well, then it must use the same version used also in openPMD, otherwise it might happen that the same symbol defined by toml11/nlohmann_json might exist in different implementations (different versions) in libopenPMD.so and the compiled binary of the downstream. The linker might then jump between implementations, causing undefined behavior.

**Attempted solution in this PR:**

1. By default, do not use internally shipped dependencies, but look for an installation of toml11/nlohmann_json in the system. (not yet implemented)
2. If using an internally shipped dependency, then install it along with openPMD such that these versions become available in the software environment.
3. Downstream codes can then use the same logic to agree on common versions of toml11/nlohmann_json.

**Alternatives:**

* Maybe include toml11 / nlohmann_json into their own namespaces in openPMD instead?

TODO:

- [ ] Look for nlohmann_json/toml11 in the system software environment by default
- [ ] Implement this logic also for workflows using openPMD_toml11_src / openPMD_json_src
- [ ] Add a flag to opt-out from this. (Yes, opt-**out**, this can only be a viable solution for the described issue if we can assume that an openPMD installation will share its versions of toml11 / nlohmann_json by default)
- [ ] Open PRs at toml11/nlohmann_json for binary symbol versioning with `inline namespace`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1076,14 +1076,24 @@ if(openPMD_INSTALL)
         RUNTIME DESTINATION ${openPMD_INSTALL_BINDIR}
         INCLUDES DESTINATION ${openPMD_INSTALL_INCLUDEDIR}
     )
-    install(CODE "
-        execute_process(
-            COMMAND ${CMAKE_COMMAND} \
-                --build \"${CMAKE_BINARY_DIR}\" \
-                --target fetchedtoml11-install \
-                --target fetchednlohmann_json-install
-        )
-    ")
+    if(TARGET fetchednlohmann_json-install)
+        install(CODE "
+            execute_process(
+                COMMAND ${CMAKE_COMMAND} \
+                    --build \"${CMAKE_BINARY_DIR}\" \
+                    --target fetchednlohmann_json-install
+            )
+        ")
+    endif()
+    if(TARGET fetchedtoml11-install)
+        install(CODE "
+            execute_process(
+                COMMAND ${CMAKE_COMMAND} \
+                    --build \"${CMAKE_BINARY_DIR}\" \
+                    --target fetchedtoml11-install
+            )
+        ")
+    endif()
     # install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} --build \"${CMAKE_BINARY_DIR}\" --target toml11_install)")
     # install(CODE "execute_process(COMMAND ${CMAKE_COMMAND} --build \"${CMAKE_BINARY_DIR}\" --target json_install)")
     if(openPMD_HAVE_PYTHON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1076,6 +1076,16 @@ if(openPMD_INSTALL)
         RUNTIME DESTINATION ${openPMD_INSTALL_BINDIR}
         INCLUDES DESTINATION ${openPMD_INSTALL_INCLUDEDIR}
     )
+    install(CODE "
+        execute_process(
+            COMMAND ${CMAKE_COMMAND} \
+                --build \"${CMAKE_BINARY_DIR}\" \
+                --target fetchedtoml11-install \
+                --target fetchednlohmann_json-install
+        )
+    ")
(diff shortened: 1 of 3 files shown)


cmake/dependencies/toml11.cmake
```diff
@@ -44,6 +47,7 @@ function(find_toml11)
         #mark_as_advanced(FETCHCONTENT_SOURCE_DIR_FETCHEDtoml11)
         mark_as_advanced(FETCHCONTENT_UPDATES_DISCONNECTED)
         #mark_as_advanced(FETCHCONTENT_UPDATES_DISCONNECTED_FETCHEDtoml11)
+
```
franzpoeschel:
```suggestion
```

psychocoderHPC:
@franzpoeschel Please take care that dependency fetching from the web is on some systems not possible due to restrictions that login nodes or compute nodes are not allowed to download things from the web.
@chillenzer has experience with that because he solved the issue already in https://github.com/alpaka-group/mallocMC

franzpoeschel:
> @franzpoeschel Please take care that dependency fetching from the web is on some systems not possible due to restrictions that login nodes or compute nodes are not allowed to download things from the web. @chillenzer has experience with that because he solved the issue already in https://github.com/alpaka-group/mallocMC

Dependency fetching is already part of our build system, this PR does not change that, but rather adds an install step. Offline builds are fully supported (and we use them for distribution via package managers), see CMake variables:
```
openPMD_USE_INTERNAL_TOML11
openPMD_USE_INTERNAL_JSON
openPMD_toml11_src
openPMD_json_src
```

CMakeLists.txt
```diff
@@ -1076,6 +1076,24 @@ if(openPMD_INSTALL)
...
         INCLUDES DESTINATION ${openPMD_INSTALL_INCLUDEDIR}
     )
+    if(TARGET fetchednlohmann_json-install)
+        install(CODE "
+            execute_process(
+                COMMAND ${CMAKE_COMMAND} \
+                    --build \"${CMAKE_BINARY_DIR}\" \
+                    --target fetchednlohmann_json-install
+            )
+        ")
```
ax3l, line 1086: So far, I have intentionally turned the install of sub-projects off. Instead of this code, we could set `JSON_Install ON`:

https://github.com/nlohmann/json/blob/v3.12.0/CMakeLists.txt#L48

CMakeLists.txt
```diff
@@ -1076,6 +1076,24 @@ if(openPMD_INSTALL)
...
+    endif()
+    if(TARGET fetchedtoml11-install)
+        install(CODE "
+            execute_process(
+                COMMAND ${CMAKE_COMMAND} \
+                    --build \"${CMAKE_BINARY_DIR}\" \
+                    --target fetchedtoml11-install
```
ax3l, line 1093: `TOML11_INSTALL` is the equivalent option there:
https://github.com/ToruNiina/toml11/blob/v4.4.0/CMakeLists.txt#L27-L28

CMakeLists.txt
chillenzer, line 1086: Different question: Why would you hack together a manual installation here? Isn't it conceptually closer to what you want to express to link to json/toml11 publicly?

franzpoeschel: See Axel's comment above; this did not work because I was not aware about CMake options in nlohmann-json/toml11 which we set to suppress installations.
I'm only leaving this PR open rn because we might in future add nlohmann-json/toml11 functionality to our public API, in which case we would need to install these libraries somehow along with the main library. There is at the moment no concrete plan to work on this further.

franzpoeschel:
Closing this for now, might reuse the code later.

PR #1768, opened 2025-07-11
Snapshots API: Fill missing bits, harmonize API

franzpoeschel:
* Adds Python API
* Adds synchronous/linear access modes for write side, removes confusing parameter from `Series::snapshots()` call

TODO:

- [x] Update documentation in different places (examples!)
- [x] Use new API in tests and examples more
- [x] ~~Should we automatically `open()` Iterations when using `READ_RANDOM_ACCESS`, `snapshot()` and deferred parsing?~~ I have a prototype for this implemented on https://github.com/franzpoeschel/openPMD-api/tree/use-snapshots-in-tests, but let's move it out to another PR


ax3l (approved):
Fantastic, thank you!

PR #1771, opened 2025-07-16
Deactivate Span API by default in BP5

franzpoeschel:
Workaround for this bug https://github.com/ornladios/ADIOS2/issues/4586.
Since BP5 can write directly from user buffers, this change (essentially activating our fallback implementation) is not too costly.

TODO:

- [x]  Wait for an answer in the issue linked above. The current draft assumes that this will be fixed in upcoming versions, but let's be sure of that.

### DIFF
--- a/src/IO/ADIOS/ADIOS2IOHandler.cpp
+++ b/src/IO/ADIOS/ADIOS2IOHandler.cpp
@@ -1247,6 +1247,11 @@ void ADIOS2IOHandlerImpl::getBufferView(
     Writable *writable, Parameter<Operation::GET_BUFFER_VIEW> &parameters)
 {
     // @todo check access mode
+    /*
+     * We will check BP5 again below. BP5 fundamentally supports the Span API,
+     * but runs into this bug https://github.com/ornladios/ADIOS2/issues/4586,
+     * so we treat the Span API as opt-in there.
+     */
     std::string optInEngines[] = {"bp4", "bp5", "file", "filestream"};
     if (std::none_of(
             begin(optInEngines),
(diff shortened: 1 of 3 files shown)


guj:
This bug was fixed in BP5.  Should span be allowed again in BP5?

franzpoeschel:
> This bug was fixed in BP5. Should span be allowed again in BP5?

It is allowed based on the discovered ADIOS2 version, any version newer than 2.10.2 will use the Span API by default again:
```cpp
#if (                                                                          \
    ADIOS2_VERSION_MAJOR * 1000 + ADIOS2_VERSION_MINOR * 10 +                  \
    ADIOS2_VERSION_PATCH) <= 2102
            ||
            // Deactivate the Span API in BP5 by default due to this bug
            // https://github.com/ornladios/ADIOS2/issues/4586,
            // fixed by https://github.com/ornladios/ADIOS2/pull/4587
            this->realEngineType() == "bp5"
#endif
```

Do you know if the error you saw is related to this?

guj:
The one from master is 2.10.0.735 from bpls -V
so that's why span was deactivated.

src/IO/ADIOS/ADIOS2IOHandler.cpp
```diff
@@ -1270,7 +1276,17 @@ void ADIOS2IOHandlerImpl::getBufferView(
         return;
     case UseSpan::Auto:
         if (switchAdios2VariableType<detail::HasOperators>(
-                parameters.dtype, name, ba.m_IO))
+                parameters.dtype, name, ba.m_IO)
+#if (                                                                          \
+    ADIOS2_VERSION_MAJOR * 1000 + ADIOS2_VERSION_MINOR * 10 +                  \
+    ADIOS2_VERSION_PATCH) <= 2102
```
ax3l, line 1282: @franzpoeschel do I read correctly that the span-based API is broken for ADIOS2 of all versions unless ADIOS2 in a version newer than 2.10.2 is used? In other words, we cannot use it just yet in production (unless we use ADIOS2 development) and we do not need any further changes to this check here (with https://github.com/ornladios/ADIOS2/pull/4587 merged)?

ax3l: Ah, see your comment above. Exactly

PR #1774, opened 2025-08-04
Optimize include time, push definitions to object files

### LINKED ISSUE #5984: PhysicalParticleContainer.cpp takes a long time to compile
I compiled a clean build of WarpX with 50 threads using HIP clang and ninja (which is 25% faster than the default cmake) and generated a trace using [these](https://gist.github.com/jrhemstad/811d1e41af9685670241fa472170ef32) instructions. The `PhysicalParticleContainer.cpp` file accounts for a large chunk of the time due to taking the longest of all files and starting rather late (I suspect because it comes late in alphabetical order). I tried to profile the file individually using `-ftime-trace` however, this only traces the host compiler and not the offload GPU compiler, which takes most of the time.

It would be good to split some of the large kernels like push and AddPlasma out of `PhysicalParticleContainer.cpp` into separate new .cpp files to greatly reduce compile times when using many threads.

Image

franzpoeschel:
Fix https://github.com/BLAST-WarpX/warpx/issues/5984#issuecomment-3021083493 

Including openPMD currently takes too much time. The main culprit seems to be our use of `std::variant` for all datatypes in the Datatype enum.

This PR is somewhat macro-heavy to keep down compile times on our end.

- [x] Do not use `std::variant` in public headers of `openPMD::Variant`, move all use of it to object files, use `std::any` publically. This breaks the API, but I think that this part of the API was rarely used. 
- [x] Move most storeChunk/loadChunk implementations to object files
- [ ] Add an extra header to bring back the removed functionality?
- [x] Instantiate more things in object files, removing them from public includes

It is somewhat likely that `std::variant` was very costly due to things like `Attribute::get<T>()`: The requested type `T` must be matched against the actual type `A`, leading to 38²=1444 instantiations. If naively pushing templates to the object files, this leads to very big compiled binaries.  Some short-circuiting logic (pre-filling matrixes of which types are actually convertible and only instantiating in those cases when they are) helped in there.

** Results **

The cut on compile times is significant. The SerialIOTests (which have gotten too huge and take ages to compile) used to compile in ~ a minute, now it's half a minute:
Bildschirmfoto vom 2025-08-05 19-18-25
Bildschirmfoto vom 2025-08-05 19-18-12

For a more normal example, I'll compile 10_streaming_write, which includes `openPMD.hpp` and does some basic transforma
Bildschirmfoto vom 2025-08-05 19-23-17
tions.  Down to 2 seconds from 5 seconds:
Bildschirmfoto vom 2025-08-05 19-18-46

Funny enough, the library size slightly shrinks from 9.8MB (Clang) / 8.5MB (g++) to 8.5MB (Clang) / 7.6MB (g++), despite pulling definitions into object files.
My guess is that most definitions were instantiated anyway, but the additional logic to cut away type conversion instantiations brings some size benefit.


franzpoeschel:
Headers whose implementations could be moved to a .cpp file:

include/openPMD/Span.hpp
include/openPMD/backend/PatchRecordComponent.hpp
✓ include/openPMD/backend/Container.hpp 
✓ include/openPMD/backend/BaseRecord.hpp
✓ include/openPMD/auxiliary/UniquePtr.hpp // except those templates with a Del parameter
✓ include/openPMD/auxiliary/Memory.hpp
include/openPMD/auxiliary/StringManip.hpp
include/openPMD/auxiliary/ShareRawInternal.hpp
include/openPMD/auxiliary/OutOfRangeMsg.hpp
include/openPMD/auxiliary/Environment.hpp

include/openPMD/Datatype.hpp // but keep constexpr implementations public

*some single functions:*
include/openPMD/ParticleSpecies.hpp
include/openPMD/backend/Attributable.hpp
include/openPMD/backend/BaseRecordComponent.hpp

ax3l (approved):
Awesome, thank you for digging this!

PR #1780, opened 2025-08-22
Doc: First Write `np.float64`

ax3l:
Newer numpy removed the alias for `np.float`, so use an explicit type now.


franzpoeschel (approved):
Thank you!

PR #1783, opened 2025-09-09
Clang Tidy and Sanitizer: Use clang-19 on Ubuntu 24.04

franzpoeschel:
* Updates Spack to version 1.0.1 for the tooling runs
* Updates the tooling runs to Ubuntu 24.04 with clang-19
* Updates the Spack config files to >= v.1.0.0 format
* Applies some of the new hints found by clang-tidy

### DIFF
--- a/.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
+++ b/.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
@@ -49,14 +49,16 @@ spack:
       variants: ~fortran
       providers:
         mpi: [openmpi]
-        c: [clang]
-        cxx: [clang]
+        c: [llvm]
+        cxx: [llvm]
         fortran: [gcc]
         gcc-runtime: [gcc-runtime@13.3.0]
+      prefer:
+      - "%llvm"
 
-    clang:
+    llvm:
       externals:
-      - spec: clang@19.1.1 target=x86_64
+      - spec: llvm@19.1.1 target=x86_64
         prefix: /
         paths:
           cc: /usr/bin/clang-19
--- a/.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
+++ b/.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
@@ -43,6 +43,8 @@ spack:
       variants: ~fortran
       providers:
         mpi: [openmpi]
+        c: [clang@19.1.1]
+        cxx: [clang@19.1.1]
       compiler: [clang@19.1.1]
 
   compilers:
--- a/.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
+++ b/.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
@@ -43,7 +43,7 @@ spack:
       variants: ~fortran
       providers:
         mpi: [openmpi]
-      compiler: [clang@19.0.0]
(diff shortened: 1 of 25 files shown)


.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
```diff
@@ -14,54 +14,78 @@ spack:
...
+      externals:
+        - spec: gcc-runtime@13.3.0
+          prefix: /usr
 
   config:
     build_jobs: 2
 
   mirrors:
-    E4S: https://cache.e4s.io
+    spack-public: https://mirror.spack.io
```
franzpoeschel, line 91: Please check the change of mirror URL, cache.e4s.io is outdated

.github/workflows/tooling.yml
```diff
@@ -36,32 +37,31 @@ jobs:
...
+      env: {CC: mpicc, CXX: mpic++, OMPI_CC: clang-19, OMPI_CXX: clang++-19, CXXFLAGS: -Werror, OPENPMD_HDF5_CHUNKS: none, OPENPMD_TEST_NFILES_MAX: 100}
       run: |
         sudo ln -s "$(which cmake)" /usr/bin/cmake
         eval $(spack env activate --sh .github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/)
         spack install
         SOURCEPATH="$(pwd)"
         share/openPMD/download_samples.sh build
         export LDFLAGS="${LDFLAGS} -fsanitize=address,undefined -shared-libsan"
-        export CXXFLAGS="${CXXFLAGS} -fsanitize=address,undefined -shared-libsan"
+        export CXXFLAGS="${CXXFLAGS} -fsanitize=address,undefined -shared-libsan -DOMPI_SKIP_MPICXX"
```
franzpoeschel, line 64: Please check `-DOMPI_SKIP_MPICXX`. For some reason, this run pulls in the C++ bindings of MPI and then fails. This can be avoided with this preprocessor define, but there might be a better solution. I see that we have some finicky config in `CMakeLists.txt` on whether we link with `MPI::MPI_C` or `MPI::MPI_CXX`. Changing one to the other made no difference in my tests.

franzpoeschel (commented):
Two notes below for review @ax3l

.github/workflows/tooling.yml
ax3l: I found in the past that surprisingly, the C API of MPI sometimes depends on the (outdated) C++ bindings (e.g., BullMPI, which is an OpenMPI variant)

.github/ci/spack-envs/clangtidy_nopy_ompi_h5_ad2/spack.yaml
ax3l: yes, perfect! :)

.github/workflows/tooling.yml
ax3l: We need `C CXX` but not `MPICXX`
https://cmake.org/cmake/help/latest/module/FindMPI.html

ax3l: There is a CMake option `MPI_CXX_SKIP_MPICXX`
> Add some definitions that will disable the MPI-2 C++ bindings. Currently supported are MPICH, Open MPI, Platform MPI and derivatives thereof, for example MVAPICH or Intel MPI.

Let's add that! :)

ax3l: We will do that in a follow-up PR, so we can patch it out if it causes issues after all.

ax3l (approved):
Thank you, splendid! :rocket: :sparkles:

PR #1785, opened 2025-09-16
CMake: Skip MPICXX

ax3l:
We do not use the MPI-2 only MPI C++ bindings, but some implemenations (esp. OpenMPI) still create an internal dependency on it, which is unnecessary and complicates tooling and deployments.

Follow-up to #1783


franzpoeschel (approved):
Thank you!

PR #1789, opened 2025-09-30
External block storage for JSON/TOML backend via file system or S3

franzpoeschel:
TODO

- [ ] error handling
- [ ] can we use ADIOS2 as a block storage backend?
- [ ] do not always try creating the bucket, only if we need to
- [x] JSON merging (MPI)
- [x] async IO
- [ ] (probably for later PR) non-matching load operations
- [x] merge script: read file names from stdin, not from command line args
- [ ] ensure that restarting will not overwrite old blocks
- [ ] JSON backend: provide backend-specific implementation for Span API: we profit from using unique_ptr types in async AWS
- [ ] do we need some handling for runningIndex in the JSON files?


src/IO/JSON/JSONIOHandlerImpl.cpp
```diff
@@ -270,38 +484,27 @@ auto JSONIOHandlerImpl::retrieveDatasetMode(
...
-                    skipWarnings = true;
+                    if (access::writeOnly(m_handler->m_backendAccess))
+                    {
+                        parse_external_mode(
+                            std::move(mode), std::nullopt, configLocation, res);
+                    }
+                    else
+                    {
+                        // sic! initialize the deferred json config as a new
+                        // tracing object
```
franzpoeschel, line 525: TODO: Return warning upon unused keys

PR #1791, opened 2025-10-13
Added a simple btd example

guj:
This is a simple version of BTD. While writing this, it looks like something unexpected is observed.

to reproduce: 
1. compile and run 16_btd_write_parallel 
2. use more than one rank (2 is ok)
3. use -s (span) with -b (mpi barrier before flush)
4. run "mpirun -n 2 16_btd_write_parallel -s " works, however "mpirun -n 2 16_btd_write_parallel -s -b " hangs.  It is not able to finish through the storeChunk call. 

@franzpoeschel  if you have time, please verify.

### DIFF
--- a/examples/16_btd_write_parallel.cpp
+++ b/examples/16_btd_write_parallel.cpp
@@ -101,10 +101,10 @@ void setupMeshComp(
             Dataset dataset = Dataset(datatype, global_extent);
 
             curr_mesh_comp.resetDataset(dataset);
-            if (m_verbose)
+            if (m_verbose and !currRank)
             {
-                cout << "Rank : " << currRank << " Prepared a Dataset [" << ff
-                     << "/" << cc << "] of size " << dataset.extent[0] << " x "
+                cout << "AllRanks: Prepared a Dataset [" << ff << "/" << cc
+                     << "] of size " << dataset.extent[0] << " x "
                      << dataset.extent[1] << " x " << dataset.extent[2]
                      << " and Datatype " << dataset.dtype
                      << " iteration=" << w.whichSnapshot << '\n';
@@ -121,23 +121,24 @@ void doFlush(
 
     if (m_barrier_at_flush)
     {
-        if (m_verbose)
-            std::cout << " Barrier at doFlush(), rank:" << currRank
-                      << std::endl;
+        if (m_verbose and !currRank)
+            std::cout << "AllRanks: Barrier at doFlush() ";
+
         MPI_Barrier(MPI_COMM_WORLD);
     }
 
     else
     {
-        if (m_verbose)
-            std::cout << " At doFlush(), rank:" << currRank << std::endl;
+        if (m_verbose and !currRank)
+            std::cout << "AllRanks: At doFlush(), ";
     }
 
-    series->iterations[w.whichSnapshot].seriesFlush(
-        "adios2.engine.preferred_flush_target = \"buffer\"");
-
-    if ((w.whichBuffer % m_diskFlushFrequency) == 0)
+    if ((m_diskFlushFrequency > 0) &&
+        (w.whichBuffer % m_diskFlushFrequency) == 0)
     {
+        if (m_verbose and !currRank)
+            std::cout << " To disk!" << std::endl;
+
         if (m_adiosFlattenSteps)
         {
             series->iterations[w.whichSnapshot].seriesFlush(
@@ -148,6 +149,14 @@ void doFlush(
             series->iterations[w.whichSnapshot].seriesFlush(
                 R"(adios2.engine.preferred_flush_target = "disk")");
         }
+    }
+    else
+    {
+        if (m_verbose and !currRank)
+            std::cout << " To  buffer." << std::endl;
+
+        series->iterations[w.whichSnapshot].seriesFlush(
+            "adios2.engine.preferred_flush_target = \"buffer\"");
     }
 }
 
@@ -183,17 +192,19 @@ void doWork(
         Offset chunk_offset = {
             m_blockX * (m_buffers.size() - w.whichBuffer), 0, 0};
         Extent chunk_extent = {m_blockX, m_blockY, m_blockZ};
-        if (m_verbose)
-        {
-            cout << "Rank: " << currRank << " At snapshot:" << w.whichSnapshot
-                 << " buffer " << w.whichBuffer << "  seed: " << seed;
-            cout << "    box: " << chunk_offset[0] << ", " << chunk_offset[1]
-                 << ", " << chunk_offset[2] << std::endl;
-        }
 
         // prepare data block value
-        auto value = double(
-            seed + currRank + 0.1 * w.whichSnapshot + 100 * w.whichBuffer);
+        double value =
+            seed + w.whichSnapshot + (double)10 * (double)w.whichBuffer;
+
+        if (m_verbose)
+        {
+            cout << "  Rank: " << currRank << " At snapshot:" << w.whichSnapshot
+                 << " buffer " << w.whichBuffer << "  seed: " << seed;
+            cout << "    box: " << chunk_offset[0] << ", " << chunk_offset[1]
+                 << ", " << chunk_offset[2] << "  " << field_name << "/"
+                 << comp_name << std::endl;
+        }
 
         auto numElements = size_t(m_blockX) * m_blockY * m_blockZ;
         auto input = std::shared_ptr<double[]>(new double[numElements]);
@@ -264,7 +275,7 @@ void doConfig(int argc, char *argv[], int currRank)
             if (i + 1 < argc)
             {
                 int value = std::atoi(argv[++i]);
-                if (value > 0)
+                if (value >= 0)
                     m_diskFlushFrequency = value;
                 else if (0 == currRank)
                     std::cerr << "Error: -d value must be a positive integer. "
@@ -287,13 +298,15 @@ void doConfig(int argc, char *argv[], int currRank)
 
     if (0 == currRank)
     {
-        std::cout << " Configuration: \n\t[-v verbose] =" << m_verbose
-                  << "\n\t[-s span] =" << m_span
-                  << "\n\t[-b barrier_at_flush] =" << m_barrier_at_flush
-                  << " \n\t[-d diskFlushAfterNumbuffer] = "
+        std::cout << " Configuration: \n\t[-v verbose: default(no)] = "
+                  << m_verbose << "\n\t[-s use_span: default(no)] = " << m_span
+                  << "\n\t[-b mpi_barrier_at_flush: default(no)] = "
+                  << m_barrier_at_flush
+                  << " \n\t[-d diskFlushAfterNumbuffer: default(4) 0 means "
+                     "flush after every storechunk)] = "
                   << m_diskFlushFrequency
-                  << " \n\t[-f adiosFlattenSteps] = " << m_adiosFlattenSteps
-                  << std::endl;
+                  << " \n\t[-f adiosFlattenSteps: default=no] = "
+                  << m_adiosFlattenSteps << std::endl;
     }
 }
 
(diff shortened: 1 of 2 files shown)


franzpoeschel:
Yep, I see the hangup, too. Will have a look.

franzpoeschel:
Hey Junmin, the issue in here is that the Span-based storeChunk API is collective, which was not properly documented so far. Fixing that with #1794.

For your code, this means that all ranks must call storeChunk() when using the Span API. The following patch should do that:

```diff
diff --git a/examples/16_btd_write_parallel.cpp b/examples/16_btd_write_parallel.cpp
index a7b420ba2..65a602d94 100644
--- a/examples/16_btd_write_parallel.cpp
+++ b/examples/16_btd_write_parallel.cpp
@@ -204,6 +204,10 @@ void doWork(
                 input.get(), input.get() + numElements, spanBuffer.data());
         }
     }
+    if (m_span)
+    {
+        mymesh.storeChunk<double>({0, 0, 0}, {0, 0, 0}).currentBuffer();
+    }
 }
 
 int main(int argc, char *argv[])
```

guj:
> Hey Junmin, the issue in here is that the Span-based storeChunk API is collective, which was not properly documented so far. Fixing that with #1794.
> 
> For your code, this means that all ranks must call storeChunk() when using the Span API. The following patch should do that:
> 
> ```diff
> diff --git a/examples/16_btd_write_parallel.cpp b/examples/16_btd_write_parallel.cpp
> index a7b420ba2..65a602d94 100644
> --- a/examples/16_btd_write_parallel.cpp
> +++ b/examples/16_btd_write_parallel.cpp
> @@ -204,6 +204,10 @@ void doWork(
>                  input.get(), input.get() + numElements, spanBuffer.data());
>          }
>      }
> +    if (m_span)
> +    {
> +        mymesh.storeChunk<double>({0, 0, 0}, {0, 0, 0}).currentBuffer();
> +    }
>  }
>  
>  int main(int argc, char *argv[])
> ```

Oh, I did not know that. 
Then why is the barrier call cause the hanging?  while without the barrier call, store chunk with span succeeded?

franzpoeschel:
> > Hey Junmin, the issue in here is that the Span-based storeChunk API is collective, which was not properly documented so far. Fixing that with #1794.
> > For your code, this means that all ranks must call storeChunk() when using the Span API. The following patch should do that:
> > ```diff
> > diff --git a/examples/16_btd_write_parallel.cpp b/examples/16_btd_write_parallel.cpp
> > index a7b420ba2..65a602d94 100644
> > --- a/examples/16_btd_write_parallel.cpp
> > +++ b/examples/16_btd_write_parallel.cpp
> > @@ -204,6 +204,10 @@ void doWork(
> >                  input.get(), input.get() + numElements, spanBuffer.data());
> >          }
> >      }
> > +    if (m_span)
> > +    {
> > +        mymesh.storeChunk<double>({0, 0, 0}, {0, 0, 0}).currentBuffer();
> > +    }
> >  }
> >  
> >  int main(int argc, char *argv[])
> > ```
> 
> Oh, I did not know that. Then why is the barrier call cause the hanging? while without the barrier call, store chunk with span succeeded?

Okay, I take it back, the Span API can be used non-collectively, but there is currently a bug in `Iteration::open()`, fixed by #1794. 
So you can roll back the last commit and use the fix on #1794 instead.

guj:
> > > Hey Junmin, the issue in here is that the Span-based storeChunk API is collective, which was not properly documented so far. Fixing that with #1794.
> > > For your code, this means that all ranks must call storeChunk() when using the Span API. The following patch should do that:
> > > ```diff
> > > diff --git a/examples/16_btd_write_parallel.cpp b/examples/16_btd_write_parallel.cpp
> > > index a7b420ba2..65a602d94 100644
> > > --- a/examples/16_btd_write_parallel.cpp
> > > +++ b/examples/16_btd_write_parallel.cpp
> > > @@ -204,6 +204,10 @@ void doWork(
> > >                  input.get(), input.get() + numElements, spanBuffer.data());
> > >          }
> > >      }
> > > +    if (m_span)
> > > +    {
> > > +        mymesh.storeChunk<double>({0, 0, 0}, {0, 0, 0}).currentBuffer();
> > > +    }
> > >  }
> > >  
> > >  int main(int argc, char *argv[])
> > > ```
> > 
> > 
> > Oh, I did not know that. Then why is the barrier call cause the hanging? while without the barrier call, store chunk with span succeeded?
> 
> Okay, I take it back, the Span API can be used non-collectively, but there is currently a bug in `Iteration::open()`, fixed by #1794. So you can roll back the last commit and use the fix on #1794 instead.

I just checked,  and it works.  will you merge your fix?

franzpoeschel:
I have merged it now, you can rebase this branch

guj:
@ax3l @franzpoeschel  can we merge this one?

guj:
@franzpoeschel  I adjusted the examples, and looks like it is worth to tryout the following commands:   
mpirun -n 4 ./bin/16_btd_write_parallel     -s  -v -d 4  

It will finish successfully (h5 or bp) however,  if you try "h5ls -r" for each of the h5 files,  not all of them show datasets.  And those files are not readable by h5dump. 

mpirun -n 4 ./bin/16_btd_write_parallel     -s  -v -d 0   
This gives good h5 files. 
The difference is with -d 0, basically  a flush  is called after each store chunk.  Otherwise it will call flush after a few store chunk calls. 
I believe this has something to do with the umbrella span storechunk() support. I followed the pointers they are valid up to H5Dwrite. So not sure what the root cause is. 

To reduce run time, you can change line 65 to use  one field "B" instead of three fields "B" "j" "E".

examples/16_btd_write_parallel.cpp
```diff
@@ -0,0 +1,400 @@
+/* Copyright 2017-2021 Fabian Koller, Axel Huebl
```
ax3l:
```suggestion
/* Copyright 2017-2025 Junmin Gu, Axel Huebl, Franz Poeschel
```

examples/16_btd_write_parallel.cpp
```diff
@@ -0,0 +1,400 @@
...
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <openPMD/openPMD.hpp>
+
+#include <mpi.h>
+
+#include <iostream>
+#include <memory>
+#include <vector> // std::vector
```
ax3l:
```suggestion
#include <vector>
```

examples/16_btd_write_parallel.cpp
```diff
@@ -0,0 +1,400 @@
...
+ * You should have received a copy of the GNU General Public License
+ * and the GNU Lesser General Public License along with openPMD-api.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <openPMD/openPMD.hpp>
+
+#include <mpi.h>
+
+#include <iostream>
+#include <memory>
```
ax3l, line 27:
```suggestion
#include <map>
#include <memory>
```

examples/16_btd_write_parallel.cpp
```diff
@@ -0,0 +1,400 @@
...
+    {
+        if (0 == mpi_rank)
+            std::cout << "========== I/O with " << ext
+                      << " ========== " << std::endl;
+        try
+        {
+            std::string options = "";
+            if (m_adiosFlattenSteps)
+                options = R"(adios2.engine.parameters.FlattenSteps = "on")";
+            std::unique_ptr<Series> series = std::make_unique<openPMD::Series>(
```
ax3l, line 369: You could store the series just as a variable and pass it around by reference, but ok so as well if you try to emulate some pattern we use in codes?

examples/16_btd_write_parallel.cpp
```diff
@@ -0,0 +1,400 @@
...
+    double seed)
+{
+
+    bool const first_write_to_iteration =
+        !series->iterations.contains(w.whichSnapshot);
+
+    auto meshes = series->iterations[w.whichSnapshot].meshes;
+
+    // is this the trouble maker?
+    series->iterations[w.whichSnapshot].open();
```
ax3l, line 179: @franzpoeschel should we use
```suggestion
    series->snapshots()[w.whichSnapshot].open();
```
now?

ax3l:
@franzpoeschel can you please take a look at the [flush call issues](https://github.com/openPMD/openPMD-api/pull/1791#issuecomment-3439935240) described by @guj?

examples/16_btd_write_parallel.cpp
```diff
@@ -0,0 +1,401 @@
...
+        if (m_verbose)
+        {
+            cout << "  Rank: " << currRank << " At snapshot:" << w.whichSnapshot
+                 << " buffer " << w.whichBuffer << "  seed: " << seed;
+            cout << "    box: " << chunk_offset[0] << ", " << chunk_offset[1]
+                 << ", " << chunk_offset[2] << "  " << field_name << "/"
+                 << comp_name << std::endl;
+        }
+
+        auto numElements = size_t(m_blockX) * m_blockY * m_blockZ;
```
ax3l, line 210: @franzpoeschel should we add a helper function to calculate he number of elements from `chunk_offset` and `chunk_extent` (as used in `storeChunk` below)?

Something like
```C++
std::size_t numElements = openPMD::count_chunk_elements(chunk_offset, chunk_extent);
```

examples/16_btd_write_parallel.cpp
franzpoeschel: Ideally, yes

examples/16_btd_write_parallel.cpp
franzpoeschel: This probably depends on `chunk_extent` only? But yeah, I have manually written that often enough, a helper would be nice.

ax3l: Hm, not sure what I meant here, this is independent of the offset...

guj:
> This is a simple version of BTD. While writing this, it looks like something unexpected is observed.
> 
> to reproduce:
> 
> 1. compile and run 16_btd_write_parallel
> 2. use more than one rank (2 is ok)
> 3. use -s (span) with -b (mpi barrier before flush)
> 4. run "mpirun -n 2 16_btd_write_parallel -s " works, however "mpirun -n 2 16_btd_write_parallel -s -b " hangs.  It is not able to finish through the storeChunk call.
> 
> @franzpoeschel if you have time, please verify.

just verified that this bug is gone

guj:
> @franzpoeschel I adjusted the examples, and looks like it is worth to tryout the following commands: mpirun -n 4 ./bin/16_btd_write_parallel -s -v -d 4
> 
> It will finish successfully (h5 or bp) however, if you try "h5ls -r" for each of the h5 files, not all of them show datasets. And those files are not readable by h5dump.
> 
> mpirun -n 4 ./bin/16_btd_write_parallel -s -v -d 0 This gives good h5 files. The difference is with -d 0, basically a flush is called after each store chunk. Otherwise it will call flush after a few store chunk calls. I believe this has something to do with the umbrella span storechunk() support. I followed the pointers they are valid up to H5Dwrite. So not sure what the root cause is.
> 
> To reduce run time, you can change line 65 to use one field "B" instead of three fields "B" "j" "E".

Also I retried this one. 
---- file 1 & 3  has metadata  error:  (only has info of B fields, not E/j)
> h5dump -H   ../samples/16_btd_0000003.h5 
h5dump error: internal error (file h5dump.c:line 1440)
> h5ls -r  ../samples/16_btd_0000003.h5   
/                        Group
/data                    Group
/data/3                  Group
/data/3/fields           Group
/data/3/fields/B         Group
/data/3/fields/B/x       Dataset {192, 64, 64}
/data/3/fields/B/y       Dataset {192, 64, 64}
/data/3/fields/B/z       Dataset {192, 64, 64}

examples/16_btd_write_parallel.cpp
guj: As it gets created and destroyed in main(), this way is cleaner to me

examples/16_btd_write_parallel.cpp
guj: for writing, more straightforward to use iterations[]?

guj:
> @franzpoeschel can you please take a look at the [flush call issues](https://github.com/openPMD/openPMD-api/pull/1791#issuecomment-3439935240) described by @guj?

The branch is up to date, and you can reproduce h5 issue with: " mpirun -n 4  ./bin/16_btd_write_parallel -v -s" which invokes span, and files 0 2 are good but files 1 & 3 are not.  " mpirun -n 3  ./bin/16_btd_write_parallel -v -s"  will be ok, as well as " mpirun -n 4  ./bin/16_btd_write_parallel -v -s -t 3". In the later two cases, rank 0 processed the first buffer of all files.  Not sure why that makes difference in span call of h5 files. If needed, I am happy to enforce this test to use span only with bp.  what do you think?

Issue #1792, opened 2025-10-13
opmd matplotlib.pylab crash (depending on import order)


pgrete:
**Describe the bug**
Code crashes on import when openpmd is imported before matplotlib

**To Reproduce**
Compile-able/executable code example to reproduce the problem:

```bash
(py-oct25) pgrete@login10 ~ $ python -c "import openpmd_api as opmd; import matplotlib.pylab as plt; print('nope')"
Segmentation fault (core dumped)
(py-oct25) pgrete@login10 ~ $ python -c "import matplotlib.pylab as plt; import openpmd_api as opmd; print('yay')"
yay
(py-oct25) pgrete@login10 ~ $ python -c "import matplotlib as mpl; import openpmd_api as opmd; import matplotlib.pylab as plt; print('yay')"
yay
```

**Expected behavior**
No crash

**Software Environment**
 - version of openPMD-api: current dev
 - installed openPMD-api via: from source
 - machine: Frontier
 - name and version of Python implementation: CPython 3.11

franzpoeschel:
I remember having had a very similar issue a while ago with Lammps and openPMD-api (https://github.com/mala-project/mala/issues/476#issuecomment-2838818757). The same error pattern: Depending on import order, either a segfault or not.
In the related issue, the reason for the error was mainly a packaging error: Both packages were compiled against incompatible implementations of the C++ stdlib. Since the ABI of the stdlib is in parts identical, the loader jumped from lib A to lib B, or from B to A, depending which it saw first (hence the error only if you import things in a certain order).

The shortterm solution is probably to ensure a consistent build environment for both packages.
@ax3l We should maybe think of a way to not reexpose stdlib symbols?

pgrete:
I see. Thanks for the quick pointer.

This helped me to work around the problem. I didn't use the python binding from my openpmd install (from source), but now installed the openpmd version from pip.

But practically, how would I fix this if I depend on features only enable/disabled in source builds (assuming I don't want to build an entire python/matplotlib/... stack from source)?

franzpoeschel:
> But practically, how would I fix this if I depend on features only enable/disabled in source builds (assuming I don't want to build an entire python/matplotlib/... stack from source)?

Since it seems to be openPMD injecting itself into Matplotlib in this case, probably not very easily..
[...]

PR #1793, opened 2025-10-14
Symbol visibility: Hidden by default

franzpoeschel:
This revisits #165 as an attempt to fix software stack issues such as #1792 where binary-incompatible versions of e.g. the C++ stdlib are used by openPMD-api and other packages.

TODO:

- [ ] We now also have an Export.hpp header, merge this with Visibility.hpp (stolen from #165)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,6 +137,7 @@ openpmd_option(MPI            "Parallel, Multi-Node I/O for clusters"     AUTO)
 openpmd_option(HDF5           "HDF5 backend (.h5 files)"                  AUTO)
 openpmd_option(ADIOS2         "ADIOS2 backend (.bp files)"                AUTO)
 openpmd_option(PYTHON         "Enable Python bindings"                    AUTO)
+openpmd_option(SYMBOLHIDING   "Enable symbol hiding in libraries" AUTO)
 
 option(openPMD_INSTALL               "Add installation targets"             ON)
 option(openPMD_INSTALL_RPATH         "Add RPATHs to installed binaries"     ON)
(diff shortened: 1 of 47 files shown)


CMakeLists.txt
```diff
@@ -509,6 +538,18 @@ list(APPEND _msvc_options
...
 
+target_compile_definitions(openPMD PUBLIC
+    $<BUILD_INTERFACE:OPENPMD_BUILDPHASE=1>)
+
+# symbol hiding
+if(openPMD_HAVE_SYMBOLHIDING)
+    set_target_properties(openPMD PROPERTIES
+        CXX_VISIBILITY_PRESET hidden
+        VISIBILITY_INLINES_HIDDEN ON
+    )
```
ax3l, line 549: We did this for the old ADIOS1 libraries if you like to take a look:
https://github.com/openPMD/openPMD-api/blob/0.14.5/CMakeLists.txt#L555-L571

CMakeLists.txt
```diff
@@ -509,6 +538,18 @@ list(APPEND _msvc_options
 )
 target_compile_options(openPMD PUBLIC ${_msvc_options})
 
+target_compile_definitions(openPMD PUBLIC
+    $<BUILD_INTERFACE:OPENPMD_BUILDPHASE=1>)
+
+# symbol hiding
+if(openPMD_HAVE_SYMBOLHIDING)
+    set_target_properties(openPMD PROPERTIES
```
ax3l, line 546: This will not work for C++ users, unless you explicitly export all our public C++ APIs.

Already, we should not be exporting C++ stdlib symbols, I think. I think that mixing different C++ stdlibs generally does not work in the same process, just how the dynamic loader works and populates symbols. But this is not needed in most cases, e.g., on HPC one does instead compile with a consistent set of compilers/modules.

franzpoeschel: > Already, we should not be exporting C++ stdlib symbols, I think

Right, I see that this PR affects only symbols for our IO backends, which can be theoretically linked fully internally. I'll close.

PR #1794, opened 2025-10-14
Fix Iteration::open(), needed for correct use of Span API

franzpoeschel:
Seen on #1791, fixes:

* an optimization introduced by #1741 
* `Iteration::open()` in random-access mode (did not open files, leading to later hangups due to non-collective file opening)

### DIFF
--- a/docs/source/details/mpi.rst
+++ b/docs/source/details/mpi.rst
@@ -30,7 +30,6 @@ Functionality                Behavior           Description
 ``::resetDataset`` [1]_ [2]_ *backend-specific* declare, write
 ``::makeConstant`` [3]_      *backend-specific* declare, write
 ``::storeChunk`` [1]_        independent        write
-``::storeChunk`` [5]_        collective         Span-based overloads
 ``::loadChunk``              independent        read
 ``::availableChunks`` [4]_   collective         read, immediate result
 ============================ ================== ================================
@@ -49,8 +48,6 @@ Functionality                Behavior           Description
 
 .. [4] We usually open iterations delayed on first access. This first access is usually the ``flush()`` call after a ``storeChunk``/``loadChunk`` operation. If the first access is non-collective, an explicit, collective ``Iteration::open()`` can be used to have the files already open.
        Alternatively, iterations might be accessed for the first time by immediate operations such as ``::availableChunks()``.
-
-.. [5] The Span-based ``storeChunk`` API calls return a backend-allocated pointer, requiring flush operations. These API calls hence inherit the collective properties of flushing. Use store calls with zero-size extent if a rank does not contribute anything.
 
 .. warning::
 
--- a/docs/source/details/mpi.rst
+++ b/docs/source/details/mpi.rst
@@ -30,6 +30,7 @@ Functionality                Behavior           Description
 ``::resetDataset`` [1]_ [2]_ *backend-specific* declare, write
 ``::makeConstant`` [3]_      *backend-specific* declare, write
 ``::storeChunk`` [1]_        independent        write
+``::storeChunk`` [5]_        collective         Span-based overloads
 ``::loadChunk``              independent        read
 ``::availableChunks`` [4]_   collective         read, immediate result
 ============================ ================== ================================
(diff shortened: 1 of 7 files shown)


docs/source/details/mpi.rst
```diff
@@ -49,6 +50,8 @@ Functionality                Behavior           Description
 .. [4] We usually open iterations delayed on first access. This first access is usually the ``flush()`` call after a ``storeChunk``/``loadChunk`` operation. If the first access is non-collective, an explicit, collective ``Iteration::open()`` can be used to have the files already open.
        Alternatively, iterations might be accessed for the first time by immediate operations such as ``::availableChunks()``.
 
+.. [5] The Span-based ``storeChunk`` API calls return a backend-allocated pointer, requiring flush operations. These API calls hence inherit the collective properties of flushing. Use store calls with zero-size extent if a rank does not contribute anything.
```
ax3l: cc @guj 

Oh, but that has some severe implications for particle flushes and codes with dynamic load balancing, where one does have a non-equal number of calls to `storeChunk`.

* Fields: generally, non-equal number of calls per MPI rank https://github.com/BLAST-WarpX/warpx/pull/6123
* Particles: some ranks write zero particles, some call multiple times https://github.com/BLAST-WarpX/warpx/pull/3200

This makes usage very complicated for codes like WarpX.

@franzpoeschel Maybe we can relax the constraint t keep `storeChunk` indepdnent - instead put more logic in `resetDataset` or so?

franzpoeschel: Alternative:

* Introduce `RecordComponent::open()` to do all necessary collective preparations
* Check that `SkeletonOnly` flush does nothing collective

ax3l: Discussed:
- collective component only comes in for the first `storeChunk` (that means only particles would have a problem in most cases)
- there is some confusion in the table there between the MPI concepts: collective/independent (described) and synchronizing #1796 (not described)

ax3l: Franz comment: Collective _might_ only be the ADIOS engine open -- we probably do not need at all to delay this and could do this collectively and blocking in Series open...?

To double check if `Iteration::open()` really opens the engine ([x-ref](https://github.com/BLAST-WarpX/warpx/pull/6123/files#r2429977745))

guj: is collective a performance boost requirement:

ax3l: No, right now for span it is even a correctness requirement.

Franz will check if the collective aspect can be handled in `Iteration::open` instead. Actual I/O happens in flush, but `storeChunk` with Span needs some resources of the engine (and maybe variable) defined from ADIOS to work.

franzpoeschel: Okay, we can keep the current API:

* `Iteration::open()` must be used in parallel setups to ensure collective file access
* This was not working properly in random-access mode, now fixed
* The Span API can then be used non-collectively

franzpoeschel: Oh and:

* Due to the bug mentioned above, current releases require a workaround: Each rank must contribute at least one `storeChunk()` operation that may be empty.

guj: This fix works for me.

guj: > Oh and:
> 
> * Due to the bug mentioned above, current releases require a workaround: Each rank must contribute at least one `storeChunk()` operation that may be empty.

so span should require the next release version?

franzpoeschel: That's probably for the best, yeah. For older versions, you can alternatively include the workaround described above.

ax3l:
Thank you so much @franzpoeschel and @guj , this is great!

Let's do a new release soon? :)
Maybe after we have the issue with HDF5 span fallbacks figured out? https://github.com/BLAST-WarpX/warpx/pull/6123#discussion_r2337584301

franzpoeschel:
> Let's do a new release soon? :)

Sure, I think that there is not a lot to do any more. PRs that I would still like to get in are:
#1661
#1601

PR #1795, opened 2025-10-14
Upgrade Musllinux runner

franzpoeschel:
Recent h5py release requires HDF5 package newer than that shipped in our Musllinux runner


ax3l (approved):
Thank you!

PR #1806, opened 2025-11-14
Minor fixes, mostly obstructive warnings

franzpoeschel:
* openpmd-pipe currently prints a warning about rank tables when used on a file encoded Series
* opening a Series in `READ_LINEAR` and directly closing it again will print a warning because the deconstructor will try flushing, but no IO access has been made yet


ax3l (approved):
Thank you!

Wow, GCC14 has a lot of regressions o.0

PR #1808, opened 2025-11-21
CI: Upgrade to macos-14

franzpoeschel:
macos-13 EOL

- [x] macos-14
- [x] Using Python venv for now, maybe use brew install, did not work first try though and I did not bother figuring out why

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -44,8 +44,8 @@ jobs:
         cmake --build build --parallel 3
         ctest --test-dir build --verbose
 
-  appleclang14_py:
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yml
```diff
@@ -44,16 +44,18 @@ jobs:
         cmake --build build --parallel 3
         ctest --test-dir build --verbose
 
-  appleclang14_py:
-    runs-on: macos-13
+  appleclang18_py:
+    runs-on: macos-15
```
ax3l: Let's use 14 for now, we have latest covered above.

PR #1814, opened 2025-12-04
DataFrame: Add Attribute Columns

### LINKED ISSUE #1813: `.to_df()` with Particle Species Attributes
Looking at the code for ImpactX data plotting over `s`:
https://github.com/BLAST-ImpactX/impactx/issues/1216#issuecomment-3537614225

I realized that adding an attribute of a particle species as a column can be a quite common request, e.g., for each iteration/snapshot of an openPMD series, we store reference particle information of ImpactX directly on the particle species group and could use this as a dataframe column.

Expanding the API to allow something along these lines:
```py
series = io.Series("diags/openPMD/monitor.h5", io.Access.read_only)
df = series.to_df("beam", attributes=["s_ref"])

# add s column
iter_list = list(series.iterations)
```

and likewise on the `it.particles["beam"].to_df(attributes=["s_ref"])` level underneath could be very helpful to simplify the user experience.

ax3l:
Optionally add particle species attributes as extra columns. This is super useful when dealing with openPMD extensions or custom attributes, e.g., for accelerator physics. In the latter case, we store reference particle information on the particle species group (changes per iteration/snapshot).

Close #1813

### DIFF
--- a/examples/11_particle_dataframe.py
+++ b/examples/11_particle_dataframe.py
@@ -38,26 +38,27 @@ if __name__ == "__main__":
     s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
     electrons = s.snapshots()[400].particles["electrons"]
 
-    # all particles
-    df = electrons.to_df()
+    # all particles, extra column for "particleShape" attribute
+    #                (from ED-PIC extension)
+    df = electrons.to_df(attributes=["particleShape"])
     print(type(df) is pd.DataFrame)
     print(df)
 
     # only first 100 particles
-    df = electrons.to_df(np.s_[:100])
+    df = electrons.to_df(slice=np.s_[:100])
     print(df)
 
     # all particles over all steps
-    df = s.to_df("electrons")
+    df = s.to_df("electrons", attributes=["particleShape"])
     print(df)
 
     if found_cudf:
         # all particles - to GPU
-        cdf = cudf.from_pandas(electrons.to_df())
+        cdf = cudf.from_pandas(electrons.to_df(attributes=["particleShape"]))
         print(cdf)
 
         # all particles over all steps - to GPU
-        cdf = s.to_cudf("electrons")
+        cdf = s.to_cudf("electrons", attributes=["particleShape"])
         print(cdf)
 
     # Particles
@@ -67,7 +68,7 @@ if __name__ == "__main__":
         # pickle capabilities, so we test this here:
         dask.config.set(scheduler='processes')
 
-        df = electrons.to_dask()
+        df = electrons.to_dask(attributes=["particleShape"])
         print(df)
 
         # check chunking of a variable
--- a/docs/source/analysis/pandas.rst
+++ b/docs/source/analysis/pandas.rst
@@ -50,6 +50,17 @@ One can also combine all iterations in a single dataframe like this:
    # like before but with a new column "iteration" and all particles
    print(df)
 
+Additionally, one can add additional openPMD particle species attributes, e.g.,
+from the `ED-PIC <https://github.com/openPMD/openPMD-standard/blob/1.1.0/EXT_ED-PIC.md#particle-records-macroparticles>`__ extension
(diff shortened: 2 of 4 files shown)


examples/11_particle_dataframe.py
```diff
@@ -38,26 +38,27 @@
...
-    df = electrons.to_df()
+    # all particles, extra column for "particleShape" attribute
+    #                (from ED-PIC extension)
+    df = electrons.to_df(attributes=["particleShape"])
     print(type(df) is pd.DataFrame)
     print(df)
 
     # only first 100 particles
-    df = electrons.to_df(np.s_[:100])
+    df = electrons.to_df(slice=np.s_[:100])
```
franzpoeschel, line 48: Just to make sure, this change is not API breaking, you are just adding the `slice=` keyword argument here for clarity purposes?

examples/11_particle_dataframe.py
```diff
@@ -38,26 +38,27 @@
     s = io.Series("../samples/git-sample/data%T.h5", io.Access.read_only)
     electrons = s.snapshots()[400].particles["electrons"]
 
-    # all particles
-    df = electrons.to_df()
+    # all particles, extra column for "particleShape" attribute
+    #                (from ED-PIC extension)
+    df = electrons.to_df(attributes=["particleShape"])
```
franzpoeschel, line 43: With "attribute" in here, are you referring not to e.g. an ADIOS/HDF5 attribute, but to a record (component) name?
I have seen the term be used in the latter sense, so I am not fully sure here.

ax3l: What I refer to is an attribute stored on the particle species:
image

examples/11_particle_dataframe.py
ax3l: Unfortunately, it is API breaking for the slice argument.
We could add attributes after slice to make it compatible. I am not sure how widely this example here might be used.

examples/11_particle_dataframe.py
ax3l: Maybe I need a more detailed inline comment?

franzpoeschel: Maybe just use one of the base standard's attributes, so people might relate to the example more closely?

examples/11_particle_dataframe.py
franzpoeschel: I will leave the decision up to you, I just wanted to be sure we are aware of any breakages

examples/11_particle_dataframe.py
ax3l: That was also my initial inclination. But I think we do not have any [in the base standard](https://github.com/openPMD/openPMD-standard/blob/1.1.0/STANDARD.md#particle-records).

Thus I picked one [from the ED-PIC extension](https://github.com/openPMD/openPMD-standard/blob/1.1.0/EXT_ED-PIC.md#additional-attributes-for-the-group-of-each-particle-species)... Not super ideal, but best I came up with.

examples/11_particle_dataframe.py
ax3l: Yeah, good question. We do not use parameters to `to_df()` in ParaView or ImpactX or openPMD-viewer... probably because it is still relatively new. I could move attributes (a bit ugly, but ok!) behind slice to make it fully non-breaking. Let's decide in our weekly meeting.

ax3l: Looking through our [dependents on GitHub](https://github.com/openPMD/openPMD-api/network/dependents), the `to_df()` function is relatively popular. I cannot find in public usage someone using arguments with the call yet, but it is possible someone does processing of large particle files with it in their notebooks and slices with an unnamed `slice=` argument passed in...

ax3l: Added backwards compatibly logic now to `particles_to_dataframe`

examples/11_particle_dataframe.py
franzpoeschel: Ah right
I guess then the current example is probably fine

franzpoeschel (approved):

PR #1821, opened 2025-12-17
Doc: Examples per Version

### LINKED ISSUE #1805: Examples: Use Correct Download
We currently link the [examples](https://openpmd-api.readthedocs.io/en/0.16.1/usage/examples.html) to `dev`.

That is of course unstable and we should use `:download:` (or so) instead, to refer to the current version of the docs+source.

ax3l:
Link the examples in the manual corresponding to the current version of the manual. Before this, we always linked to `dev`, which can cause confusion when working with a stable release.

[Preview here.](https://openpmd-api--1821.org.readthedocs.build/en/1821/usage/examples.html)

Close #1805

### DIFF
--- a/docs/source/usage/11_particle_dataframe.py
+++ b/docs/source/usage/11_particle_dataframe.py
@@ -1,0 +1,1 @@
+../../../examples/11_particle_dataframe.py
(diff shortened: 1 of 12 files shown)


franzpoeschel (approved):
Thank you, LGTM!

PR #1834, opened 2026-01-14
Write scientific defaults upon close (Iteration::close or Series::close)

### LINKED ISSUE #1439: Scientific default values
The implementation of scientific default attribute values (i.e. those prescribed as mandatory by the openPMD-standard) is currently relatively ad-hoc, and while it works ok so far, it is neither really flexible nor extensible.

Current situation: Default attributes are set in the constructors of objects. They are set to values that often don't really make sense (e.g., for a 3D grid, setting the `axisLabels` to `{"x"}` by default makes no sense). Having these defaults set in the constructor makes the class inflexible to reuse and causes issues in the backends because of duplicate attribute definitions (first the default value, then the user-specified one).

Suggested redesign: 

1. Separate logic that specifies default values from object model. I imagine that we add a separate logic that collects the standard-defined attribute logic in one central place. This could be an abstract class, implemented by multiple derived classes for multiple versions of the standard. The central methods would be something like `virtual readDefaults(Attributable &) = 0;` and `virtual writeDefaults(Attributable &) = 0;`.
2. Instead of at construction time, defaults would be written at the time of `Series::close()` and `Iteration::close()` (not destructors, that causes too many issues down the road). This would postpone this logic to be as late as possible. The logic could take into consideration the dimensionality of meshes, that are then defined.
3. Reading defaults would happen at the same time as now (during parsing). It would deal leniently with missing / wrong values, e.g. by just setting the default, or ignoring it.
4. The implementation would no longer be based (exclusively) on the class type, but rather on the value of `Attributable::myPath()`. This increases reusability of classes.

To avoid merge conflicts, this should start **after** #1154 has been merged, or at least start based on that branch.

franzpoeschel:
Close #1439 
This PR focuses on Iteration-level attributes (including attributes included below Iteration level)
Needed for introducing a synchronous mode

TODO:

- [x] Write side changes
- [x] Read side changes: Move attribute initializations to new ScientificDefaults logic as well, be more lenient about missing attributes 
- [x] Maybe expose the `finalize()` method publically, so users can request openPMD to set defaults now
- [x] Reduce use of templates, size of ScientificDefaults.o
- [x] Check places from where readDefaults is called
- [x] particlePatches being weird, check them
- [x] clean up declarations and definitions
- [x] Decide if defaults should be populated upon `flush()` (benefit: early readers get readable datasets) or upon `close()` (benefit: we do populate too eagerly, users might add defaults late)
- [x] Do sth about `populateDefaultMetadata()`


franzpoeschel:
TODO: Move this check into finalize, maybe only print a warning

```cpp
            throw std::runtime_error(
                "A Record can not be written without any contained "
                "RecordComponents: " +
                name);
```

ax3l:
@franzpoeschel is this for 0.18.0? :)

franzpoeschel:
> @franzpoeschel is this for 0.18.0? :)

Yup, this is still under development

ax3l:
@franzpoeschel sorry small conflict in `CoreTest.cpp`, will review the rest in the meantime :)

src/RecordComponent.cpp
```diff
@@ -720,6 +694,16 @@ void RecordComponent::verifyChunk(
...
 
+void RecordComponent::scientificDefaults_impl(bool write, OpenpmdStandard)
+{
+    using namespace internal;
+    auto float_types = get_float_types();
+    auto const wor = write ? WriteOrRead::Write : WriteOrRead::Read;
+
+    defaultAttribute(*this, "unitSI")
+        .template withSetter<RecordComponent>(1.0, &RecordComponent::setUnitSI)
+        .withReader(float_types, require_type<double>())(wor);
```
franzpoeschel, line 709: TODO: verify that this properly implements the previous logic for `require_unit_SI`

src/Iteration.cpp
```diff
@@ -509,80 +534,7 @@ void Iteration::read_impl(std::string const &groupPath)
...
-        throw error::ReadError(
-            error::AffectedObject::Attribute,
-            error::Reason::UnexpectedContent,
-            {},
-            "Unexpected Attribute datatype for 'timeUnitSI' (expected double, "
-            "found " +
-                datatypeToString(
-                    Attribute(Attribute::from_any, *aRead.m_resource).dtype) +
-                ")");
+    internal::ScientificDefaults::readDefaults(IOHandler()->m_standard);
```
ax3l: Codex:
>   1. src/Iteration.cpp:537, src/Mesh.cpp:497/564, src/Record.cpp:200
     Read-side defaults are populated before readAttributes(ReadMode::FullyReread), but FullyReread clears m_attributes at src/backend/Attributable.cpp:415. Defaults synthesized for missing time, dt, mesh  metadata, record unitDimension, etc. are immediately discarded. Move the final readDefaults() after   readAttributes(FullyReread), or avoid clearing synthesized defaults.

src/Mesh.cpp
```diff
@@ -439,170 +494,7 @@ void Mesh::read()
...
-                error::Reason::UnexpectedContent,
-                {},
-                "Unexpected Attribute datatype for 'gridUnitSI' "
-                "(expected double, found " +
-                    datatypeToString(
-                        Attribute(Attribute::from_any, *aRead.m_resource)
-                            .dtype) +
-                    ")");
-    }
+    internal::ScientificDefaults::readDefaults(IOHandler()->m_standard);
```
ax3l: Codex:
>   1. src/Iteration.cpp:537, src/Mesh.cpp:497/564, src/Record.cpp:200
     Read-side defaults are populated before readAttributes(ReadMode::FullyReread), but FullyReread clears  m_attributes at src/backend/Attributable.cpp:415. Defaults synthesized for missing time, dt, mesh  metadata, record unitDimension, etc. are immediately discarded. Move the final readDefaults() after  readAttributes(FullyReread), or avoid clearing synthesized defaults.

src/Mesh.cpp
```diff
@@ -669,7 +561,7 @@ void Mesh::read()
 
     std::move(homogenizeExtents).homogenize(*this);
 
-    readBase();
+    internal::ScientificDefaults::readDefaults(IOHandler()->m_standard);
```
ax3l: Codes: see line 497 comment

src/Record.cpp
```diff
@@ -192,11 +197,38 @@ auto Record::read() -> internal::HomogenizeExtents
         }
     }
 
-    readBase();
+    internal::ScientificDefaults::readDefaults(IOHandler()->m_standard);
```
ax3l: Codex:
>   1. src/Iteration.cpp:537, src/Mesh.cpp:497/564, src/Record.cpp:200
     Read-side defaults are populated before readAttributes(ReadMode::FullyReread), but FullyReread clears  m_attributes at src/backend/Attributable.cpp:415. Defaults synthesized for missing time, dt, mesh  metadata, record unitDimension, etc. are immediately discarded. Move the final readDefaults() after  readAttributes(FullyReread), or avoid clearing synthesized defaults.

src/Iteration.cpp
```diff
@@ -236,6 +254,13 @@ bool Iteration::closedByWriter() const
     }
 }
 
+void Iteration::visitHierarchy(HierarchyVisitor &v, bool recursive)
```
ax3l, line 257: Codex:
>   2. src/Iteration.cpp:257, src/Series.cpp:3567, src/ParticleSpecies.cpp:30
     The new recursive flag is not consistently honored. Iteration::visitHierarchy(false) still visits meshes
     and particles, Series::visitHierarchy(false) still visits iterations, and ParticleSpecies::visitHierarch
     y(false) still visits particlePatches. This contradicts the public API docs in Attributable.hpp:417/452
     and makes populateMissingMetadata(false) / custom visitors surprising. Guard child traversal with if (recursive).

src/Series.cpp
```diff
@@ -3556,6 +3564,12 @@ void Series::close()
     m_attri.reset();
 }
 
+void Series::visitHierarchy(HierarchyVisitor &v, bool recursive)
```
ax3l, line 3567: Codex:
>   2. src/Iteration.cpp:257, src/Series.cpp:3567, src/ParticleSpecies.cpp:30
     The new recursive flag is not consistently honored. Iteration::visitHierarchy(false) still visits meshes
     and particles, Series::visitHierarchy(false) still visits iterations, and ParticleSpecies::visitHierarch
     y(false) still visits particlePatches. This contradicts the public API docs in Attributable.hpp:417/452
     and makes populateMissingMetadata(false) / custom visitors surprising. Guard child traversal with if (recursive).

src/ParticleSpecies.cpp
```diff
@@ -20,15 +20,19 @@
...
-#include "openPMD/Series.hpp"
-#include "openPMD/auxiliary/DerefDynamicCast.hpp"
 #include "openPMD/backend/Writable.hpp"
 
 #include <algorithm>
 #include <iostream>
 
 namespace openPMD
 {
+void ParticleSpecies::visitHierarchy(HierarchyVisitor &v, bool recursive)
```
ax3l, line 30: Codex:
>   2. src/Iteration.cpp:257, src/Series.cpp:3567, src/ParticleSpecies.cpp:30
     The new recursive flag is not consistently honored. Iteration::visitHierarchy(false) still visits meshes
     and particles, Series::visitHierarchy(false) still visits iterations, and ParticleSpecies::visitHierarch
     y(false) still visits particlePatches. This contradicts the public API docs in Attributable.hpp:417/452
     and makes populateMissingMetadata(false) / custom visitors surprising. Guard child traversal with if (recursive).

franzpoeschel: fixed

src/Series.cpp
franzpoeschel: fixed

src/Iteration.cpp
franzpoeschel: fixed

src/Record.cpp
franzpoeschel: Issue actually existed before, but was valid. Fixed

src/Mesh.cpp
franzpoeschel: fixed

src/Mesh.cpp
franzpoeschel: fixed

src/Iteration.cpp
franzpoeschel: fixed

src/RecordComponent.cpp
franzpoeschel: Answer: Yes, it does:
```cpp
void PatchRecordComponent::scientificDefaults_impl(
    internal::WriteOrRead wor, OpenpmdStandard)
{
    using namespace internal;

    // We don't require unitSI for PatchRecordComponent
    //
    // addParentDefaults<RecordComponent, write>();
    //
    // But we still set it when writing. Doesnt hurt and some readers might
    // expect it.
    defaultAttribute(*this, "unitSI")
        .template withSetter<PatchRecordComponent>(
            1.0, &PatchRecordComponent::setUnitSI)
        // Do NOT add a reader here.
        // .withReader(numerical_types, require_type<double>())
        (wor);
}
```

ax3l (approved):

PR #1832, opened 2026-01-13
Fix AppVeyor 64bit build

ax3l:
The CMake Generator line for 64bit build was outdated and does not include the architecture anymore.

---

Use force to address for the `legacy_datasets.tar`
```
Move-Item : Cannot create a file when that file already exists.
At C:\projects\openpmd-api\share\openPMD\download_samples.ps1:30 char:1
+ Move-Item -Path legacy_datasets\* samples\git-sample\legacy\
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : WriteError: (C:\projects\ope...5.0-hdf5-plugin:DirectoryInfo) [Move-Item], IOException
    + FullyQualifiedErrorId : MoveDirectoryItemIOError,Microsoft.PowerShell.Commands.MoveItemCommand
```

PR #1845, opened 2026-01-23
Use force-overwrite in Windows download_samples script

franzpoeschel:
Separating this out of #1832 so we finally get a working CI again

Use force to address for tge `legacy_datasets.tar`
```
Move-Item : Cannot create a file when that file already exists.
At C:\projects\openpmd-api\share\openPMD\download_samples.ps1:30 char:1
+ Move-Item -Path legacy_datasets\* samples\git-sample\legacy\
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : WriteError: (C:\projects\ope...5.0-hdf5-plugin:DirectoryInfo) [Move-Item], IOException
    + FullyQualifiedErrorId : MoveDirectoryItemIOError,Microsoft.PowerShell.Commands.MoveItemCommand
```

### DIFF
--- a/share/openPMD/download_samples.ps1
+++ b/share/openPMD/download_samples.ps1
@@ -24,10 +24,10 @@ Invoke-WebRequest https://github.com/openPMD/openPMD-example-datasets/raw/566b35
 7z.exe x -r example-3d-bp4.tar
 7z.exe x -r legacy_datasets.tar.gz
 7z.exe x -r legacy_datasets.tar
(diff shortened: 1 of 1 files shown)


franzpoeschel:
Ah, the upgrade is necessary for fixing? Will close then.

PR #1851, opened 2026-02-19
Python: Fix keep_alive specifications, add tests for keep_alive

franzpoeschel:
Follow-up to #1849. keep_alive is now finally in a working state.

TODO:

- [x] formatting

### DIFF
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -2026,6 +2026,9 @@ class APITest(unittest.TestCase):
         # Cleaner: write.close()
         # But let's keep this instance to test that that workflow stays
         # functional.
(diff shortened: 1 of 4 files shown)


test/python/unittest/API/APITest.py
```diff
@@ -2026,6 +2026,9 @@ def makeAvailableChunksRoundTrip(self, ext):
         # Cleaner: write.close()
         # But let's keep this instance to test that that workflow stays
         # functional.
+        # Need to delete everything as garbage collection will keep `write`
+        # alive as long as E_x is around.
+        del E_x
         del write
```
franzpoeschel, line 2032: this is slightly API-breaking
but realistically, the previous behavior was unspecified behavior (relying on deleting a Series, while its dependent object are still hanging around)

PR #1852, opened 2026-02-19
Add Ruff formatter to pre-commit for Python

franzpoeschel:
TODO:

- [x] Remove old checks that conflict with ruff
- [x] Either enforce line length 79 in ruff, or be fine with its default of 88
- [x] Revert my changes on APITest.py, let the bot do the commit


.pre-commit-config.yaml
```diff
@@ -85,13 +85,20 @@ repos:
...
+#     name: isort (python)
+
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.15.1
   hooks:
-  - id: isort
-    name: isort (python)
+  - id: ruff-check
+    args: [ --line-length, "79", --fix ]
+  - id: ruff-format
```
ax3l, line 101: We could add a few more flags for the linter:
https://github.com/AMReX-Codes/pyamrex/blob/789dc8c4f1bca202219072e6a9801f060fc4ba3a/.pre-commit-config.yaml#L74-L83

Then move options like `--line-length 79` to a toml config file
https://github.com/AMReX-Codes/pyamrex/blob/development/pyproject.toml

That way, local runs of ruff linter (outside of pre-commit) will have the same configs.

.pre-commit-config.yaml
```diff
@@ -85,13 +85,20 @@ repos:
...
-# Sorts Python imports according to PEP8
-# https://www.python.org/dev/peps/pep-0008/#imports
-- repo: https://github.com/pycqa/isort
-  rev: 7.0.0
+# # Sorts Python imports according to PEP8
+# # https://www.python.org/dev/peps/pep-0008/#imports
+# - repo: https://github.com/pycqa/isort
+#   rev: 7.0.0
+#   hooks:
+#   - id: isort
```
ax3l, line 93: Yes this one we remove, replaced by ruff

.pre-commit-config.yaml
franzpoeschel: Ruff apparently also looks for configuration directly inside `pyproject.yaml`, so that would keep things simple. I moved the line length flags there for now.
EDIT: Ah, I just saw that this is the style you linked to. Then let's go with that, yeah. Should we just copy the config from there or should we adapt some things?

ax3l: I think the logic there is good for us to copy for now, unless we see an issue with it.

.pre-commit-config.yaml
ax3l: Commented logic below:

Flake8: This can be removed as well
Black below: also remove.

.pre-commit-config.yaml
```diff
@@ -85,13 +85,20 @@ repos:
   - id: pycln
     name: pycln (python)
```
ax3l, line 86: This should probably be included in `ruff` already, so we should be able to remove it, too. To check.

.pre-commit-config.yaml
ax3l, line 66:
```suggestion
# clang-format
```

franzpoeschel:
@ax3l 0.17.1 is out, so this is ready for review / merge

.github/workflows/source.yml
```diff
@@ -20,7 +20,9 @@ jobs:
     - name: PEP8
       run: |
         python3 -m pip install -U flake8
-        python3 -m flake8 --exclude=thirdParty .
+        # E203: checks that ':' is not preceded by a space
+        # But the Ruff formatter has better logic for that
+        python3 -m flake8 --extend-ignore=E203 --exclude=thirdParty .
```
ax3l: Let's remove the flake8 completely

.pre-commit-config.yaml
ax3l: pls remove black, too

pyproject.toml
```diff
@@ -7,3 +7,6 @@ requires = [
     "pybind11>=2.13.0,<3.0.0"
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.ruff]
+line-length = 79
```
ax3l:
```toml
[tool.ruff.format]
docstring-code-format = true

[tool.ruff.lint.isort]
known-first-party = ["openpmd_api", "openpmd_api_cxx"]
```

ax3l (approved):
LGTM 🎉

PR #1854, opened 2026-02-26
Harmonize Datatype equality checks

franzpoeschel:
## currently
We override `operator==(Datatype, Datatype)` such that it checks Datatype equality by representation on the platform. However, in some places, our `loadChunk` / `storeChunk` API still manually implements this check, with slight implementation differences. 

## with this pr
This PR merges those differences into the `operator==()` implementation and uses that consistently.

Also:

- Adds a missing type check for store_chunk / load_chunk in our Python API
- Adds non-template type comparison operators

- [x] Maybe add a further overload to `Dataset` constructor, to avoid this weirdness:
    ```python
        e_chargeDensity_x.reset_dataset(DS(np.dtype(np.int_), [10]))
    ```
    EDIT: done separately on #1856

### DIFF
--- a/include/openPMD/backend/PatchRecordComponent.hpp
+++ b/include/openPMD/backend/PatchRecordComponent.hpp
@@ -161,7 +161,7 @@ template <typename T>
 inline void PatchRecordComponent::store(uint64_t idx, T data)
 {
     Datatype dtype = determineDatatype<T>();
-    if (dtype != getDatatype())
+    if (!isSame(dtype, getDatatype()))
     {
         std::ostringstream oss;
         oss << "Datatypes of patch data (" << dtype << ") and dataset ("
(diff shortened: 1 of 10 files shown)


include/openPMD/backend/PatchRecordComponent.hpp
```diff
@@ -160,10 +161,7 @@ template <typename T>
 inline void PatchRecordComponent::store(uint64_t idx, T data)
 {
     Datatype dtype = determineDatatype<T>();
-    if (dtype != getDatatype() && !isSameInteger<T>(getDatatype()) &&
-        !isSameFloatingPoint<T>(getDatatype()) &&
-        !isSameComplexFloatingPoint<T>(getDatatype()) &&
-        !isSameChar<T>(getDatatype()))
+    if (dtype != getDatatype())
```
franzpoeschel:
```suggestion
    if (!isSame(dtype, getDatatype()))
```

PR #1855, opened 2026-03-02
create_directories: preserve sticky and setgid permissions

franzpoeschel:
Ref. https://github.com/ComputationalRadiationPhysics/picongpu/issues/5638

### DIFF
--- a/src/auxiliary/Filesystem.cpp
+++ b/src/auxiliary/Filesystem.cpp
@@ -96,6 +96,44 @@ std::vector<std::string> list_directory(std::string const &path)
     return ret;
 }
 
+std::string get_parent(std::string const &path)
+{
+    std::string parent = path;
+    size_t pos = parent.find_last_of(directory_separator);
+    if (pos != std::string::npos)
+    {
+        parent = parent.substr(0, pos);
+        if (parent.empty())
+            parent = "/";
+    }
+    else
+    {
+        parent.clear();
+    }
+    return parent;
+}
+
+mode_t get_permissions(std::string const &path)
+{
+    std::string parent = get_parent(path);
+    if (parent.empty())
+    {
+        return 0;
+    }
+    if (!directory_exists(parent))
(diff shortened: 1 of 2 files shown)


src/auxiliary/Filesystem.cpp
```diff
@@ -106,10 +144,11 @@ bool create_directories(std::string const &path)
         return CreateDirectory(p.c_str(), nullptr);
     };
 #else
-    mode_t mask = umask(0);
-    umask(mask);
```
franzpoeschel, line 110: Why do we retrieve the umask just to apply it manually without any modifications? That's what the system does implicitly anyway?

franzpoeschel: Is that supposed to make the function thread-safe against modification of the umask on other threads? If yes, this does not work due to the double call to `umask()`, also it is irrelevant for MPI-parallel deletion.

ax3l: Agreed. Let's make sure that if someone creates dirs and files with openPMD, it behaves the same as a terminal `mkdir` and `touch` with respect to `umask` and other inherited properties.

It's possible the current implementation is overly verbose and forgets setgid/sticky bits

franzpoeschel: Discussed yesterday: Our current implementation probably assumes that the umask is overwritten by the `mode_t mode` parameter of `mkdir()`, but that is not the case:

```
       The  argument mode specifies the mode for the new directory (see inode(7)).  It is modified by the process's umask in the usual way: in the absence
       of a default ACL, the mode of the created directory is (mode & ~umask & 0777).  Whether other mode bits are honored for the created  directory  de-
       pends on the operating system.  For Linux, see VERSIONS below.
```

Hence, our manual umask handling (1) is unnecessary and (2) is not thread safe due to temporarily manipulating the umask.

PR #1856, opened 2026-03-02
Numpy Datatype: Enable np.dtype() type conversions

franzpoeschel:
This makes it possible to use things like `np.int_` or `"double"` in the `Dataset` constructor, which currently has to be written as `np.dtype(np.int_)` and `np.dtype("double")`. Check also examples and tests.
Affects `Datatype()` constructor as well as `RecordComponent.make_empty()`.

Future: Derive Dataset definition from Numpy array directly

### DIFF
--- a/examples/10_streaming_write.py
+++ b/examples/10_streaming_write.py
@@ -40,7 +40,7 @@ if __name__ == "__main__":
         electronPositions.set_attribute("comment", "I'm a comment")
 
         length = 10
-        local_data = np.arange(i * length, (i + 1) * length, dtype=np.dtype("double"))
+        local_data = np.arange(i * length, (i + 1) * length, dtype="double")
         for dim in ["x", "y", "z"]:
             pos = electronPositions[dim]
             pos.reset_dataset(io.Dataset(local_data.dtype, [length]))
@@ -60,7 +60,7 @@ if __name__ == "__main__":
         # temperature has no x,y,z components, so skip the last layer:
         temperature_dataset = temperature
         # let's say we are in a 3x3 mesh
-        temperature_dataset.reset_dataset(io.Dataset(np.dtype("double"), [3, 3]))
+        temperature_dataset.reset_dataset(io.Dataset("double", [3, 3]))
         # temperature is constant
         temperature_dataset.make_constant(273.15)
 
--- a/examples/10_streaming_write.py
+++ b/examples/10_streaming_write.py
@@ -60,7 +60,7 @@ if __name__ == "__main__":
         # temperature has no x,y,z components, so skip the last layer:
         temperature_dataset = temperature
         # let's say we are in a 3x3 mesh
-        temperature_dataset.reset_dataset(io.Dataset("double", [3, 3]))
+        temperature_dataset.reset_dataset(io.Dataset(local_data.dtype, [3, 3]))
         # temperature is constant
         temperature_dataset.make_constant(273.15)
 
--- a/include/openPMD/binding/python/Numpy.hpp
+++ b/include/openPMD/binding/python/Numpy.hpp
@@ -98,6 +98,20 @@ inline Datatype dtype_from_numpy(pybind11::dtype const dt)
     }
 }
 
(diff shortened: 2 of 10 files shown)


examples/10_streaming_write.py
```diff
@@ -63,7 +63,7 @@
         temperature_dataset = temperature
         # let's say we are in a 3x3 mesh
         temperature_dataset.reset_dataset(
-            io.Dataset(np.dtype("double"), [3, 3]))
+            io.Dataset("double", [3, 3]))
```
ax3l: Can we even support
```suggestion
            io.Dataset(local_data.dtype, [3, 3]))
```
or even (but slightly more confusing):
```suggestion
            io.Dataset(local_data, [3, 3]))
```
?

ax3l: For non-MPI runs, we could even support

```suggestion
            io.Dataset(full_data)
```

include/openPMD/binding/python/Numpy.hpp
```diff
@@ -98,6 +98,64 @@ inline Datatype dtype_from_numpy(pybind11::dtype const dt)
...
+                    "dtype_from_numpy: Cannot convert from object type to "
+                    "datatype without numpy, failed importing: ") +
+                e.what());
+        }
+        pybind11::object create_dtype;
+        try
+        {
+            create_dtype = numpy.attr("dtype");
+        }
+        catch (std::exception const &e)
```
ax3l: Maybe check if it has the attribute and then throw? Avoids one catch

include/openPMD/binding/python/Numpy.hpp
```diff
@@ -98,6 +98,64 @@ inline Datatype dtype_from_numpy(pybind11::dtype const dt)
     }
 }
 
+inline Datatype dtype_from_numpy(pybind11::object const &dt)
```
franzpoeschel: check if bare object is fine instead of const&

src/binding/python/Dataset.cpp
```diff
@@ -37,10 +37,11 @@ void init_Dataset(py::module &m)
                 py::arg("extent"),
                 py::arg("options") = "{}")
             .def(
-                py::init([](py::dtype dt, Extent e, std::string options) {
-                    auto const d = dtype_from_numpy(std::move(dt));
-                    return new Dataset{d, std::move(e), std::move(options)};
-                }),
+                py::init(
+                    [](py::object const &dt, Extent e, std::string options) {
```
franzpoeschel: check if bare object instead of const&

include/openPMD/binding/python/Numpy.hpp
franzpoeschel: I have removed those try-catch blocks which should only throw if the numpy interfaces change. This leaves those two blocks that check if numpy is installed and for any errors when calling `numpy.dtype()` with user-supplied objects.

ax3l (approved):
LGTM - looks bw compatible. Please fix the one CodeQL error it brought up and then ready :)

include/openPMD/binding/python/Numpy.hpp
```diff
@@ -239,4 +277,5 @@ inline pybind11::dtype dtype_to_numpy(Datatype const dt)
         throw std::runtime_error("dtype_to_numpy: Invalid Datatype!");
     }
 }
+
```
franzpoeschel:
```suggestion
```

src/binding/python/Dataset.cpp
```diff
@@ -25,6 +25,7 @@
 #include "openPMD/binding/python/auxiliary.hpp"
 
 #include <string>
+#include <utility>
```
franzpoeschel:
```suggestion
```

src/binding/python/RecordComponent.cpp
```diff
@@ -51,6 +51,7 @@
 #include <string>
 #include <tuple>
 #include <type_traits>
+#include <utility>
```
franzpoeschel:
```suggestion
```

Issue #1859, opened 2026-03-04
Overhead with variable-based `bp5` series


aeriforme:
**Describe the bug**
Performance kill when opening/closing a variable-based `bp5` series with many iterations.

**To Reproduce**
Standalone reproducer in Python:

```python
import openpmd_api as io
from tqdm import tqdm
adios2_cfg = r'''
{
  "iteration_encoding": "variable_based",
  "adios2": {
    "engine": {
      "type": "bp5",
      "parameters": {
        "StatsLevel": "0"
      }
    }
  }
}
'''
OUTPUT_PATH = "myseries.bp5"
series = io.Series(OUTPUT_PATH, io.Access_Type.create_linear, adios2_cfg)
N = 1000000
for i in tqdm(range(N)):
    it = series.snapshots()[i]
    it.close()
```

**Expected behavior**
I expect much less overhead just to open/close the iterations, and not to change in time.

**Software Environment**
 - version of openPMD-api: 0.17.0
 - installed openPMD-api via: conda-forge
 - operating system: 
 - machine: Perlmutter at NERSC - login nodes 
 - name and version of Python implementation: CPython 3.11.14
 - version of ADIOS2: 2.11.0
 - name and version of MPI: N/A

**Additional context**
The overhead builds up as the loop moves through the iterations. 
Starts around ~1k iterations/s and goes down to ~200 iterations/s after ~100k iterations.

ax3l:
Thank you for posting the reproducer, @aeriforme!

ccing @guj and @franzpoeschel : let us benchmark this deeper. I am surprised this is so slow:
- 1ms per open when we start seems excessive
- linear slowdown over time as we add more snapshots is probably a data container issue that we can solve, too

What do you think?

franzpoeschel:
I am off-work this week, but will have a look next week. Achieving linear performance with the number of Iterations should be a goal for this release cycle, but we are not there yet.
I think, with the current release, we have this goal down inside the ADIOS2 backend. For our frontend, we will have to move to a workflow where old Iterations are fully erased from our datastructures upon opening a new one. This has not been done yet, since the details of handle invalidation can get tricky.

One suspicion: `indexOf()` inside the Iteration flushing logic is linear in the number of Iterations. We should be able to optimize this; after 100K Iterations, the cost for this operation may become dominant.

guj:
At the mean time, can you see how long it takes to load the file with ADIOS python library? 
e.g.  
[...]

PR #1860, opened 2026-03-05
attempt to cache the index and avoid some indexOf() call 

guj:
Hope this resolves issue #1859

### DIFF
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -136,7 +136,7 @@ namespace internal
          * of the owning map entry instead of a linear scan in
          * <code>Series::indexOf()</code>.
          */
-        uint64_t m_iterationIndex = 0;
+        std::optional<uint64_t> m_iterationIndex = 0;
 
         /**
          * Information on a parsing request that has not yet been executed.
@@ -265,7 +265,7 @@ public:
      */
     uint64_t getCachedIterationIndex() const
     {
-        return get().m_iterationIndex;
+        return *get().m_iterationIndex;
     }
     /**
      * @brief Has the iteration been parsed yet?
--- a/include/openPMD/Iteration.hpp
+++ b/include/openPMD/Iteration.hpp
@@ -130,6 +130,15 @@ namespace internal
         StepStatus m_stepStatus = StepStatus::NoStep;
 
         /**
+         * Cached copy of the key under which this Iteration lives in
+         * <code>Series::iterations</code>.  Populated when the iteration
+         * object is created/inserted.  This allows constant-time lookup
+         * of the owning map entry instead of a linear scan in
+         * <code>Series::indexOf()</code>.
+         */
+        uint64_t m_iterationIndex = 0;
+
+        /**
          * Information on a parsing request that has not yet been executed.
          * Otherwise empty.
          */
@@ -246,6 +255,18 @@ public:
      */
     bool closed() const;
 
+    /**
+     * @brief Get the cached iteration index.
+     *        This is the key under which this iteration is stored in the
+     *        Series::iterations map. Used for testing the index caching
+     *        optimization.
+     *
+     * @return The cached iteration index.
+     */
+    uint64_t getCachedIterationIndex() const
+    {
+        return get().m_iterationIndex;
+    }
     /**
      * @brief Has the iteration been parsed yet?
               If not, it will contain no structure yet.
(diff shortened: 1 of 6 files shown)


include/openPMD/backend/ContainerImpl.tpp
```diff
@@ -201,13 +211,23 @@ template <typename T, typename T_key, typename T_container>
...
 auto Container<T, T_key, T_container>::insert(value_type &&value)
     -> std::pair<iterator, bool>
 {
-    return container().insert(value);
+    auto res = container().insert(value);
+    if constexpr (std::is_same_v<T, Iteration>)
+    {
+        res.first->second.get().m_iterationIndex = res.first->first;
+    }
+    return res;
```
franzpoeschel: I think that this logic might be better placed inside `linkHierarchy()`

include/openPMD/Iteration.hpp
```diff
@@ -129,6 +129,15 @@ namespace internal
...
         StepStatus m_stepStatus = StepStatus::NoStep;
 
+        /**
+         * Cached copy of the key under which this Iteration lives in
+         * <code>Series::iterations</code>.  Populated when the iteration
+         * object is created/inserted.  This allows constant-time lookup
+         * of the owning map entry instead of a linear scan in
+         * <code>Series::indexOf()</code>.
+         */
+        uint64_t m_iterationIndex = 0;
```
franzpoeschel: Can you use a `std::optional<uint64_t>` here, please? Otherwise, we will at some point get bugs that the flushing logic silently goes into the wrong Iterations

src/Series.cpp
```diff
@@ -2604,11 +2604,27 @@ std::string Series::iterationFilename(IterationIndex_t i)
...
+    {
+        auto it = series.iterations.find(idx);
+        if (it != series.iterations.end() &&
+            &it->second.Attributable::get() == &iteration.Attributable::get())
+        {
+            return it;
+        }
+        // if the cached index is stale (shouldn't happen), fall back
+    }
+    // fallback to linear scan for safety
```
franzpoeschel: This fallback becomes relevant only if we have a bug in our flushing logic. Throwing an exception here should be fine, otherwise the performance of this becomes too hard to reason about (fast as long as things are written correctly, slow for the remaining cases)

include/openPMD/backend/ContainerImpl.tpp
```diff
@@ -19,8 +19,8 @@
  * If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "openPMD/Iteration.hpp" // needed for index caching in operator[]
```
franzpoeschel: Pre-declaring should be fine, but note my comment below, this entire logic is better placed in `linkHierarchy()`

franzpoeschel (changes requested):
Hey Junmin, thank you for the PR. Some comments below.

src/Series.cpp
franzpoeschel: I've pushed a commit that implements this. The fallback was not used anywhere in the tests.

include/openPMD/Iteration.hpp
franzpoeschel: I've pushed a commit that does this now

include/openPMD/Iteration.hpp
```diff
@@ -246,6 +255,18 @@ class Iteration : public Attributable
...
 
+    /**
+     * @brief Get the cached iteration index.
+     *        This is the key under which this iteration is stored in the
+     *        Series::iterations map. Used for testing the index caching
+     *        optimization.
+     *
+     * @return The cached iteration index.
+     */
+    uint64_t getCachedIterationIndex() const
```
franzpoeschel: This is currently part of the public Iteration API. I think it's enough to keep this private? Also could you please move the implementation to `Iteration.cpp`, to avoid future linker issues

guj: ok

include/openPMD/backend/ContainerImpl.tpp
guj: oh, didn't know, let me see

guj: I think I don't know linkHierarchy() enough to decide. It looks like a big change to me.  Maybe merge this pull request and  you make a new pull request and work on it?

franzpoeschel: I put the logic into `GenerationPolicy` now, `linkHierarchy` was not adequate upon closer looking. With this, I think we can merge.

franzpoeschel (approved):

PR #1866, opened 2026-03-26
HDF5: Close operations also upon failure

franzpoeschel:
Basically our entire HDF5 implementation is a resource leak in the error case. This PR uses RAII to implement a deferred cleanup operation (similar to how Go handles resource freeing)

TODO:

- [x] Add a test for this: Do some kind of nonexistent attribute read, the HDF5 handle will stay open, HDF5 will fail upon writing to the Series for a second time
- [x] fix this for the entire remaining HDF5 implementation......
- [x] Verify the vibe-coded commit

### DIFF
--- a/include/openPMD/auxiliary/Defer.hpp
+++ b/include/openPMD/auxiliary/Defer.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <functional>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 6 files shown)


include/openPMD/auxiliary/Defer.hpp
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
ax3l, line 1:
```suggestion
/* Copyright 2017-2026 Franz Poeschel, Axel Huebl
 *
 * This file is part of openPMD-api.
 *
 * openPMD-api is free software: you can redistribute it and/or modify
 * it under the terms of of either the GNU General Public License or
 * the GNU Lesser General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * openPMD-api is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License and the GNU Lesser General Public License
 * for more details.
 *
 * You should have received a copy of the GNU General Public License
 * and the GNU Lesser General Public License along with openPMD-api.
 * If not, see <http://www.gnu.org/licenses/>.
 */
#pragma once
```

PR #1873, opened 2026-04-20
Do not initially emplace any value for the cached Iteration index

franzpoeschel:
this avoids subtle bugs
followup to #1860


guj (approved):
nice catch!

PR #1874, opened 2026-04-21
Use most recent main branch of toml11

franzpoeschel:
Needed for these commits

* https://github.com/ToruNiina/toml11/pull/291
* https://github.com/ToruNiina/toml11/pull/295

Up for discussion: This disables use of `std::source_location` by default in openPMD. There are some compilers that have trouble with it, and it's a feature for developers, there is no reason to ship it to users, especially for an internal library.


ax3l (approved):
That LGTM :)

PR #1878, opened 2026-04-28
Python: reacquire GIL for deallocation operations

franzpoeschel:
This allows releasing the GIL for flushing, making it possible to run a data loading job in another Python thread

TODO:

- [x] Do the same for all other flush operations (Iteration::close, Attributable::seriesFlush)
- [x] Release 0.17.1 yay or nay? yay is defensible :D

### DIFF
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -324,7 +324,6 @@ struct StoreChunkFromPythonArray
         Offset const &offset,
         Extent const &extent)
     {
-        a.inc_ref();
         void *data = a.mutable_data();
         // here, we store an owning handle in the lambda capture so that
         // temporary and lost-scope variables stay alive until we flush
--- a/src/binding/python/RecordComponent.cpp
+++ b/src/binding/python/RecordComponent.cpp
@@ -331,9 +331,11 @@ struct StoreChunkFromPythonArray
         // note: this does not yet prevent the user, as in C++, to build
         // a race condition by manipulating the data that was passed
         std::shared_ptr<T> shared(
-            (T *)data, [owning_handle = a.cast<py::object>()](T *) {
+            (T *)data,
+            [owning_handle =
+                 std::make_optional(a.cast<py::object>())](T *) mutable {
                 py::gil_scoped_acquire need_the_gil_for_this;
-                owning_handle.~object();
+                owning_handle.reset();
             });
         r.storeChunk(std::move(shared), offset, extent);
     }
@@ -355,9 +357,11 @@ struct LoadChunkIntoPythonArray
         // note: this does not yet prevent the user, as in C++, to build
         // a race condition by manipulating the data that was passed
         std::shared_ptr<T> shared(
-            (T *)data, [owning_handle = a.cast<py::object>()](T *) {
+            (T *)data,
+            [owning_handle =
+                 std::make_optional(a.cast<py::object>())](T *) mutable {
                 py::gil_scoped_acquire need_the_gil_for_this;
-                owning_handle.~object();
+                owning_handle.reset();
             });
         r.loadChunk(std::move(shared), offset, extent);
     }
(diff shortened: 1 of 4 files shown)


src/binding/python/Series.cpp
```diff
@@ -35,6 +35,7 @@
 
 #include <optional>
 #include <pybind11/attr.h>
+#include <pybind11/gil.h>
```
ax3l, line 38: Just checked: `#include <pybind11/gil.h>` causes no ODR issues if included in some TUs and not others.

src/binding/python/RecordComponent.cpp
```diff
@@ -324,15 +325,17 @@ struct StoreChunkFromPythonArray
         Offset const &offset,
         Extent const &extent)
     {
-        a.inc_ref();
```
ax3l, line 327: Why is that removed now? because `.reset()` blocks now until `a` is handled?

src/binding/python/RecordComponent.cpp
```diff
@@ -324,15 +325,17 @@ struct StoreChunkFromPythonArray
...
         // a race condition by manipulating the data that was passed
         std::shared_ptr<T> shared(
-            (T *)data, [owning_handle = a.cast<py::object>()](T *) {
-                // no-op
+            (T *)data,
+            [owning_handle =
+                 std::make_optional(a.cast<py::object>())](T *) mutable {
+                py::gil_scoped_acquire need_the_gil_for_this;
+                owning_handle.reset();
             });
```
ax3l, line 339: I am confused what happens here. In `dev` it was a no-op, now it does aquire the gil and call reset. The do string is the same 😅 what's happening :D

src/binding/python/RecordComponent.cpp
franzpoeschel: That was a bug I noticed while working on this PR, I extracted this to https://github.com/openPMD/openPMD-api/pull/1879. 
With https://github.com/openPMD/openPMD-api/pull/1775, I removed explicit ref-counting and replaced it by owning Pybind references. The unmatched `inc_ref()` is basically a memory leak.

src/binding/python/RecordComponent.cpp
franzpoeschel: Before this PR, `owning_handle` is auto-destroyed by RAII, that's why the lambda itself is a no-op. But since destroying a Python handle latches into the Python interpreter, we need to do it while holding the GIL, so I moved the destruction inside the function body.

src/binding/python/Series.cpp
franzpoeschel: Ah, my editor inserted these, I can quickly remove them again

src/binding/python/Series.cpp
```diff
@@ -283,22 +284,8 @@ not possible once it has been closed.
             .def(
                 "__getitem__",
                 [](Snapshots &s, Series::IterationIndex_t key) {
```
ax3l, line 286: Codex saw this:
>   - src/binding/python/Series.cpp:286: the new unconditional py::gil_scoped_release also covers SnapshotWorkflow::RandomAccess. In
    that path s[key] reaches RandomAccessIteratorContainer::operator[], which can mutate the underlying container by inserting a
    missing iteration. That was previously serialized by the GIL and is not a flushing/blocking path. I’d keep the old switch and
    only release the GIL for the synchronous workflow.

src/binding/python/Series.cpp
ax3l: But we use `py::gil_scoped_release` below, so the include is good.

src/binding/python/Series.cpp
ax3l: Careful: `s[key]` in C++ inserts an element if missing on first access.

franzpoeschel: This bug existed before this PR, but we can use this PR to fix it

franzpoeschel: Update: Not a bug, false positive. This binds `Snapshots`, an opaque C++ map type, not managed by Python. Python only accesses this through methods, so the interpreter is not bothered by container state manipulation.

PR #1880, opened 2026-05-18
Release 0.17.1

franzpoeschel:
https://github.com/openPMD/openPMD-api/milestone/17

## TODO

- [x] Remaining items from the milestone
- [x] Docs
- [x] toml11, maybe use recent master commit?

### Python

```python
import openpmd_api
```

### C++

```C++
#include <iostream>

int main() {
    // query compile-time macros:

    return 0;
}
```

### CLI

```bash
openpmd-ls --version
```


ax3l (approved):
Looks great, I'll tag the release now.

PR #1896, opened 2026-06-24
`setup.py`: Windows `openPMD_CMAKE_...` Env Vars

ax3l:
We pass `openPMD_CMAKE_<CMakeVar>=<value>` environment variables through as `-D` flags to CMake configure.

Windows environment variable are all CAPS, so also accept the all-caps variant of the prefix, then the `openPMD_` CMake option names are case-sensitive,
so restore the `openPMD_...` prefix that upper-casing would destroy. All suffixes of our options are
luckily `..._ALL_CAPS` anyway.

Noticed in #1892 that our `openPMD_CMAKE_openPMD_USE_ADIOS2='ON'` env var was ignored on Windows, defaulting to `AUTO`, which skipped ADIOS2 since we had a transitive dependency issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -173,7 +173,7 @@ CMAKE_INTERPROCEDURAL_OPTIMIZATION = os.environ.get(
 extra_cmake_args = []
 # Pass openPMD_CMAKE_<CMakeVar>=<value> environment variables through as -D
 # flags. Windows environment variable are all CAPS, so also accept the
-# all-caps variant of the prefix, then the openPMD_ CMake option names are case-
+# all-caps variant of the prefix, then the openPMD_ CMake option names are case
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -171,13 +171,23 @@ def build_extension(self, ext):
 
 # extra CMake arguments
 extra_cmake_args = []
+# Pass openPMD_CMAKE_<CMakeVar>=<value> environment variables through as -D
+# flags. Windows environment variable are all CAPS, so also accept the
+# all-caps variant of the prefix, then the openPMD_ CMake option names are case-
```
ax3l:
```suggestion
# all-caps variant of the prefix, then the openPMD_ CMake option names are case
```

PR #1902, opened 2026-06-28
HDF5: harden wasm read type-detection against H5Tequal errors

ax3l:
H5Tequal returns a tri-state htri_t (>0 equal, 0 not equal, <0 ERROR), but the HDF5 backend's read type-detection used it as a plain bool, so a <0 error counted as "equal".

In the Pyodide/wasm wheel, openPMD's manually-built 80-bit long-double types (m_H5T_LONG_DOUBLE_80_LE, m_H5T_CLONG_DOUBLE_80_LE) end up as INVALID type ids (the handler destructor fails to H5Tclose them). Comparing any type against an invalid id makes H5Tequal return <0, which short-circuited the detection chain and mis-detected EVERY attribute/dataset -- including the 'openPMD' format-version string -- as LONG_DOUBLE, so reading any openPMD HDF5 file on wasm failed with "Unexpected Attribute datatype for 'openPMD' (expected string, found LONG_DOUBLE)". Writing was unaffected, which is why it went unnoticed.

Guard every comparison against these custom types with an explicit `> 0`, in the dataset reader and both the scalar- and vector-attribute readers, so an H5Tequal error means "not equal" and types fall through correctly. Safe on all platforms: where the types are valid H5Tequal never returns <0, so `> 0` matches the prior behavior and only the error case changes.

Note: this hardens the READER. Why the 80-bit type is invalid specifically in the Pyodide side-module build is a separate issue -- a minimal *static* wasm build of the same HDF5 1.14.6 constructs the type fine (long double is 128-bit there) and H5Tequal works -- and resolving it is what would enable reading genuine 80-bit long-double *data* on wasm.

### DIFF
--- a/src/IO/HDF5/HDF5IOHandler.cpp
+++ b/src/IO/HDF5/HDF5IOHandler.cpp
@@ -1618,16 +1618,16 @@ void HDF5IOHandlerImpl::openDataset(
             else if (H5Tequal(dataset_type, H5T_NATIVE_DOUBLE))
                 d = DT::DOUBLE;
             else if (
-                H5Tequal(dataset_type, H5T_NATIVE_LDOUBLE) ||
-                H5Tequal(dataset_type, m_H5T_LONG_DOUBLE_80_LE))
+                H5Tequal(dataset_type, H5T_NATIVE_LDOUBLE) > 0 ||
+                H5Tequal(dataset_type, m_H5T_LONG_DOUBLE_80_LE) > 0)
                 d = DT::LONG_DOUBLE;
-            else if (H5Tequal(dataset_type, m_H5T_CFLOAT))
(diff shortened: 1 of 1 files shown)


ax3l:
Potential follow-up PR:
  > In the vector-attribute reader, m_H5T_CFLOAT, m_H5T_CDOUBLE, and m_H5T_CLONG_DOUBLE
  > still use bare H5Tequal(...) checks at src/IO/HDF5/HDF5IOHandler.cpp:3121, 3127, and
  > 3133, while the dataset reader now uses > 0 for the same custom compound types. Not
  > needed for the reported 80-bit wasm failure if those handles are guaranteed valid, but
  > using > 0 here too would make the hardening consistent and avoid the same class of
  > issue if any custom compound type comparison ever returns <0.

ax3l:
added

PR #1903, opened 2026-06-28
CI: WASM Pyodide

ax3l:
CI test for WASM Pyodide.

Ensure co-loading with `h5py`, even when both use different HDF5 versions, works.


.github/workflows/wasm.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: Build wheel + run the Python suite in Pyodide
+      run: python -m cibuildwheel --platform pyodide --output-dir wheelhouse
+      env:
+        CIBW_BUILD: "cp314-pyodide_wasm32"
+        CIBW_BEFORE_BUILD_PYODIDE: "bash .github/ci/wasm_deps.sh"
+        CIBW_ENVIRONMENT_PYODIDE: >
+          openPMD_CMAKE_openPMD_USE_HDF5='ON'
+          openPMD_CMAKE_openPMD_USE_ADIOS2='OFF'
+          openPMD_USE_MPI='OFF'
+          HDF5_USE_STATIC_LIBRARIES='ON'
```
ax3l, line 48: Both might need openPMD_CMAKE_ prefix

test/python/unittest/API/APITest.py
```diff
@@ -2333,7 +2333,10 @@ def testScalarHdf5Fields(self):
 
         series_read_write = io.Series(file, io.Access.read_write)
         E_x = series_read_write.iterations[0].meshes["E"]["x"]
-        E_x[:] = np.array([45])
+        # match the component's own dtype: h5py wrote the scalar with its
+        # native int (int32 on wasm32) while np.array([45]) defaults to int64,
+        # so store the component's dtype rather than assume the two agree.
```
ax3l: Simplify
```suggestion
```

.github/workflows/wasm.yml
ax3l: Oh, special case we handle: https://github.com/openPMD/openPMD-api/blob/092db098feba3e5f7d56346c5e2be0aa1a1ec125/setup.py#L156-L157

CMakeLists.txt
```diff
@@ -647,12 +647,21 @@ if(openPMD_HAVE_PYTHON)
...
     # the module init symbol so each extension keeps its dependencies private.
-    # Windows does not auto-export and Emscripten links a single namespace,
-    # so neither needs this.
+    # Windows does not auto-export, so it needs nothing. Emscripten needs BOTH
+    # halves: -sSIDE_MODULE=2 stops Pyodide's default -sSIDE_MODULE=1 from
+    # whole-archiving and re-exporting every symbol, and the statically linked
+    # deps must additionally be built -fvisibility=hidden (.github/ci/wasm_deps.sh)
+    # so their definitions are DSO-local -- only then can wasm-ld relax our GOT
+    # references to HDF5 into direct calls into our own copy, instead of binding
+    # to a co-loaded second HDF5 (h5py, ImpactX).
```
```suggestion
    # deps must additionally be built -fvisibility=hidden
    # so their definitions are DSO-local. Only then can wasm-ld relax our GOT
    # references to HDF5 into direct calls into our own copy, instead of binding
    # to a co-loaded second HDF5 (h5py, ImpactX).
```

CMakeLists.txt
```diff
@@ -647,12 +647,21 @@ if(openPMD_HAVE_PYTHON)
...
     if(APPLE)
         target_link_options(openPMD.py PRIVATE
             "LINKER:-exported_symbol,_PyInit_openpmd_api_cxx")
-    elseif(NOT WIN32 AND NOT EMSCRIPTEN)
+    elseif(EMSCRIPTEN)
+        target_link_options(openPMD.py PRIVATE
+            "SHELL:-sSIDE_MODULE=2"
+            "SHELL:-sEXPORTED_FUNCTIONS=_PyInit_openpmd_api_cxx")
```
ax3l, line 663: It turns out #1900 + matching the openPMD and HDF5 versions in ImpactX wheels with openPMD-api wheels was the fix here.

I am not currently aware of a functional way to properly hide symbols in WASM.

`EXPORTED_FUNCTIONS` is probably still sensible to mark which functions we need at all to expose to JS.

PR #1904, opened 2026-06-28
Wheels 0.17.1.post4

ax3l:
- [x] bump version
- [x] rebase against wheels branch
- [x] Win mutext issue seen in https://github.com/BLAST-ImpactX/impactx-wheels/pull/26
- [x] pull in whatever we find in #1903
- [x] skip RTD PR builds on `wheels` for good 

## `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` (`CXXFLAGS)`

VS 2022 17.10 made `std::mutex`'s constructor constexpr, leaving its internals NULL until first use. We `--exclude msvcp*.dll` from the wheel, so the SYSTEM `msvcp140.dll` runs; if it predates 17.10 it NULL-derefs in `Mtx_destroy` -> `import openpmd_api` access-violates on win64 (surfaced when co-loaded under the ImpactX wheel's test). The define reverts to the non-constexpr constructor, ABI-compatible with any `msvcp140`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ env:
     .github/setup-py-version-post3.patch
     .github/setup-py-windows-cmake-options.patch
     .github/python-hide-symbols.patch
+    .github/python-wasm-side-module.patch
     .github/hdf5-dont-atexit-emscripten.patch
     .github/hdf5-read-h5tequal-tristate.patch
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 env:
   SRC_REF: '0.17.1'
   COMMON_PATCHES: >
-    .github/setup-py-version-post3.patch
+    .github/setup-py-version-post4.patch
     .github/setup-py-windows-cmake-options.patch
     .github/python-hide-symbols.patch
     .github/python-wasm-side-module.patch
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -193,11 +193,19 @@ jobs:
           openPMD_USE_MPI='OFF'
           HDF5_USE_STATIC_LIBRARIES='ON'
           ZLIB_USE_STATIC_LIBS='ON'
+        # _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR (CXXFLAGS): VS 2022 17.10 made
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -193,11 +193,19 @@ jobs:
...
+        # msvcp140.dll runs; if it predates 17.10 it NULL-derefs in Mtx_destroy ->
+        # `import openpmd_api` access-violates on win64 (surfaced when co-loaded
+        # under the ImpactX wheel's test). The define reverts to the non-constexpr
+        # constructor, ABI-compatible with any msvcp140.
         CIBW_ENVIRONMENT_WINDOWS: >
           HDF5_USE_STATIC_LIBRARIES='ON'
           ZLIB_USE_STATIC_LIBS='ON'
           openPMD_CMAKE_openPMD_USE_HDF5='ON'
           openPMD_CMAKE_openPMD_USE_ADIOS2='${{ matrix.env.ADIOS2 || 'ON' }}'
+          CXXFLAGS='/D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR'
```
ax3l, line 203: Should also be set for library_builders.bat

ax3l: done

.github/workflows/build.yml
```diff
@@ -193,11 +194,19 @@ jobs:
           openPMD_USE_MPI='OFF'
           HDF5_USE_STATIC_LIBRARIES='ON'
           ZLIB_USE_STATIC_LIBS='ON'
+        # _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR (CXXFLAGS): VS 2022 17.10 made
+        # std::mutex's constructor constexpr, leaving its internals NULL until
+        # first use. We --exclude msvcp*.dll from the wheel (below), so the SYSTEM
+        # msvcp140.dll runs; if it predates 17.10 it NULL-derefs in Mtx_destroy ->
+        # `import openpmd_api` access-violates on win64 (surfaced when co-loaded
+        # under the ImpactX wheel's test). The define reverts to the non-constexpr
+        # constructor, ABI-compatible with any msvcp140.
```
ax3l: Simplify

```suggestion
        # _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR for portability to older runtimes
```

PR #1909, opened 2026-07-03
Wheels: ARM64 builders for Linux & Windows

ax3l:
Travis-CI is dead; it was the only builder for Linux aarch64 (and ppc64le) wheels. GitHub now offers free [native ARM runners for public repos](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-git), so move Linux ARM wheel production into build.yml and add Windows ARM64 (new platform).

build.yml:
- Linux aarch64 on ubuntu-24.04-arm: full parity with x86_64 (cp310-cp314, many+musl, ADIOS2 ON), reuses library_builders.sh unchanged (native aarch64 needs no SIMD/cross tweaks; blosc2 auto-selects NEON, HDF5 builds natively).
- Windows ARM64 on windows-11-arm: cp311-cp314 (CPython win_arm64 exists only from 3.11+, no PyPy). First cut ships HDF5 + JSON only (ADIOS2 OFF), like the win32/i686 legs, to avoid an unproven ADIOS2-on-ARM64 build.
- Smoke test: derive the ADIOS2 expectation from matrix.env.ADIOS2 instead of process bitness, so a 64-bit ARM wheel without ADIOS2 passes; also makes the i686/win32 legs explicit.
- Retarget the retired Travis aarch64/ppc64le matrix comments.

library_builders.bat:
- Gate the ADIOS2 dependency chain (SQLite3, ZFP, C-Blosc2, ADIOS2) behind openPMD_CMAKE_openPMD_USE_ADIOS2==ON (visible in the before-build step); keep zlib + HDF5 unconditional. Windows ARM64 builds only zlib + HDF5; AMD64 unchanged (win32 stops building the ADIOS2 stack it already discarded).

.travis.yml:
- Sync to the current post4 patch and cibuildwheel 4.1.0 (kept as reference; Travis no longer runs).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,7 @@ env:
   COMMON_PATCHES: >
     .github/setup-py-version-post4.patch
     .github/setup-py-windows-cmake-options.patch
+    .github/setup-py-windows-arm64-platform.patch
     .github/python-hide-symbols.patch
     .github/python-wasm-side-module.patch
     .github/hdf5-dont-atexit-emscripten.patch
@@ -68,6 +69,7 @@ jobs:
             build: "cp311-* cp312-* cp313-* cp314-*"
             env:
               ADIOS2: "OFF"
+              CMAKE_GENERATOR_PLATFORM: "ARM64"
 
           - os: macos-15-intel
             arch: "x86_64"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,7 @@ jobs:
         #   platform     CIBW_PLATFORM                 (default auto)
         #   build        CIBW_BUILD                    (default *, i.e. all)
         #   patches      space-separated setup patches (default: Windows one)
+        #   artifact     artifact-name suffix override (default: build || arch)
         #   sdist        also build + upload the sdist artifact (default: off)
         include:
           - os: ubuntu-22.04
@@ -67,6 +68,7 @@ jobs:
           - os: windows-11-arm
             arch: "ARM64"
             build: "cp311-* cp312-* cp313-* cp314-*"
+            artifact: "ARM64"   # build has '*'/spaces; not a valid artifact name
             env:
               ADIOS2: "OFF"
               CMAKE_GENERATOR_PLATFORM: "ARM64"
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -38,11 +38,15 @@ jobs:
             env:
               ADIOS2: "OFF"
 
-          # builds faster on Travis-CI:
-          #- os: ubuntu-20.04
-          #  arch: "aarch64"
+          # Linux ARM64 (aarch64): native GitHub-hosted runner, free for public
+          # repos (cibuildwheel builds aarch64 natively, no QEMU). Full parity with
+          # the x86_64 leg above: cp310-cp314, many + musl, ADIOS2 ON.
+          # (Formerly built on Travis-CI, which is retired.)
```
ax3l:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -63,6 +67,17 @@ jobs:
               CMAKE_GENERATOR_PLATFORM: "Win32"
               ADIOS2: "OFF"   # 32-bit: no 32-bit ADIOS2 CI; BP read segfaults (see i686)
 
+          # Windows ARM64: native GitHub-hosted runner, free for public repos
+          # (cibuildwheel ARM64 support is still experimental). CPython win_arm64
+          # exists only from 3.11+, so no cp310 and no PyPy here. First cut ships
+          # HDF5 + JSON only (ADIOS2 OFF) to avoid an unproven ADIOS2-on-ARM64 build;
+          # library_builders.bat then compiles only zlib + HDF5. See i686/win32.
```
ax3l:
```suggestion
```

library_builders.bat
```diff
@@ -288,10 +288,20 @@ exit /b 0
...
 call :build_hdf5
-call :build_adios2
+
+rem ADIOS2 and its exclusive dependencies (SQLite3, ZFP, C-Blosc2) are only built
+rem when openPMD is configured with ADIOS2 (openPMD_CMAKE_openPMD_USE_ADIOS2=ON, set
+rem via CIBW_ENVIRONMENT_WINDOWS and visible here in the before-build step). Targets
+rem that ship HDF5 + JSON only -- Windows ARM64 and win32 (x86) -- skip this whole
+rem chain, which avoids an unproven ADIOS2-on-ARM64 build and trims the required
+rem tooling to curl + PowerShell + CMake. HDF5 depends only on zlib, so it stays
+rem unconditional above.
```
ax3l:
```suggestion
rem ADIOS2 and its exclusive dependencies
```

.github/workflows/build.yml
```diff
@@ -63,6 +65,14 @@ jobs:
               CMAKE_GENERATOR_PLATFORM: "Win32"
               ADIOS2: "OFF"   # 32-bit: no 32-bit ADIOS2 CI; BP read segfaults (see i686)
 
+          - os: windows-11-arm
+            arch: "ARM64"
+            build: "cp311-* cp312-* cp313-* cp314-*"
+            artifact: "ARM64"   # build has '*'/spaces; not a valid artifact name
```
ax3l:
```suggestion
            artifact: "WinArm64"
```

.github/workflows/build.yml
```diff
@@ -7,6 +7,7 @@ env:
   COMMON_PATCHES: >
     .github/setup-py-version-post4.patch
     .github/setup-py-windows-cmake-options.patch
+    .github/setup-py-windows-arm64-platform.patch
```
ax3l, line 10: TODO: This still needs a PR to `dev`!

ax3l: Done: #1910

PR #1914, opened 2026-07-09
Two little fixes for scientific defaults

franzpoeschel:
Follow-up to #1834, affects only dev.


src/backend/scientific_defaults/ConfigAttribute.cpp
```diff
@@ -109,7 +109,7 @@ void ConfigAttribute::read()
                 write_to_stderr(attribute) << "'. Expected one of ";
                 auxiliary::write_vec_to_stream(
                     std::cerr, type_unmatched.expectedDatatypes)
-                    << " or convertible to such a type." << std::endl;
```
franzpoeschel, line 112: Matches at this point are precise. Datatype conversions may optionally occur in later postprocessing (see `require_type` inside `ProcessParsedAttribute.hpp`).

PR #1916, opened 2026-07-13
Test: Attribute Only Iterations

ax3l:
Check that we can write and read iterations that only contain attributes. Should work with ADIOS2 BP5, HDF5 and text backends.

### DIFF
--- a/test/SerialIOTest.cpp
+++ b/test/SerialIOTest.cpp
@@ -2491,6 +2491,100 @@ TEST_CASE("bool_test", "[serial]")
     }
 }
 
+void attributes_only_over_iterations_test(
+    std::string const &file_ending, IterationEncoding const it_encoding)
+{
+    std::string name = "../samples/attributes_only_over_iterations_";
+    if (it_encoding == IterationEncoding::fileBased)
+        name.append("f_%T");
+    else if (it_encoding == IterationEncoding::groupBased)
+        name.append("g");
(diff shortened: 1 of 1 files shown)


test/SerialIOTest.cpp
```diff
@@ -2491,6 +2491,100 @@ TEST_CASE("bool_test", "[serial]")
     }
 }
 
+void attributes_only_over_iterations_test(
```
franzpoeschel, line 2494: Could you please move new test code to separate files under test/Files_SerialIO` to not blow up `SerialIOTests.cpp` further :)

test/SerialIOTest.cpp
```diff
@@ -2491,6 +2491,100 @@ TEST_CASE("bool_test", "[serial]")
...
+        }
+        write.close();
+    }
+
+    /*
+     * Read the attributes back and verify that the per-iteration attribute
+     * carries the value written for its iteration.
+     */
+    {
+        Series read(name, Access::READ_ONLY);
```
franzpoeschel, line 2536: Since the parsing procedures for `READ_ONLY` and `READ_LINEAR` are internally very different, it might be worth to run the same read test also for `READ_LINEAR`:

```cpp
        Series read(name, Access::READ_LINEAR);
        read.parseBase();
```

franzpoeschel (changes requested):

Issue #1919, opened 2026-07-16
Pickle: Mesh points at wrong series after deserializing


pordyna:
**Describe the bug**
When serializing and deserializing a series object in python and doing the same for a `MeshRecord` (or `MeshRecordComponent` ?) both will successfully deserialize. But, although calling `load_chunk` on the deserialized mesh will work, calling `flush` on the deserialized series won't actually load the data. 
There is no proper error message, but the buffer will contain random data. When serializing with some parallelization libraries like `joblib` workers can fail with messages like:
```
Fatal Python error: gilstate_tss_set: failed to set current tstate (TSS)
Python runtime state: finalizing (tstate=0x00000000017f9e18)
```

**To Reproduce**
`uv run deserialize_bug.py`
[deserialize_bug.py](https://github.com/user-attachments/files/30080371/deserialize_bug.py)

**Expected behavior**
Since the `MeshRecordComponent` and `MeshRecord` both have `add_pickle` both methods in the example should work

**Software Environment**
 See attached file

pordyna:
@franzpoeschel this is not really urgent,  just pinging in case you haven't seen it yet.

franzpoeschel:
> [@franzpoeschel](https://github.com/franzpoeschel) this is not really urgent, just pinging in case you haven't seen it yet.

Ah thanks, I had forgotten about it
Will check

franzpoeschel:
I see that something is going wrong, but I have not yet understood why
Will continue

franzpoeschel:
Okaayyy, I understand why this happens, but this might get ugly
Our serialization/deserialization strategy is to save only the filename + the location of the object within that file (e.g. /400/meshes/rho), and then upon deserialization init a new Series and point it to the right object again.
Consequence: After deserializing, `series`, `it` and `mesh` are three entirely different openPMD Series objects. Meaning that `series.flush()` won't do things that you enqueued into `mesh`. 

You can work around this by adding a line `mesh.series_flush()`, but the pitfall remains.. No clue how to easily solve that at the moment

pordyna:
If there is no easy solution maybe you should think about removing pickle capabilities from everything except for series. And just force the user to rebuild everything from series.

franzpoeschel:
> If there is no easy solution maybe you should think about removing pickle capabilities from everything except for series. And just force the user to rebuild everything from series.

That would be the last resort, and we will have to consider it if there is no other solution
As this would break a lot of production use cases, I would like to avoid it
[...]

PR #1922, opened 2026-08-04
Pickle: Consider Series identity when unpickling

franzpoeschel:
This tries fixing the bug reported in #1919 by @pordyna by storing and reading additionally the immutable internal Series memory location. If the location of an unpickled Series is identical to the memory location of a previously unpickled Series, then that Series is reused.

Notes:

* ~~The cache of unpickled Series objects is still `thread_local` at the moment. This keeps thread safety at the cost of restricting this bugfix to uses in the same thread.~~
    **Suggestion:** When unpickling from the same `Series` object, users should expect to share the same internal data references. So it might be fine to use a per-process cache instead of a per-thread cache.
* ~~The cache is never emptied, so this can be a memory leak.~~ We *could* maybe add a scan operation that erases closed `Series` objects from the cache?
* Since our pickling strategy is relatively basic, we will probably keep running into issues like this, as use cases get more complex.

- [x] tqdm as test dependency
- [x] check if we can automatically manage memory, this currently relies on `Series::close()`
- [x] we can ditch the make_owning logic with this
- [x] More testing?
- [ ] This fix is a bit too experimental for a bugfix release. For an eventual bugfix release, we could move the `thread_local Series` cache variable from (template) function scope to global scope. This would not fix the entire problem, but it would be a fix for use cases that only use one `Series`.

### DIFF
--- a/include/openPMD/binding/python/Pickle.hpp
+++ b/include/openPMD/binding/python/Pickle.hpp
@@ -51,7 +51,9 @@ namespace openPMD
  * For this, the pickle structure contains as first entry the internal
  * (immutable) SharedAttributable pointer address of the `Series` object
  * referenced by any handle. When unpickling, this is used to restore shared
- * handles in accordance with their original reference structure.
+ * handles in accordance with their original reference structure. The pointers
+ * themselves are not restored (this would not be possible), but they are used
+ * as equivalence classes.
  */
 struct unpickled_series
 {
--- a/include/openPMD/binding/python/Pickle.hpp
+++ b/include/openPMD/binding/python/Pickle.hpp
@@ -29,6 +29,7 @@
 
 #include <cstdint>
 #include <exception>
+#include <memory>
 #include <shared_mutex>
 #include <string>
 #include <sys/types.h>
@@ -39,10 +40,11 @@ namespace openPMD
 {
 struct unpickled_series
 {
-    std::map<uintptr_t, Series> m_series_by_former_id;
+    std::map<uintptr_t, std::weak_ptr<Series>> m_series_by_former_id;
     std::shared_mutex m_mutex;
 
-    auto get(uintptr_t id, std::string const &filename) -> Series &;
+    auto get(uintptr_t id, std::string const &filename)
+        -> std::shared_ptr<Series>;
 };
 
 /*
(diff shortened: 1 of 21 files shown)


test/python/unittest/API/APITest.py
```diff
@@ -2523,6 +2523,48 @@ def get_component_only():
...
+                dens = mesh[:]
+            self.series.flush()
+            return dens
+
+    def testPickleSeriesIdentity(self):
+        # This tests the bug reported in
+        # https://github.com/openPMD/openPMD-api/issues/1919
+        # The code is adapted from the reproducer in there.
+        try:
+            from tqdm.contrib.concurrent import process_map
```
pordyna: Do you add `tqdm` as a requirement for your CI? Otherwise, this will never be tested. You could maybe also rewrite it with `Pool().map()` from `multiprocessing` if you want to avoid the extra dependency.

franzpoeschel: Not yet, I have added it as a TODO
I want to give the CI one try on the test without fix anyway, so it will come up latest at that point

ax3l: pytest has @skipif that we can use for missing imports.

ax3l: I agree we can probably reproduce this with multiprocessing from the Python stdlib

include/openPMD/binding/python/Pickle.hpp
```diff
@@ -27,13 +27,29 @@
...
+#include <sys/types.h>
 #include <tuple>
 #include <vector>
 
 namespace openPMD
 {
+struct unpickled_series
+{
+    std::map<uintptr_t, Series> m_series_by_former_id;
+    std::shared_mutex m_mutex;
```
ax3l, line 68: An idea:
Usually when we create series in Python, the Python interpreter owns the lifetime. Now after pickle, there is a C++ object owning the lifetime.

There might be a way to attach the extra data/counter/id to the Python object dynamically and reference it (or the series that is de-serialized first) between each other in Python.

franzpoeschel: The last commit tries sth like that

test/python/unittest/API/APITest.py
franzpoeschel: @ax3l Pyodide does not seem to have multiprocessing. I don't know if it's just not in our CI run or if it's legitimately not there. The easiest way out would be to just add a `try...except` block around the import again.

include/openPMD/binding/python/Pickle.hpp
franzpoeschel: Lifetime is now managed from the Python side, with tests for different workflows.

test/python/unittest/API/APITest.py
franzpoeschel: pyodide now excluded from tests
note that multiprocessing seems to be in pyodide, but it's broken

PR #1935, opened 2026-09-04
Experimental: Nix packaging

franzpoeschel:
Pros:

- one further deployment target
- developers can enter a reproducible development shell with all necessary dependencies using `nix develop`
- reproducible software environment for CI runs
- openPMD tooling available on-the-fly without installation via `nix run <flake-location>#ls` (or pipe)
- HDF5 and ADIOS2 are also deployed there

Todo:

- [ ] Add an integration test for those tests that are too complex for the Nix build sandbox


ax3l (approved):
Cool!

I actually would also like to add a pixi recipe to the repo, which I find is easier to use.

PR #1939, opened 2026-09-11
Fix Python `StatefulIterator`

### LINKED ISSUE #1938: Python StatefulIterator is missing __iter__, breaking iterator consumers
(prepared with Codex)

`StatefulIterator` returned by `iter(series.read_iterations())` implements `__next__` but does not implement `__iter__`. Consequently, `iter(iterator)` raises `TypeError` instead of returning the iterator itself. This breaks consumers that request an iterator from an existing iterator.

We encountered this in the list comprehension that reads BeamMonitor output in [ImpactX PR #1381](https://github.com/BLAST-ImpactX/impactx/pull/1381). The failure happens in the aliased-monitor case, before a copied monitor is constructed; it is in the Python reader rather than the ImpactX writer.

## Minimal reproduction (no ImpactX needed)

```python
from pathlib import Path
from tempfile import TemporaryDirectory
import openpmd_api as io

with TemporaryDirectory() as directory:
    path = str(Path(directory) / "iterator.json")
    writer = io.Series(path, io.Access.create)
    writer.iterations[0].set_time(0.0)
    writer.iterations[1].set_time(0.5)
    writer.close()

    reader = io.Series(path, io.Access.read_linear)
    iterator = iter(reader.read_iterations())
    print(type(iterator).__name__)           # StatefulIterator
    print(hasattr(iterator, "__next__"))    # True
    print(hasattr(iterator, "__iter__"))    # False
    print(next(iterator).time)              # 0.0
    try:
        assert iter(iterator) is iterator  # TypeError: ...StatefulIterator... is not iterable
    finally:
        reader.close()
```

I reproduced this on Linux with Python 3.12.9 and openPMD-api 0.17.1. The explicit iterator-protocol reproduction above does not depend on comprehension implementation details.

The ImpactX CI failure was with Python 3.13.15 and the PyPI package `openpmd-api==0.17.1.post4`, at:

```python
positions = [
    iteration.particles["beam"].get_attribute("s_ref")
    for iteration in series.read_iterations()
]
```

[Failing CI traceback](https://github.com/BLAST-ImpactX/impactx/actions/runs/34552335983/job/103117663158) (`tests/python/test_element_copy.py:117`). A [second CI job](https://github.com/BLAST-ImpactX/impactx/actions/runs/34552336063) fails the same way. CPython has also tracked related comprehension behavior in [python/cpython#128161](https://github.com/python/cpython/issues/128161); implementing the iterator protocol makes the binding robust regardless of those implementation details.

## Expected behavior and suggested fix

`iter(iterator) is iterator`, and the original list comprehension should work. In `src/binding/python/Series.cpp`, the `StatefulIteratorPythonAdaptor` binding currently registers only `__next__`. Please also bind `__iter__` to return `self`, preserving Python identity and lifetime (for example, a lambda accepting and returning `py::object`).

A regression test should check both `iter(iterator) is iterator` and consumption of the resulting iterator, in addition to `for iteration in series.read_iterations()`.

## Workaround in ImpactX

An explicit `for` loop avoids the additional iterator conversion in affected comprehensions, without weakening the output assertions:

```python
positions = []
for iteration in series.read_iterations():
    positions.append(iteration.particles["beam"].get_attribute("s_ref"))
series.close()
```

The standalone JSON reproduction reads `[0.0, 0.5]` successfully with this loop on the local environment above.

ax3l:
Fix #1938

### DIFF
--- a/src/binding/python/Series.cpp
+++ b/src/binding/python/Series.cpp
@@ -310,6 +310,7 @@ not possible once it has been closed.
     finalize_container<Snapshots>(snapshots, /* skip_getitem = */ true);
 
     py::class_<StatefulIteratorPythonAdaptor>(m, "StatefulIterator")
+        .def("__iter__", [](py::object self) { return self; })
         .def(
             "__next__",
             [](StatefulIteratorPythonAdaptor &iterator) {
--- a/src/binding/python/Series.cpp
+++ b/src/binding/python/Series.cpp
@@ -310,6 +310,12 @@ not possible once it has been closed.
     finalize_container<Snapshots>(snapshots, /* skip_getitem = */ true);
 
     py::class_<StatefulIteratorPythonAdaptor>(m, "StatefulIterator")
+        /*
+         * Note: This does NOT copy at C++ level, but at Python level (note the
+         * parameter type py::object) and is hence a shallow copy. The returned
+         * iterator is the same instance as the provided. This is the correct
+         * choice for stateful shared iterators.
+         */
         .def("__iter__", [](py::object self) { return self; })
         .def(
             "__next__",
--- a/test/python/unittest/API/APITest.py
+++ b/test/python/unittest/API/APITest.py
@@ -2010,7 +2010,9 @@ class APITest(unittest.TestCase):
             io.Access_Type.read_only,
             jsonConfig,
         )
-        for it in read.read_iterations():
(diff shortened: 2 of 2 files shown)


franzpoeschel (approved):
Added a small comment, because I was initially confused why this works.

